PR #7, opened 2017-10-27
Update README.md - step-by-step example for compiling on Ubuntu

nshiff:
Provide step-by-step instructions for compiling on Ubuntu.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,11 @@ Run make and hope it works. You might need to edit the paths in the Makefile.
     cd leela-zero/src
     sudo apt install libboost-all-dev libopenblas-dev opencl-headers ocl-icd-libopencl1 ocl-icd-opencl-dev 
     make
+    cd ..
+    wget https://sjeng.org/zero/best.txt.zip
+    unzip https://sjeng.org/zero/best.txt.zip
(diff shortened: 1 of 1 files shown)


WuSiying315:
But what I ran  src/leelaz --help  and then I get back an error message, no such file or directory
I want to know what went wrong

nshiff:
@WuSiying315 Please try the following:

    # Display your current working directory.
    # It will probably be like /home/WuSiying315/Documents/leela-zero
    # At least, it should end with "leela-zero" 
    pwd

    # Double check by displaying directories and files in your current directory
    # You should see src, COPYING, README.md
    ls -l

    # If you are not there, use the cd command to enter the leela-zero directory. 
    # This is analogous to visiting a folder in the graphical File Explorer.

    # Double check that the leela binary has been created.
    # If it is not there ... then something is wrong. ;-)
    ls -l src/leela-zero

    # If it is there, then you should be able to run.
    src/leela-zero

Hopefully that is helpful. If you need further assistance, I recommend creating an "Issue," as those represent problems either with setup or with the application itself. That way it would be a little easier for people to keep track of your specific issue. And, it will help other people in the future!

PR #8, opened 2017-10-27
Add scale layer after bn

mzmssg:
In Caffe, a BatchNorm layer followed by a Scale layer is the fully same operation as original paper. Reference:[Caffe BatchNorm Details](http://caffe.berkeleyvision.org/doxygen/classcaffe_1_1BatchNormLayer.html#details)

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ weights.txt file on the command line with the -w option.
 
 # Compiling
 
-## Requirements
+Requirements:
 
 * GCC, Clang or MSVC, any C++14 compiler
 * boost 1.58.x or later (libboost-all-dev on Debian/Ubuntu)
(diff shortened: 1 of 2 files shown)


gcp:
A few things:

a) Although the original paper recommended scale layers, it's not clear these are always beneficial. It's not specified in the Zero paper which variant was used, but (IIRC) TensorFlow - the framework in use at at Google - now defaults to BatchNorm layers without scale.

b) NVIDIA-Caffe supports it by simply adding a "scale_bias: true" parameter in the layer, there is no need for a separate layer. This is important when using cuDNN because it can accelerate the combined layers.

c) It's pointless to just change the layer description without also adjusting the DCNN computations in the program.

Unless there's some indication that "Zero" used BatchNorm layers with the Scale/Bias part, or that this provides better performance, there's no point in changing this, and in any case just changing the protocol description isn't enough.

PR #27, opened 2017-11-09
Merge and save summaries 

godmoves:
Following changes are commited:
1. Class `StatisticsCollector` is added to aggregate the accuracy/cost over batches.
(I usually use tf1.3, so I don't know whether tf1.4 has that function built-in)
2. `merge` & `FileWriter` is added to merge and save the summries
3. Summry parts are moved together
4. Factor 1/4 is moved to the defination of mse to make sure summries record the right value.
5. The model saving folder is changed to `leelamodel/leelaz-model` to avoid too many models in the main folder.

(I think the 2000-step check should be test part, if not please tell me.)
Summaries slow down the training, maybe we should save them every 100 steps 
I test the code on a Ubuntu 16.04

### DIFF
--- a/training/tf/tfprocess.py
+++ b/training/tf/tfprocess.py
@@ -64,12 +64,12 @@ class TFProcess:
             tf.nn.softmax_cross_entropy_with_logits(labels=self.y_,
                                                     logits=self.y_conv)
         self.policy_loss = tf.reduce_mean(cross_entropy)
-        tf.summary.scalar('policy_loss', self.policy_loss)
 
         # Loss on value head
+        # Google's paper scales MSE by 1/4 to a [0, 1] range, so do the same to
+        # get comparable values.
         self.mse_loss = \
-            tf.reduce_mean(tf.squared_difference(self.z_, self.z_conv))
-        tf.summary.scalar('mse_loss', self.mse_loss)
+            tf.reduce_mean(tf.squared_difference(self.z_, self.z_conv)) / 4.0
 
         # Regularizer
         regularizer = tf.contrib.layers.l2_regularizer(scale=0.0001)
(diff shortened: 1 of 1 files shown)


godmoves:
Mistakes

training/tf/tfprocess.py
```diff
@@ -64,12 +64,12 @@ def __init__(self, next_batch):
...
         self.policy_loss = tf.reduce_mean(cross_entropy)
-        tf.summary.scalar('policy_loss', self.policy_loss)
 
         # Loss on value head
+        # Google's paper scales MSE by 1/4 to a [0, 1] range, so do the same to
+        # get comparable values.
         self.mse_loss = \
-            tf.reduce_mean(tf.squared_difference(self.z_, self.z_conv))
-        tf.summary.scalar('mse_loss', self.mse_loss)
+            tf.reduce_mean(tf.squared_difference(self.z_, self.z_conv)) / 4.0
```
gcp: Changing this here means the loss function that is being minimized will change too - not what we want, I think. You can probably create a self.display_loss that is scaled so the optimizer can keep the right one.

training/tf/tfprocess.py
```diff
@@ -91,12 +91,21 @@ def __init__(self, next_batch):
...
-        tf.summary.scalar('accuracy', self.accuracy)
 
         self.avg_policy_loss = None
         self.avg_mse_loss = None
         self.time_start = None
 
+        # Summary part
+        self.test_stats = StatisticsCollector()
+        self.test_writer = tf.summary.FileWriter(
+                    os.path.join(os.getcwd(), "leelalogs/test"), self.session.graph)
```
gcp: Can you indent according to PEP8?

training/tf/tfprocess.py
```diff
@@ -91,12 +91,21 @@ def __init__(self, next_batch):
...
+        self.train_writer = tf.summary.FileWriter(
+                    os.path.join(os.getcwd(), "leelalogs/train"), self.session.graph)
+        self.train_summary = tf.summary.merge([
+                    tf.summary.scalar('policy_loss', self.policy_loss), 
+                    tf.summary.scalar('train_mse_loss', self.mse_loss)])
```
gcp: For consistency they should both have train_ or neither.

training/tf/tfprocess.py
```diff
@@ -131,28 +138,28 @@ def process(self, batch_size):
             if self.time_start:
                 elapsed = time_end - self.time_start
                 speed = batch_size * (100.0 / elapsed)
+            # train_summary = tf.summary.merge([
+            #                 tf.summary.scalar('train_policy_loss', self.avg_policy_loss),
+            #                 tf.summary.scalar('train_mse_loss', self.avg_mse_loss)])
+            # self.train_writer.add_summary(self.session.run(train_summary), steps)
```
gcp: No code that is commented out please.

training/tf/tfprocess.py
```diff
@@ -301,3 +308,36 @@ def construct_net(self, planes):
         h_fc3 = tf.nn.tanh(tf.add(tf.matmul(h_fc2, W_fc3), b_fc3))
 
         return h_fc1, h_fc3
+
+
+class StatisticsCollector(object):
```
gcp: This code appears to be lifted out of another program without attribution or conserving the license. If this is from @brilee Mugo, we can use it because Mugo's license is compatible with ours (Apache 2.0 and GPL3). But it should be moved to a separate source file so you properly attribute the copyright and mention the license. Probably also need to update leela-zero's README.md which lists the licenses.

godmoves:
Code updated, save the training every 100 steps to avoid slowing down the training speed.

roy7:
StatisticsCollector may be coming from here?

https://github.com/brilee/MuGo/blob/master/policy.py

godmoves:
@roy7 Yes, but it is not necessary to use it as there is a easy way to collect the summaries, and it is removed

PR #28, opened 2017-11-09
new CMakeFilelist to compile Leela with cmake (works in Linux)

marcocalignano:
Hi this Cmake config works for me what do you think?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,11 @@ if(OPENCL_FOUND)
   TARGET_LINK_LIBRARIES(leelaz ${OPENCL_LIBRARIES})
 endif(OPENCL_FOUND)
 
+if(ZLIB_FOUND)
+  INCLUDE_DIRECTORIES(${ZLIB_INCLUDE_DIRS})
+  TARGET_LINK_LIBRARIES(leelaz ${ZLIB_LIBRARIES})
+endif(ZLIB_FOUND)
+
 SET(IncludePath "CMAKE_CURRENT_SOURCE_DIR}/src")
 SET(SrcPath "${CMAKE_CURRENT_SOURCE_DIR}/src")
 
@@ -51,8 +56,6 @@ FILE(GLOB leelaz_SRC "${SrcPath}/*.cpp")
 ADD_EXECUTABLE(leelaz ${leelaz_SRC})
 
 TARGET_LINK_LIBRARIES(leelaz pthread)
-TARGET_LINK_LIBRARIES(leelaz m)
-TARGET_LINK_LIBRARIES(leelaz z)
 
 SET(CMAKE_CXX_FLAGS "-Wall -Wextra -pipe -O3 -g -ffast-math -flto -march=native -std=c++14 -DNDEBUG")
 SET(CMAKE_EXE_LINKER_FLAGS "-flto -g")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,23 @@
+/*
+    This file is part of Leela Zero.
+    Copyright (C) 2017 Marco Calignano
+    Leela Zero is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    Leela Zero is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+    You should have received a copy of the GNU General Public License
+    along with Leela Zero.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
 CMAKE_MINIMUM_REQUIRED (VERSION 2.6)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
+CMAKE_MINIMUM_REQUIRED (VERSION 2.6)
+
+#add_definitions(-DOEM_TYPE=${OEM})
+
+#Required Packages
+PROJECT(leelaz)
+
+SET (BOOST_MIN_VERSION "1.58.0")
+FIND_PACKAGE(Boost ${BOOST_MIN_VERSION} REQUIRED)
```
gcp: nit: Spacing is inconsistent.

CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+FIND_PACKAGE( BLAS REQUIRED )
+FIND_PACKAGE( ZLIB REQUIRED )
+FIND_PACKAGE( OpenCL REQUIRED )
+
+SET(IncludePath "CMAKE_CURRENT_SOURCE_DIR}/src")
+SET(SrcPath "${CMAKE_CURRENT_SOURCE_DIR}/src")
+
+INCLUDE_DIRECTORIES(${IncludePath})
+INCLUDE_DIRECTORIES(/usr/include/openblas)
```
gcp: Instead of hardcoding this, wouldn't it need to use the output from the FIND_PACKAGE(BLAS) we did earlier?

CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+
+SET(SOURCES
+    "${SrcPath}/FastBoard.cpp"
```
gcp: Can this be automated via some kind of glob? Everything in src/*.cpp should be same to compile.

CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+    "${SrcPath}/TTable.cpp"
+    "${SrcPath}/UCTNode.cpp"
+    "${SrcPath}/UCTSearch.cpp"
+    "${SrcPath}/Utils.cpp"
+    "${SrcPath}/Zobrist.cpp"
+)
+
+ADD_EXECUTABLE(leelaz ${SOURCES})
+
+TARGET_LINK_LIBRARIES(leelaz openblas)
```
gcp: Similar remark as above, shouldn't this use the output from the FIND_PACKAGE thing? What if the user has ATLAS or another BLAS lib?

CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+TARGET_LINK_LIBRARIES(leelaz OpenCL)
+TARGET_LINK_LIBRARIES(leelaz pthread)
+TARGET_LINK_LIBRARIES(leelaz boost_program_options)
+TARGET_LINK_LIBRARIES(leelaz m)
+TARGET_LINK_LIBRARIES(leelaz z)
+
+SET(CMAKE_CXX_FLAGS "-Wall -Wextra -pipe -O3 -g -ffast-math -flto -march=native -std=c++14 -DNDEBUG")
```
gcp: These aren't going to work if the compiler is not GCC or clang.

CMakeLists.txt
```diff
@@ -5,48 +5,38 @@ CMAKE_MINIMUM_REQUIRED (VERSION 2.6)
...
-TARGET_LINK_LIBRARIES(leelaz OpenCL)
+
+FILE(GLOB leelaz_SRC "${SrcPath}/*.cpp")
+
+ADD_EXECUTABLE(leelaz ${leelaz_SRC})
+
 TARGET_LINK_LIBRARIES(leelaz pthread)
-TARGET_LINK_LIBRARIES(leelaz boost_program_options)
 TARGET_LINK_LIBRARIES(leelaz m)
 TARGET_LINK_LIBRARIES(leelaz z)
```
gcp: This can use the output from FIND_PACKAGE ZLIB probably?

CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+CMAKE_MINIMUM_REQUIRED (VERSION 2.6)
+
+#add_definitions(-DOEM_TYPE=${OEM})
```
gcp: Also get rid of this line.

gcp:
CMake Error at CMakeLists.txt:23 (FIND_PACKAGE):
  By not providing "FindOPENCL.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "OPENCL", but
  CMake did not find one.

  Could not find a package configuration file provided by "OPENCL" with any
  of the following names:

    OPENCLConfig.cmake
    opencl-config.cmake

  Add the installation prefix of "OPENCL" to CMAKE_PREFIX_PATH or set
  "OPENCL_DIR" to a directory containing one of the above files.  If "OPENCL"
  provides a separate development package or SDK, be sure it has been
  installed.

gcp:
I guess the problem is that my CMake is 3.5.1, and yours is newer, when those packages might've been added by default.

I pushed a bunch of fixes to "next", but now the problem is that I need to tell it I prefer getting OpenBLAS rather than whatever other BLAS it is picking up. Or at least some environment variable to check so I can disable OpenBLAS specific code if that is not the BLAS it picks out.

gcp:
Okay, I think I got most of this, including finding OpenBLAS specifically.

PR #30, opened 2017-11-10
new CMakeFilelist to compile Leela with cmake

marcocalignano:
So now I took everything I could from the package search. The FindBlas script does not provide a Libriries variable so it need to be insert manually
For the compiler and OS selection I can only test on Linux.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,65 @@
+
+#   This file is part of Leela Zero.
+#   Copyright (C) 2017 Marco Calignano
+#   Leela Zero is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


gcp:
Can you rebase and send a pull request to the "next" branch? I fixed various issues from the first request there. I see you have some further improvements here, we'll have to combine them one way or the other.

marcocalignano:
So I cannot rebase I probably have to fork again. I saw you add cmake modules for OpenCL I solved that requiring cmake 3.1 in the cmake fole so all the modules can be deleted from the repository

marcocalignano:
I did it check the the new pull request Adding find thread removing ...

PR #49, opened 2017-11-14
Update README.md

olbrichj:
compiled leelaz needs to be in the same directory as autogtp

### DIFF
--- a/autogtp/README.md
+++ b/autogtp/README.md
@@ -17,3 +17,5 @@ Run:
 
     qmake
     make
+    copy the compiled leelaz to the same directory as autogtp
+    ./autogtp


OmnipotentEntity:
This is wrong though.  leelaz just needs to be in the current path.  You can (and I have) cd into the src directory and run autogtp via relative path.  Or add leelaz to the $PATH variable.

WhiteHalmos:
I just put it on my path and remove the `./` in autogtp before compiling.

gcp:
The instructions in that section are about compiling, not running, so this should go in a separate section. I'll make a change similar as you propose.

PR #51, opened 2017-11-14
autogtp: some error messages

marcocalignano:


killerducky:
Thanks marco, I was able to figure out what I was doing wrong with your error messages!

marcocalignano:
I am glad I could help!

PR #58, opened 2017-11-16
Update autogtp README.md with apt install command

sethtroisi:
Updated Requirements:
QRegularExpression requires Qt5,  https://github.com/vaibhavpandeyvpz/apkstudio/issues/10

### DIFF
--- a/autogtp/README.md
+++ b/autogtp/README.md
@@ -12,7 +12,7 @@ the SGF and training data at the end of the game.
 * gzip and gunzip
 
 ## Example of compiling - Ubuntu
-    sudo apt install qt5-default
+    sudo apt install qt5-default qt5-qmake
     qmake
     make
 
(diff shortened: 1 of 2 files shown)


sethtroisi:
gcp, I modified to use the "qmake -qt5" per https://github.com/gcp/leela-zero/issues/66, alternatively I can try to do #ifdef on which qt version is present.

gcp:
I did a check and QStringLiteral is also Qt5 only, so I guess Qt4 isn't really supported any more.

pmontrasio:
`sudo apt install qt5-default` didn't make it into the README of the autogtp directory on the main branch yet. This could stop some people from compiling and running autogtp.

gcp:
Is qt5-default needed if you give the correct qmake command?

sethtroisi:
qt5-default causes qtbase5-dev to be installed which is what's actually satisfying the but one of them is still needed

PR #69, opened 2017-11-16
Add command line option to keep self-play .sgf files

### LINKED ISSUE #68: Running autogtp doesn't save .sgf files
I ran autogtp (v0.4) for a while on my Windows machine. The console prints the individual moves and after a game ends, mentions the dumping and sending the file to the server. However, this file is nowhere to be found - I assume the script automatically removes it or it's only in-memory.

It would be great to see the actual .sgf files accumulating in a folder, to be able to take a look at what kind of games are produced during the training.

tux3:
Can be enabled with -k, or --keep-sgf.
Also adds simple Qt-provided help and version options.

Fix #68

### DIFF
--- a/autogtp/main.cpp
+++ b/autogtp/main.cpp
@@ -20,6 +20,7 @@
 #include <QtCore/QTimer>
 #include <QtCore/QTextStream>
 #include <QtCore/QStringList>
+#include <QCommandLineParser>
 #include <QProcess>
 #include <QFile>
 #include <QDir>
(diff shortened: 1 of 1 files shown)


killerducky:
Maybe it should keep the data file also, not just the sgf?

gcp:
The server code should be resilient against this (we detected duplicate/resubmitted games). But yes, it seems like a good idea to dump into a separate directory. @tux3 can you change this?

tux3:
Sure, I'll make the output path an option and copy uncompressed files then. Unless you'd rather just have it dump into ./sgf/ or something.

tux3:
I added the change and rebased on next, tell me if it looks good!

PR #74, opened 2017-11-17
print timing information

sethtroisi:
Example output: https://imgur.com/a/g6jBP

### DIFF
--- a/autogtp/main.cpp
+++ b/autogtp/main.cpp
@@ -176,6 +176,20 @@ bool run_one_game(QTextStream& cerr, const QString& weightsname) {
     return true;
 }
 
+template<typename Clock>
+void print_timing_info(QTextStream& cerr, int games_played,
+                       std::chrono::time_point<Clock> start,
+                       std::chrono::time_point<Clock> game_start,
+                       std::chrono::time_point<Clock> game_end) {
+    auto game_time_s = std::chrono::duration_cast<std::chrono::seconds>(game_end - game_start);
+    auto total_time_s = std::chrono::duration_cast<std::chrono::seconds>(game_end - start);
+    auto total_time_min = std::chrono::duration_cast<std::chrono::minutes>(total_time_s);
+    cerr << games_played << " games_played in "
+         << total_time_min.count() << " minutes = "
+         << total_time_s.count() / games_played << " seconds/game, last game "
+         << game_time_s.count() << " seconds\n\n";
+}
+
 int main(int argc, char *argv[])
 {
     QCoreApplication app(argc, argv);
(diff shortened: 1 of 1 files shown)


autogtp/main.cpp
```diff
@@ -50,26 +51,27 @@ bool fetch_best_network_hash(QTextStream& cerr, QString& nethash) {
...
     if (server_expected > AUTOGTP_VERSION) {
-        cerr << endl;
         cerr << "Server requires client version " << server_expected
              << " but we are version " << AUTOGTP_VERSION << endl;
         cerr << "Check https://github.com/gcp/leela-zero for updates." << endl;
         exit(EXIT_FAILURE);
-    } else {
-        cerr << " (OK)" << endl;
+    }
+    if (games_played == 0) {
```
sethtroisi: I think this is fairly useless output (except the first time) but I can change it to (games_played % 10 == 0) if you want it to appear more frequently but not everytime

SebastienFaurite:
It would be cool if we had some benchmarking like the number of move per minute

sethtroisi:
I'm working on that in a follow up commit :) but I like to split commits into the smallest pieces that can be pulled at a time

autogtp/main.cpp
gcp: These can change at any time, so I don't like this. If you think it is too spammy, it should probably remember the original value and report if there's been a change. The current patch looks like the worst of both worlds.

autogtp/main.cpp
```diff
@@ -50,26 +51,27 @@ bool fetch_best_network_hash(QTextStream& cerr, QString& nethash) {
...
     nethash = outhash;
     return true;
 }
 
-bool fetch_best_network(QTextStream& cerr, QString& netname) {
+bool fetch_best_network(QTextStream& cerr, QString& netname, long games_played) {
     if (QFileInfo::exists(netname)) {
-        cerr << "Already downloaded network." << endl;
+        if (games_played == 0) {
+            cerr << "Already downloaded network." << endl;
```
gcp: Same remark, there's no reason why this would only apply to the first game.

autogtp/main.cpp
```diff
@@ -196,20 +198,28 @@ int main(int argc, char *argv[])
 #else
     QTextStream cerr(stderr, QIODevice::WriteOnly);
 #endif
-
     cerr << "autogtp v" << AUTOGTP_VERSION << endl;
 
+    auto start = std::chrono::high_resolution_clock::now();
     auto success = true;
-    auto games_played = 0;
+    long games_played = 0;
```
gcp: Please keep the coding style: auto games_played = long{0}; (or use a literal suffix)

autogtp/main.cpp
```diff
@@ -196,20 +198,28 @@ int main(int argc, char *argv[])
...
         success &= run_one_game(cerr, netname);
         success &= upload_data(cerr, netname);
         games_played++;
-        cerr << games_played << " games played." << endl;
+
+        auto game_end = std::chrono::high_resolution_clock::now();
+        auto game_time_s = std::chrono::duration_cast<std::chrono::seconds>(game_end - game_start);
+        auto total_time_s = std::chrono::duration_cast<std::chrono::seconds>(game_end - start);
+        cerr << games_played << " games_played in "
+             << (total_time_s.count() / 60) << " minutes = "
```
gcp: You can probably avoid the magic number here with std::chrono::minutes (but it's ok like this too).

autogtp/main.cpp
sethtroisi: If the best network updates then this will be skipped (because the file won't exist)and the logging at line 108 will print showing an update.

autogtp/main.cpp
sethtroisi: You're right, I'll make it remember. 
I think it's a little spammy because the best network hash can also be found in the curl line which is printed with the same frequency

autogtp/main.cpp
sethtroisi: I use total minutes here, and seconds below so I need to convert between them (or use a bulky cast which I'm not a fan of)

sethtroisi: done.

sethtroisi:
I refactored the function out to try and keep the loop in main cleaner.

Issue #64, opened 2017-11-16
autogtp could pile up more work on my GPU


FlorinAndrei:
I've a Titan X Pascal. When running autogtp.exe the memory usage on the GPU is a few percent points, GPU load is less than 10%, and power consumption is under 25% - all these measured with the GPU-Z app.

I think the software should pile up more work if the GPU can take it.

gcp:
The current network is tiny because there is very little learning data yet, so the program ends up being completely CPU-bound on a fast GPU.

This will "fix" itself after a few epochs.

FlorinAndrei:
Hmm. CPU usage is not too high either - 30 to 40% on an AMD FX-8350 eight-core.

sbbdms:
You can create multiple instances of AutoGTP according to the CPU & GPU usage of your machine.

gcp:
You can launch a second instance, it defaults to only 2 threads. (Some of the CPU usage you already see is from the GPU driver)

Sivous:
How long will it take to update the neural network?

gcp:
I'm not sure what you mean. It gets updated when it beats the old one by a significant %. It's not possible to know in advance how long this takes.

sbbdms:
Maybe he means the interval of http://zero.sjeng.org/networks/.
The author mentioned that he will probably dump every 25k games or so. See #1

gcp:
Latest release can play multiple games at once with the -g option.

Issue #57, opened 2017-11-16
Why do games slow down as they increase in number of moves?


CamWagner:
At the start of a self-play game, the moves are pretty quick. Why does each turn take longer as the game progresses? (if anything, I would expect them to speed up...)

(Also wow this game is 650+ moves long and still going)

gcp:
The network right now is the smallest possible size, so all kinds of things that are normally completely negligible influence the runtime now. There is nothing that would speed up the program near the endgame (it does not do Monte Carlo game playouts) and some things that slow it down: state tracking to detect superko and a narrower search tree which means that there are more nodes to investigate to find the best move to search.

gcp:
tux3 fixed the largest CPU bottleneck (state tracking for superko) so I think this is no longer a perceptible issue.

PR #89, opened 2017-11-18
Improve leelaz search performance 3-4x

tux3:
This fixes the search getting slower with each move played due to large unecessary copies of the game state dominating the CPU time and using all the memory bandwidth.
Now it's as fast after 200 or 600 moves as in the opening.

This should help with #57 and #64 since (at least for me) it's now very fast and uses 100% GPU with only one or two instances.

### DIFF
--- a/src/GameState.cpp
+++ b/src/GameState.cpp
@@ -37,7 +37,7 @@ void GameState::init_game(int size, float komi) {
     KoState::init_game(size, komi);
 
     game_history.clear();
-    game_history.push_back(*this);
+    game_history.emplace_back(std::make_shared<KoState>(*this));
 
     m_timecontrol.set_boardsize(board.get_boardsize());
     m_timecontrol.reset_clocks();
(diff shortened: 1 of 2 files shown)


tux3:
For comparison, here is the top functions (by self time) before this patch:

snap-43

The 3 highlighted are just doing copies of the game state, which takes more and more time with each move played.
And after:

snap-44

The top functions (still by self time) are now from OpenCL and OpenBLAS, which is much better!
This is just a fix for the biggest issue, so there are probably more opportunities for optimizations further down the list, but once the network gets larger the GPU time will probably dominate anyways.

WhiteHalmos:
This is awesome! Nice optimization!

gcp:
Thanks, this looks too clean a workaround not to take.

The underlying problem (i.e. where this n^2 behavior comes from) could be fixed by truncating the GameState at the last 16 positions before copying it into the search loops in UCTSearch.

PR #90, opened 2017-11-18
Add "selfplay" option

OmnipotentEntity:
Add "selfplay" option, whose sole purpose is to output the correct data in the SGF file.

There's probably a better way to handle this.  But I didn't want to screw with the Fastboard enum just in case something else was relying on this.

### DIFF
--- a/autogtp/Game.cpp
+++ b/autogtp/Game.cpp
@@ -35,7 +35,7 @@ Game::Game(const QString& weights, QTextStream& out) :
 #endif
     cmdLine.append(" -g -q -n -d -m 30 -r 0 -w ");
     cmdLine.append(weights);
-    cmdLine.append(" -p 1000 --noponder");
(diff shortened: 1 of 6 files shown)


gcp:
Seems pretty involved for something so simple. What about "printsgf_self" instead of a command line option?

Also do we really care so much? The main thing is that one can identify which network played a game. If the "Human" thing bothers someone it could be fixed up with a script afterwards.

Not too keen on taking this.

OmnipotentEntity:
That's a much better option and one that I hadn't considered.  But it's not a huge deal, especially if you don't care overly much.

Let me know if you'd like me to submit a pull request for that option, or if you'd rather I just dropped it.

gcp:
This can be fixed at a few other levels (the ones that have knowledge about who the opponent actually is), so I think an engine-level fix is not ideal. Given that he impact is purely cosmetic, I'm going to close this.

Issue #70, opened 2017-11-16
A non GPU (or only CPU) version?


emdio:
So, I'm quite sure if this hasn't been done it's either impossible or worthless, but just in case, wouldn't be possible to develop an only CPU version for computers that don't have a compatible graphic card?

emdio:
Well, it seems here there's a turnaround to make leela-zero work without GPU

https://www.reddit.com/r/baduk/comments/7dg219/psa_you_can_help_training_leelazero_without_a/

gcp:
Running on the CPU is very slow for a large network, running it through OpenCL on the CPU is far, far worse.

https://github.com/gcp/leela-zero/blob/9064f6facc452262b3eba12a55f9ff2ca3486dfd/src/Network.cpp#L450

It's not very hard to fill in this part, all the supporting code for convolutions, batchnorm etc is there, and the residual part is a simple add. But I didn't do it because it's a waste.

fishcu:
I'm the author of that reddit post. It was more of a test of feasibility rather than practicability.
Still, the difference in performance between that old i5 (2nd gen) and my new machine with an i7 (7th gen) and a GTX 1060 is only about 1 order of magnitude! If 1 order of magnitude more people can run the distributed training, it is still worth enabling them to do so -- why do you consider it a waste?

gcp:
This is because the initial network is very small. Until the learning data accumulates a little, the program is effectively CPU limited and the GPU part (or not) doesn't really matter. But this won't last very long...

fishcu:
You are right, that's something I was aware of but forgot to mention in my little experiment. I have no idea how that will change though.

Do you have any idea how much you'd expect a CPU-only trainer to slow down in the near future? Another order of magnitude?

gcp:
From experience with regular Leela it is about a factor 10x-30x. That is with OpenBLAS convolutions. With OpenCL-on-CPU it would be far worse.

fishcu:
I'll keep my old machine with OpenCL-on-CPU running. Then we will find out how slow it really gets!

nauzykaa:
[...]

PR #104, opened 2017-11-19
Simplify Im2col for the filter_size = 1 case

sethtroisi:
im2Col is overly complicated*, I'm a proponent of You Ain't Gonna Need It.
let's replaced this potentially buggy code with an assert(filter_size = 1) + copy till a larger filter size is needed at that point this can be reverted or less buggy code written.

*uses many different for loop styles, has a host of extra variables, performs an unsafe signed to unsigned conversion and ultimately doesn't do anything.

### DIFF
--- a/src/Im2Col.h
+++ b/src/Im2Col.h
@@ -28,11 +28,12 @@ template <unsigned int filter_size>
 void im2col(const int channels,
             const std::vector<float>& input,
             std::vector<float>& output) {
-    assert
+    // Not implemented
+    exit(EXIT_FAILURE);
 }
 
 template <>
-void im2col_test(const int channels,
+void im2col<1>(const int channels,
             const std::vector<float>& input,
             std::vector<float>& output) {
     constexpr unsigned int boardsize = 19;
--- a/src/Im2Col.h
+++ b/src/Im2Col.h
@@ -19,46 +19,23 @@
 #ifndef IM2COL_H_INCLUDED
 #define IM2COL_H_INCLUDED
 
-#include "config.h"
+#include <cassert>
 #include <vector>
-#include <algorithm>
-#include "Utils.h"
 
 template <unsigned long filter_size>
 void im2col(const int channels,
             const std::vector<float>& input,
             std::vector<float>& output) {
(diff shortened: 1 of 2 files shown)


gcp:
Don't let the people in issue #70 hear you.

gcp:
I wonder what makes you say the cast is unsafe. I specifically checked the guarantees for this conversion.

gcp:
Given that im2col is templated on filter size, can't you just add a specialization for == 1 that does the copy?

sethtroisi:
I wrote a version that cleans up what I don't like about the current code. I'll add that to this pull request and then template this for one.

sethtroisi:
Template specialization on functions is complicated: https://stackoverflow.com/a/10695059/459714

gcp:
Your link talks about partial specialization. im2col has only 1 template parameter.

sethtroisi:
Fixed, and implementation looks very clean now :)

Sorry that all my pull requests require extra guidance

gcp:
I pulled the specialization because this looks like the most useful for me.

As for the rest of this pull request: although I don't think providing a full CPU implementation is a priority right now, I also see no need to remove working code and make it harder for anyone who wants to attempt to make one. I took a look at your code review and I disagree on most remarks. In a CPU implementation this code is performance critical, so things like downward counting loops, and avoiding the dual comparison for the row (which is where the cast is used) have measurable performance impact. The code was written this way for a reason!

sethtroisi:
I think the specialization gives an 8% speedup (as measured by 3x50000 netbench tests)

gcp:
Very nice!

sethtroisi:
Can you explain what im2col should do when filter_size > 1?

I've tried my hardest to understand the code, I've tried to do my due diligence but I don't get it,
I really think the current code would generate the wrong result (and I'm super passionate for some reason about fixing it)

gcp:
https://petewarden.com/2015/04/20/why-gemm-is-at-the-heart-of-deep-learning/
https://arxiv.org/pdf/1410.0759.pdf

PR #113, opened 2017-11-20
Add milliseconds per move timing info

sethtroisi:
https://i.imgur.com/X0aQV68.png

### DIFF
--- a/autogtp/Production.cpp
+++ b/autogtp/Production.cpp
@@ -55,6 +55,7 @@ void ProductionWorker::run() {
                 return;
             }
             game.readMove();
+            (*m_movesMade)++;
         } while (game.nextMove() && m_state == RUNNING);
         switch(m_state) {
         case RUNNING:
@@ -88,12 +89,15 @@ void ProductionWorker::run() {
     } while (m_state != FINISHING);
 }
 
-void ProductionWorker::init(const QString& gpuIndex, const QString& net) {
+void ProductionWorker::init(const QString& gpuIndex,
+                            const QString& net,
+                            QAtomicInt* movesMade) {
     m_option = " -g -q -d -n -m 30 -w ";
     if (!gpuIndex.isEmpty()) {
         m_option.prepend(" --gpu=" + gpuIndex + " ");
     }
     m_network = net;
+    m_movesMade = movesMade;
     m_state = RUNNING;
 }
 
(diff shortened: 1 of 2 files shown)


kingvash:
Can this be pulled? It's useful in testing performance (where number of games is way to coarse)

autogtp/Game.h
```diff
@@ -36,6 +36,7 @@ class Game : QProcess {
     bool waitForMove() { return waitReady(); }
     bool readMove();
     bool nextMove();
+    int  getMoveNum();
```
gcp: nit: there's an extra space here

autogtp/Production.cpp
```diff
@@ -37,16 +37,18 @@ void ProductionWorker::run() {
              game.readMove();
          } while (game.nextMove());
          QTextStream(stdout) << "Game has ended." << endl;
+         int movesMade = 1;
```
gcp: The way this value is set as the default makes me suspicious the error handling around here is wonky. Notably, should we call emit resultReady when we can't do a getScore()?  Ignoring that problem (which exists in the current code), can't we already just call game.getMoveNum() here?

sethtroisi:
Changed calculation slightly, now calculates time / all_moves (from finished and partial games)

this make the output much more stable (instead of bouncing around passed on how many outstanding moves had been calculated but where part of games that hadn't finished), and more useful as a real performance evaluation (as compared to the single instance netbench)

1 game(s) played in 1 minutes = 95 seconds/game, 372 ms/move, last game took 94 seconds.
12 game(s) played in 14 minutes = 73 seconds/game, 381 ms/move, last game took 163 seconds.
22 game(s) played in 28 minutes = 78 seconds/game, 396 ms/move, last game took 135 seconds.
36 game(s) played in 47 minutes = 78 seconds/game, 390 ms/move, last game took 114 seconds.

autogtp/Production.cpp
```diff
@@ -178,16 +182,22 @@ void Production::getResult(const QString& file, float duration) {
...
+    if (m_movesMade == 0 || m_gamesPlayed == 0) {
+        return;
+    }
     auto game_end = std::chrono::high_resolution_clock::now();
     auto total_time_s =
         std::chrono::duration_cast<std::chrono::seconds>(game_end - m_start);
     auto total_time_min =
         std::chrono::duration_cast<std::chrono::minutes>(total_time_s);
+    auto total_time_millis =
+        std::chrono::duration_cast<std::chrono::milliseconds>(total_time_s);
```
sethtroisi: I should really use (game_end - m_start) to calculate total_time_millis then cast that to total_time_s and total_time_min but the net effect is in the 4 or 5th sig fig.

PR #114, opened 2017-11-20

marcocalignano:


marcocalignano:
@gcp can you revert the last commit and take mine. I check the version inside the production Object. You do not need to check inside every Game object you instantiate but only at an higher level when the network is checked

gcp:
The check is for the Leela Zero version, not the autogtp version vs the server requested version. I don't think that's in your code yet.

I think it's best that I hold off on merging pull requests for a few days until we can get this merged? Else you'll be rebasing constantly as this is a huge patch.

autogtp/sprt.cpp
```diff
@@ -0,0 +1,210 @@
...
+#include <cmath>
+#include <QtGlobal>
+
+class BayesElo;
+class SprtProbability;
+
+class BayesElo
+{
+	public:
+		BayesElo(double bayesElo, double drawElo);
```
sethtroisi: Gives that Go rarely has draws this can be dramatically simplified.

I've uploaded an old python implementation I used here
https://gist.github.com/sethtroisi/786b99e37f64cd034ff0baec649ca2b5

If you don't want to do this and are willing I'll jump in tomorrow and write a c++ version

gcp: He doesn't need a C++ version, he already has one.

marcocalignano:
ok so I rebase this one last time to integrate the version check and then I make a new pull request
on next

autogtp/validation.cpp
```diff
@@ -0,0 +1,176 @@
...
+    m_gamesThreads(gpus*games),
+    m_games(games),
+    m_gpus(gpus),
+    m_gpusList(gpuslist),
+    m_gamesPlayed(0),
+    m_firstNet(firstNet),
+    m_secondNet(secondNet)
+{
+    m_statistic.initialize(25.0,35.0,0.05,0.05);
+    m_statistic.addGameResult(Sprt::Win);
```
sethtroisi: Why do you add this initial win / loss / draw?

autogtp/validation.cpp
```diff
@@ -0,0 +1,176 @@
...
+        return;
+    }
+    m_syncMutex.lock();
+    m_gamesPlayed++;
+    m_statistic.addGameResult(result);
+    Sprt::Status status = m_statistic.status();
+    if(status.result != Sprt::Continue) {
+        quitThreads();
+        QTextStream(stdout) << "The first net is ";
+        QTextStream(stdout) <<  ((status.result ==  Sprt::AcceptH0) ? "wrost " : "better ");
```
sethtroisi: worse

PR #118, opened 2017-11-20

marcocalignano:


.gitignore
```diff
@@ -3,3 +3,5 @@ leelaz
 *.d
 .vscode/settings.json
 training/tf/venv
+autogtp/*.sgf
```
WhiteHalmos: Mind moving these down to autogtp/.gitignore?

PR #123, opened 2017-11-21

marcocalignano:
Please merge before i get new conflicts!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ leelaz
 *.d
 .vscode/settings.json
 training/tf/venv
+autogtp/*.sgf
+autogtp/*.gz
--- a/autogtp/Game.cpp
+++ b/autogtp/Game.cpp
@@ -51,6 +51,9 @@ void Game::error(int errnum) {
         case Game::WRONG_GTP:
             output << "Error in GTP response." << endl;
             break;
+        case Game::LAUNCH_FAILURE:
+            output << "Could not talk to engine after launching." << endl;
+            break;
         default:
             output << "Unexpected error." << endl;
             break;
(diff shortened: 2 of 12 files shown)


marcocalignano:
Opps I think there is an error in the licence header of Game.cpp

gcp:
It doesn't compile either :) I'll fix and clean it up. Warning: force-pushes to next branch incoming.

marcocalignano:
here compiles fine what is your problem?

gcp:
std::tuple has an explicit constructor, so your changes to the minimum version aren't legal, see e.g.: https://stackoverflow.com/questions/32084706/why-initialization-doesnt-work-for-tuple

I guess some compilers might accept it nevertheless.

marcocalignano:
ok so my compiler accept and the right sintax is without the '=' ?

gcp:
Yes. I already fixed it, will force-push the next branch a few times with some style fixups too.

marcocalignano:
ok when you are finished I do the same on my Fork so we are on the same page for the next developmen :)

gcp:
./autogtp  -g 4 -k games

Game has ended.
Score: B+353.5
Winner: black

Game has ended.
Score: B+353.5
Winner: black

Game has ended.
Score: B+353.5
Winner: black

Game has ended.
Score: B+353.5
Winner: black

gcp:
The games also look fairly short on average.

gcp:
I will stop force-pushing on next now, so it's safe to pull. But I think there is some debugging to do.

marcocalignano:
All the sgf files contain only the first move, even if genmove works because we can output the move but somehow it is not registered in the game log.

marcocalignano:
I might have found the problem testing now ...

marcocalignano:
Game data d7b53e2eb316153e5ef48e4ba79161e3267c1aaf82d3ede20348bd7f0319af5b stored in database

Game data 7f000e0562e794b2a6f1eb0a086613f9595d5ebfcf7d9162910bb8736d74a58c stored in database

Can you check these?

gcp:
They are both short and have the exact same result.

(PS. It's better to just disable the curl submission of these broken results, and comment out the code that deletes the files, then you can check locally)

In leela's commandline, you can use loadsgf file.sgf and then do commands like "final_score" etc.

marcocalignano:
In my test seems everything OK I try to upload a couple of file so you can check them

marcocalignano:
Game data 67d76c833387c739da910f722b30d2380fd400d540464d786f1f135aace6bed1 stored in database

Game data 3e8cbddb8febea86880e5e282eaf58a2046c3f6110af25aea1922d1d91e15c89 stored in database

Check these please.

marcocalignano:
I checked them I would like only a second opinion!

PR #129, opened 2017-11-21
autogtp: bugfix remove double color change before move

marcocalignano:
This line got duplicated due to one too many rebase ;)

### DIFF
--- a/autogtp/Game.cpp
+++ b/autogtp/Game.cpp
@@ -212,7 +212,6 @@ bool Game::readMove() {
     } else {
         m_passes = 0;
     }
-    m_blackToMove = !m_blackToMove;
     return true;
 }
 


marcocalignano:
This should fix also the validation mode of autogtp so you can test SPRT again

gcp:
Trying this now.

Remarks:
1) It would be good to see the actual score.
2) No need to dump training data during validation
3) I am playing 2 games at once and noticed:

Score: W+16.5
Winner: white
Writing 0cee33e3a14f435ebdb377ad84f41731.sgf
Writing 1004ff42c47942e8b6dbc4c78e76c09d.sgf
Dumping 0cee33e3a14f435ebdb377ad84f41731.txt
Stopping engine.
1 games played.
Status: 0 Log-Likelyhood Ratio -0.0526068 Lower Bound -2.94444 Upper Bound2.94444

Score: W+61.5
Winner: white
Writing 98311c73f5c8489c8e2f6ea135d6fff2.sgf
Writing 80c46ef05d9f448e8b64a399d5960464.sgf
Dumping 98311c73f5c8489c8e2f6ea135d6fff2.txt
2 (L3) Stopping engine.
2 games played.
Status: 0 Log-Likelyhood Ratio -0.0154951 Lower Bound -2.94444 Upper Bound2.94444

White won both games, but Log-Likelyhood Ratio (LLR) dropped in both instances. Was the engine playing white in both games? (I guess it alternates per thread?)

marcocalignano:
> It would be good to see the actual score.

what do you mean actual score? of the game or a percentage of winning games?

gcp:
This :)

https://github.com/gcp/leela-zero/commit/956e478421158fbb09e33906de85703cf03fa2d7#diff-3746371f9b46ece1340a2a556a47a633R159

marcocalignano:
Ok, I did some test with more debug and I cannot confirm then problem you are haveing with the alternate color. Can you add this line after line 60 in Validation.cpp and show me the output again?

QTextStream(stdout) << "result " << result << " expected " <<  m_expected << " emit " << ((result == m_expected) ? Sprt::Win : Sprt::Loss) << endl;

gcp:
Colors are probably fine, I misread the output (and deleted that part of the question here).

170 (Q1) 3 wins, 33 losses
36 games played.
Status: 0 LLR -1.03203 Lower Bound -2.94444 Upper Bound 2.94444

I was a bit surprised SPRT was still so far from stopping the test despite this result. I need to play a bit with it.

gcp:
I'm playing with this now (it behaves correctly with drawElo=0):
https://github.com/lucasart/sprt

SPRT termination is much stricter than for example an error bound over Elo margins:
./sprt 0 100 10 50000 0 25 35 0.9
BayesElo Elo    %Pass   Avg     Q90%
0.00    -0.00   0.0000  1196    1728
10.00   10.00   0.0000  1794    2770
20.00   20.00   0.0032  3585    6276
30.00   30.00   0.5023  10629   21838
40.00   40.00   0.9979  3573    6238
50.00   50.00   1.0000  1805    2780
60.00   60.00   1.0000  1208    1743
70.00   70.00   1.0000  910     1257
80.00   80.00   1.0000  732     981
90.00   90.00   1.0000  613     804
100.00  100.00  1.0000  529     679

It makes more sense to set the margin [0, 35] instead of [25, 35]:
./sprt 0 100 10 50000 0 0 35 0.9
BayesElo Elo    %Pass   Avg     Q90%
0.00    -0.00   0.0485  537     1039
10.00   10.00   0.2200  781     1576
20.00   20.00   0.6092  872     1794
30.00   30.00   0.8961  652     1299
40.00   40.00   0.9791  441     827
50.00   50.00   0.9964  317     558
60.00   60.00   0.9993  245     408
70.00   70.00   0.9998  199     322
80.00   80.00   1.0000  168     263
90.00   90.00   1.0000  146     223
100.00  100.00  1.0000  129     193

The amount of games required to make that determination is more in line with AGZ paper. You guarantee strongly the new network is not worse, and you are very unlikely to "miss" an improvement if it is at least 35 Elo better.

gcp:
A small value of drawElo doesn't really change the results much so the existing issue of having to set a prior of 1 draw isn't a big problem.

marcocalignano:
so you are saying with elo [0,35] will be finished in ~500 games with a probability of 95%?

gcp:
No, termination time always depends on the actual difference between the two networks:

```
/sprt 0 50 5 50000 0 0 35 0.95
BayesElo Elo    %Pass   Avg     Q95%
0.00    -0.00   0.0485  537     1289
5.00    5.00    0.1075  654     1647
10.00   10.00   0.2200  781     2007
15.00   15.00   0.3989  871     2273
20.00   20.00   0.6092  872     2295
25.00   25.00   0.7852  779     2017
30.00   30.00   0.8961  652     1638
35.00   35.00   0.9527  534     1288
40.00   40.00   0.9791  441     1023
45.00   45.00   0.9907  370     823
50.00   50.00   0.9964  317     680
```

Worst case you need 2300 games, if the network is 20-25 Elo better. (Q95% at +20 Elo) But the test terminates faster if it is worse (0 Elo) or clearly better. If the test is still going on after 400+ games it makes sense to move on to the next network. It just means the improvement is on the threshold.

gcp:
Basically if the network is between -32 Elo and +67 Elo, then there's a 5% chance you need over 400 games.

On averge you expect to need 400 games if it is between -8 Elo and +43 Elo.

gcp:
Advantage of SPRT is most easily understood like this:

```
./sprt -300 300 50 50000 0 0 35 0.95
BayesElo Elo    %Pass   Avg     Q95%
-300.00 -300.00 0.0000  40      51
-250.00 -250.00 0.0000  45      59
-200.00 -200.00 0.0000  52      72
-150.00 -150.00 0.0000  65      95
-100.00 -100.00 0.0000  90      143
-50.00  -50.00  0.0000  154     276
0.00    -0.00   0.0485  537     1289
50.00   50.00   0.9964  317     680
100.00  100.00  1.0000  129     219
150.00  150.00  1.0000  83      126
200.00  200.00  1.0000  63      90
250.00  250.00  1.0000  52      71
300.00  300.00  1.0000  46      61
```

If there's a big swing you need far less than 400 games and can stop early in a sound manner that doesn't risk accidentally taking a worse network.

marcocalignano:
OK did you try [0,35] yet?

gcp:
Yes, most of the output above is with [0, 35].

[25, 35] is not a good idea because you can end up spending playing thousands of games proving the network was "only" 24 Elo better instead of 25.

marcocalignano:
I meant have you tried [0,35] in autogtp yet

gcp:
https://github.com/gcp/leela-zero/commit/4fe03cf31b6071fde7a4d26c93373508b3f110ab

I will for next test. Don't see why it wouldn't work. Implementation in cutechess was well tested.

marcocalignano:
did you test production with multiple GPUs?

gcp:
No, not yet. I only have one such machine.

marcocalignano:
we could ask qysnn  he has 4 GPUs

marcocalignano:
And could you test the new  SPRT with [0,35]

Issue #93, opened 2017-11-19
curl --retry


roy7:
Should curl --retry be used by autogtp in case server has an outage, it might try reconnecting for a while before autogtp aborts? There are various options that can be passed to curl, perhaps increase the # of tries over 5 since it'll do exponential backup by default.

And/or, should autogtp resume after a curl failure?

gcp:
Sounds like a good idea yes.

roy7:
Going to close this suggestion per the discussion at https://github.com/gcp/leela-zero/pull/130 and the more advanced things happening there.

PR #130, opened 2017-11-21
Retry in case of network error instead of quitting

Ttl:
Refactored pull request #95. Now checks the curl exit code too. The goal is to keep the clients running waiting for the server to come back if there is a server outage.

I removed the offline game playing from the commit to keep it smaller.

@marcocalignano can you take a look too?

fetchBestNetwork() is called even if fetchBestNetworkHash() reported that the hash has not changed to make sure that the file exists if there were previously any issues with the download.

### DIFF
--- a/autogtp/Production.cpp
+++ b/autogtp/Production.cpp
@@ -45,16 +45,18 @@ void ProductionWorker::run() {
          auto end = std::chrono::high_resolution_clock::now();
          auto gameDuration =
             std::chrono::duration_cast<std::chrono::seconds>(end - start).count();
-         emit resultReady(game.getFile(), gameDuration);
+         emit resultReady(game.getFile(), gameDuration, m_index);
     } while (true);
 }
 
-void ProductionWorker::init(const QString& gpuIndex, const QString& net) {
+void ProductionWorker::init(const QString& gpuIndex, const QString& net,
+                            const int index) {
     m_option = " -g -q -n -d -m 30 -r 0 -w ";
     if(!gpuIndex.isEmpty()) {
         m_option.prepend(" -gpu=" + gpuIndex + " ");
     }
     m_network = net;
+    m_index = index;
 }
 
 Production::Production(const int gpus,
@@ -82,7 +84,7 @@ void Production::startGames() {
     QString myGpu;
     for(int gpu = 0; gpu < m_gpus; ++gpu) {
         for(int game = 0; game < m_games; ++game) {
-            auto thread_index = gpu * m_games + game;
+            int thread_index = gpu * m_games + game;
             connect(&m_gamesThreads[thread_index],
                     &ProductionWorker::resultReady,
                     this,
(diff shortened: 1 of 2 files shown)


marcocalignano:
- I would put the struct NetworkException inside the Class Production
- I would call fetchBestNetwork()  even if fetchBestNetworkHash() return true only in the begininnig in Stargames() , later If fetchBestNetworkHash() return true you do not need to make the clients waiting if the server is down you have your network, plus you can wait only with the thread that currently finished the game the othere already have their network and cannot change them while running. So i would write a Production::updateNetwork(bool force) function where force is true in the beginning and false later.

Ttl:
fetchBestNetwork() checks at the beginning if the file exists and if it does it will return immediately, so it barely costs anything to call it if the network exists. Do you still feel like the if should be added? I thought it was better to actually check that the file exists instead of trusting the fetchBestNetworkHash() return value.

I'll do the changes and test the sender() change later today after work (~10 hours).

marcocalignano:
Then move the test of "if the file exsist" in the fetchBestNetworkHash() function and use fetchBestNetwork() only for the download.

Ttl:
I moved the NetworkExpection under Production.

Regarding the force argument, do you want the program to download the network again at the startup even if it already exists? The current implementation doesn't download it again if the file already exists and this pull request doesn't change that.

roy7:
I've seen some people complain that a failed weights download will prevent autogtp from working until they manually remote the corrupt weights. If it's easy enough to calculate the same sha256 hash we're using to name the networks with, perhaps at first startup calculate the hash on the weights file and if it doesn't match best-network then re-download it? That should clean up a corrupt download, right?

Ttl:
I made a separate pull request #143 for checking the hash.

Ttl:
Resolved conflicts.

Also included a fix for hash checking pull request I just uncovered because curl exit code is checked. Apparently Curl doesn't overwrite if the server gives the filename.

roy7:
@ttl Do you think it'd make sense to give curl the --retry option I mentioned in https://github.com/gcp/leela-zero/issues/93 or is it better if it fails to just immediately run another game? I can close https://github.com/gcp/leela-zero/issues/93 if it doesn't make sense given this pull request.

Ttl:
This pull request added a retry function outside of curl for best-network-hash and best-network calls. There shouldn't be a need for curl to try to retry too.

Curl retry option could be used for game uploading as that's not being retried right now.

roy7:
So if server is down when a /submit it done, it'll autogtp won't catch that and it'll exit?

Ttl:
No it won't exit. The upload just fails and the game is lost. Next it calls best-network-hash which has the retry functionality. If the server is down it will keep trying to get the hash for few hours before giving up.

I don't think it's a big deal since only one game is lost.

roy7:
Ahh ok. As long as we're past the point autogtp will exit and people stop running games, that's great. Hopefully /next can go out to a new release sometime soon. :) The server process restarts automatically any time we change the code, and although it's fast to restart, there's a chance someone will get an error during that second or two. If so the old autogtp will just stop running which makes me uncomfortable when working on things.

It probably isn't as easy as just adding --retry to the submit curl command, which I was just looking at, since curl's result when retrying may confuse the success/error checking in autogtp. I should probably spend my time continuing server work vs learning how autogtp works.

roy7:
I tried simply adding --retry for a local test but it caused uploading to fail (or at least autogtp reported a failure). So I'll not worry about it, as you say, it'll just lose 1 game and not stop running.

> Writing 62e76d6932a94f1cbb02535146f2de8e.sgf
> Dumping 62e76d6932a94f1cbb02535146f2de8e.txt
> Stopping engine.
> 1 game(s) played in 7 minutes = 467 seconds/game, last game took 463 seconds.
> curl --retry -F networkhash=223737476718d58a4a5b0f317a1eeeb4b38f0c06af5ab65cb9d76d68d9abadb6 -F clientversion=5 -F sgf=@62e76d6932a94f1cbb02535146f2de8e.sgf.gz -F trainingdata=@62e76d6932a94f1cbb02535146f2de8e.txt.0.gz http://zero.sjeng.org/submit
> Upload failed. Curl Exit code: 2
> Continuing...

PR #138, opened 2017-11-22
Bugfixes

marcocalignano:


autogtp/Production.cpp
```diff
@@ -45,16 +45,17 @@ void ProductionWorker::run() {
          auto end = std::chrono::high_resolution_clock::now();
          auto gameDuration =
             std::chrono::duration_cast<std::chrono::seconds>(end - start).count();
-         emit resultReady(game.getFile(), gameDuration);
+         emit resultReady(game.getFile(), gameDuration, m_index);
     } while (true);
 }
 
-void ProductionWorker::init(const QString& gpuIndex, const QString& net) {
+void ProductionWorker::init(const QString& gpuIndex, const QString& net, const int &index) {
```
gcp: No point in passing an int per reference unless you want to modify it. (For objects like strings it's different, it avoids copies)

marcocalignano:
I am an Embedded System developer by Day I save all the copy i can ;)

PR #143, opened 2017-11-22
Check the network file hash

### LINKED ISSUE #128: client didn't check the checksum of network file
If the download of value network file is incomplete, the client will quit immediately after start engine. 
There is no check or retry for the network file. Every time you execute the client, it will fail.

Ttl:
Uses picosha2 (https://github.com/okdshin/PicoSHA2) for generating the hash.

Fixes #128

### DIFF
--- a/autogtp/Production.cpp
+++ b/autogtp/Production.cpp
@@ -18,6 +18,7 @@
 
 #include <QDir>
 #include <QFileInfo>
+#include <QCryptographicHash>
 #include "Production.h"
 #include "Game.h"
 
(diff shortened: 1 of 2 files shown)


gcp:
FWIW Qt also has support for this http://doc.qt.io/qt-5/qcryptographichash.html

That said no objection over adding a single MIT licensed header-only lib.

marcocalignano:
That won't 'ruin' GPL 3?

Ttl:
I'll redo this with the Qt function. Just a sec.

Ttl:
I updated the commit.

gcp:
MIT and X11 license are compatible with GPLv3. There is already other MIT/X11 code in the tree.

marcocalignano:
On line 187 of Production.cpp, can't you delete the file and return false so the network is downloaded again?
I think is better then exiting

Ttl:
If that check fails there are some serious problems. If the file can't be opened for reading I'm not so sure that deleting it is going to succeed either.

marcocalignano:
so if QFileInfo::exists finds it but cannot open for reading, probably we miss permission so we cannot even delete it but you can still try and of deleting fail you can exit. (maybe)

PR #151, opened 2017-11-23
autogtp: bugfix for leelaz gpu command line option

marcocalignano:


marcocalignano:
Are you going to release the multigame version, so maybe we see a bit more games coming in?

PR #161, opened 2017-11-24
AutoGTP: Fix the GPU command line option for Validation mode.

akdtg:
" -gpu=" => " --gpu="


gcp:
@marcocalignano Is that option actually used anywhere? Looking in the code, m_option appears to be unused, instead the options are hardcoded again in run().

marcocalignano:
I did a new pull request with both fixes. Thanks @atdtg! Another bug found.

gcp:
I merged the other pull request.

PR #177, opened 2017-11-25
Small fix in Game.cpp

godmoves:


earthengine:
I spotted this out independently; however, if compiled with VC++ with my pull request, the file name of the exe will be different though.

PR #182, opened 2017-11-26
Changed NN in-memory representation from fp32 to fp16

ihavnoid:
This is a patch that uses half fp for the weights and the input/outputs.  They are converted immediately to fp32 once the GPU loads them, and all arithmetic is done in fp32.  This gives some (10~15%) performance improvement using a GTX1080.

Not sure if this is worth merging to master, but putting a pull request since it may worth thinking...

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,7 @@ https://github.com/BlinkDL/BlinkDL
 
 # License
 
-The code is released under the GPLv3 or later, except for ThreadPool.h and
-cl2.hpp, which have specific licenses (compatible with GPLv3) mentioned in
+The code is released under the GPLv3 or later, except for ThreadPool.h, half.hpp
+and cl2.hpp, which have specific licenses (compatible with GPLv3) mentioned in
 those files.
+
--- a/src/Im2Col.h
+++ b/src/Im2Col.h
@@ -25,7 +25,7 @@
 
 template <unsigned long filter_size>
 void im2col(const int channels,
-            const std::vector<half>& input,
+            const std::vector<net_t>& input,
             std::vector<float>& output) {
     constexpr unsigned int height = 19;
     constexpr unsigned int width = 19;
@@ -35,7 +35,7 @@ void im2col(const int channels,
     constexpr unsigned int output_h = height + 2 * pad - filter_size  + 1;
     constexpr unsigned int output_w = width + 2 * pad - filter_size + 1;
 
-    const half* data_im = input.data();
+    const net_t* data_im = input.data();
     float* data_col = output.data();
 
     for (int channel = channels; channel--; data_im += channel_size) {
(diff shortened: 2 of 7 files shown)


Ttl:
I tried the changes and they seem to work. I also get 10% performance increase on GTX1050 and the strength of the program seems same as before.

What's the source for the half.hpp header? It doesn't have any license information.

ihavnoid:
half.hpp is from:

http://half.sourceforge.net/

MIT License - is it okay?

Ttl:
MIT license is compatible with GPL, there is no problem with it.

Can you restore the license comments to the beginning of half.hpp?

ihavnoid:
Strange, I thought I didn't strip out any copyright info, but apparently it seems I did.  Will restore it and update it.

sethtroisi:
On my CPU limited computer (i5-2400 + GTX 1080 Ti) I only see a 6% speedup.

Before:
44 game(s) played in 141 minutes = 192 seconds/game, 17957 moves = 471 ms/move

After:
44 game(s) played in 135 minutes = 185 seconds/game, 18433 moves = 442 ms/move

isty2e:
I tried to compile this on my mac, but ended up with the following error:

> In file included from Network.cpp:33:
> ./Im2Col.h:39:11: error: cannot initialize a variable of type 'half_float::half *' with an rvalue of type 'value_type *' (aka 'float *')
>     half* data_col = output.data();

What would be the problem and the solution?

ihavnoid:
I see the problem here.  I will push something new, but I wonder how it did work before... :(

isty2e:
Yes, it seems like working fine now.

ihavnoid:
@sethtroisi - It is possible that it adds some overhead to CPU because this implementation does the final fp16->fp32 conversion from the CPU.  Still it is not clear if doing it from GPU will improve performance since it will increase the amount of data moving from GPU to CPU.  I am trying to see if there is any other potential optimization though.

isty2e:
A 15% performance gain for me. However its Elo rating should be checked against fp32 version, I suppose?

gcp:
>Not sure if this is worth merging to master, but putting a pull request since it may worth thinking...

I would consider merging it if the same source can support both float and half. The cleanest solution is likely templating a bunch of stuff on type, and then adding a USE_HALF in config.h. There's ways to use DEFINE's in OpenCL kernels that should make it workable to substitute the type in them.

But not all that easy, unfortunately.

ihavnoid:
Well, I did make changes to support things with an ifdef.  Please see 74ac481.

Points to check:
- datatype naming
- coding style
- glaring bugs (hope none)

At least made sure it generates the first move properly on both modes.

gcp:
Can you resend this pull request but on the "next" branch instead of "master"?

PR #187, opened 2017-11-26
Add weights files and qt generated files to gitignore

sethtroisi:


OmnipotentEntity:
Another nit to pick: Can you add the CMake generated files to the .gitignore as well?

They are:
CMakeCache.txt
CMakeFiles/
cmake_install.cmake

It also generates a Makefile but I'm unsure how to tell .gitignore to only ignore that file in the top directory...

sethtroisi:
I think /Makefile would do the job. I'll update it.

autogtp/.gitignore
```diff
@@ -2,4 +2,9 @@
...
 /autogtp
 /*.sgf
-/*.gz
\ No newline at end of file
+/*.gz
+/moc_*.cpp
+/moc_*.h
+
+# Weight files
+/[0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f]*[0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f]
```
sethtroisi: technically I should repeat the pattern 64 times but I think if the file starts with 5 hex characters and ends with 5 more it's probably a weight file

marcocalignano:
It depends of where you execute cmake. 
Usually it is not good to execute cmake in the main directory, but you create a new directory like build and you execute cmake inside it. All the TEMPORARY files and the end binary are done in build so the change you did in .gitignore is useless. If I choose a directory build_leelaz as building directory then even having build in .gitignore won't help.

sethtroisi:
That's totally true and I can revert any portion of this change if you or gcp don't like it

My reason for adding it was 
1. It's a gitignore file, I use it to make my life easier. OmnipotentEntity and I have both see the extra untracked files in our git status and having no untracked files feel much more zen.
2. if you build in build or build_leelaz then git lists only one entry, the untracked folder, instead of 4 files

Tell me what you'd like and I'll make the change.

marcocalignano:
It is not what @gcp likes or what I like.
It is that you like to do things in a way that can be different from mine. So you need a different .gitignore then me. 
We can decide to write a .gitignore that is compatible with all the people contributing to this project, but then every different one will have a different pull request.
For sure, you work inside your own fork, so make your .gitignore there. (In some project the .gitignore is not even added to the repository so every developer has her/his own).

For the main repository we can establish a praxis (providing the description in the readme) that if you use Cmake you should use a certain subdirectory for the build and this subdirectory can be ignored. 
But this is just my opinion.

sethtroisi:
I just learned the word praxis and I really like it.

I'm new to using GitHub to coordinate a multi person project. I'll search around on Google and see if I can't understand how to manage common bits of local code more effectively.

In the mean time I've reverted all the changes that aren't the convention (resulting from running code in the readme.md).

autogtp/.gitignore
OmnipotentEntity: You can just do `[0-9a-f]{64}` no?

gcp:
I believe it's as good as standard to run cmake in a separate "build" directory.

i.e. 
```
mkdir build
cd build
cmake ..
```

autogtp/.gitignore
sethtroisi: No, sadly .gitignore uses a glob pattern match , instead of a full regex

https://stackoverflow.com/questions/33189437/explain-gitignore-pattern-matching

sethtroisi:
I've reverted the cmake portion of this PR leaving entries for:
weights.txt, autogtp weights, and qt moc files.

Would you like me to add "/build" back to .gitignore?

PR #190, opened 2017-11-27
Optimizations to Network.cpp and OpenCL.cpp

sethtroisi:
## Benchmark

'weights' is weights.txt from 'best_v1.txt.zip' (6 x 128)
'4d9a5cf' is 62k network (4 x 32)

| network | compile mode | GPU RAM | netbench result |
| --- | --- | ---:| ---:|
| 4d9a5cf | default | 195MiB | 1739 n/s |
| 4d9a5cf | debug | 195MiB | 1432 n/s |
| weights | default | 203MiB |  686 n/s |
| weights | debug | 203MiB |  602 n/s |

##  OpenCl.cpp + Network.cpp (smaller arrays)

| network | compile mode | GPU RAM | netbench result | rough speed up |
|---|---|---:|---:|---:|
| 4d9a5cf | default | 149MiB |  2239 n/s | +29%
| 4d9a5cf | debug | 149MiB | 2020 n/s |
| weights | default | 159MiB |  769 n/s | +12%
| weights | debug | 159MiB |  711 n/s | 

-----

## Spot Check (requires disabling RANDOM_ROTATION)

test command: `
make -j3 && echo -e "d3 n19\nd9 o18\nd18 r15\ne4 e5\nf4 f5\ng4 g5\ngenmove black" | sed 's/^\(.*\) \(..[0-9]\)/play black \1\nplay white \2/' | ./leelaz -w <NETWORK> --noponder -p 5000 -n |& grep -A6 '^\(NN eval\|[0-9]* visits\)'`

| patched | Network | NN eval | speed | PV |
| --- | :--- | :--- | :---| ---: |
| NO | 4d9a5cf | NN eval=0.298897 | 2146 n/s | L17 ->      88 (V: 32.71%) (N:  0.9%) PV: L17 B1 J17 |
| NO | weights | NN eval=0.540105 | 889 n/s |   R4 ->    1695 (V: 58.92%) (N: 28.7%) PV: R4 P4 P3 O3 Q3 O4 R6 K4 H3
| YES | 4d9a5cf | NN eval=0.298897 |  2488 n/s (+16%) |  L17 ->      88 (V: 32.71%) (N:  0.9%) PV: L17 B1 J17
| YES | weights | NN eval=0.540105 |  1004 n/s  (+13%) |  R4 ->    1697 (V: 58.93%) (N: 28.7%) PV: R4 P4 P3 O3 Q3 O4 R6 K4 H3

### DIFF
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -85,7 +85,7 @@ std::array<float, 1> ip2_val_b;
 
 void Network::benchmark(GameState * state) {
     {
-        int BENCH_AMOUNT = 1600;
+        int BENCH_AMOUNT = 20000;
         int cpus = cfg_num_threads;
         int iters_per_thread = (BENCH_AMOUNT + (cpus - 1)) / cpus;
 
@@ -277,8 +277,8 @@ void convolve(const std::vector<float>& input,
     im2col<filter_size>(channels, input, col);
 
     // Weight shape (output, input, filter_size, filter_size)
-    // 96 22 5 5
-    // outputs[96,19x19] = weights[96,22x9] x col[22x9,19x19]
+    // 96 22 3 3
+    // outputs[96,19x19] = weights[96,22x3x3] x col[22x3x3,19x19]
     // C←αAB + βC
     // M Number of rows in matrices A and C.
     // N Number of columns in matrices B and C.
@@ -404,13 +404,14 @@ Network::Netresult Network::get_scored_moves(
 Network::Netresult Network::get_scored_moves_internal(
     GameState * state, NNPlanes & planes, int rotation) {
     assert(rotation >= 0 && rotation <= 7);
-    constexpr int channels = INPUT_CHANNELS;
-    assert(channels == planes.size());
+    constexpr int input_channels = INPUT_CHANNELS;
+    assert(input_channels == planes.size());
     constexpr int width = 19;
     constexpr int height = 19;
-    constexpr int max_channels = MAX_CHANNELS;
-    std::vector<float> input_data(max_channels * width * height);
-    std::vector<float> output_data(max_channels * width * height);
+
+    auto convolve_channels = int(conv_pol_w.size() / conv_pol_b.size());
+    std::vector<float> input_data(convolve_channels * width * height);
+    std::vector<float> output_data(convolve_channels * width * height);
     std::vector<float> policy_data_1(2 * width * height);
     std::vector<float> policy_data_2(2 * width * height);
     std::vector<float> value_data_1(1 * width * height);
@@ -419,12 +420,14 @@ Network::Netresult Network::get_scored_moves_internal(
     std::vector<float> softmax_data((width * height) + 1);
     std::vector<float> winrate_data(256);
     std::vector<float> winrate_out(1);
-    for (int c = 0; c < channels; ++c) {
+
+    float* input = input_data.data();
+    for (int c = 0; c < input_channels; ++c) {
+        int board_pos = 0;
         for (int h = 0; h < height; ++h) {
             for (int w = 0; w < width; ++w) {
-                auto rot_idx = rotate_nn_idx(h * 19 + w, rotation);
-                input_data[(c * height + h) * width + w] =
-                    (float)planes[c][rot_idx];
+                auto rot_idx = rotate_nn_idx(board_pos++, rotation);
+                *(input++) = planes[c][rot_idx];
             }
         }
     }
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -405,7 +405,7 @@ Network::Netresult Network::get_scored_moves_internal(
     GameState * state, NNPlanes & planes, int rotation) {
     assert(rotation >= 0 && rotation <= 7);
     constexpr int input_channels = INPUT_CHANNELS;
-    assert(planes.size() == input_channels);
+    assert(input_channels == planes.size());
     constexpr int width = 19;
     constexpr int height = 19;
 
@@ -415,6 +415,7 @@ Network::Netresult Network::get_scored_moves_internal(
 
     std::vector<net_t> input_data(input_size);
     std::vector<net_t> output_data(output_size);;
+
     std::vector<float> policy_data_1(2 * width * height);
     std::vector<float> policy_data_2(2 * width * height);
     std::vector<float> value_data_1(1 * width * height);
@@ -423,12 +424,14 @@ Network::Netresult Network::get_scored_moves_internal(
     std::vector<float> softmax_data((width * height) + 1);
     std::vector<float> winrate_data(256);
     std::vector<float> winrate_out(1);
+
+    // Data layout is input_data[(c * height + h) * width + w]
+    net_t* input_ptr = input_data.data();
     for (int c = 0; c < input_channels; ++c) {
         for (int h = 0; h < height; ++h) {
             for (int w = 0; w < width; ++w) {
                 auto rot_idx = rotate_nn_idx(h * 19 + w, rotation);
-                input_data[(c * height + h) * width + w] =
-                    net_t(planes[c][rot_idx]);
+                *(input_ptr++) = net_t(planes[c][rot_idx]);
             }
         }
     }
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -404,18 +404,15 @@ Network::Netresult Network::get_scored_moves(
 Network::Netresult Network::get_scored_moves_internal(
     GameState * state, NNPlanes & planes, int rotation) {
     assert(rotation >= 0 && rotation <= 7);
-    constexpr int input_channels = INPUT_CHANNELS;
-    assert(input_channels == planes.size());
     constexpr int width = 19;
     constexpr int height = 19;
-
-    const auto output_channels = int(conv_pol_w.size() / conv_pol_b.size());
+    constexpr int input_channels = INPUT_CHANNELS;
+    assert(input_channels == planes.size());
     const auto input_size = input_channels * width * height;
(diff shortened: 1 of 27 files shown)


ihavnoid:
Amazing.  So, the optimization was due to simply instantiating a smaller buffer which resulted in subsequent stages getting to do less work.  Correct?

gcp:
It potentially avoids the index calculation but I think the std::vector default initializes the elements so this avoids initializing the unused part.

gcp:
I don't think you need MAX_CHANNELS in any case actually, it looks like input_data can be sized according to the planes size, as the OpenCL code probably only copies stuff out of it. So this should give you some further speedup.

src/OpenCL.cpp
```diff
@@ -512,11 +514,13 @@ void OpenCL_Network::convolve(int filter_size, int channels, int outputs,
...
     }
 
-    // Input channel grouping
+    // Input channel grouping in multiples of 8
     int channelGroup = 8;
     int channelShift = 3;
+
     // Input layer is not a multiple of 8
-    if (channels == 18) {
+    if (channels % 8 != 0) {
```
sethtroisi: I thought this was slightly better semantically but I can revert if you like the hard coded 18 (or I can replace with Network::INPUT_CHANNELS)

sethtroisi:
I updated the PR and added quite a lot of debug output to the description.

TL;DR is much faster on smaller networks, a little faster on large networks, uses 25% less VRAM

src/OpenCL.cpp
gcp: It's fine.

src/Network.cpp
```diff
@@ -404,13 +404,14 @@ Network::Netresult Network::get_scored_moves(
...
+    constexpr int input_channels = INPUT_CHANNELS;
+    assert(input_channels == planes.size());
     constexpr int width = 19;
     constexpr int height = 19;
-    constexpr int max_channels = MAX_CHANNELS;
-    std::vector<float> input_data(max_channels * width * height);
-    std::vector<float> output_data(max_channels * width * height);
+
+    auto convolve_channels = int(conv_pol_w.size() / conv_pol_b.size());
+    std::vector<float> input_data(convolve_channels * width * height);
```
gcp: Is this actually correct? The point of my previous remark was that this is only ever equal to the input channels, and now you're resizing it back to the same size as the outputs, so that would be too big? Thereby throwing a bunch of your speedup away?

src/OpenCL.cpp
```diff
@@ -394,30 +394,30 @@ void OpenCL_Network::forward(const std::vector<float>& input,
     constexpr size_t one_plane = width * height * sizeof(float);
 
     opencl.ensure_thread_initialized();
-    const size_t midSize = one_plane * Network::MAX_CHANNELS;
     const size_t inSize = sizeof(float) * input.size();
+    const size_t channels = input.size() / (width * height);
```
gcp: Maybe name it input_channels for clarity.

src/OpenCL.cpp
```diff
@@ -394,30 +394,30 @@ void OpenCL_Network::forward(const std::vector<float>& input,
...
     const size_t finalSize = m_layers.back().outputs * one_plane;
 
+    // Input layers in groups of 2 or channels in groups of 8.
+    const size_t maxOutputs = std::max(size_t(Network::INPUT_CHANNELS >> 1),
+                                       channels >> 3);
+
     if (!opencl_thread_data.m_buffers_allocated) {
-        size_t alloc_midSize = one_plane * Network::MAX_CHANNELS;
```
gcp: FWIW the reason this was historically hardcoded was that the first layer could have less outputs than the subsequent layers, so sizing based on the first call didn't work. But for the Zero design it works fine.

sethtroisi:
I ended up reverting some of your changes from 8a01db302b7b5f7ee7f52322f050b053832c8c46 but I think the solution came out cleaner

src/Network.cpp
```diff
@@ -406,10 +406,13 @@ Network::Netresult Network::get_scored_moves_internal(
...
     constexpr int width = 19;
     constexpr int height = 19;
-    assert(INPUT_CHANNELS == planes.size());
-    std::vector<net_t> input_data;
-    auto convolve_channels = conv_pol_w.size() / conv_pol_b.size();
-    std::vector<net_t> output_data(convolve_channels * width * height);
+    constexpr int input_channels = INPUT_CHANNELS;
+    assert(input_channels == planes.size());
+    const auto input_size = input_channels * width * height;
+    const auto output_channels = int(conv_pol_w.size() / conv_pol_b.size());
```
gcp: Don't need the cast here. (It's counterproductive: wouldn't you rather end up with a size_t?)

src/Network.cpp
```diff
@@ -406,10 +406,13 @@ Network::Netresult Network::get_scored_moves_internal(
...
+    const auto output_size = output_channels * width * height;
+    std::vector<net_t> input_data(input_size);
```
gcp: This undoes part of the optimization! It forces the array to be default-initialized even though we will overwrite it instantly. 

The previous code doesn't initialize the array, but reserves enough place ahead of time (so there is only a single allocation), and then emplace_backs the elements into it, which is equivalent to the pointer code, but (IMHO) better C++ style.

src/Network.cpp
```diff
@@ -419,12 +422,12 @@ Network::Netresult Network::get_scored_moves_internal(
     std::vector<float> winrate_data(256);
     std::vector<float> winrate_out(1);
     // Data layout is input_data[(c * height + h) * width + w]
-    input_data.reserve(INPUT_CHANNELS * width * height);
-    for (int c = 0; c < INPUT_CHANNELS; ++c) {
+    net_t* input_ptr = input_data.data();
```
gcp: No need to revert to raw pointer use here. Yes, you can't use emplace_back (or push_back) any more but that is because the array is already initialized, which is actually not what you want.

src/Network.cpp
```diff
@@ -406,10 +406,13 @@ Network::Netresult Network::get_scored_moves_internal(
...
-    std::vector<net_t> input_data;
-    auto convolve_channels = conv_pol_w.size() / conv_pol_b.size();
-    std::vector<net_t> output_data(convolve_channels * width * height);
+    constexpr int input_channels = INPUT_CHANNELS;
+    assert(input_channels == planes.size());
+    const auto input_size = input_channels * width * height;
+    const auto output_channels = int(conv_pol_w.size() / conv_pol_b.size());
+    const auto output_size = output_channels * width * height;
+    std::vector<net_t> input_data(input_size);
+    std::vector<net_t> output_data(output_size);;
```
gcp: Stray semicolon.

src/OpenCL.cpp
```diff
@@ -411,35 +411,37 @@ void OpenCL_Network::forward(const std::vector<net_t>& input,
...
-        size_t alloc_mergeSize = one_plane *
-            Network::MAX_CHANNELS * (Network::MAX_CHANNELS / 2);
+        auto maxInBufferSize = 0;
+        auto maxMergeSize = 0;
+        for (auto& layer : m_layers) {
+            auto channelGroups = layer.channels / (layer.channels % 8 ? 2 : 8);
+            maxMergeSize = std::max(
+                maxMergeSize, int(layer.outputs * channelGroups));
```
gcp: You can avoid the extra casts by using  ``` std::max<int> ```or ```std::max<size_t>```

src/OpenCL.cpp
```diff
@@ -411,35 +411,37 @@ void OpenCL_Network::forward(const std::vector<net_t>& input,
...
-    const size_t inSize = sizeof(net_t) * input.size();
-    const size_t finalSize = m_layers.back().outputs * one_plane;
 
     if (!opencl_thread_data.m_buffers_allocated) {
-        size_t alloc_midSize = one_plane * Network::MAX_CHANNELS;
-        size_t alloc_mergeSize = one_plane *
-            Network::MAX_CHANNELS * (Network::MAX_CHANNELS / 2);
+        auto maxInBufferSize = 0;
+        auto maxMergeSize = 0;
+        for (auto& layer : m_layers) {
```
gcp: can be ```const auto&```

src/OpenCL.cpp
```diff
@@ -411,35 +411,37 @@ void OpenCL_Network::forward(const std::vector<net_t>& input,
...
+            auto channelGroups = layer.channels / (layer.channels % 8 ? 2 : 8);
+            maxMergeSize = std::max(
+                maxMergeSize, int(layer.outputs * channelGroups));
+            maxInBufferSize = std::max(maxInBufferSize, int(layer.channels));
+        }
+        const size_t alloc_inSize = one_plane *  maxInBufferSize;
```
gcp: Might as well use auto here for all new code but it's up to you.

sethtroisi: "always use const" is on the gcp styleguide

src/OpenCL.cpp
sethtroisi: I didn't know, thanks

src/Network.cpp
sethtroisi: Yep, you're totally correct about reserve / fill constructor. 

I think that emplace_back and push_back get optimized to the same code but I'll check with assembler / profiler at some later date.

src/Network.cpp
sethtroisi: I don't write a lot of c++, I'm sorry that you keep having to walk me (and others) though c++ basics and your preferred style.

src/OpenCL.cpp
OmnipotentEntity: `const auto` is a thing.

PR #192, opened 2017-11-28
small optimization: simplify get_pv

sethtroisi:
Add a new function to get the "best" child (as determined by NodeComp)

this avoids sorting and linking all children x2

Intel VTune says impact is ~0.5% CPU


gcp:
Looks great. Impact is little but avoiding the resorting is good because the fact that the node with the highest prior isn't necessarily first in the list can make some search experiments more annoying.

PR #196, opened 2017-11-28
Support mini-batch, CPU, Cuda, CuDNN by using Theano running in a seperated process

NTT123:
This is my hack to run the feed forward Net in a separated python process. Leela zero C++ then talks to Leela zero python using UDP connections. Sending UDP packets of the input data. Receiving UDP packets of Output of the Net.

This allows us to combine many threads (and processes) of Leela C++ in a single feed forward batch.

Moreover, using Theano allows us to automatically support multiple backends (CPU, Cuda, CUDNN, ...)

My simple benchmarks show ~ 2-3 X speed up compared to Leela C++ OpenCL. (I have a slow Nvidia GPU card, therefore not much benefit from mini-batches.)

I'm sure that this is an ugly hack but I cannot wait longer to share it with you.


gcp:
Thanks. It makes sense that batching would provide a good speedup, especially combined with cuDNN (I assume you have that backend active?), which is why I put that in the README.

Unfortunately the odds of merging this as is are obviously zero, as it makes distribution of the clients immeasurably more difficult. It's probably better to make a CUDA version and then put optional cuDNN support in directly (just using CUDA, with no batching and no cuDNN, won't provide a performance benefit - it's the same speed as the OpenCL code). But also note, that you cannot redistribute cuDNN(!). So it's not possible to make a bundle with it and the user would always have to go make an NVIDIA account themselves.

Another thing to check would be the influence of blocking the UCT search until the batch is finished. This would theoretically make the program weaker because the search is much less efficient, although likely the speed difference makes more than up for it. For the use case of generating training data, it would have to be all the games played in parallel. But it's more difficult to implement this cleanly.

gcp:
My recommendation would be to make a separate github repo and include setup instructions. But make sure the leelaz versions are *not* using multi-threading, or the games will be weaker than intended. Just running multiple instances in parallel would still work with this UDP setup, right?

NTT123:
Yes, it works with multiple instances.

gcp:
Can you change

https://github.com/gcp/leela-zero/blob/master/src/Leela.cpp#L55
and
https://github.com/gcp/leela-zero/blob/master/src/GTP.cpp#L67

to make sure this is never ever run with threads != 1?

NTT123:
Yes, `#define MAX_CPUS 1` in the `config.h` is enough.

fishcu:
@NTT123 I ran your code this night (still on this commit: https://github.com/gcp/leela-zero/pull/196/commits/b285717fb20792cc4ddac038c01806f9e6551dbc). Everything worked fine and I generated 5 games!

I found that for this i5 CPU, batch size = 1 or = 2 works best. With batch size = 4, it is significantly slower and CPU usage drops from 100% to about 80%. There must be some bottleneck introduced.

Using theano and your code is about a 10x speedup over running OpenCL on the CPU. Great work!!

NTT123:
Great. You better pull my latest commit as I switched to TCP instead of UDP because UDP may cause problem in the output when UDP packets were missed!

fishcu:
Just pulled the latest commit, but now it's not working anymore. The server output reads (I tried starting autogtp twice):
THEANO_FLAGS=blas.ldflags='-L/usr/lib/ -lblas,device=cpu,dnn.conv.algo_fwd=time_once,floatX=float32' python server.py 292k.txt 2
Leela Zero TCP Neural Net Service
Detecting the number of residual layers...
64 channels...
5 blocks...
Batch size = 2
Serving on ('127.0.0.1', 9999)
Connection from ('127.0.0.1', 33679)
Connection from ('127.0.0.1', 55851)


The autogtp output reads:
./autogtp 
autogtp v4
Best network hash: 92c658d7325fe38f0c8adbbb1444ed17afd891b9f208003c272547a7bcb87909
Required client version: 4 (OK)
Already downloaded network.
Engine has started.
Infinite thinking time set.


CPU usage is at 0%, so nothing is going on. Any ideas?

NTT123:
With Batchsize=2, you need 2 Leela instances running in parallel.

fishcu:
My bad, just realized this. Thanks!!

NTT123:
Btw, could you please try my last commit which now allows auto reloading the network from server? Thanks.

fishcu:
Using urllib.request breaks python 2 compatibility. 
When I replace urllib.request.urlopen with urllib2.urlopen as suggested [here](https://stackoverflow.com/questions/3745771/urllib-request-in-python-2-7), I run into this error when starting the server:
THEANO_FLAGS=blas.ldflags='-L/usr/lib/ -lblas,device=cpu,dnn.conv.algo_fwd=time_once,floatX=float32' python server.py 2
Leela Zero TCP Neural Net Service

Loading lastest network
Hash: 92c658d7325fe38f0c8adbbb1444ed17afd891b9f208003c272547a7bcb87909
Downloading weights
  File "server.py", line 82, in <module>
    txt     = downloadBestNetworkWeight(nethash)
  File "server.py", line 33, in downloadBestNetworkWeight
    open(hash + ".txt", "w").write(txt).close()
AttributeError: 'NoneType' object has no attribute 'close'

I guess the interface between urllib.request.urlopen and urllib2.urlopen is slightly different. Trying to fix this now...

fishcu:
It worked on 2nd try, I guess it was just some bad return from the request. Maybe this should be handled in code.

NTT123:
It is my fault calling close function after write.

NTT123:
Ok, it now should support py2 and `Ctrl + C` again.

fishcu:
It works great!!

NTT123:
@fishcu Please consider pulling the latest commit as the earlier version could stuck when not enough data is sent.

sbf2000:
Report: Ubuntu 16.04 ,gpupy0.7.5, theaon1.0.0,CUDA 9.0 ,cuDNN7.0,  test batch size from 1 to 8, no speed up.

NTT123:
Hmm. Did you turn the `USE_SERVER_TEST` on ?  I just try with Ubuntu 17.10 and the same your versions of your packages and get about  ~2.5x speed up with batch-size = 4.

Possibly, speed up only for weak Nvidia GPUs as mine (quadro k2000).

sbf2000:
turn  USE_SERVER_TEST on ,compile error. I use Nvidia GTX1080ti .

NTT123:
I confirmed that this patch is not speed thing up with fast GPUs. I still don't know the season why. Possibly, because of my python code in the middle.

NTT123:
On the other hand, the good news is that the current python version uses less GPU than OpenCL. and my calculation show that IF we can get it use `say` 90% GPU, the *possible* speedup is about `2`.

sbf2000:
Yes,GPU only used  about 40%,CPU 15%.

sethtroisi:
I was also unable to get a speed up from a GTX1080 on an early version of this code (I'm excited to test again this weekend).
I noticed It was certainty using significantly less GPU % per move BUT latency in waiting for the large number of leela instances to sync was cancelling that out I think

NTT123:
I'm working on a shared memory version between C++ and Python. Hope that will solve the sync delay problem.

sbf2000:
I tested  your new code , set batch size  = 2 , GPU  usage 24% ; set batch size =8 ,GPU  usage 30% .

NTT123:
Thanks, your results is similar to mine. Multi-thread does not help us in this case.

fishcu:
Hi, I ran your code for 2 days before updating to newest commit. Everything worked stable.

With the newest commit, a few things again break python 2 compatibility:
- queue is called **Q**ueue in python 2
- socketserver => SocketServer

After fixing these just to try to run it, I get the following error:
THEANO_FLAGS=blas.ldflags='-L/usr/lib/ -lblas,device=cpu,dnn.conv.algo_fwd=time_once,floatX=float32' python server.py 1
Leela Zero TCP Neural Net Service

Loading latest network
Hash: 223737476718d58a4a5b0f317a1eeeb4b38f0c06af5ab65cb9d76d68d9abadb6
Downloading weights
Done!
 64 channels and 5 blocks

Compling the latest neural network
Done!

Batch size = 1
('\nListening at', 9999)

Starting a thread for auto updating latest weights
('A new connection from', ('127.0.0.1', 33911))
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 33911)
  File "/usr/lib/python2.7/SocketServer.py", line 596, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 331, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "server.py", line 251, in __init__
    super().__init__(request, clientaddr, server)
TypeError: super() takes at least 1 argument (0 given)
----------------------------------------

NTT123:
Hi @fishcu, the last commit is sadly my unsuccessful effort for getting speed up on fast GPUs with multi threads. I just run a `git revert` to the commits before multi thread effort.  Please pull the latest commit, it should work as before.

NTT123:
We now use shared memory and semaphores to sync between C++ and Python. This version gives about x2 speedup on Quadro P2000 with `batch size = 20`, even though, it uses only ~ 40-50% GPU in my tests.

The performance seems to be bounded by CPU as too many Leela are running.

ljn917:
I tried the most recent ipc branch (commit dce7146), but I got the following error after the 1st game. The server.py is dead locked, even does not respond to Ctrl+C.

This one is about 2x SLOWER than OpenCL (Geforce 770M)!!!!

```
1 game(s) played in 49 minutes = 2960 seconds/game, last game took 2960 seconds.
Best network hash: 223737476718d58a4a5b0f317a1eeeb4b38f0c06af5ab65cb9d76d68d9abadb6
Required client version: 4 (OK)
Already downloaded network.
Engine has started.
*ERROR*: Could not talk to engine after launching.
```

NTT123:
@ljn917 Thanks. I now know there is a bug at the end of the game when leela is closed and created again. At the moment, could you try with large batch size, e.g. `batch size = 8, 16 ...` , any improvement?

ljn917:
@NTT123 Yes, I will try, but I guess the lower GPU utilization could mean some bottleneck/delay in data transfer. At least, the speed should be comparable with OpenCL for the same batch size.

ljn917:
@NTT123  Looks like we need some exception handling when checking best network.
```
Exception in thread Thread-1:
  File "/usr/lib64/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/ljn/Downloads/leela-zero/leela-zero_ntt123/ipc/nn.py", line 222, in run
    newhash = getLatestNNHash()
  File "/home/ljn/Downloads/leela-zero/leela-zero_ntt123/ipc/nn.py", line 20, in getLatestNNHash
    txt = urllib.request.urlopen("http://zero.sjeng.org/best-network-hash").read().decode()
  File "/usr/lib64/python3.6/urllib/request.py", line 223, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib64/python3.6/urllib/request.py", line 532, in open
    response = meth(req, response)
  File "/usr/lib64/python3.6/urllib/request.py", line 642, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/lib64/python3.6/urllib/request.py", line 570, in error
    return self._call_chain(*args)
  File "/usr/lib64/python3.6/urllib/request.py", line 504, in _call_chain
    result = func(*args)
  File "/usr/lib64/python3.6/urllib/request.py", line 650, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 502: Bad Gateway
```

ljn917:
I think minibatch works better. Batch size = 4 increases my GPU utilization to ~85%. Time per move increases only 10~20% compared with batch size = 1.

The issue is still with end of game. Once a game in the batch finishes, the server hangs up, and all the others autogtps are suspended.

NTT123:
Thanks. Will fix the hang up problem soon.

ljn917:
Can we subscribe more autogtp instances than batch size, and make server.py asynchronous? We can solve I/O bottleneck if there is any.

NTT123:
At the moment, no. Server looking for exactly `batch-size` semaphores released to run the network.

We can run two mini-batch in parallel though. But, currently, it requires to edit semaphore's name and shared memory's name in the code.

NTT123:
The last commit fixed the hangup problem.

ljn917:
It works. I did some performance benchmark. Time per autogtp instance = time per move / batch size. It looks like we need very large mini batch size.

batch size       time per move(s)      time per autogtp instance(s)
   4                                 7.2                         1.8
  16                              18                            1.1
128                              77                            0.6

NTT123:
how much speed up do you get at best compared to OpenCL?

ljn917:
OpenCL is about 2.5s per move

NTT123:
Did you use multiple OpenCL instances or just one?

ljn917:
Just one

NTT123:
Then, it will not fair to compare the results of OpenCL and CuDNN.

ljn917:
Batch size = 1 for cuda is about same as opencl.

ljn917:
@NTT123 Can you take a look what happened here?
1. There is "gzip: 767772662a9d495bbeb569c65028f9b0.sgf: No such file or directory"
2.  Is curl called twice? The second should not be there.

BTW, the sgf is uploaded successfully. http://zero.sjeng.org/view/88e58a635c1f41a40545e02f10a6b73ca9b72a8fa92674f577dc57ccb3cff81b

```
autogtp v4
Best network hash: 223737476718d58a4a5b0f317a1eeeb4b38f0c06af5ab65cb9d76d68d9abadb6
Required client version: 4 (OK)
Already downloaded network.
Engine has started.
Infinite thinking time set.
1 (H17) 2 (N10) 3 (J5) 4 (D5) 5 (G9) 6 (O1) 7 (C16) 8 (R15) 9 (D2) 10 (G4) 11 (S18) 12 (J16) 13 (pass) 14 (pass) Game has ended.
Score: W+7.5
Winner: white
Writing 2fd6bb02c61c4e3fb30c07ec6b55f688.sgf
Dumping 2fd6bb02c61c4e3fb30c07ec6b55f688.txt
Stopping engine.
curl -F networkhash=223737476718d58a4a5b0f317a1eeeb4b38f0c06af5ab65cb9d76d68d9abadb6 -F clientversion=4 -F sgf=@2fd6bb02c61c4e3fb30c07ec6b55f688.sgf.gz -F trainingdata=@2fd6bb02c61c4e3fb30c07ec6b55f688.txt.0.gz http://zero.sjeng.org/submit
gzip: 767772662a9d495bbeb569c65028f9b0.sgf: No such file or directory
Game data 974352fc38ed15bbba6da7940f594f7b58c3bb50e17d3563f0860eb53b4e4764 stored in database
curl -F networkhash=223737476718d58a4a5b0f317a1eeeb4b38f0c06af5ab65cb9d76d68d9abadb6 -F clientversion=4 -F sgf=@767772662a9d495bbeb569c65028f9b0.sgf.gz -F trainingdata=@767772662a9d495bbeb569c65028f9b0.txt.0.gz http://zero.sjeng.org/submit
1 game(s) played in 17 minutes = 1053 seconds/game, last game took 1053 seconds.
Best network hash: 223737476718d58a4a5b0f317a1eeeb4b38f0c06af5ab65cb9d76d68d9abadb6
Required client version: 4 (OK)
Already downloaded network.
Engine has started.
Infinite thinking time set.
1 (T11) 2 (A5) 3 (B11) 4 (A3) 5 (L2) 6 (H8) 7 (O2) 
```

NTT123:
It likes you have two autogtp running in your console. Or, the results from other autogtp are shown together.

ljn917:
No, I redirected stdout to files.

ljn917:
Only one of the 128 autogtps shows this, and it is the first one. Let me run it longer and see what will happen.

NTT123:
Hmm, current version requires you to restart `python server.py` when  you *manually* kill autogtp  and leelaz with `Ctrl + C`. Because, when be forced exit, leelaz cannot clear up its *slot* in shared memory. Then, when you run a next autogtp, all slot are full causing problem.

If you let everything run in order:

  1. start python server.py
  2. start `batch size`  autogtp

It should be OK.

ljn917:
I did that because I was running a different batch size.

You are right that the second curl output comes from a different autogtp, but I don't understand why it happens....

I started 128 `./autogtp &> output_$i`. i=1 has a piece of curl output from i=4, but all the other curl outputs are fine....

ljn917:
I know what is going on. autogtp: main.cpp: upload_data() search for all *.sgf files, and try to upload them all.....

ljn917:
@NTT123 I tried to implement 2 mini batches, but it does not help.

NTT123:
@ljn917 You mean rename the shared memory and semaphore?

Can I ask your speed up when running multiple OpenCL?

I just want to know whether CuDNN is actually helpful or not.

ljn917:
OpenCL is faster when it is 4. I cannot run 16 OpenCL.... I have issue like #226

NTT123:
So, you're saying that OpenCL is faster than CuDNN batch-size=128? 
:(

ljn917:
I cannot run OpenCL when batch size > 4

ljn917:
@NTT123 Sorry, I did the math wrong..... When batch size = 4, CUDA is faster than OpenCL

ljn917:
```
batch size       time per move(s)      time per autogtp instance(s)
OpenCL:
  1                          2.5                          2.5
  4                          9                            2.25
CUDA:
  1                          3                            3
  4                          7.2                          1.8
 16                         18                            1.1
128                         77                            0.6
```

ljn917:
The new minibatch version is considerably slower...  ~5 times slower if # minibatch = 1, 20% slower if # minibatch = 2.

I had another implementation at https://github.com/ljn917/leela-zero/tree/udpserver (without renaming semaphore, please check if it is correct), but it does not provide performance benefit either. It has about the same performance as the original code without minibatch.

I think it would be better to put the minibatch version into a different branch and keep the original version.

NTT123:
@ljn917. Thanks. The last commit should not slow thing down (at least in case of 1 minbatch) as I only add a few lines of code for supporting rename. What are your parameters ? num_instances? batch_size?

ljn917:
I run `server.py 8 2` twice with different names, and started 8x2 autogtp with those names.

For the 1 minibatch, I just run `server.py 16 1` with 16 autogtp

NTT123:
@ljn917, the second parameter is the `batch size`. You can get similar result to the old version with `server.py 1 1`,  `server.py 2 2`, or, `server.py 4 4`.

If 128 is your best batch size, I suggest to use `python server.py 256 128`. As 128 processes running on GPU while other 128 processes prepare input for the next batch.

In case your GPU usage is not 100% with 1 batch, then, you can run `2, 3` batches in parallel, depend on your GPU.

ljn917:
@NTT123 Oh, I see. So, I should run `server.py 16 8` twice with 2 different names, and run 8x2 autogtp. Is it right?

NTT123:
@ljn917, if you run `server.py 16 8` **twice** then,  you need `2x 2 x 8` autogtp. The first `2x8` instances will use the first name, and the other `2x8` instances use the second name.

fishcu:
Do you plan to make your branch compatible with the performance improvements that the "next" branch brings over "master"? Or is there no difference because you are using theano anyways?

NTT123:
@fishcu, it is very easy to merge with "next" branch. We can get benefits from improvements 
 in Leela Monte Carlo Playouts (not neural network, though)

NTT123:
Merged with v0.7.

sethtroisi:
Another data point from my GTX 970

|server|games tested|ms/move|
|----|----|----:|
|./autogtp -g 1 | 165| 1681|
|./autogtp -g 1 | 135| 1239|
|python server.py 4 1 | 21 | 1190|
|python server.py 8 1| 8 | 1136|

with `server.py 4 1` and `server.py 8 1` I get 80% GPU load

Should I try increasing batchsize?
using more instances?

Any useful information I can provide?

NTT123:
@sethtroisi you should try `server.py 16 8`, `server.py 32 16` instead. I think you could get better results. 

You may also explore other configurations quickly by timing a few beginning moves, say,  2 or 3 moves. I find it's a quite accurate estimate of the average time per move.

sethtroisi:
This is fantastic, I get 270% speed on my 1080 TI and my GPU is now limited by my CPU (CPU usage is ~100%, GPU ~40%)

I get to come home to this
779 game(s) played in 1147 minutes = 88 seconds/game, 199 ms/move, last game took 5514 seconds.

instead of this

350 game(s) (245 self played and 105 matches) played in 1194 minutes = 204 seconds/game, 542 ms/move, last game took 798 seconds.

I'm working on reducing CPU usage in leela, at which point I think I think I can get another 20-30% improvement (meaning 3-3.5x the next branch)

zhu-jz:
@NTT123 mentioned it would be difficult to merge the next branch because of the match of different networks. Is there any easy solution now?

NTT123:
@zhu-jz, I'm working on supporting cuda/cudnn directly at this `next` branch https://github.com/NTT123/leela-zero/tree/next . However, it's currently slower than opencl.

sethtroisi:
I need a better CPU, I'm able to get 3x default performance (180ms/move, game ~70 seconds) on my 1080  after some CPU performance tweaks (primarily this [branch](https://github.com/gcp/leela-zero/compare/next...sethtroisi:x__no_history)) but I'm still only at less than 50% GPU usage

roy7:
What CPU are you on @sethtroisi?

sethtroisi:
I'm currently using an i5-2400, the 1080 doesn't fit in my other computer :/

roy7:
If I can catch my server when it isn't busy sometime, would I just clone your repo and run autogtp for a bit as a benchmark vs the main repo? I have a 1700x + 1080 Ti.

sethtroisi:
My most recent commits have slowed it down a bit but you might try pulling https://github.com/sethtroisi/leela-zero/tree/udp_server,

I added a print statement every 2 minutes, it takes 3-4 or them for timing to stabilize (startup takes a couple of seconds but it gets amortized after a few minutes)

NTT123:
@sethtroisi, it now computes `softmax` on GPU. However, CPU usage is the same as before on my PC.

bood:
@sethtroisi You can use NC6 instance on MS/Azure for free for 10 days, it has 6 E5 CPU cores and a K80 card.

BTW: how do you pin point the CPU cost code? by using some kind of profiler?

sethtroisi:
@bood I'm using Intel vtune, it has a bunch of [views](https://software.intel.com/sites/default/files/did_feeds_images/185F93CE-C0A0-45C3-8E06-BD27191AE3DC/185F93CE-C0A0-45C3-8E06-BD27191AE3DC-imageId=84e3b017-302c-46d1-bf5a-ff758b3679bb.png) for understanding where the CPU spent time.

Alderi-Tokori:
Does this implementation support multiple GPUs? If so, I might want to take a look at it. If I can reproduce the 270% speed increase of @sethtroisi, it might help me a lot.

NTT123:
@Alderi-Tokori, it does not use multiple GPUs in a shared min-batch. However, you can run multiple min-batches, each min-batch on a GPU.

roy7:
Did this end up getting merged since it's a closed PR, or did it shuffle elsewhere in github?

bood:
I think @NTT123 closed it himself since @gcp suggests it's not going to be
merged. You can find it in his repo, and I think @sethtroisi also made some
changes in his repo too.

NTT123:
@roy7: what @bood said is true.There are also some compatibility problems with the `next` branch as current implementation does not support running evaluation games.

MartinDevelopment:
It seems like this is not for running on CPU only, it is like 20x(if it isn't more) slower than using the normal next branch.

jjoshua2:
Does this need to be rebased to .12 or is it ok to run still? Trying to set up cuda and cudnn...

NTT123:
@jjoshua2 I didn't update it lately as there isn't much interest in this approach (CuDNN). IMO, recent improvements on the master/next branch is a better solution.

jillybob:
Given the increasing in network size and computational power to make visits. I am very interested in using CuDNN for the latest networks. @NTT123 are you able to have a cuDNN option in a fork of LeelaZero?

PR #199, opened 2017-11-28
Update Production.cpp to fix dependency

NimbyDagda:
Fixed compile time error because cmath isn't included (dependency for std::pow)


marcocalignano:
What was the error?

marcocalignano:
There is not  std::pow in Production.h then you should put the include in the  Production.cpp

NimbyDagda:
OK I can do it that way if you prefer.

PR #200, opened 2017-11-28
Add optional autogtp param for specifying leelaz path

apetresc:
I am packaging leela-zero up for AUR to make it easy to set up on Arch Linux. In that context, I cannot assume the user's current working directory, so autogtp's behaviour of requiring leelaz in the same directory is a blocker.

This PR simply adds an optional argument to autogtp, `-p /path/to/leelaz`, which can allow autogtp and leelaz to be installed to a standard path like `/usr/bin` by packagers.

### DIFF
--- a/autogtp/Game.cpp
+++ b/autogtp/Game.cpp
@@ -19,10 +19,10 @@
 #include <QUuid>
 #include "Game.h"
 
-Game::Game(const QString& weights, QTextStream& out) :
+Game::Game(const QString& weights, QTextStream& out, const QString& leela_path) :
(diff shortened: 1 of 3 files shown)


gcp:
This needs rebasing and a more clever solution for Windows.

earthengine:
I guess this will work fine under Windows. Will try it tonight.

gcp:
This needs rebasing.

marcocalignano:
Rebasing on /master or on /next ?

gcp:
On next.

Issue #173, opened 2017-11-25
Questions about FastBoard


barrybecker4:
Thanks for contributing this project. It's really great.
I went through the FastBoard code today and have some questions. It would be great if you could add some comments directly to the code that would address these questions. I am sure that a few words of explanation will help others that might try to understand it.

What is the difference between count_pliberties and count_rliberties? Specifically, what to the "p" and "r" stand for?

What is the significance of 16384? m_libs[MAXSQ] is initialized to it. After some guessing I deduced that it is 2^14,  but why? Is this value the same regardless of the size of the board?

 Does m_libs hold the number of liberties of the parent string at each board position? 

 What does ss stand for in fast_ss_suicide?
 
 How does the s_eyemask work?

Instead of MAXSQ being used throughout the code, shouldn't you use m_maxsq? That way everything should work for dimensions other than 19x19. Maybe I'm wrong and it already does work for other dimensions.
 
How should we interpret the value that resides in m_neighbours[i]? I don't think it is literally the number of neighbours, but I'm not sure how it is used.
 
How does count_neighbours work? In particular can you add a comment to this line to explain what it does?
   ` return (m_neighbours[v] >> (NBR_SHIFT * c)) & 7;  `
			
In display_board, the code that draws the "a b c ..." letters at top and bottom is duplicated and could be refactored out into a separate method.  I'll make a pull request for this once I can build on windows.

Thanks!

fishcu:
> What is the difference between count_pliberties and count_rliberties? Specifically, what to the "p" and "r" stand for?

I have not gone through the code in detail, but to make an educated guess, I would say these refer to "pseudo liberties" and "real" (?) liberties.
With pseudo liberties, you count all neighboring empty cells to a group as a liberty. When placing stones adjacent to that group, you add one pseudo liberty for neighboring empty cells, and you remove one pseudo liberty for each neighboring group. There may be duplicate pseudo liberties with certain "bent" shapes. However, the test for 0 liberties remains correct, and IIRC, there is also a simple "atari" check.
Real liberties, OTOH, correspond to actual liberties of the group. I have not checked the code but would assume they're implemented using union-find.

The difference is speed, pseudo liberties have O(1) adding and removal, and real liberties have the inverse Ackermann function for that, but in practice there is a significant speed difference.

barrybecker4:
[...]

PR #203, opened 2017-11-28
Fast board refactor and comments

barrybecker4:
@gcp, I added some comments based on your responses in #173 and did a simple refactor so that the code to print the column header labels is not duplicated. Hope you can review. I am mostly a java, javascript, and scala programmer these days, but in the distant past I used c++. The PR is mostly about me learning and trying to get familiar with using MSVC. I'd feel a bit more comfortable making changes if there were some unit tests. Maybe I will open a separate issue for that.


.gitignore
```diff
@@ -2,4 +2,13 @@ leelaz
 *.o
 *.d
 training/tf/venv
-weights.txt
\ No newline at end of file
+weights.txt
+.idea
```
gcp: Remove the .idea (we had a discussion that we're not going to put everyone's favorite IDE/editor in here) and duplicate the VS2015 parts to VS2017.

.gitignore
```diff
@@ -2,4 +2,13 @@ leelaz
...
+/msvc/packages
+/msvc/VS2015/x64/Debug
+/msvc/VS2015/x64/Release
+/msvc/x64/Debug
+/msvc/x64/Release
+/msvc/leela-zero2015.VC.db
+/msvc/leela-zero2015.VC.VC.opendb
+/msvc/.vs/leela-zero2015/v14/.suo
```
gcp: These last 3 look unfamiliar to me.

src/FastBoard.h
```diff
@@ -154,6 +165,7 @@ class FastBoard {
     int update_board_eye(const int color, const int i);
     int in_atari(int vertex);
     bool fast_in_atari(int vertex);
+	void print_column_labels(int size, std::string padding = " ");
```
gcp: nit: incorrect indent

src/FastBoard.cpp
```diff
@@ -38,19 +38,22 @@ const int FastBoard::BIG;
 const int FastBoard::PASS;
 const int FastBoard::RESIGN;
 
+/* indicates that an empty square is never in atari. A large value that fits in a short */
+const int MAX_LIBS = 16384;  // 2^14
```
gcp: Move the definition to the header for consistency. It should be a constexpr as well. Also it's not really a max value. INF_LIBS?

src/FastBoard.cpp
```diff
@@ -200,10 +210,13 @@ void FastBoard::reset_board(int size) {
     }
 
     m_parent[MAXSQ] = MAXSQ;
-    m_libs[MAXSQ]   = 16384;    /* we will subtract from this */
+    m_libs[MAXSQ]   = MAX_LIBS;    /* we will subtract from this */
     m_next[MAXSQ]   = MAXSQ;
 }
 
+/*
+    @return true if placing the specified colored stone at the specified position would be suicide
```
gcp: IMHO this comment is entirely superfluous.

src/FastBoard.cpp
```diff
@@ -255,17 +268,30 @@ bool FastBoard::is_suicide(int i, int color) {
     }
 }
 
+/*
+    Liberties on a single point - called pseudo liberties.
```
gcp: If you're going to add an explanation of pseudoliberties, they should probably go with the respective datastructure in the header, rather than a random function using them. (The explanation is also not very clear, I think)

src/FastBoard.cpp
```diff
@@ -255,17 +268,30 @@ bool FastBoard::is_suicide(int i, int color) {
...
+    Count neighbours of color c at vertex v.
+    The border of the board has fake neighours of both colors.
+*/
 int FastBoard::count_neighbours(const int c, const int v) {
     assert(c == WHITE || c == BLACK || c == EMPTY);
     return (m_neighbours[v] >> (NBR_SHIFT * c)) & 7;
 }
 
+/* 
+    @return true if playing the specified single stone (ss) is suicidal.
```
gcp: true if playing at the given location is a single-stone suicide.

src/FastBoard.cpp
```diff
@@ -392,7 +421,10 @@ std::vector<bool> FastBoard::calc_reach_color(int col) {
     return bd;
 }
 
-// Needed for scoring passed out games not in MC playouts
+/*
+   The score is from the point of view of the black player. A negative score means white is leading.
+   @return score needed for scoring passed out games not in Monte-Carlo play-outs
```
gcp: It returns the area score. The explanation why this is needed is not related to the return value, so put it in the function description.

src/FastBoard.cpp
```diff
@@ -427,6 +462,9 @@ int FastBoard::estimate_mc_score(float komi) {
     return bsc-wsc-((int)komi)+1;
 }
 
+/*
+    @return final Monte-Carlo play-out score
```
gcp: Again, looks like a superfluous comment.

src/FastBoard.cpp
```diff
@@ -418,6 +450,9 @@ int FastBoard::get_stone_count() {
     return m_totalstones[BLACK] + m_totalstones[WHITE];
 }
 
+/*
+    @return estimate of the Monte-Carlo play-out score
```
gcp: That's what the function name says.

src/FastBoard.cpp
```diff
@@ -455,15 +493,8 @@ float FastBoard::final_mc_score(float komi) {
...
-    myprintf("\n   ");
-    for (int i = 0; i < boardsize; i++) {
-        if (i < 25) {
-            myprintf("%c ", (('a' + i < 'i') ? 'a' + i : 'a' + i + 1));
-        } else {
-            myprintf("%c ", (('A' + (i-25) < 'I') ? 'A' + (i-25) : 'A' + (i-25) + 1));
-        }
-    }
-    myprintf("\n");
+	myprintf("\n");
```
gcp: nit: incorrect spacing

This problem is repeated all over. I think your editor is misconfigured. Also if you can, try avoid lines longer than 80 characters.

src/FastBoard.cpp
```diff
@@ -565,6 +581,22 @@ void FastBoard::display_liberties(int lastmove) {
     myprintf("\n\n");
 }
 
+void FastBoard::print_column_labels(int size, std::string padding) {
+	myprintf("   ");
```
gcp: nit: inconsistent indentation with the rest of the program

padding isn't change so it can be passed by const reference.

src/FastBoard.cpp
```diff
@@ -716,6 +748,9 @@ int FastBoard::update_board_fast(const int color, const int i, bool & capture) {
     return -1;
 }
 
+/*
+    @return true if surrounded on 4 sides by the specified color and there are enough diagonals to avoid false eye.
```
gcp: This is the good kind of comment 👍

src/FastBoard.cpp
```diff
@@ -565,6 +581,22 @@ void FastBoard::display_liberties(int lastmove) {
     myprintf("\n\n");
 }
 
+void FastBoard::print_column_labels(int size, std::string padding) {
+	myprintf("   ");
+	for (int i = 0; i < size; i++) {
+		if (i < 25) {
```
gcp: I think Leela Zero is limited to maximum 25 sized boards so the >= 25 case can probably go.

src/FastBoard.cpp
```diff
@@ -1000,6 +1036,14 @@ int FastBoard::string_size(int vertex) {
     return m_stones[m_parent[vertex]];
 }
 
+/*
+   Actual (real) liberties of the string the specified vertex position belongs to.
+   Implemented using union-find.
```
gcp: That's rather misleading. The calculation is done incrementally, union-find is only used for locating the parent vertex. And even so, I don't think it's the pure union find that requires recursing on find-parent.

src/FastBoard.cpp
```diff
@@ -1000,6 +1036,14 @@ int FastBoard::string_size(int vertex) {
...
+   The difference is speed, pseudo liberties have O(1) adding and removal,
+   and real liberties have the inverse Ackermann function for that,
+   but in practice there is a significant speed difference.
```
gcp: This is just wrong. There's a difference between keeping track of the liberties and keeping track of the parent, and the latter is not the speed critical part.

.gitignore
barrybecker4: OK. done.

.gitignore
barrybecker4: I was going to remove them from .gitignore, but then I did not see how to exclude them from the commit using VS. I removed them and pushed from the cmd line.

src/FastBoard.cpp
barrybecker4: Moved it to .h and renamed it to INF_LIBS.

src/FastBoard.cpp
barrybecker4: I was basing the comment on what @fishcu said, and which you indicated was a correct description. There is no data structure for pseudoliberties that I can find - only the count_pliberties method which calls count_neighbors(EMPTY, i). I reduced the comment to just a single line until a better description can be found.

src/FastBoard.cpp
barrybecker4: Updated the comment to your suggestion.

src/FastBoard.cpp
barrybecker4: I removed the @return tag and updated the function description.

src/FastBoard.cpp
barrybecker4: OK. I removed the comment, but it was not automatically obvious to me when I first read the code what _mc_ stood for.

src/FastBoard.cpp
barrybecker4: Added the const qualifier.

src/FastBoard.cpp
barrybecker4: Yes. I did not have VS configured correctly so it was mixing in tabs. I think I have it configured correctly now. All tabs replaced with spaces and I turned on visible white-space to avoid future occurrences. I will try to keep lengths below 80, but that seems pretty low, and there are many lines that exceed that in the original.

src/FastBoard.cpp
barrybecker4: OK. I removed that case.

src/FastBoard.cpp
barrybecker4: Glad to see I got something right.

src/FastBoard.cpp
barrybecker4: I was going off of what @fishcu said. Reduced it to just a single line until a better description can be found.

src/FastBoard.cpp
barrybecker4: removed it.

src/FastBoard.h
barrybecker4: Fixed. No more tabs.

barrybecker4:
I made changes based on review comments, and merged with the latest. It seems that a bunch of code was removed recently.

gcp:
This unfortunately got overtaken by the large number of changes (and simplifications in FastBoard) being pulled in. Not sure how feasible it is to try another rebase.

barrybecker4:
I tried building the next branch in Visual studio 2015 and 2017 this morning, but I am getting the following error for autogtp. Not sure how to resolve it.

1>leela-zero.vcxproj -> E:\projects\leela-zero\msvc\x64\Debug\leela-zero.exe
1>Done building project "leela-zero.vcxproj".
2>------ Build started: Project: autogtp, Configuration: Debug x64 ------
2>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Microsoft.CppBuild.targets(387,5): warning MSB8028: The intermediate directory (x64\Debug\) contains files shared from another project (leela-zero.vcxproj).  This can lead to incorrect clean and rebuild behavior.
2>Performing Custom Build Tools
2>The system cannot find the file specified.
2>Performing Custom Build Tools
2>The system cannot find the file specified.
2>Moc'ing Job.h...
2>The system cannot find the path specified.
2>Moc'ing Management.h...
2>The system cannot find the path specified.
2>Moc'ing Worker.h...
2>The system cannot find the path specified.
2>Generate moc_predefs.h
2>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Microsoft.CppCommon.targets(171,5): error MSB6006: "cmd.exe" exited with code 2.
2>Done building project "autogtp.vcxproj" -- FAILED.
========== Build: 1 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========

isty2e:
Is #583 somehow relevant to the building failure?

gcp:
Going to close this for now. A new attempt is going to require quite some rebasing anyway.

Issue #135, opened 2017-11-22
When will resignation be enabled?


isty2e:
Apparently current LZ doesn't have any good idea about counting, and that has been the reason of disabling resignation for self-play games. A few questions regarding this:

1. When will be the point of allowing it to resign? LZ would probably remain weak for a long time, which also suggests that its winrate estimation will remain not that trustworthy. Even if it learns how to catch a stone, its prediction on winrates will not improve that much, I believe.
2. Will the AGZ-like approach, namely disallowing resignation for 10% of the games, be suitable at the early stage?
3. By allowing resignation, the individual games will be shorter in general, and there would be less board-filling games. Will it somehow affect the trained network?

gcp:
AGZ pulled statistics from the games first to set a suitable resignation threshold. It's possible to obtain the same during validation tests, and hence see if it makes sense to try resigning, but this needs some tooling to be written first.

isty2e:
If it makes sense, we probably should allow resignation as soon as possible. LZ is in its early stage, and refreshing games faster with updated networks would be a virtue. And it will presumably be trained less with nearly fully filled board, which might be desirable too.

gcp:
This assumes the bot is estimating reliable who is winning. That's far from obvious. And if it's not, it needs to see the game played out and realize that it's wrong.

isty2e:
Yeah, so that's why setting a proper resignation threshold depending on the network is important, if I understand correctly. Given the data of self-play games, how long will it take to estimate the resignation threshold for a trained NN?

gcp:
You can't, see above. It needs tooling during the validation test.

isty2e:
Oh well... I got it.

Ttl:
I wrote some Python code to analyze the resign rate from the completed self-play games played with the newest network. I loaded completed games to leelaz, ran it with one playout and recorded the NN eval value for each position.

![min_winner_nn_eval](https://user-images.githubusercontent.com/544240/33237777-b9340cba-d285-11e7-8ad8-952c8d0c270e.png)

The above plot is the minimum NN eval value for the winner during the game. If the resign rate is set above the winner's minimum rate then there won't be an incorrect resignation.

Some statistics I calculated:

```
Dataset size 145 games
Minimum observed evaluation for winner 0.00377
Minimum evaluation in game for winner mean=0.20014442069, std=0.0866228330411
Minimum evaluation in game for loser mean=0.0271061655172, std=0.0644117728729
[...]

PR #208, opened 2017-11-29
consider removing unused code in FullBoard, FastBoard, FastState, KoState

sethtroisi:
Most of this code seems to be for Leela to construct complex features which aren't needed by Leela-Zero

Maybe these have value if someone want's to do analysis (like https://github.com/gcp/leela-zero/issues/135) or for future use, but it also has cost (when people try to understand them, reformat code)...

I was waiting for my own analysis and it was a quick and easy PR even if it immediately gets closed working as intended

### DIFF
--- a/src/FastBoard.cpp
+++ b/src/FastBoard.cpp
@@ -105,44 +105,6 @@ void FastBoard::set_square(int x, int y, FastBoard::square_t content) {
     set_square(get_vertex(x, y), content);
 }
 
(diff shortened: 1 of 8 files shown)


gcp:
I think you deleted a tad too much because it doesn't compile any more.

sethtroisi:
I removed one tiny getter method to many :)

gcp:
You can probably construct a begin() and end() for that so it can be used with range-based for loops.

PR #236, opened 2017-12-03
Handle the situation when maximum tree size has been reached.

### LINKED ISSUE #230: leelaz on Windows exits after ~80k playouts
On my configuration (Windows 10 + GTX 1080), leelaz stops after around 80000 playouts. To reproduce:

```
 ./leelaz.exe -w 223737476718d58a4a5b0f317a1eeeb4b38f0c06af5ab65cb9d76d68d9abadb6

time_settings 0 3600 1
genmove b
```

Leela runs the search and after the line
```
Playouts: 84573, Win: 65.99%, PV: S15 C17 S13 N2
```
it exits, presumably crashing. Similar thing happens in other positions as well.

If you could generate a debug build, I could run it and attach the stack trace.

eterevsky:
resolve #230

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -44,6 +44,7 @@ using namespace Utils;
 bool cfg_allow_pondering;
 int cfg_num_threads;
 int cfg_max_playouts;
+int cfg_max_tree_size;
 int cfg_lagbuffer_cs;
 int cfg_resignpct;
 int cfg_noise;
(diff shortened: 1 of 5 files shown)


gcp:
Very nice but see my remarks in #230.

gcp:
I implemented this slightly differently. Feel free to do a new pull request if it turns out the other suggested approach was better.

eterevsky:
You left out the part that made the limit configurable. Can I add a PR to put it back in? It seems stupid not to use memory if you have it.

PR #237, opened 2017-12-03
Add command-line options to specify time control settings.

eterevsky:
This is useful in the case when the GTP UI can't set those settings via a GTP command.

### DIFF
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -35,6 +35,10 @@
 
 using namespace Utils;
 
+static int maintime = 3600;
+static int byotime = 0;
+static int byostones = 25;
+
 void license_blurb() {
     printf(
         "Leela Zero  Copyright (C) 2017  Gian-Carlo Pascutto\n"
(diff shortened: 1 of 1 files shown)


gcp:
This is a pretty hard "wontfix" from my side. Setting the time control is the responsibility of the GTP controller or GUI. Setting it in the engine **cannot** work correctly due to how the protocol is constructed, as I suspect you're well aware: there's no way to communicate the current clock time back to the GUI, only the other way around. It's also a bad idea to let the engines keep track of time for any GUI that wants to play engine matches - hopefully the reason why is obvious. If the GTP UI can't set the timecontrol correctly, it's much better to fix the GUI or develop better ones.

I realize that as a stopgap measure it can be useful to have this in the engine, but that just lets the idea permeate that this is an acceptable way of doing things. It is not! It is a completely broken idea, and not how GTP was designed to work - and GTP's design here is correct, for reasons mentioned in the first paragraph.

I would rather take a hard stance here to get this message across, which I think is important if we want to hope the quality of GTP UI's and GTP controllers (which is indeed pretty poor right now) ever improves. And it will need to, if the Go users ever want to have the kind of engine/UI interoperability that chess engines now enjoy.

roy7:
Since Sabaki is open source, perhaps someone could add some better time controls to it for everyone to benefit from. :)

eterevsky:
Makes sense. I was thinking of Sabaki, yes. :)

gcp:
In Sabaki you can set the timecontrol via giving a time_settings command in GTP shell. It's not very user friendly but messing with command line options isn't really all that better in that regard.

eterevsky:
> In Sabaki you can set the timecontrol via giving a time_settings command in GTP shell.

You can't if you're running a game between two engines.

PR #238, opened 2017-12-04
Change first-play-urgency.

killerducky:
Set first-play-urgency to parent's net_eval instead of 1.1f. The AGZ document isn't clear on what to do in this case, but it's reasonable to assume the children's value will be similar to the parent's. This allows the search to exploit killer moves earlier, such as responding to pass by also passing, and ignore very bad moves, such as passing when it will lose.

```
leelaz -d -w /home/aolsen/networks/585k.txt -p 1000 --noponder -r 0 -t 1

cmd: play b q16
cmd: play w d4
cmd: play b pass
cmd: genmove w

Before:
pass ->     641 (V: 100.00%) (N:  0.4%) PV: pass
  E6 ->       1 (V: 17.05%) (N:  0.3%) PV: E6 N2
 D17 ->       1 (V: 16.75%) (N:  0.2%) PV: D17 N18
  L4 ->       1 (V: 16.73%) (N:  0.3%) PV: L4 N2
  F3 ->       1 (V: 16.71%) (N:  0.3%) PV: F3 N2

After:
pass ->     741 (V: 100.00%) (N:  0.355%) PV: pass 
  F3 ->       1 (V: 16.89%) (N:  0.330%) PV: F3 B13
  C7 ->       1 (V: 16.83%) (N:  0.256%) PV: C7 N2
  R8 ->       1 (V: 16.67%) (N:  0.336%) PV: R8 S13
  M5 ->       1 (V: 16.65%) (N:  0.297%) PV: M5 S13
```

Running without "-t 1" makes the effect smaller, but still very significant.

I also tested setting first-play-urgency to 0. Some people on the computer-go mailing list suggested the AGZ document implies this. But I saw the search was far too narrow when using that.

ETA: I had a bug in my first commit, pushed a fix and updated the stats above. Not as big a difference as before but still good.


killerducky:
Some more experiments, the effect is larger when the number of playouts is smaller, such as when we reach a node further down in the tree that has a killer reply:

Before (using 1.1f):

playouts | pass picked | %
-----|----------|---
50      | 1     | 2%
100     | 1     | 1%
625     | 266   | 43%
1000    | 641   | 64%
1600    | 1241  | 78%   

After (using parent's net_eval):

playouts | pass picked | %
-----|----------|---
50    | 30     | 60%  
100   | 80     | 80%  
625   | 572    | 92%  
1000  | 741    | 74%  
1600  | 1246   | 78%

isty2e:
This will allow easier exploitation, but may be detrimental for exploration. How about running some test matches with it? And I don't think FPU=0 will be beneficial, by the way.

Also it is interesting that the pass percentage decreases from 92% to 78%... How does it evolve over playouts? Does it converge?

sethtroisi:
I'm testing this vs the existing code right now.
So far results are even (3 to 3) but I'll update in two hours when ~60 games have been played.

killerducky:
(I edited the table to add a few more points)

@sethtroisi thanks, I'm testing too but your machine is faster. So far black won every single game. It might be interesting to test with an older net that has less bias to one color. 

@isty2e Once all children have been visited once the results will be identical because the children take their own NN value instead of whatever first-play-urgency we use. The difference is in the beginning this way exploits the killer move before getting around to test all possible alternatives. Exactly when that happens depends on the distribution of the policy priors and the values. In this example you can see around 500 is the peak of early exploiting of pass, and it starts exploring more of the unvisited children. In this example it's a coincidence that that last unvisited child is somewhere around 1600, and that's why the tables match up from there on.

For large values the policy prior terms starts getting smaller and the value term dominates. In this case the value of pass is very high compared to all other choices so the search continues to exploit that.

isty2e:
@killerducky Yeah, but what I wanted to say is that especially for small number of playouts moves that aren't considered good in the parent node will be less explored, since there is always a tradeoff between exploitation and exploration. That can be considered good in general, but not sure if its strength will be improved when the NN is weak like in the current LZ. The test will let us know which is better.

gcp:
To avoid issues with weirdness of the current net you can use the supervised net in the README.

This is a good thing to investigate. It's missing from both AlphaGo papers and it's actually one of the things I've asked Aja about.

killerducky:
My results overnight with the 595k net. I did two tests, with and without the randomness options (-n -m 30), see results below. Since @sethtroisi is doing more testing I have stopped this and I'm retesting the resign threshold on the new 595k net.


```
leelaz-1000 v leelaz-1000-fpu (76/500 games)
board size: 19   komi: 7.5
                  wins              black          white     avg cpu
leelaz-1000         38 50.00%       38 100.00%     0 0.00%    182.17
leelaz-1000-fpu     38 50.00%       38 100.00%     0 0.00%    184.64
                                    76 100.00%     0 0.00%

leelaz-r-1000 v leelaz-r-1000-fpu (75/500 games)
board size: 19   komi: 7.5
                    wins              black         white     avg cpu
leelaz-r-1000         36 48.00%       35 92.11%     1 2.70%    167.73
leelaz-r-1000-fpu     39 52.00%       36 97.30%     3 7.89%    170.50
                                      71 94.67%     4 5.33%
```

I'm using gomill ringmaster, here's the ctl file:
```
competition_type = "playoff"

board_size = 19
komi = 7.5
leelaz = "/home/aolsen/projects/leela-zero-utils/leela-zero/src/leelaz"
leelaz_fpu = "/home/aolsen/projects/test_first_move_and_puct/leela-zero/src/leelaz"

players = {
    "leelaz-1000"       : Player("%s -g -q -d -r 0 -w /home/aolsen/networks/585k.txt -p 1000 --noponder" % leelaz),
    "leelaz-1000-fpu"   : Player("%s -g -q -d -r 0 -w /home/aolsen/networks/585k.txt -p 1000 --noponder" % leelaz_fpu),
    "leelaz-r-1000"     : Player("%s -g -q -d -r 0 -n -m 30 -w /home/aolsen/networks/585k.txt -p 1000 --noponder" % leelaz),
    "leelaz-r-1000-fpu" : Player("%s -g -q -d -r 0 -n -m 30 -w /home/aolsen/networks/585k.txt -p 1000 --noponder" % leelaz_fpu),
}

matchups = [
    Matchup("leelaz-1000"  , "leelaz-1000-fpu"  , alternating=True, number_of_games=500),
    Matchup("leelaz-r-1000", "leelaz-r-1000-fpu", alternating=True, number_of_games=500),
]

```

killerducky:
Just another thought on white's poor performance: The impact of using 1.1 for first play urgency is higher on the player who is losing, because net_evals will come back with less than 0.5, and the policy priors will have a more difficult time exploiting any promising move until all possible moves have been tried once.

By contrast, the winning side will have net_evals greater than 0.5. Since these will be closer to 1.1 it's easier for the policy priors to convince search to exploit promising moves earlier.

This could create a feedback loop where once the network gets sour on white, white starts to play worse. Also the shape of the policy prior for white will get flatter, which exacerbates the problem even more, etc, etc.

Alderi-Tokori:
I thought the paper said they initialized the value at 0 (rather than 1.1)? Regardless, I think it's a good idea to carry on the parent's value.

gcp:
It might be worthwhile to try 0 and re-tune the PUCT parameter. The AGZ paper indeed says initial Q_s_a is 0. The old paper didn't mention it.

This is an unintended difference, and the impact could be rather big because it'll make the tree search more selective on the priors.

isty2e:
Oh, there is the sentence here:

>The leaf node is expanded and each edge (s_L, a) is initialized to {N(s_L, a) = 0, W(s_L, a) = 0, Q(s_L, a) = 0, P(s_L, a) = p_a}.

killerducky:
The AGZ paper is ambiguous because Select (Figure 2a) happens before Expand and evaluate (Figure 2b). So during the Select step the leaf node has not been expanded, and Q(s_L, a) does not exist. 

Also note they initialize N(s_L,a)=0, and Q(s, a) is not the same term as Q(s_L, a). The equation for Q(s, a) is given on page 5 and it includes a 1/N(s,a) term, which gives a divide by zero error.

Imagine a position where it's white's move and the previous node had a net_eval=0.2. The policy priors are uniform, so no move is preferred. You select the first move and it gives a net_eval=0.1. Intuitively we should be disappointed in that result and prefer another move. But if you initialize first play urgency to 0, the algorithm will think a value of 0.1 is great and try to exploit it.

It's the same but reverse for first play urgency 1.1. Again suppose the previous node had net_eval=0.2. If we find a move that has a net_eval 0.3 we should be happy and try to exploit it. But instead the algorithm is tuned to be disappointed with anything less than 1.1 and try to explore more.

The Q term should use the best approximation for the expected return of exploiting a particular move. The puct term should be the one used to tune the balance between explore and exploit. The best prior we have for Q of an unexpanded node is the parent's net_eval.

gcp:
>given on page 5 includes a 1/N(s,a) term, which gives a divide by zero error

The complete formula they give later on gives it as .../1+N(s,a) so that's not actually an issue.

killerducky:
> The complete formula they give later on gives it as .../(1+N(s,a)) so that's not actually an issue.

Where do you see that? I see on page 26 Select Figure 2a has a formula for U(s, a) that includes that term. But I don't see anything for Q(s,a) other than the one on page 5 with the 1/N(s,a). 

Anyways I think we agree the paper is not precise on this point.

isty2e:
Yes, Q and U are divided by N and 1+N, respectively in both papers, and for Q I believe dividing by N is the only sensible choice.

gcp:
Yeah for the average it indeed makes no other sense. Seems like this part of AGZ is still "undefined" then.

isty2e:
I also feel uncertain when I read the following sentence:

>Each node s in the search tree contains edges (s, a) for all legal actions a ∈ A(s).

Could this actually mean that FPU=0 since the list of legal actions are restricted?

Alderi-Tokori:
I don't see Q(s,a) being divided by N(s,a) anywhere in the paper. They define Q(s,a) as the mean action value so Q(s,a) = W(s,a) / N(s,a), but at first, when N(s,a) is still 0, then it's the initialization phase where Q(s,a) = 0.

Are we looking at different versions of the same paper?

isty2e:
@Alderi-Tokori I think we're talking about the same thing but some miscommunication is involved here... The division by N just implies that Q(s, a) = 1/N(s, a) ∑_{s′|s, a→s′} V(s′) and not anything else, so I assume we're on the same line?

Alderi-Tokori:
@isty2e Ah I see, good catch.

killerducky:
I chatted with @Alderi-Tokori on OGS, now I agree with him that the AGZ paper uses initial value for Q(s,a)=0. My confusion was because I was thinking only of nodes, as the LeelaZ code is implemented. But the AGZ paper defines nodes and edges.

Search Algorithm, Expand and evaluate:

> The leaf node sL is added to a queue for NN evaluation... The leaf node is expanded and each edge (sL,a) is initialised to {N=0,W=0,Q=0,P=0}. 

So every leaf node already has 362 edges initialised with Q=0. I was stuck thinking the Q value would be one per node, not 362 edges with one node.

Earlier in the paper, it talks about what happens after the NN evaluation is done:

> Each edge (s,a) traversed in the simulation is updated to increment its visit count N(s,a), and to update it's action-value to the mean evalution over these simulations, Q(s,a)=1/N(s,a)*sum(V(s')).

killerducky:
@sethtroisi do you have updated test results? Could you maybe try running another test changing the first play urgency to 0?

Maybe we should try putting in the current version for now and do more testing to pick between this and using 0? I'm worried that continuing with 1.1 will continue to make the loser (white) play worse. Or perhaps I should just learn to be more patient.  :)

sethtroisi:
@killerducky I'll try to run 100 games of each of these configs
0 vs 1.1 and 0 vs 0.5

Results in ~8 hours

gcp:
I am doing a tuning run for cfg_puct with Q(s, a) = 0 and surprisingly small values like 0.8 may be good.

gcp:
I'm running with 1600 playouts, the supervised weights file, FPU=0 and PUCT=0.95, no randomness.

It's a complete blowout so far:

```
leelaz-fpu v leelaz-0 (10/1000 games)
board size: 19   komi: 7.5
             wins               black         white       avg cpu
leelaz-fpu      0   0.00%       0   0.00%     0   0.00%    814.52
leelaz-0       10 100.00%       4 100.00%     6 100.00%    825.49
                                4  40.00%     6  60.00%
```

I am not surprised at all either. I observed earlier that that not investigating moves with a bad policy prior until a lot of playouts have happened makes the program stronger: https://github.com/gcp/leela-zero/blob/b07992672f78840f0ff3265ed7f930b8f970b563/src/UCTNode.cpp#L378

It didn't occur to me that this was how the AGZ algorithm was supposed to behave.

The difference is probably a lot less on our current networks because the policy networks aren't as good. But I am seeing 8-9% prediction rate already, so it probably isn't nothing either. I'll run that test, too.

sethtroisi:
FPU-zero win percentage is >78% with 95% confidence[1] that's **220 elo**!

More speculation:
95% confidence with 11-1 implies >66% win rate equals >110 elo
95% confidence with 12-0 implies >78% win rate equals >220 elo
95% confidence with 15-0 would be 260 elo
95% confidence with 32-0 would be 400 elo

[1] If win changes was <78% then there'd be less than a 5% change of this happening
(0.78)^12 * (1-0.78)^0 = 0.0507
[2] (.66^11 * (1-.66)^1 * 12) + .66^12 = 0.049

sethtroisi:
I'm running with 1000 playouts, the supervised weights file, no policy noise randomness, and no resigning.

"-fpu-one-point-one" is default (FPU=1.1 and PUCT=2.8)
"-fpu-zero-puct-point-eight" is FPU=0 and PUCT=0.8

isty2e:
@sethtroisi Isn't gcp running it with PUCT=0.95, instead of 0.8?

sethtroisi:
@isty2e, yes I'll try that after I let this run for 10 games. then I'll try FPU=0.5

gcp:
I also ran a match with the current 6f weights file, and it went completely the other way, i.e. FPU=1.1 won 55-0 against FPU=0.

This is a bit problematic.

gcp:
PUCT=0.8 vs 0.95 make no big difference. My testing shows it's close to optimal in the 0.5...1.5 range or so.

isty2e:
How about just making the change and proceeding? There can be some types of networks that fits more to FPU=0 and FPU=1.1, though not sure about it...

gcp:
Rebasing the next branch messed up this pull, but I'm fixing that right now. I think the code is buggy:

```
UCTNode::get_eval(int tomove)

return m_init_eval;
```

As the eval is stored from black's point of view, it has to be mirrored if tomove is white.

killerducky:
> As the eval is stored from black's point of view, it has to be mirrored if tomove is white.

Oh, that's the real fix for the problem I saw. But I think you need to undo my incorrect fix:
https://github.com/gcp/leela-zero/pull/238/commits/675a7796dfd5dae36dbaf9bdd5ae2337ecf8aac5

isty2e:
I have thought about it, but if I were to decide, I'll probably set FPU=0. Though it might not be beneficial at the moment, if it can be better later, we can probably bear with it.

gcp:
Weak networks

```
leelaz-fpu v leelaz-par (154/1000 games)
board size: 19   komi: 7.5
leelaz-fpu     42 27.27%       39  50.65%     3   3.90%    445.73
leelaz-par    112 72.73%       74  96.10%     38 49.35%    444.69
                               113 73.38%     41 26.62%
```

Good networks

```
leelaz-fpu v leelaz-par (15/1000 games)
board size: 19   komi: 7.5
             wins               black         white       avg cpu
leelaz-fpu      0   0.00%       0   0.00%     0   0.00%    796.49
leelaz-par     15 100.00%       7 100.00%     8 100.00%    790.25
                                7  46.67%     8  53.33%

```

Using the parent value is the clear winner.

gcp:
Merged in 87c658f23ef516a474aeb1b82533d24f80823271.

luweiagi:
> Just another thought on white's poor performance: The impact of using 1.1 for first play urgency is higher on the player who is losing, because net_evals will come back with less than 0.5, and the policy priors will have a more difficult time exploiting any promising move until all possible moves have been tried once.
> 
> By contrast, the winning side will have net_evals greater than 0.5. Since these will be closer to 1.1 it's easier for the policy priors to convince search to exploit promising moves earlier.
> 
> This could create a feedback loop where once the network gets sour on white, white starts to play worse. Also the shape of the policy prior for white will get flatter, which exacerbates the problem even more, etc, etc.

I think what you said is wrong, no matter loss < 0.5 or win > 0.5, both of them would pass through all the child nodes first before go deep for exploitation, when fpu is set to 1.1, because whether 0.1 or 0.9 both < 1.1, they are no difference for 1.1.

PR #241, opened 2017-12-04
Add optional iterations parameter to netbench GTP command

sethtroisi:
the 1600 default is nice for understanding how long it might take for one move under normal circumstances.

But sometimes I want to run more (or less) when testing on different platforms (GTX 1080, my old 970, and historic 560) and when profiling test code. it would be nice not to have to modify Network.cpp as often as I find myself doing :)

-----


sethtroisi:
That was fast :)

PR #243, opened 2017-12-04
Fixed some trailing whitespace, indenting, and formatting

sethtroisi:
Looks like some of this was fixed in 

https://github.com/gcp/leela-zero/commit/1dbad5bb4ed18e37770de901c90cfdd57c15201d


gcp:
I got these in the followups.

PR #253, opened 2017-12-05
OS auto detection for leelaz compile

bood:
So you don't have to do extra modifications when compiling on MacOS

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,8 +93,6 @@ https://github.com/KhronosGroup/OpenCL-Headers/tree/master/opencl22/)
     git clone https://github.com/gcp/leela-zero
     cd leela-zero/src
     brew install boost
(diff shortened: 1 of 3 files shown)


gcp:
Seems fine but can you actually get a working Windows version out of the Makefile? If everyone uses the MSVC project files then we don't need that part.

bood:
Makefile on windows is tricky. MSVC has some support thru NMAKE but making a single makefile working with both GNU make and NMAKE could be not that pleasant.

Better approach is probably use `CMake` (noticing there are already some cmake files in the repo, deprecated?). I can look into that some time later.

gcp:
Then I'd suggest stripping the OS_WIN part out and I'll pull.

bood:
Fair enough. Just removed.

PR #257, opened 2017-12-05
small cleanups

sethtroisi:


autogtp/Production.cpp
```diff
@@ -117,26 +118,25 @@ Production::Production(const int gpus,
 }
 
 bool Production::updateNetwork() {
-    auto retries = 0;
```
sethtroisi: I'm trying to learn what c++ styles you like and I didn't know here,
I prefer for loops when the loop count is constant (baring early return / break).
I'll revert this if you don't agree (and stick with the do while that you seem to prefer in many places)

autogtp/Production.cpp
```diff
@@ -21,6 +21,7 @@
 #include <QDir>
 #include <QFileInfo>
 #include <QThread>
+#include <QMutex>
```
sethtroisi: I included QMutex in each file that uses the name (which seems to be what you do) instead of relying on <file>.h or worse some other class to include it.

gcp: Relying on the .h that matches the .cpp is fine and probably better style.

autogtp/Production.cpp
gcp: I don't really care.

(Note that I didn't wrote most of this code :-)

autogtp/Production.cpp
sethtroisi: updated to have rely on includes in .h

sethtroisi:
updated readers

PR #261, opened 2017-12-05
autogtp: write result on resignation

marcocalignano:


roy7:
I believe the standard for SGF score result for resigns is B+R or B+Resign, you should use that please.

PR #262, opened 2017-12-05
autogtp: write result on resignation

marcocalignano:


roy7:
Should the trailing space be there after the Resign scores?

Issue #265, opened 2017-12-06
Proposal: Disable pass for 10% of the game


ljn917:
This is inspired by the AlphaGo Zero paper, in which resignation was disabled for 10% of the game. Many people may argue that the net should eventually learn white cannot pass just hoping black also passes, but disabling pass can accelerate learning.

isty2e:
It is essentially putting human knowledge in it, that passing is bad in general. Moreover, without pass the game should end with resign or reaching maximum number of moves (723), which might hinder fast game generation.

ljn917:
@isty2e I agree that it is putting human knowledge. I am running a local version that does not allow passing before 300 moves. Considering it is only 10%, it won't be that bad, and we can disable it after white winning rate increases.

Currently, my doubt is whether white have a chance to win or not with no pass, because the net is completely on the black side.

isty2e:
@ljn917 Regarding the black winrate issue, as I mentioned in other issues, it might be an issue with multithreaded MCTS, which isn't as strong as intended (single threaded) version. I really wish I could test it by myself...

ljn917:
@isty2e If it is MCTS issue, why does it have a bias on one color? If you are right, I think it means the multithread MCTS has a bug. It would be better disable it for now.

isty2e:
@ljn917 I'm not sure why, and in fact no one has tested it yet unfortunately so we don't know if that is even true. Anyway, it is not a bug, since when multithreaded MCTS becomes always weaker given a fixed number of playouts. Nonetheless, I do think that it is better to disable multithreading.

ljn917:
I  would like to withdraw my doubt because I just saw white won after playing through the end.

ljn917:
@isty2e I think there were some testing in other discussions, but I don't have a good understanding why it is weaker or has a bias. I have not tried to change the num of threads either.

Usually, people expect that multithreaded program should behave exactly like single threaded program, except the differences in speed. I don't know what is the issue here.

earthengine:
This is like what I proposed in another thread. I suggested that passing early should be considered resign in the rule perspective, so it only affects the results and causes a immediate end of the game. The benifit of this is this "knowledge" is in the rule level, and can be considered an undocumented rule already applied to human games (think of you are playing with your friends and the game is just beginning and you have something else to do. Then you just tell your friend you are not going to take the next move and the game ends).

isty2e:
[...]

PR #266, opened 2017-12-06
nopass 10%

ljn917:
10% games will be played without passing before 300 moves as proposed in #265. `--nopassbefore N` is added to Leela Zero to control passing. This argument only changes the behavior of genmove.

I hope that no pass will generate more white wins and reduce the bias towards black in the training dataset.

As @isty2e pointed out, it is essentially adding human knowledge. I hope that we can remove this from training later once the net is balanced.

Currently, I only run 13 self-played games and 2 of them are white wins after played through end game. I don't know whether this is enough for correcting the imbalance. If no, the percentage (10%) may be increased.


gcp:
Won't merge this. AGZ doesn't talk about this heuristic, and I'm not going to add random ones to this effort.

PR #283, opened 2017-12-08
New client that receive task from server.

marcocalignano:
Most important point is:
Autogtp cannot be used for local validation anymore so the option -c and -n are gone.
For that task there is a new application called Validation in the directory validation. 

Since on my computer it take day to test it till the end I ask you to test on your computer and see if it gives the same output and the same result. 

For all contributor that improve autogtp in the last days: I have tried to do the rebase the best I could, but some of ther feature you introduced may be gone, please check and if you say they are gone just let me know. 
Note all the improvement regarding SPRT and results output for the competition mode are in the new application 'validation' not in autogtp anymore.


validation/moc_Validation.cpp
```diff
@@ -0,0 +1,224 @@
+/****************************************************************************
+** Meta object code from reading C++ file 'Validation.h'
```
gcp: This shouldn't be checked in.

autogtp/Game.cpp
```diff
@@ -34,7 +34,7 @@ Game::Game(const QString& weights, const QString& opt) :
 #endif
     m_cmdLine.append(opt);
     m_cmdLine.append(weights);
-    m_cmdLine.append(" -p 1600 --noponder");
+   // m_cmdLine.append(" -p 1600 --noponder");
```
gcp: Remove this.

autogtp/Job.h
```diff
@@ -0,0 +1,73 @@
...
+        FINISHING
+    };
+    enum {
+        Production = 0,
+        Validation
+    };
+    Job(QString gpu);
+    ~Job() = default;
+    virtual Result execute() = 0;
+    virtual void init(const QStringList &l) { m_option = l[0] + m_gpu + " -g -q -d -w "; }
```
gcp: Maybe add -t 1 here too?

isty2e:
The "dumbpass" field is false by default, as commented [here](https://github.com/gcp/leela-zero/issues/249#issuecomment-350133627). Is it okay?

gcp:
No, "dumpass" should be true.

isty2e:
I didn't mention the others, because except for threads and dumbpass the options for self-play games will be determined by the server anyways, I suppose.

autogtp/Job.h
marcocalignano: -t 1 is added from the json parser if the json option 'threads' is not found

marcocalignano:
Ok, good to go?

gcp:
I need to push one more bugfix release out of the door and then we'll pull this.

marcocalignano:
The leela version is now in Management.cpp

Ttl:
I'm unable to build validation ("Validation.cpp:25:30: error: ‘min_leelaz_version’ was not declared in this scope").

marcocalignano:
@ttl will be fixed with #302

Issue #282, opened 2017-12-07
Possible bug in UCTNode::dirichlet_noise ?


amirlb:
I was reading the code for fun, and saw something strange: the child_cnt variable is not incremented during the last loop in the method (lines 229-235 in https://github.com/gcp/leela-zero/blob/master/src/UCTNode.cpp).

If I'm reading it correctly, this would make the noise not take Dirichlet form but be uniform among all the actions, with the uniform constant between 0 and 0.25, determined by a one-variable Dirichlet distribution.

Is this OK?

isty2e:
It seems it is not uniform, because of child = child->m_nextsibling; is there, if I understand correctly. I had it printed and it was not, perhaps someone can confirm this.

killerducky:
I think you're right. I'm working on something else closely related right now so I will try fixing this as well.

sethtroisi:
I agree it looks like you just need to add 
```  child_cnt++;``` on 235

isty2e:
Oh wait, I was so sleepy that I was not looking into it carefully and was printing score instead of eta_a. Yes you're right, and this can be problematic indeed.

isty2e:
Oh you're right. In my previous (deleted) comment I was somehow printing score instead of eta_a for debugging.

gcp:
This warrants another release.

PR #287, opened 2017-12-08
Fix bug in dirichlet noise, re-normalize legal move policy priors.

killerducky:
Dirichlet noise was only using the first value of the vector.
Policy priors should be re-normalized after removing illegal moves
so that they are on the same scale as dirichlet noise.

Fixes https://github.com/gcp/leela-zero/issues/282

I also have a debug version of the same thing here that prints more information.
https://github.com/killerducky/leela-zero/commit/c5801d3263bb45a8bbdaa0721eb59adc5d50800d

I took a position on move 293, the legal moves add up to 0.92 so the re-normalization isn't too bad there, but it might be worse in other positions, leading to the dirichlet noise overwhelming the policy priors.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -112,17 +112,26 @@ bool UCTNode::create_children(std::atomic<int> & nodecount,
     FastBoard & board = state.board;
     std::vector<Network::scored_node> nodelist;
 
+    auto legal_sum = 0.0f;
     for (auto& node : raw_netlist.first) {
         auto vertex = node.second;
         if (vertex != FastBoard::PASS) {
             if (vertex != state.m_komove
                 && !board.is_suicide(vertex, board.get_to_move())) {
                 nodelist.emplace_back(node);
+                legal_sum += node.first;
             }
         } else {
             nodelist.emplace_back(node);
+            legal_sum += node.first;
         }
     }
+
+    // re-normalize after removing illegal moves.
+    for (auto& node : nodelist) {
+        node.first /= legal_sum;
+    }
+
     link_nodelist(nodecount, nodelist, net_eval);
 
     return true;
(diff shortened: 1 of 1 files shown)


bood:
In general, what is the impact of this bug? Making the first 30 moves more random than we wanted to? If I understood correctly.

killerducky:
You mentioned 30 moves, that is different. The dirichlet noise applies for the entire game. It's supposed to randomly give a bonus to some of the moves' policy prior. In this way some moves that normally wouldn't be prioritized get looked at earlier in the search. But instead the same bonus was given to all moves, which is similar to not having the noise at all. The overall effect is we might not find new tactics if the Net is already convinced they won't work.

The re-normalize part ensures the noise doesn't overwhelm the policy priors in cases where the network put a lot of the policy prior into illegal moves. The dirichlet noise is supposed to be mixed in with the policy priors at a 25/75 ratio. But that ratio won't hold without re-normalization.

killerducky:
BTW I got the idea for re-normalizing legal moves from the latest AlphaZero paper.

isty2e:
I am asking this because I do not have a thorough knowledge about the whole code, but is there any possibility that some quantities other than the noise can be summed to the normalized scores? I think it is hardly the case considering the MCTS algorithm, but if there is a possibility it is safer to de-normalize scores back?

src/UCTNode.cpp
```diff
@@ -112,17 +112,26 @@ bool UCTNode::create_children(std::atomic<int> & nodecount,
...
             }
         } else {
             nodelist.emplace_back(node);
+            legal_sum += node.first;
         }
     }
+
+    // re-normalize after removing illegal moves.
+    for (auto& node : nodelist) {
+        node.first /= legal_sum;
```
gcp: In theory this shouldn't be zero as this loop won't run if there are no legal moves, but if the policy prior value for a move underflows, you might still end up dividing by zero. This bit us in the dirichlet noise generation. So I'd add the same guard.

killerducky: Fixed.

amirlb:
This is starting to get really insignificant, but I think the probabilities won't be normalized in superko situations

PR #296, opened 2017-12-08
Avoided creation of extra cl::buffer and vectors

sethtroisi:
cl::buffers only does a swallow copy during copy constructor so no performance impact but I think code is cleaner.


gcp:
Not sure I the const_iterator passing, and then doing array style indexing into that (which you can do because it happens to be a vector) is any cleaner...

The emplace_back thing at the start is OK.

sethtroisi:
Updated for comment

#### Performance results
**Before**
Example netbench: Leela:  1600 evaluations in  1.64 seconds -> 975 n/s
mean 954n/s from 10 runs, stddev = 18.25

**After**
Example netbench Leela: Leela:  1600 evaluations in  1.53 seconds -> 1045 n/s
mean 1049n/s from 10 runs, stddev = 14.93

----

alias testOne='echo "netbench" | ./leelaz -w ../weights.txt |& grep "evaluations"'
results() { cat results_$1.txt | cut -d' ' -f 10 > results_$1.csv; cat results_$1.csv; R -q -e "x <- read.csv('results_$1.csv', header = F); summary(x); sd(x[ , 1])"; }
testMany() { make clean && make -j4 && for i in $(seq 1 $1); do testOne >> results_$2.txt; done && results $2; }

sethtroisi:
I changed the iterators per your comment, would be nice if you could take a 2nd look

PR #301, opened 2017-12-08
Remove Linked List from UCTNode

sethtroisi:
Vector replacement is much cleaner.

Thanks for adding seed! I had a much uglier solution to test :)

Tested with:

testGenmove() { echo "genmove black" | ./leelaz -s 123 -t 1 -w ../weights.txt -p 20000 --noponder &> $1; }

cd src

git checkout pull301
make clean && make -j8 && testGenmove new_data
git checkout next
make clean && make -j8 && testGenmove next_data 

wc next_data new_data 
diff  next_data new_data


sethtroisi:
This is a late night attempt to help out with [[1]](https://github.com/gcp/leela-zero/issues/230#issuecomment-350044519) and [[2]](https://github.com/gcp/leela-zero/issues/230#issuecomment-348905664)

sethtroisi:
echo -e "time_settings 0 1 0\ngenmove black" | /usr/bin/time -v ./leelaz -t 1 -p 100000 --noponder -w ../autogtp/4d9a5cfc39babcb3d02e399ab59aee65d02162bc371ab793645deca80363971a 

before:	Maximum resident set size (kbytes): 2898492 / 10000 ~= 29 kilobytes / playout
after: 	Maximum resident set size (kbytes): 3306208 / 10000 ~= 33 kilobytes / playout

so even though we don't store two pointers (m_firstchild, m_nextsibling, the overhead from unreserved capacity in vectors is probably bad)

src/UCTNode.cpp
```diff
@@ -150,41 +141,35 @@ void UCTNode::link_nodelist(std::atomic<int> & nodecount,
...
 
         if (move != FastBoard::PASS) {
             KoState mystate = state;
             mystate.play_move(move);
 
             if (mystate.superko()) {
-                UCTNode * tmp = child->m_nextsibling;
-                delete_child(child);
-                child = tmp;
+                childIter = m_children.erase(childIter);
```
gcp: Classic iterator invalidation bug here :-)

All existing end() are no longer valid after this point.

src/UCTNode.cpp
```diff
@@ -268,28 +241,10 @@ void UCTNode::randomize_first_proportionally() {
...
-            m_firstchild = child->m_nextsibling;
-            m_firstchild->m_nextsibling = old_first;
-            // Point through our nextsibling ptr
-            child->m_nextsibling = old_next;
-            return;
-        }
-        child_cnt++;
-        child = child->m_nextsibling;
-    }
+    std::iter_swap(begin(m_children), begin(m_children) + index);
```
gcp: This got a bit simpler indeed.

src/UCTNode.cpp
```diff
@@ -473,16 +407,18 @@ void UCTNode::sort_root_children(int color) {
...
     }
 
-    // reverse sort, because list reconstruction is backwards
     std::stable_sort(rbegin(tmp), rend(tmp), NodeComp());
+    std::reverse(begin(tmp), end(tmp));
 
-    m_firstchild = nullptr;
+    // I didn't get the same result with
+    // std::stable_sort(begin(tmp), end(tmp), NodeComp());
+    // maybe because of stable?
```
gcp: It's not stable if the sort order changes :-)

src/UCTNode.cpp
```diff
@@ -473,16 +407,18 @@ void UCTNode::sort_root_children(int color) {
...
 
-    for (auto& sortnode : tmp) {
-        link_child(std::get<3>(sortnode));
+    m_children.clear();
+    for (const auto& sortnode : tmp) {
```
gcp: Isn't it possible to replace all of this by a single sort call with a custom node comparator now? The deconstruction was done because sorting a linked list is tricky. But now it's a vector.

src/UCTNode.cpp
```diff
@@ -501,33 +437,32 @@ UCTNode::sortnode_t get_sortnode(int color, UCTNode* child) {
...
 UCTNode* UCTNode::get_best_root_child(int color) {
     LOCK(get_mutex(), lock);
-    assert(m_firstchild != nullptr);
+    assert(!m_children.empty());
 
     NodeComp compare;
-    auto child = m_firstchild;
-    auto best_child = get_sortnode(color, child);
-    while (child != nullptr) {
+    auto best_child = get_sortnode(color, m_children[0]);
```
gcp: m_children.front()

src/UCTNode.cpp
```diff
@@ -501,33 +437,32 @@ UCTNode::sortnode_t get_sortnode(int color, UCTNode* child) {
...
     }
     return std::get<3>(best_child);
 }
 
 UCTNode* UCTNode::get_first_child() const {
-    return m_firstchild;
+    if (m_children.empty()) {
+        return nullptr;
+    }
+    return m_children.front();
```
gcp: Bit surprised this doesn't need a &.

src/UCTNode.cpp
```diff
@@ -501,33 +437,32 @@ UCTNode::sortnode_t get_sortnode(int color, UCTNode* child) {
...
 }
 
-UCTNode* UCTNode::get_sibling() const {
-    return m_nextsibling;
+const std::vector<UCTNode*> UCTNode::get_children() const {
```
gcp: This should probably return a reference, not copy the vector.

src/UCTSearch.cpp
```diff
@@ -113,16 +113,14 @@ void UCTSearch::dump_stats(KoState & state, UCTNode & parent) {
     // sort children, put best move on top
     m_root.sort_root_children(color);
 
-    UCTNode * bestnode = parent.get_first_child();
 
-    if (bestnode->first_visit()) {
+    auto children = parent.get_children();
+    if (children[0]->first_visit()) {
```
gcp: .front()

src/UCTNode.h
```diff
@@ -83,8 +83,8 @@ class UCTNode {
 
     // Tree data
     std::atomic<bool> m_has_children{false};
-    UCTNode* m_firstchild{nullptr};
-    UCTNode* m_nextsibling{nullptr};
+    std::vector<UCTNode*> m_children;
```
gcp: So, one reason to use a vector is to get rid of the pointers. Unfortunately, due to the use of atomics, we can't move the nodes into the container (boo!). So we have to use pointers. But we ARE able to use std::unique_ptr inside the vector.

And using them would solve the gigantic memory leak that I think you have right now. You're not freeing the nodes, are you?

And actually, if we can avoid having to sort the nodes at any point, then you could actually store the nodes directly in the vectors. You removed one of those sorts already. Maybe it's possible to get rid of the second one too.

src/UCTNode.cpp
oni-link: Erased element not deleted.

gcp: That too but that's a general issue with this code. It either has to move the nodes into the vector or use unique_ptrs.

oni-link: `std::stable_partition()` could be used to reorder the nodes in the vector:
```c++
    auto iter = std::stable_partition(begin(m_children), end(m_children),
                                      [&state](const auto &child) {
                                          int move = child->get_move();

                                          if (move != FastBoard::PASS) {
                                              KoState mystate = state;
                                              mystate.play_move(move);
                                              return !mystate.superko();
                                          }
                                          return true;
                                      });
```
nodes that need to be removed are in the range `[iter,end(m_children))`.

src/UCTNode.h
sethtroisi: I got rid of all references but emplace_back still require the move constructor :(

src/UCTNode.cpp
sethtroisi: Yes, I've been trying to do the minimum change possible in my pull requests (so that I don't get distracted by fixing random not important things and to make it easier for you to verify the effect of the change).

src/UCTNode.cpp
```diff
@@ -412,133 +349,79 @@ UCTNode* UCTNode::uct_select_child(int color) {
...
-UCTNode::sortnode_t get_sortnode(int color, UCTNode* child) {
-    auto visits = child->get_visits();
-    return UCTNode::sortnode_t(
-        visits == 0 ? 0.0f : child->get_eval(color),
-        visits,
-        child->get_score(),
-        child);
+    std::stable_sort(begin(m_children), end(m_children),
+                     NodeComp(color));
+    std::reverse(begin(m_children), end(m_children));
```
sethtroisi: This can be verified by the command in comment number two

sethtroisi:
Fixed and ready for another look.

src/UCTNode.cpp
```diff
@@ -412,133 +349,75 @@ UCTNode* UCTNode::uct_select_child(int color) {
...
-
-        child = child->m_nextsibling;
-        // Make sure we are at a valid successor.
-        while (child != nullptr && !child->valid()) {
-            child = child->m_nextsibling;
+            best = child.get();
         }
     }
 
     assert(best != nullptr);
```
oni-link: The old code allowed to return `nullptr` when all children were invalid.
In debug mode we would now assert.

sethtroisi: IMHO best can never be nullptr, there will always be one valid child (pass).

as a smoke test I've also tested a couple of games in debug mode without any asserts being thrown

sethtroisi: With a 2nd look I'm confident this behavior hasn't changed, the assert was there before. the loop accesses evaluates the same children

src/UCTNode.cpp
oni-link: I think @gcp mentioned it before, but why not reverse the sort order in `NodeComp` so that `std::reverse` could be removed?

src/UCTNode.cpp
```diff
@@ -412,133 +349,75 @@ UCTNode* UCTNode::uct_select_child(int color) {
...
-
-    bool operator()(const UCTNode::sortnode_t a, const UCTNode::sortnode_t b) {
-        // One node has visits, the other does not
-        if (!std::get<1>(a) && std::get<1>(b)) {
-            return false;
-        }
-
-        if (!std::get<1>(b) && std::get<1>(a)) {
-            return true;
+    NodeComp(int color) : m_color(color) {};
```
oni-link: Should an `explicit` constructor be used to prevent implicit conversion?

src/UCTNode.h
```diff
@@ -21,17 +21,20 @@
...
 
 #include "SMP.h"
 #include "GameState.h"
 #include "Network.h"
 
 class UCTNode {
 public:
-    using sortnode_t = std::tuple<float, int, float, UCTNode*>;
+    // Should this have a pointer?
+    using sortnode_t = std::tuple<float, int, float, std::unique_ptr<UCTNode>>;
```
oni-link: Could be removed, is not used anymore.

src/UCTNode.cpp
```diff
@@ -150,41 +137,34 @@ void UCTNode::link_nodelist(std::atomic<int> & nodecount,
     }
 
     // sort (this will reverse scores, but linking is backwards too)
```
oni-link: Comment could be updated. No more linking as in linked list.

src/UCTNode.cpp
```diff
@@ -150,41 +137,34 @@ void UCTNode::link_nodelist(std::atomic<int> & nodecount,
...
 
             if (mystate.superko()) {
-                UCTNode * tmp = child->m_nextsibling;
-                delete_child(child);
-                child = tmp;
+                // TODO delete comment before merge.
+                // Instead of removing invalid node just mark node invalid.
+                // This logic is already used in UCTSearch.cpp, I want to
+                // revisit ko moves next so it can be cleaned up more then.
+                child->invalidate();
```
oni-link: Could not removing a child make problems if anyone depends on a check with `has_children()`?

src/UCTNode.cpp
```diff
@@ -150,41 +137,34 @@ void UCTNode::link_nodelist(std::atomic<int> & nodecount,
...
                 continue;
```
oni-link: `continue;` could be removed.

src/UCTNode.cpp
```diff
@@ -268,28 +236,10 @@ void UCTNode::randomize_first_proportionally() {
         return;
     }
 
+    assert(m_children.size() >= index);
```
oni-link: This test never fails.
`m_children` has always the same size as `accum_vector`. So `index` is always a valid index into `m_children`.

src/UCTNode.cpp
sethtroisi: It doesn't sort the same because of stable.

2 1a 1b 5 3
sort_stable => 1a 1b 2 3 5
reverse => 5 3 2 1b 1a

sort_stable (reverse compare) => 5 3 2 1a 1b

all the elements that are equal will be in the wrong order.

I thought that by using rbegin and rend I could fix this but I haven't made it work

src/UCTNode.cpp
sethtroisi: has_children instead looks at m_has_children (see L~315) which measures if the expand happened.
plus there will ALWAYS be at least one valid child (the pass child), except in get_no_pass_child (which returns a potentially null pointer to UCTNode

src/UCTNode.h
sethtroisi: Thanks for this (and the other comments), I do appreciate a 2nd pair of fresh eyes looking at the change :)

src/UCTNode.cpp
sethtroisi: updated

src/UCTNode.cpp
sethtroisi: updated

src/UCTNode.cpp
sethtroisi: I generally use asserts to test my assumptions by adding statements I believe will always to be true if the code is working but which might get broken from bugs, changing code. If you feel strongly I can remove the assert.

src/UCTNode.cpp
sethtroisi: My understanding from http://en.cppreference.com/w/cpp/language/explicit
is that adding explicit will prevent copy initialization (which would be fine in this case).

but I defer to a c++ expert (you or gcp)

sethtroisi:
This is ready for review.

I re-ran the two analysis from above and a new one 
`$ echo "auto" | ./leelaz -s 123 -t 1 -p 500 --noponder -w ../autogtp/4d9a5cfc39babcb3d02e399ab59aee65d02162bc371ab793645deca80363971a  |& tee next_game`

src/UCTNode.cpp
gcp: I like the use of asserts because it document assumed invariants and serves as testing at the same time. Keep that stuff.

gcp:
Thanks for your efforts here. I'll have a careful look when I'm back home (which will be a few days).

PR #303, opened 2017-12-08
autogtp: fixing default value for any boolean option

marcocalignano:
This fixes the problem with 'noise : false'

### DIFF
--- a/autogtp/Management.cpp
+++ b/autogtp/Management.cpp
@@ -133,8 +133,10 @@ QString Management::getOption(const QJsonObject &ob, const QString &key, const Q
 
 QString Management::getBoolOption(const QJsonObject &ob, const QString &key, const QString &opt, bool defValue) {
     QString res;
-    if (ob.contains(key) && ob.value(key).toString().compare("true", Qt::CaseInsensitive)) {
-        res.append(opt + " ");
+    if (ob.contains(key)) {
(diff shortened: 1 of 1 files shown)


roy7:
I was concerned about storing booleans in the database as "false" native boolean since what if it turned into 0 or something, so I do cast it to a String internally all of the time.

marcocalignano:
@roy7 so what would you like to have?

roy7:
My C++ skills are weak, I have no idea, I'm just letting you know I pass options.noise as a String per our discussion that you want all options as strings. :)

marcocalignano:
@roy7 yes the type is string but the string content you can choose :smile:

marcocalignano:
@roy you could save it as a string in the DB?

roy7:
I do save it as a string. :)

bood:
It seems a bit odd to me that you use all values as string types.
IMO bool/number types are better choice when the options are bool/numbers.

gcp:
C++ doesn't support JSON at all, he's "lucky" that Qt has support classes for it :-)

bood:
No easy library either? (Probably not with reflection)

But in general, I think we can still learn something from java.

I would suggest write a class representing the whole `json` object and fill it with the json received. Then just manipulate the `leelaz` options according to the class filled earlier. Could make the code cleaner maybe? It's up to you though :-) @marcocalignano

PR #307, opened 2017-12-08
autogtp: sending random seed back to the server during file uploading

marcocalignano:


marcocalignano:
@gcp this merge send back the random seed to the server I think it is important now that the server send the seeds.

PR #313, opened 2017-12-09
Minor changes to VS2017 solution

qysnn:
I made some changes to the solution setting so that the it will work out of the box with VS2017. I also updated the solution so that it works with recent updates of next branch and fixed a bug that causes crash on Windows. But unfortunately I failed to make the solution automatically add dependencies to output dir as @earthengine did. So you will need to add them manually. 
Tested on Windows 10 and Ubuntu 16.04


zediir:
Thanks. This one works for me. Current next gives me the error "error MSB6006: "cmd.exe" exited with code 1"

zediir:
I get this error when I try to build it

autogtperror

qysnn:
@zediir I would suggest you to update with the newest commits and try to debug within Visual Studio so that you will have a better track of what's going wrong.

zediir:
Yeah no idea what happened. I changed the project qt version to msvc2017_64 and it worked and then changed it back to $(DefaultQtVersion) and it works again.

gcp:
@earthengine can you take a look?

earthengine:
I think this should be good, except that there should be some changes in the VC2015 solution as well.

qysnn:
@earthengine Thank you for your review. 
I switch it to Windows SDK 10 because it is the default SDK if you install VS2017, Windows SDK 8.1 will not be installed by default.
I also noticed that Qt is not using QtDll but I didn't change anything about it. It is the default behaviour I guess.

gcp:
I pulled the other request which I believe included this.

PR #315, opened 2017-12-09
Implements reuse of search subtree as described in AlphaGo Zero paper…

sp522:
… (Methods,Search Algorithm,Play section):

   "The search tree is reused at subsequent time-steps: the child node corresponding to the played action becomes the new root node; the subtree below this child is retained along with all its statistics, while the remainder of the tree is discarded."

MCTS continues the search with subtree's root node, skipping neural network evaluations on previously expanded nodes.  More efficiently using current turn's resources on unexplored nodes/paths.

Compatible with commands at GTP layer.  Ponder will continue to expand the search tree for opportunistic reuse on next enemy move.  Legal play moves trigger the search tree culling.  Clear_board/undo moves resets search tree.

Edited:  TTable serves additional purpose as explained by @gcp, for transpositions in unexplored paths 
<del>Disable TTable calls, reused subtree contains all the info TTable syncs.</del> Re-enabled

Experiments show improved visit counts vs configured playouts.  No signs of memory leaks.
Attached a debug log showing behavior of visits, nodes pruned/preserved each iteration for a -p 1600 run.
[log.reuse.debug.txt](https://github.com/gcp/leela-zero/files/1544456/log.reuse.debug.txt)

### DIFF
--- a/src/FastBoard.h
+++ b/src/FastBoard.h
@@ -60,6 +60,8 @@ public:
     */
     static constexpr int RESIGN = -2;
 
+    static constexpr int ROOT = -3;
+
     /*
         possible contents of a square
     */
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -65,6 +65,34 @@ std::string cfg_logfile;
 FILE* cfg_logfile_handle;
 bool cfg_quiet;
 
+UCTNode * root_subtree;
+
+void cull_tree(UCTNode ** root, int move, bool reset=false) {
+  if (reset || (move == FastBoard::RESIGN) || (*root == nullptr)) {
+    delete *root; *root = nullptr;
+  }
+  else {
+    //advance root to subtree of move
+    UCTNode * subtree = (*root)->remove_child(move);
+    if (subtree != nullptr) {
+      UCTNode * oldRoot = *root;
+      subtree->subtree_adjust_children();
+      *root = subtree;
+      myprintf("[SUBTREE REUSE] nodes: deleted[%d] preserved[%d]\n",
+	       oldRoot->calc_subtree_size(),
+	       subtree->calc_subtree_size());
+      delete oldRoot;
+      //deletes all old history except subtree of move
+    }
+    else {
+      myprintf("[SUBTREE REUSE] nodes: deleted all[%d] no subtree for move[%d]\n",
+	       (*root)->calc_subtree_size(),
+               move);
+      delete *root; *root = nullptr;
+    }
+  }
+}
+
 void GTP::setup_default_parameters() {
     cfg_allow_pondering = true;
     int num_cpus = std::thread::hardware_concurrency();
@@ -272,6 +300,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
     } else if (command.find("clear_board") == 0) {
         Training::clear_training();
         game.reset_game();
+        cull_tree(&root_subtree,0,true);
         gtp_printf(id, "");
         return true;
     } else if (command.find("komi") == 0) {
@@ -299,6 +328,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
             gtp_printf(id, "");
         } else if (command.find("pass") != std::string::npos) {
             game.play_pass();
+            cull_tree(&root_subtree, FastBoard::PASS);
             gtp_printf(id, "");
         } else {
             std::istringstream cmdstream(command);
@@ -314,6 +344,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
                     gtp_fail_printf(id, "illegal move");
                 } else {
                     gtp_printf(id, "");
+                    cull_tree(&root_subtree, game.get_last_move());
                 }
             } else {
                 gtp_fail_printf(id, "syntax not understood");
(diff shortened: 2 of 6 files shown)


gcp:
Isn't this strictly worse than using a TTable, which achieves the same thing but recognizes transpositions?

sp522:
@gcp The current implementation of TTable uses a direct mapped bucket indexed by board hash, caching the visits and blackevals of a node.  On "cache hits", values are synchronized at the start/end of play_simulation.  However, there are some flaws to this particular implementation.
1. Performance: Nodes that hit the TTable still require the neural network evaluation step to expand since TTable doesn't track node connectivity (child and siblings).
2. Correctness: Child nodes created aren't sync'ed with TTable.  On non-first visits to a node, uct_select_child compares these new nodes' "first play urgency" against potential TTable synced sibling node(at least one or more) on the same tree level that were TTable synced.  This implementation bug biases the search towards sync'ed nodes heavily.  This might be why our inquiries into Q init value was dismissed (init Q(s,a)=0 works fine as stated with subtree reuse).

Subtree reuse does not have the above issues.
TTable implementation can likely be tweaked to contain more info to avoid network eval, and fixed to sync new child nodes, but at that point it's virtually the same as saving the entire subtree minus the goodness of culling past steps.
Now that brings us to what transposition table means in the paper, we know board positions can't repeat per the game rules.  If not caching values across moves, maybe its main use is caching neural network calls across games? (this is pure speculation).

gcp:
A transposition is an exchange of moves the leads to the same position. e.g. A1 B2 A2 is the same as A2 B2 A1.

gcp:
Ok. It occurred to me also that the TTable hits aren't detected until after the node has been selected for expansion, which is a difference.

Take a look at #301 though, you will understand why I'm not very likely to merge this right now as-is.

I'd want to do a measurement to see if we should persue it further:

a) Check performance at p=1600 compared to master (maybe use supervised network for that), see if it's better by statistically significant amount.
b) Check performance when TTable is re-enabled. The code currently enables but as explained above the TTable and subtree reuse do not achieve the same thing exactly.

gcp:
>How does this interact in autogtp training mode with the addition of dirichlet noise at the root?

Noise would have to be added after new node is set as the root. Not a fundamental problem, I think.

sp522:
@gcp I agree TTables should be re-enabled.  Subtree reuse feature is compatible with it.
My initial thought process to disable was I saw similar values being sync'ed back, but I totally overlooked the unexplored paths with transpositions that benefit from it.

gcp:
I merged #301 so a rebase now would be cool.

sp522:
Sounds good, let me merge the conflicts.

sp522:
@killerducky The effect of Dirichlet noise at the root gets diminished if the visits of the first children level don't get readjusted down (along with the accumulated black evals).  

As far as I can tell, this only needs to be done on the root level.  The google paper does not mention this detail, in fact it says the opposite if taken verbatim "retained along with all its statistics".

I support resetting the visits for children at the reused subtree's root level- I've run into a scenario where the noise discovered a node with low score/high eval.  However, it could not overcome the large inherited visit count statistics from the reused nodes.  

I've added this but TTable syncs overwrite the adjustments since those get called after.
@killerducky @gcp I think this is a good time to discuss Dirichlet noise interaction with re-used statistics (either from TTables or subtree reuse).

killerducky:
@sp522 I created #538 to discuss the interaction of Dirichlet noise and tree reuse.

sp522:
Resolved new conflicts + squashed commits

MartinDevelopment:
I combined this and https://github.com/gcp/leela-zero/pull/471#issuecomment-355280555 together and that seems to slower than using https://github.com/gcp/leela-zero/pull/471#issuecomment-355280555 alone. (I am also running LZ with https://github.com/gcp/leela-zero/pull/523#issuecomment-355393533)

sp522:
@MartinDevelopment It's due to how playouts are counted.  It may seem slower if you use moves/second as the metric.  With subtree reuse, no playouts are consumed from nodes traversed in the reused subtree -max_playouts extra leaf nodes are expanded every turn.  With nn-cache/TTables, a playout count is consumed for hits, capping each turn to (max_playouts - cache hits) leaf node expansions. #538 has some discussion on the topic.  It boils down to how we should interpret the "each MCTS used N simulations" from the paper.  Do you count each node from the subtree reuse as a sim or not?
Fundamentally, subtree reuse merely preserves the relevant parts of the tree for the next search iteration- no slowdown should be observed.

bood:
I thought #315 somehow obsoletes this PR according to @ywrt, no?

killerducky:
@gcp @sp522 what is the status of this pull? I am doing some investigation of the interaction of TTable and the shape of the policy prior for dump_training. I see cases where the shape of the tree is distorted when the root does not select children that have entries in the TTable waiting to be synced. I'm guessing this pull would solve the problem. 

If this pull is not going to be merged there are other ways to fix the issue. For example at the beginning of a new search, probe and sync the TTable for all children. 

I can provide some examples of the distortion later today.

gcp:
The idea is OK but the implementation isn't good enough. There's a ton of raw pointer manipulation which we're trying to get rid of, coding style and indentation is completely different from all other code, the manual calling of cull_tree looks extremely error prone and with NNCache it's not clear what benefit this now has given what subtree_adjust_children does.

I'll close this to make it clear it won't get merged without a large overhaul.

killerducky:
@sp522 @sethtroisi @Ttl do one of you want to try to redo this? If the tree was kept I think we wouldn't have the issues I'm trying to patch up in #581. This kind of change would be hard for me because I'm not really a C++ programmer.

PR #321, opened 2017-12-09
Write log file

### LINKED ISSUE #281: Output in ver 0.7 is unreadable
When AutoGTP play many games it displays move numbers of each game which are totaly messed, like:
F2) 340 (K19) 98 (L3) 23 (F17) 52 (P6) 190 (C3) 105 (R4) 71 (E4) 253 (B4) 341 (N19) 99 (L2) 24 (R15) 53 (R4) 191 (G8) 106 (S4) 72 (C3) 254 (R1) 342 (N16) 100 (H6) 25 (N6) 54 (P11) 192 (F5) 107 (Q4) 73 (E3) 343 (G4) 255 (A4) 101 (N3) 26 (L14) 55 (C4) 193 (G4) 74 (P15) 344 (T16) 108 (S6) 256 (P1) 102 (P11) 27 (N9) 56 (C5) 194 (F8) 75 (F5) 345 (T5) 109 (S3) 257 (O2) 103 (F5) 28 (A17) 57 (F4) 195 (F4) 346 (T1) 76 (K12) 110 (S5) 258 (A6) 104 (H16) 29 (B7) 58 (O6) 196 (H4) 347 (D2) 77 (R15) 111 (R3) 105 (G6) 259 (L19) 30 (F3) 59 (Q5) 197 (G8) 348 (N17) 78 (O18) 112 (P4) 106 (O15) 260 (M19) 31 (D5) 60 (S4) 349 (S13) 198 (E4) 79 (R16) 113 (P3) 107 (P15) 32 (F2) 261 (L18) 61 (L4) 199 (D3) 350 (J5) 80 (R17) 114 (H14) 108 (T6) 33 (G16) 262 (T3) 62 (F3) 351 (H6) 200 (F8) 81 (Q17) 115 (P6) 109 (O14)
I think this kind of output is at least confusing.

ljn917:
Each Game instance writes its own log file. In validation mode, two leelaz write two different log files.

The console (stdout) is changed to erase the previous move with a carriage return, so people won't worry about the mixed output. I don't know whether people like this idea or not... If not, I can revert it to the old fashion. The effect of the carriage return needs to be tested on Mac. Linux and windows should be fine.

Resolve #281

### DIFF
--- a/autogtp/Game.cpp
+++ b/autogtp/Game.cpp
@@ -30,7 +30,9 @@ Game::Game(const QString& weights, const QString& opt) :
     m_blackToMove(true),
     m_blackResigned(false),
     m_passes(0),
-    m_moveNum(0)
+    m_moveNum(0),
+    m_log(stdout),
+    m_islogopen(false)
 {
 #ifdef WIN32
     m_cmdLine.append(".exe");
@@ -150,7 +152,7 @@ void Game::checkVersion(const VersionTuple &min_version) {
 }
 
 bool Game::gameStart(const VersionTuple &min_version) {
-    QTextStream(stdout) << m_cmdLine << endl;
+    m_log << endl << m_cmdLine << endl;
     start(m_cmdLine);
     if (!waitForStarted()) {
         error(Game::NO_LEELAZ);
@@ -159,9 +161,9 @@ bool Game::gameStart(const VersionTuple &min_version) {
     // This either succeeds or we exit immediately, so no need to
     // check any return values.
     checkVersion(min_version);
-    QTextStream(stdout) << "Engine has started." << endl;
+    m_log << "Engine has started." << endl;
     sendGtpCommand(m_timeSettings);
-    QTextStream(stdout) << "Infinite thinking time set." << endl;
+    m_log << "Infinite thinking time set." << endl;
     return true;
 }
 
@@ -195,6 +197,10 @@ bool Game::readMove() {
         error(Game::WRONG_GTP);
         QTextStream(stdout) << "Error read " << readCount << " '";
         QTextStream(stdout) << readBuffer << "'" << endl;
+        if (m_islogopen) {
+            m_log << "Error read " << readCount << " '";
+            m_log << readBuffer << "'" << endl;
+        }
         terminate();
         return false;
     }
@@ -212,6 +218,11 @@ bool Game::readMove() {
     QTextStream(stdout) << m_moveNum << " (";
     QTextStream(stdout) << (m_blackToMove ? "B " : "W ") << m_moveDone << ") ";
     QTextStream(stdout).flush();
+    if (m_islogopen) {
+        m_log << m_moveNum << " (";
+        m_log << (m_blackToMove ? "B " : "W ") << m_moveDone << ") ";
+        m_log.flush();
+    }
     if (m_moveDone.compare(QStringLiteral("pass"),
                           Qt::CaseInsensitive) == 0) {
         m_passes++;
@@ -258,11 +269,9 @@ bool Game::getScore() {
         if (m_blackResigned) {
             m_winner = QString(QStringLiteral("white"));
             m_result = "W+Resign ";
-            QTextStream(stdout) << "Score: " << m_result << endl;
         } else {
             m_winner = QString(QStringLiteral("black"));
             m_result = "B+Resign ";
-            QTextStream(stdout) << "Score: " << m_result << endl;
         }
     } else{
         write("final_score\n");
@@ -284,13 +293,26 @@ bool Game::getScore() {
             error(Game::PROCESS_DIED);
             return false;
         }
-        QTextStream(stdout) << "Score: " << m_result;
     }
+    // users may want to see results on screen
+    QTextStream(stdout) << "Score: " << m_result << endl;
+    if (m_islogopen) {
+        m_log << endl << "Score: " << m_result << endl;
+    }
+
     if (m_winner.isNull()) {
         QTextStream(stdout) << "No winner found" << endl;
+        if (m_islogopen) {
+            m_log << "No winner found" << endl;
+        }
         return false;
     }
+
     QTextStream(stdout) << "Winner: " << m_winner << endl;
+    if (m_islogopen) {
+        m_log << "Winner: " << m_winner << endl;
+    }
+
     return true;
 }
 
@@ -303,6 +325,7 @@ int Game::getWinner() {
 
 bool Game::writeSgf() {
     QTextStream(stdout) << "Writing " << m_fileName + ".sgf" << endl;
+    m_log << "Writing " << m_fileName + ".sgf" << endl;
     return sendGtpCommand(qPrintable("printsgf " + m_fileName + ".sgf"));
 }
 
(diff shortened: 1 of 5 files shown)


marcocalignano:
So autogtp won't show anything anymore on the console for let's say 20 min till the first of the game is finished?

ljn917:
I will only show the last move on the last line.

marcocalignano:
So you copy the output that goes on the console into a log file (one for each game) so the #281 is not really solved by this change.
And like I said for #324  you can create a log file with the same name as the sgf file and the training dump file so you do not need to push your index down to game, each Game has its own unique file name.

ljn917:
I don't think #281 has a real fix, except a global lock for stdout. My carriage return tick is just hiding the mess.

My initial thought was tracking which thread is doing what and finding out potential dead threads easier, which is why I don't want to split logs. And the log is kind of duplicate with the sgf. Probably you guys have slightly different opinions on this. (?)

To remove the index in Game class, I can move log file to Worker, but I still need to pass down a QTextStream to Game.

marcocalignano:
>  I don't want to split logs

What do you mean here?

ljn917:
I mean each thread id (worker) has its own log file for all the leelaz's
running in that thread.

killerducky:
I propose the the console be used only to monitor status, and not output moves. Something like every 30 seconds or so each thread prints a short status:

Thread 1 status: Game 30, move 59.
Thread 2 status: Game 33, move 300.
etc.

In addition to the current curl command outputs etc.

For those users that want something like LeelaWatcher I propose they use the -k option and LeelaWatcher grab the latest sgf and relay it. sgf files are a much cleaner interface than trying to scrape the output of a program that is constantly changing.

marcocalignano:
> I mean each thread id (worker) has its own log file for all the leelaz's
running in that thread.

So call the log file with the same name as the sgf, data and debug files you do not need to give game the number of the thread

ljn917:
@killerducky I had similar thoughts with yours. My PR now only prints the last move on the last line. All the previous moves were erased. I can add leela instance number before the move as you suggested.

ljn917:
@marcocalignano I understand what you mean, but it is hard to track which threads are alive and which are dead. The hash cannot be easily related to threads. Also, the users may end up with lots of log files to delete.

ljn917:
I feel I need to split the log file and the console change into two pr.

ljn917:
@killerducky To support what you proposed, I think I need some console control like VT100 sequences. This is not a problem for Linux and Mac, but windows supports it only after Win 10 TH2. For earlier windows versions, something like ansicon is required. If everyone feel okay with ansicon, I can proceed with it, otherwise I need to call some APIs...

Do you have a better idea on this?

https://github.com/adoxa/ansicon

killerducky:
I was thinking more like a running log, not VT100 sequences. That would be more complicated than necessary, mess up people that are redirecting outputs, etc etc. Let's just keep it simple? 

Thread 1 status: Game 30, move 59.
Thread 2 status: Game 33, move 300.
<here there is a 30 second period of no output>
Thread 1 status: Game 30, move 79.
Thread 2 status: Game 33, move 320.
<during the 30 seconds Thread 2 finishes>
Thread 2 game 33 done
<usual output about curl, getting network, etc>
Thread 1 status: Game 30, move 100
Thread 2 status: Game 34, move 20

etc

marcocalignano:
You are trying to fix the output because it is confusing, but that output led to us finding and fixing two big bugs today. I think I want to keep the output like it is.

ljn917:
@marcocalignano Uhmm, I am just trying to provide a better output, so that things do not get mixed together.

If most of people feel this is unneccessary, I will just close this.

marcocalignano:
@ljn917 Maybe it is necessary but this output would have prevent for finding the bugs today so the information that are actually there they should stay there not be hidden away.
```
Thread 1 status: Game 30, move 59.
Thread 2 status: Game 33, move 300.
<here there is a 30 second period of no output>
Thread 1 status: Game 30, move 79.
Thread 2 status: Game 33, move 320.

```

ljn917:
@killerducky I feel I like continuous ouput more... Maybe other people have different thoughts.

marcocalignano:
Maybe you can try with QProcess::processId() number before every output and see how it looks.

marcocalignano:
@ljn917 if you want I can give my GUI play ground if you want to try at making it work.

ljn917:
@marcocalignano I agree. I am thinking to have a rolling log on the top and another per thread move log at the bottom. But it requires console control...

ljn917:
@marcocalignano I am not very good at GUI... but I can have a try.

marcocalignano:
so go on my repository and get the branch gui.
In `Worker::createJob(int type)` I connet each game with a QPlainEdit object but the direct connection is wrong you need to get a normal connection so the slot is executed by then thread and not the game thread. But then you need to wait few millisecond in `ProdutionJob::execute()` you send the signal  `sendGuiText()`
and see if this is enough for it to refresh.

ljn917:
Let me try

ljn917:
@marcocalignano I found the issue in your GUI. The problem is not the connection. The main() is blocked by a mutex to prevent it quit, so your app.exec() call does not get executed at all.... I will push a fix later.

marcocalignano:
If anyone is interested I managed to build an autogtp with a GUI.
It is ugly! It does not have any controls it starts 3 games in three threads and this threads send their move output on three different Tab panels in the main window. It was just the first attempt to synchronize GUI (not re-entrant) with the threads. My dream is to put a board on each tab and see them actually playing.
Although it sound cool this will slow down the client because after the thread send its output to the GUI the thread has to wait few milliseconds to let the GUI thread do its job.
And I am not sure we want that.
But that said if someone want to have a look it in my leelaz repo branch gui.

DanielThomas55:
>... this will slow down the client because after the thread send its
output to the GUI the thread has to wait few milliseconds to let the GUI
thread do its job.

It shouldn't have to, data is only going one direction here.

ljn917:
@marcocalignano There should be very little slowdown. The worker thread does not have to wait. The real cost is one GUI thread for event loop and a signal emission from worker thread, which is pushed into an event queue being processed in the GUI thread. So, if you still available cpu core, there should be very little slowdown.

I made a pr for @marcocalignano's gui branch yesterday. Anyone who want to test can test with the pr. It does not work without the pr.

marcocalignano:
Ok I got rid of the delay in the threads and I build a GuiReceiver wrapper that has the slot that receive the text. Now anyone know a Go Borad C++ Object Implementation (better if based on QT) that I can use to show the move when they are posted?

marcocalignano:
If anyone is interested (seems no one) I push a new commit in my gui branch. I got the board in.
A lot of house keeping to do but you can start seeing what I meant to in the previous comments.

gcp:
What should I do with this? @marcocalignano

marcocalignano:
@gcp For me is not a good solution. 
This would stop all the output to stdout and will write in a separate logfile for each game,
Autogtp would output something only when it send the file to the server I do not think we want that.

gcp:
Ok, closing.

Issue #291, opened 2017-12-08
why generate the file 022b302c296b4788820ca87f7d12dbc1.txt.debug.0.gz


l1t1:
the file isn't uploaded
<pre>

curl.exe -F networkhash=2184b7500795533b97a40cc83bdfd51d1f18183c6e76d7cca3053237e0f18c7c -F clientversion=7 -F sgf=@022b
302c296b4788820ca87f7d12dbc1.sgf.gz -F trainingdata=@022b302c296b4788820ca87f7d12dbc1.txt.0.gz http://zero.sjeng.org/sub
mit
Upload failed. Curl Exit code: 52
Continuing...

</pre>

l1t1:
the begining of the file is
<pre>
1
0.923566 0.928446 0.931919 11
0.0715592 0.0715149 0.0745254 10
0.916815 0.926541 0.934496 19
0.0789339 0.0820616 0.0918863 11
0.918546 0.921954 0.929629 12
0.0753645 0.079354 0.0864596 10
0.912243 0.918785 0.92352 62
0.0763539 0.0827979 0.0918227 10
</pre>

killerducky:
It's for the resign analysis. I agree it would be better if it was never generated, but It should get deleted, is it not?

gcp:
Is the resign analysis generated in production? It should not.

killerducky:
I can fix this tomorrow, probably add a separate dump_debug GTP command in addition to dump_training.

gcp:
Fixed in "next" branch.

PR #324, opened 2017-12-10
Create new dump_debug gtp command.

killerducky:
Only call it from ProductionJob when --debug dir is used.

This fixes https://github.com/gcp/leela-zero/issues/291

BTW I noticed dump_training is not listed in GTP list_commands. Is this on purpose to hide internal-use commands? Or should we update the list?

### DIFF
--- a/autogtp/Game.cpp
+++ b/autogtp/Game.cpp
@@ -317,6 +317,16 @@ bool Game::dumpTraining() {
     return true;
 }
 
+bool Game::dumpDebug() {
+    QTextStream(stdout) << "Dumping " << m_fileName + ".debug.txt" << endl;
+
+    if (!sendGtpCommand(qPrintable("dump_debug " +
+                        m_fileName + ".debug.txt"))) {
+        return false;
+    }
+    return true;
+}
+
 void Game::gameQuit() {
     write(qPrintable("quit\n"));
     waitForFinished(-1);
--- a/autogtp/Game.cpp
+++ b/autogtp/Game.cpp
@@ -300,31 +300,19 @@ int Game::getWinner() {
 
 bool Game::writeSgf() {
     QTextStream(stdout) << "Writing " << m_fileName + ".sgf" << endl;
-
-    if (!sendGtpCommand(qPrintable("printsgf " + m_fileName + ".sgf"))) {
-        return false;
-    }
-    return true;
+    return sendGtpCommand(qPrintable("printsgf " + m_fileName + ".sgf"));
 }
 
 bool Game::dumpTraining() {
     QTextStream(stdout) << "Dumping " << m_fileName + ".txt" << endl;
-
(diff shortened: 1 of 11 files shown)


marcocalignano:
But why all the way down to game what should the object write in the debug. I would do the debug in the object Management and if the problem come from down we could bring it up with an Result of type ERROR

marcocalignano:
I do not see any gtp command called dump_debug

marcocalignano:
Oh, I see, **you** did put the new GTP command! Ok but then you could add a command line option for leela with the filename like `--debug debug.txt` and the dump debug knows where to dump the debug or even better you do like the two function `dumpTraining` and `writeSgf`: use the random name with a different extension.
All my suggestion are a way to avoid to change all those objects only to provide a file name, the otion for autogtp is good and this option would help a lot later when we are out of the test phase. but after that we can use the same method we use for save sgf and dump training (you would have also the same name for all thrre files)

marcocalignano:
Ok so maybe you just need a Boolean value debug true or false in` Job::init() ` you do not use the QString anyway and if you want to move out of the way in a sub directory you can do it in Management (like we copy the sgf if one wants to keep them) and you can pass this Boolean from `Management` to `Job` thru the `Work::order()` function.

killerducky:
Good suggestion. I will update my code to do something like that. Also I think the new validation/production code is still in flux and being debugged. This pull request is lower priority so I think we can wait to merge this until things stabilize more.

killerducky:
@marcocalignano I updated it to use Job::init() as you suggested, it is much cleaner now! I also updated the resign_analysis.py script to handle the case were some games have -r 0 and others have resign enabled.

marcocalignano:
@killerducky Looks good. You did a refactoring and you deleted the unused parameter. So I wait with my pull request till @gtp merge this one.

autogtp/Game.cpp
```diff
@@ -317,6 +317,16 @@ bool Game::dumpTraining() {
     return true;
 }
 
+bool Game::dumpDebug() {
+    QTextStream(stdout) << "Dumping " << m_fileName + ".debug.txt" << endl;
+
+    if (!sendGtpCommand(qPrintable("dump_debug " +
+                        m_fileName + ".debug.txt"))) {
```
oni-link: Did you consider replacing this with `return sendGtpCommand(...);`?

killerducky: I kept the same style as the other functions around it. I leave it to the original author @marcocalignano to decide if he wants to change the style.

marcocalignano: @killerducky  Change it to return please.

gcp: I'll pull this next then.

killerducky: fixed.

PR #329, opened 2017-12-10
Cleanup unused m_hash_history

sethtroisi:
This is also a small optimization (~1%) given frequency of copies (https://github.com/gcp/leela-zero/pull/89).

I have some future work hoping to convert the m_ko_hash_history from vector to set (or unordered_set) but that has proven to be complex

### DIFF
--- a/src/KoState.cpp
+++ b/src/KoState.cpp
@@ -34,10 +34,7 @@ void KoState::init_game(int size, float komi) {
     FastState::init_game(size, komi);
 
     m_ko_hash_history.clear();
-    m_hash_history.clear();
-
-    m_ko_hash_history.emplace_back(board.calc_ko_hash());
-    m_hash_history.emplace_back(board.calc_hash());
+    m_ko_hash_history.emplace_back(board.get_ko_hash());
 }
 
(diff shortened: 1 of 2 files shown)


src/KoState.cpp
```diff
@@ -83,9 +65,8 @@ void KoState::play_move(int color, int vertex) {
     if (vertex != FastBoard::PASS && vertex != FastBoard::RESIGN) {
         FastState::play_move(color, vertex);
 
-        m_ko_hash_history.push_back(board.get_ko_hash());
-        m_hash_history.push_back(board.get_hash());
     } else {
-        play_pass();
+        FastState::play_pass();
```
gcp: This seems a bit weird given that the same patch changes KoState::play_pass. Intentional?

sethtroisi: intentional, I preferred both play_pass and play_move to call this method, so there's one fewer m_ko_hash_history.push_back.

sethtroisi:
Ready for a 2nd look.

other than the small change to play_pass this is a free optimization (by removing code + use memory from a fairly hot path).

Issue #198, opened 2017-11-28
LeelaZ passes at first move with network 92c658d7325fe38f0c8adbbb1444ed17afd891b9f208003c272547a7bcb87909


emdio:
So yesterday I started autogtp and later on I found out that to my surprise LeelaZ passed at his first move, and maybe more surprising, white didn't take advantage of this fact to win the game with the komi.
I thought that at this stage LeelaZ should be doing better than this, but maybe it's just normal to find this kind of behaviour once in a while.

$ ./autogtp -k ~/Documents/Leelaz
autogtp v4
Best network hash: 92c658d7325fe38f0c8adbbb1444ed17afd891b9f208003c272547a7bcb87909
Required client version: 2 (OK)
curl -s -O -J -w %{filename_effective} http://zero.sjeng.org/best-network
Curl filename: 92c658d7325fe38f0c8adbbb1444ed17afd891b9f208003c272547a7bcb87909.gz
Net filename: 92c658d7325fe38f0c8adbbb1444ed17afd891b9f208003c272547a7bcb87909
Engine has started.
Infinite thinking time set.
1 (pass) 2 (A11) 3 (Q10) 4 (M17) 5 (D8) 6 (N18) 7 (G3) 8 (A1)

evanroberts85:
The policy part of the network probably gives only a small percentage for white passing. While with enough simulations passing should end up the first choice move, 1000 playouts might not always be enough. It would be useful to do some tests.

roy7:
Also note the first 30 moves in a game have extra noise introduced to spread the learning out.

If I run leelaz normally with these weights, like:

 ./leelaz -w 92c658d7325fe38f0c8adbbb1444ed17afd891b9f208003

then I end up with these results and pass is way down the list, although the policy network per se doesn't add a lot yet to the searching.

Thinking at most 50.0 seconds...
NN eval=0.492954
Playouts: 1902, Win: 50.06%, PV: N18 N2 S7
Playouts: 5079, Win: 49.86%, PV: B7 B9 N18
Playouts: 8543, Win: 49.75%, PV: S7 M10 N2
Playouts: 12356, Win: 49.65%, PV: J16 F17 S7
Playouts: 16122, Win: 49.58%, PV: H11 F3 S13
Playouts: 18955, Win: 49.56%, PV: J8 R10 B13
Playouts: 21694, Win: 49.55%, PV: J8 R10 B13
Playouts: 25061, Win: 49.56%, PV: J8 Q15 G18
[...]

PR #336, opened 2017-12-10
FAQ: Fix hyperlink error in Tromp-Taylor rule explanation

gaieepo:
Fix hyperlink error under Tromp-Taylor rule part in FAQ markdown file, as it unexpectedly includes the Chinese characters in the rest of the paragraph.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -64,6 +64,6 @@ This is expected. Due to randomness of self-play games, once Black choose to pas
 ## 对局结果错误 ##
 ## Wrong score? ##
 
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -64,6 +64,6 @@ This is expected. Due to randomness of self-play games, once Black choose to pas
 ## 对局结果错误 ##
```
sethtroisi: If you had a second could you also like issue #198 in line 62, and change defailed to detailed

PR #339, opened 2017-12-10
autogtp: fixing game end in validation job

marcocalignano:


roy7:
Once this is merged and we roll it out so no more bad results will be coming in, I can try to do some queries on the database to fix the errors manually.

marcocalignano:
WE need to test it you found the problem with the three passes can you reproduce it?

zediir:
Could you use the random_seed parameter to test the exact same game if find a problematic one?

marcocalignano:
@zediir I guess not it is only a seed but I am not sure

roy7:
Speaking of seeds, is it fine that both black and white use the same seed?

zediir:
I would think it's fine because because they use different networks. Even if they used the same networks I guess the result would just be the same as if running on a single leelaz instance.

roy7:
Spot on @zediir. Thanks.

marcocalignano:
I tested and it is working.

autogtp/Game.h
```diff
@@ -42,6 +42,7 @@ class Game : QProcess {
     QString getMove() const { return m_moveDone; }
     QString getFile() const { return m_fileName; }
     bool setMove(const QString& m);
+    bool checkGameEnd() { return (m_resignation || m_passes > 1 || m_moveNum > (19 * 19 * 2)); }
```
sethtroisi: this is longer than 80 characters, consider splitting onto three lines (and potentially moving to .cpp)

autogtp/Job.cpp
```diff
@@ -106,6 +106,9 @@ Result ValidationJob::execute(){
            return res;
        }
        first.readMove();
+       if(first.checkGameEnd()) {
```
sethtroisi: gcp prefers "if (first.checkGameEnd()) {"

see https://github.com/gcp/leela-zero/commit/bcf0b2caa1c1918e686da13f9f34b6899559fd9e

marcocalignano:
IMHO the 80 char rule is mad. Everyone participating in this project has a decent GPU hence has a decent monitor with a decent resolution with **more** than 80 chars per line. But because of this "rule" the compiler now has to make a function call any time `checkGameEnd()` is called instead of just inline that return statement.

evanroberts85:
Some people may be using their mobile phone to monitor remotely?

Prillan:
What's wrong with writing:
```
bool checkGameEnd() { 
    return (m_resignation || m_passes > 1 || m_moveNum > (19 * 19 * 2));
}
```
I'm pretty sure that's what @sethtroisi was talking about.
> consider splitting onto three lines

marcocalignano:
@Prillan Oh well I did my changes already.

gcp:
I'm using 2 console windows side by side and keeping the line length under control helps a lot with that.

Also it's handy if you're editing over ssh from a cellphone :-)

marcocalignano:
@roy7 So now that gcp merged the bug fix you can set the resign rate like it was before and force the client min version to 10 so the people that use the branch next get up to date with the bugfixes

roy7:
Yes I can. So it's 10 and not 9? Required version currently is 8.

marcocalignano:
@roy7 yes the 10 version has the bugfixes the 9 does not!

roy7:
Version 10 now required.

PR #353, opened 2017-12-11
CMake : adding minimal Qt version 5.3

marcocalignano:


Ttl:
Did you mean to add fallback stuff to this pull request? If you add commits to branch that has open pull request those changes will be added to the pull request too.

marcocalignano:
@ttl I forgot I had this pull request open can I exclude the last commit from this pull request?

Ttl:
I would copy the fallback commits to a new branch and then force push this branch with the fallback stuff removed.

marcocalignano:
I did a new branch for them. Now can I reset the next branch to that commit on the Git-Hub web interface?

Ttl:
I don't know of any way to do it from web. You can do it locally by resetting your local next branch so that a36be79 is head and then force push.

marcocalignano:
That I know how to do it but I have to wait to be home I cannot connect via SSH today. But anyway this pullrequest conflicts so I have time to do it

marcocalignano:
So now it is fixed again

marcocalignano:
I add the "save sgf" for match games with the -k option and a workaround the fix the sgf in case White resign in a match game

gcp:
I'm a bit surprised you only have to fix the SGF for the white resigns case. Given that I'd expect Leela to write the extra [tt] move regardless of who resigns, and that the scoring always uses the board score, why wouldn't it apply in the opposite direction too?

marcocalignano:
@gcp If Black resign that means that it is the result of the "genmove black" GTP command. This command write the right result in the sgf and I always save the sgf of the Black process. 
If White resign I have to send "play white resign" GTP command to the black process and this command write a "pass" in the sgf so I have to fix it only in this case.

autogtp/Management.cpp
```diff
@@ -383,6 +383,26 @@ void Management::fetchNetwork(const QString &name) {
...
+
+void Management::archiveFiles(const QString &sgf_file, const QString &data_file, const QString &debug_file) {
+    QDir dir;
+    if (!m_keepPath.isEmpty()) {
+        QFile(sgf_file).copy(m_keepPath + '/' + sgf_file);
+    }
+    dir.remove(sgf_file);
+
+    if (!m_debugPath.isEmpty()) {
+        if(!data_file.isEmpty()) {
```
ljn917: @marcocalignano Are you sure if all the training data (data_file) correctly deleted when --debug is not present?

marcocalignano: Good catch. Thanks!

ljn917:
@marcocalignano Can you save the .sgf file instead of the .sgf.gz? Unzipping every sgf is kind of annoying.... Thanks.

marcocalignano:
@ljn917 Is it OK like that?

ljn917:
I am afraid not.... archiveFiles() deletes the sgf, which is needed by gzip (?)

ljn917:
Also, data_file is needed by curl. Things get a bit complicated here....

marcocalignano:
OK I reverted I need to think more at the solution and test it before committing. The problem I have another feature under test so I will do that later. For now Is a flow.

ljn917:
Take your time. Thanks.

marcocalignano:
The solution is easy. I have to make two function one to achive and the other to clean up. I see how long I stay up tonight ;)

marcocalignano:
Now this pull request should be complete.

autogtp/Game.cpp
```diff
@@ -317,9 +317,17 @@ bool Game::fixSgfPlayerName(QString& weightFile) {
     QString playerName("PW[Leela Zero ");
     playerName += weightFile.left(8);
     playerName += "]";
-
     sgfData.replace(re, playerName);
 
+    if(resignation) {
+        QRegularExpression oldResult("RE\\[B\\+.*\\]");
```
Ttl: I think this should also match "W+" if white resigns when white is leading according to Tromp-Taylor, but actually has dead groups and is losing when the dead groups are removed. Maybe the current network isn't strong enough for this to happen, but I think it's possible when the network is stronger.

marcocalignano: I do not understand. What do you want me to do?

Ttl: Regex doesn't match if the white is leading. For example "RE[W+3.5]" doesn't match the regex and isn't replaced.

marcocalignano:
@Ttl did you mean something like that?

Ttl:
Yes, that should work.

PR #365, opened 2017-12-12
Fix white player name in sgf

Ttl:
Replaces default `Human` player name in sgf file with `Leela Zero <network>`.

Works with match and self-play games.

Currently white network name is not stored in the match sgf file making it hard to tell afterwards which one it was.

### DIFF
--- a/autogtp/Game.cpp
+++ b/autogtp/Game.cpp
@@ -17,6 +17,9 @@
 */
 
 #include <QUuid>
+#include <QFile>
+#include <QTextStream>
+#include <QRegularExpression>
 #include "Game.h"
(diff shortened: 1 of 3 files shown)


roy7:
I was thinking about this last night in fact. For matches, since each leela has no information on the other player, one of two things would need to happen. Replace the player name in the SGF file after creation by autogtp before upload, or maybe add a custom gtp command where you can specify the opponent's name, for use when writing out SGF files.

killerducky:
gcp mentioned in another thread he prefers an external tool modify sgf instead of adding a custom gtp command.

PR #373, opened 2017-12-13
Don't die if server sends garbage

Ttl:


marcocalignano:
And what happens when the network download fails? That task stops? The
execute function return? It stay in an endless loop?

Ttl:
It calls `getWork()` again after delay.

marcocalignano:
Oh OK then!

PR #374, opened 2017-12-13
 fix save sgf 

klmsj:


marcocalignano:
Please revert the commit for the save sgf file I am waiting for the 'debug
pull request' and then I have a more complete solution.

klmsj:
ok

PR #395, opened 2017-12-16
cleanup: include what you use

sethtroisi:
I ran Google's ["Include What You Use" tool](https://include-what-you-use.org) over the src/ directory

I uploaded the generated iwyu.out to [pastebin here](https://pastebin.com/bqzsfN91), which explains why each include was added.

I had to fix a couple of includes (where private boost/ and bits/types/ headers were being included) but otherwise this was just a static analysis.

Before:

```
$ time make
real	0m43.462s
user	0m41.483s
sys	0m2.283s
```

After:

```
$ time make
real	0m41.296s
user	0m39.495s
sys	0m1.950s
```


gcp:
2 thoughts: 

- does this work correctly even if DEBUG, USE_OPENCL are enabled/disabled?
- I think I prefer it if config.h goes first because it can control the includes that are set before system/library headers are included. You already hit this in OpenCL.cpp I believe.

sethtroisi:
I move `#include "config.h"` to be above any other local includes (but sometimes below system includes).

Good things to test!
I've tested `make debug`, `make clang`, `make default`, with and without `USE_OPENCL`.

`USE_HALF` is broken from https://github.com/gcp/leela-zero/commit/01efa33c24109cacaed73af19e06d54da652fb5a

gcp:
>but sometimes below system includes

It should go above it (I know the code doesn't follow that in every place right now). You can then make the inclusion of system headers conditional on config.h, notably for non-cross-platform code.

This will need a rebase after your std::chrono work.

PR #401, opened 2017-12-16
Replace timeval and time_t with chrono

sethtroisi:
IMO chrono is much nicer than time_t and #ifdefs


gcp:
I've been wanting to do this myself. Awesome!

PR #415, opened 2017-12-18
Fallback + resend

marcocalignano:
In case of connection loss, autogtp fallback to the last valid self played configuration and save all the games.
When the connection is reestablished all the saved games are sent to the server.


gcp:
Needs a rebase.

Can you throttle the upload on reconnect to something like one game per 10 seconds?

marcocalignano:
Rebased!

PR #417, opened 2017-12-18
cleanup: Fastboard

sethtroisi:
simplify some fastboard logic

add `m_squaresize = m_boardsize + 2;` which tracks the size of the underlying board representation 
add `white_to_move()` to complicment `black_to_move()`
utilize `white_to_move()` and `black_to_move()` in a number of places


gcp:
This has AutoGTP changes intermixed.

sethtroisi:
rebased.

src/FastBoard.cpp
```diff
@@ -616,14 +611,19 @@ int FastBoard::get_prisoners(int side) {
...
+}
+
+bool FastBoard::black_to_move() const {
     return m_tomove == BLACK;
 }
 
-int FastBoard::get_to_move() {
-    return m_tomove;
+bool FastBoard::white_to_move() const {
+    return m_tomove == BLACK;
```
gcp: This looks wrong :P

sethtroisi:
merged next
fixed white_to_move
verified by comparing eval of empty board in this branch and next

src/Network.cpp
```diff
@@ -636,9 +636,8 @@ void Network::gather_features(GameState * state, NNPlanes & planes) {
     BoardPlane& black_to_move  = planes[16];
     BoardPlane& white_to_move  = planes[17];
 
-    int to_move = state->get_to_move();
-    bool whites_move = to_move == FastBoard::WHITE;
-    if (whites_move) {
+    auto to_move = state->get_to_move();
+    if (to_move == FastBoard::WHITE) {
```
gcp: state->board.white_to_move()?

sethtroisi: yes, this is even cleaner.

sethtroisi:
ready for a 2nd look

PR #418, opened 2017-12-18
Added is_move_legal to de-duplicate logic

sethtroisi:
Cleanup some duplicate logic in FastState, Training, and UCTNode

Tested by verifying several training.gz files matched sha1sum and looking at some ko positions.

`$  echo "dump_supervised all.sgf training.data" | ./leelaz -s 123 -w ../weights.txt`
`$ sha1sum training.data.*`


src/UCTNode.cpp
```diff
@@ -97,31 +97,24 @@ bool UCTNode::create_children(std::atomic<int> & nodecount,
...
     }
     eval = net_eval;
 
-    FastBoard & board = state.board;
     std::vector<Network::scored_node> nodelist;
 
     auto legal_sum = 0.0f;
     for (auto& node : raw_netlist.first) {
         auto vertex = node.second;
-        if (vertex != FastBoard::PASS) {
```
sethtroisi: I'm not sure where this checks that the vertex isn't occupied, probably a side effect of `!board.is_suicide` but I didn't find an obvious check.

Either way fastfailing if square is occupied (in is_move_legal) is probably better

gcp: https://github.com/gcp/leela-zero/blob/1894e17d8165b8f88b03f704155400091c07db7b/src/Network.cpp#L459

gcp:
Needs a rebase.

sethtroisi:
Rebased

Issue #416, opened 2017-12-18
Bias in early pass training data


killerducky:
I think there is a bias problem in the early pass training data. First my evidence of a problem, run this:

```
./leelaz -w 808ac5f35690774587d12f595d6e840801394ee30a26681e4712de1fc2acfb43
play b pass
heatmap
```
The prior for white to pass back and win is 0. I checked some older nets and it was higher but has been steadily going down. It should move towards 100%. Another problem is the prior for black to pass on the first move is holding steady at 3, it should be going down.

I believe there are two causes: 

1) Duplicate games are filtered from training data. This introduces a bias because the correct response to black pass is for white to also pass. Probably there are many of these games but they are filtered. Then all the games were the -n and -m 30 switches caused it to not pass will drown out the good game. Also even the pass/pass games themselves will have different policy targets in them, we should include them all so the NN can average them. Instead we just pick one and depending on what -n did for that one sample we will train on that noise instead of averaging out the noise.

2) Issues #375 and #359 are probably exacerbating the problem, but I think even after that's fixed the other problem will remain. Even in normal games with the current black pass prior on empty boards we will get some portion of black passes due to -n and -m 30.

Edit: Also the value target for black pass should converge on white wins, but because there is only 1 game of white passing back, and many of white just making a move, it will converge to 50/50 instead.

bood:
Good catch!

Another observation is although white prefers corner for the first move now, black still didn't get that yet. Any explanation?

jkiliani:
I don't see it as a problem really. I think what is happening is that the instances of black early passing are disappearing in the training window, and as a consequence, the network is forgetting that there's an easy (cheesy?) answer for white to this due to Tromp-Taylor scoring. This simply means that LeelaZ is no longer considering a pass as a valid move at any point which may result in instant game ending, but as as a generally losing move in early game, like any regular Go player would.

A much stronger Leela Zero may not be able to punish a black first move pass consistently by instant win, but it would punish it by making points and winning the game normally because of that. So this shouldn't really be a problem affecting the playing strength in any way.

killerducky:
> This simply means that LeelaZ is no longer considering a pass as a valid move at any point

The current network policy priors thinks passing is just as good an opening move as playing 4-4, 3-4, or 3-3. I don't call that "pass is not a valid move". If black does pass first, currently it takes more then 3000 playouts for white to realize it can win instantly by passing back.

[...]

PR #426, opened 2017-12-19
Print cfg_options_str and TimeControl in comment of sgf files.

killerducky:
This will help users reproduce games. Also helps address #416 by making short games unique as long as they used different random seeds or other command line arguments.

Time controls is a common source of confusion for users, and it's important to set correctly.

Example sgf:
```
(;GM[1]FF[4]RU[Chinese]DT[2017-12-18]SZ[19]KM[7.5]PB[Leela Zero 0.9 /home/ao]PW[Human]RE[W+7.5]
C[LeelaZ options: --noponder --playouts 1600 --weights weights --seed 15337825387906334875
LeelaZ time control: 3600s absolute]

;B[ap];W[pp])

```

Notice my user home directory is leaking in the PB field. I didn't change that code. The way autogtp works this won't happen because it uses the local directory. Also the printsgf GTP command is hard-coded to have the computer always white, human always black. Again it seems autogtp takes care of this part for us. These two things are only issues if someone is not using a GUI that writes SGFs and instead uses the GTP command. They can probably be cleaned up later.

### DIFF
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -222,6 +222,36 @@ static void parse_commandline(int argc, char *argv[], bool & gtp_mode) {
         }
     }
 #endif
+
+    auto out = std::stringstream{};
+    for (const auto& it : vm) {
+        if (it.second.defaulted()) {
+            continue;
+        }
+        auto &type = it.second.value().type();
+
+        out << "--" << it.first;
+        if (type == typeid(uint64)) {
+            out << " " << it.second.as<uint64>();
+        } else if (type == typeid(std::string)) {
+            // skip unary string options
+            if (it.second.as<std::string>() != "") {
+                // Sanitize user supplied strings
+                out << " " << it.first;
+            }
+        } else if (type == typeid(int)) {
+            out << " " << it.second.as<int>();
+        } else if (type == typeid(float)) {
+            out << " " << it.second.as<float>();
+        } else {
+            out << " unknown option type";
+        }
+        out << " ";
+    }
+    if (!vm.count("seed")) {
+        out << "--seed " << cfg_rng_seed;
+    }
+    cfg_options_str = out.str();
 }
 
 int main (int argc, char *argv[]) {
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -224,32 +224,11 @@ static void parse_commandline(int argc, char *argv[], bool & gtp_mode) {
 #endif
 
     auto out = std::stringstream{};
-    for (const auto& it : vm) {
-        if (it.second.defaulted()) {
-            continue;
-        }
-        auto &type = it.second.value().type();
-
-        out << "--" << it.first;
-        if (type == typeid(uint64)) {
-            out << " " << it.second.as<uint64>();
-        } else if (type == typeid(std::string)) {
-            // skip unary string options
-            if (it.second.as<std::string>() != "") {
-                // Sanitize user supplied strings
-                out << " " << it.first;
-            }
-        } else if (type == typeid(int)) {
-            out << " " << it.second.as<int>();
-        } else if (type == typeid(float)) {
-            out << " " << it.second.as<float>();
-        } else {
-            out << " unknown option type";
-        }
-        out << " ";
+    for (auto i=1; i < argc; i++) {
+        out << " " << argv[i];
     }
     if (!vm.count("seed")) {
-        out << "--seed " << cfg_rng_seed;
+        out << " --seed " << cfg_rng_seed;
     }
     cfg_options_str = out.str();
 }
--- a/src/SGFTree.cpp
+++ b/src/SGFTree.cpp
@@ -478,7 +478,7 @@ std::string SGFTree::state_to_string(GameState& pstate, int compcolor) {
         }
     }
 
-    header.append("\nC[LeelaZ options: " + cfg_options_str + "]");
+    header.append("\nC[" + std::string{PROGRAM_NAME} + " options:" + cfg_options_str + "]");
 
     std::string result(header);
     result.append("\n");
--- a/src/SGFTree.cpp
+++ b/src/SGFTree.cpp
@@ -402,6 +402,7 @@ std::string SGFTree::state_to_string(GameState& pstate, int compcolor) {
     header.append("DT[" + std::string(timestr) + "]");
     header.append("SZ[" + std::to_string(size) + "]");
     header.append("KM[" + str(boost::format("%.1f") % komi) + "]");
+    header.append(state->get_timecontrol().to_text_sgf());
 
     auto leela_name = std::string{PROGRAM_NAME};
     leela_name.append(" " + std::string(PROGRAM_VERSION));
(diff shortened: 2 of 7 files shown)


src/Leela.cpp
```diff
@@ -222,6 +222,36 @@ static void parse_commandline(int argc, char *argv[], bool & gtp_mode) {
...
+            }
+        } else if (type == typeid(int)) {
+            out << " " << it.second.as<int>();
+        } else if (type == typeid(float)) {
+            out << " " << it.second.as<float>();
+        } else {
+            out << " unknown option type";
+        }
+        out << " ";
+    }
```
gcp: What is the difference between this and collecting the original argv/argc?

src/SGFTree.cpp
```diff
@@ -477,6 +477,10 @@ std::string SGFTree::state_to_string(GameState& pstate, int compcolor) {
         }
     }
 
+    auto tc = state->get_timecontrol();
+    header.append("\nC[LeelaZ options: " + cfg_options_str + "\n");
```
gcp: I don't like hardcoding the program name, there's a PROGRAM_NAME define.

src/SGFTree.cpp
```diff
@@ -477,6 +477,10 @@ std::string SGFTree::state_to_string(GameState& pstate, int compcolor) {
...
+    header.append("LeelaZ time control: " + tc.to_string() + "]");
```
gcp: SGF has TM and OT properties to describe the timecontrol. We should be good citizens and use those.

src/Leela.cpp
killerducky: It was easier this way to know which options take string arguments, so I can protect user information. If I parse argv/argc the code has to be updated when new string arguments are added. Also the parser takes care of all the weird cases like extra spaces etc that can happen, what if the weights filename has a space in it, or some weird characters. I don't want to rewrite a parser for those cases. 

That said I can change it to parsing argv/argc if you want. Let me know.

src/SGFTree.cpp
killerducky: Ah right, I'll fix this and the PROGRAM_NAME tonight.

src/Leela.cpp
killerducky: Maybe I'm paranoid but for example in Windows if you copy/paste a network in a directory you might have a command line like this:
leelaz.exe -w "..\154eaebd - Copy"

And parser has to handle quotes, backslashes, spaces, and something that looks like a -Copy command line argument correctly.

killerducky: Oh sorry one more thing, I can look into using the short options instead of the long ones (-n not --noise) if you prefer to be less verbose.

killerducky: After thinking it over, almost all use cases won't have user info like a directory, so maybe it's not worth doing all this. So I changed it to just use argv verbatim.

src/SGFTree.cpp
killerducky: Fixed.

src/SGFTree.cpp
killerducky: Fixed.

killerducky:
Changes done and rebased. New example output below. Do you want me to chop the weights to 8 characters?

```
(;GM[1]FF[4]RU[Chinese]DT[2017-12-19]SZ[19]KM[7.5]TM[3600]PB[Leela Zero 0.9 808ac5f3]PW[Human]RE[W+7.5]
C[Leela Zero options: -w 808ac5f35690774587d12f595d6e840801394ee30a26681e4712de1fc2acfb43 -p 1600 --noponder --seed 17700092672977863927]

;B[pp];W[dd])
```

marcocalignano:
This will mess up the fixSgf function in autogtp. I search the same Regexp
that now is in the comment

killerducky:
@marcocalignano I tested it, it works. 
```
(;GM[1]FF[4]RU[Chinese]DT[2017-12-19]SZ[19]KM[7.5]PB[Leela Zero 0.9 657f91c7]PW[Leela Zero 0.9 657f91c7]RE[W+Resign]
C[Leela Zero options: -p 1600 -r 1 -m 30 -t 1 -d -n --noponder -s 8972309646146436749 -g -q -w 657f91c714d41ea1d0cf86ec5c2f1e69650cff3ee551dcf046b0062e4a285010]
```

I guess I don't understand QRegularExpression though because I thought the .* in this code would be greedy and gobble up the entire string until the last space. I guess it doesn't go past the newline I put before the comment?
    QRegularExpression le("\\[Leela Zero .* ");

Maybe it would be safer to change this code to use non-greedy ".*?" or can someone explain what is going on here?

marcocalignano:
@killerducky Yes it works because of this: 
` playerName = match.captured(0);`
match as all the occurrence of that RegExp but we take only the first one. If you put the comment before the player (is it allowed?) then it won't work anymore.

killerducky:
@gcp I'm new to github, do I need to press a button somewhere to notify that I've addressed your change requests?

gcp:
I think that if you push updates it knows about it. 

I'm thinking a bit about the fragility wrt @marcocalignano's comment, what to do about the nets (truncate? not duplicate because it is in player name? maybe we want to hash the actual file?) and how this is stored etc, which is why I didn't pull or deny this.

marcocalignano:
@gcp the player name is just Leela Zero in the beginning but in this line thhe network name will be appended:
https://github.com/gcp/leela-zero/blob/b1ad7561852bd9f7107b32bbf1be173b91165c0f/autogtp/Game.cpp#L322

yufanyufan:
I think game can’t be reproduced with time control, unless the number play out of each step are also recorded in the sgf.

killerducky:
Number of playouts are recorded in the command line options, e.g. "-p 1600".

yufanyufan:
When you enable time control, then the number of play out of each move is no longer a const, but depending on the hardware performence.

killerducky:
I see. That's outside the score of this pull request. For now we're focused on reproducing autogtp games. It's an interesting idea for a feature enhancement but I don't see it being used very often.

roy7:
I see this didn't make it into the new release. Perhaps the next one? :)

gcp:
I specifically didn't pull this to avoid any bad surprises with autogtp interaction, but it will be pulled into next soon.

killerducky:
What do you think about also adding something like `git describe --abbrev=4 --dirty --always --tags` in the make files, add a new GTP command verbose_version, and then also output that to the sgf file? Or send to the server as metadata?

This could help identify problems specific to people running on various next versions or with their own code.

gcp:
If the game reproduces you know they were using the right version :-)

The goal of this was to be able to reproduce the game if at all possible, and to allow repeated games to be stored on the server as long as they have a different seed. So this is enough.

There may be an argument to letting "version" return that though. Does it report a "clean" v0.10 if the user has actually checked out that tag or the corresponding revision?

killerducky:
Yes it reports cleanly, see below a few examples. With some autogtp parsing changes it would work to allow people using clean builds off next. Adding new tags will break it, maybe that's a good thing?

```
git checkout v0.10
v0.10

git checkout upstream/next
v0.10.1-1-g1bf2

echo //test >> Leela.cpp 
v0.10.1-1-g1bf2-dirty

git reset --hard
git checkout -b revtest upstream/next
git tag revtesttag
git describe --abbrev=4 --dirty --always --tags
revtesttag
```

roy7:
This isn't in 0.10.1 right? I got myself confused looking at newer games like:

http://zero.sjeng.org/view/ec4903939dea5fddfe994a166e99979bdb7636e0342dacad206b5435efd86f2e?viewer=wgo

which are on 0.10.1 but were missing params. When I do see that in game logs, it's because they are running /next I assume.

roy7:
@killerducky Did the /next special version names make it into the SGF file comments? Or was that more of a suggestion/idea that didn't get PR'ed yet?

killerducky:
git describe is an idea, but it is not in any code yet.

PR #435, opened 2017-12-19
autogtp: fix ms/move calculation

marcocalignano:


marcocalignano:
@gcp Since the resume feature is really complicated and I have no solution (and no hint from others) I think if this fix is working and no other bugs come along you can think at a new release.

gcp:
Ok, I will implement OpenCL correctness checks and then consider doing a new release.

marcocalignano:
.@gcp Before the release we need to change the URL from the test to the main server. @roy7 When?  and How?

gcp:
Whatever we do both the old and the new endpoints will have to be up for a bit. So probably the new code will have to use new endpoints (i.e. if it's zero.sjeng.org/foo now then we can't use zero-test.sjeng.org/foo because I have no way to remap that to a different version of the server code if zero-test goes to zero). Makes sense?

roy7:
Server is running almost latest test code all of the time anyway, endpoints work on both. The only thing is if we split the users then the match requests won't be sycned between the processes. But once people are no longer using zero-test that isn't an issue. Or for launch of new release we could bind the server to both ports and so have a single process running both web sites for a while so the match assignments are all in sync.

gcp:
Sounds good to me.

marcocalignano:
@gcp @roy7  we can always release the version 11 and force the client version to 11 on hte main and leave the 10 on the test.

marcocalignano:
@gcp do you make a release before Xmas? not sure how much I can help during the holidays. 
@roy7 I guess we can change the URL and the version now.

marcocalignano:
@gcp Before you make the new release I'd like to reduce the verbosity of the console output in autogtp..

PR #441, opened 2017-12-20
Unit test framework with Google Test

bood:
Attempt to fix #218 

Tested on `Mac` and `Ubuntu 16 + clang 4.0`. Still working on a Travis integration, but lack of supporting of ubuntu 16 really sucks, may end up with osx. For windows support, we need Appveyor integration, but I don't have a windows machine, so maybe someone else can work on it.

You probably need to run `git submodule update --init --recursive` first to pull the gtest submodule.

On Mac, just run
```
cmake CMakeLists.txt
make tests
./tests
```

For ubuntu, I cannot get it run with gcc, you probably need to specify the compiler with something like below:
```
sudo apt install clang-4.0 lldb-4.0
cmake -DCMAKE_CXX_COMPILER="clang++-4.0" CMakeLists.txt
```

gcc (tried both 5 and 6) is complaining some link error to vector for gtest library, maybe someone else could take a look @gcp @killerducky @sethtroisi ?

PR #442, opened 2017-12-20
Unit test framework with Google Test + Travis integration

### LINKED ISSUE #218: Intruducing cppunit?
In my previous work I suspect the GTP command "printsgf" is not working in Windows MSVC compile.

As it will takes a while to run autogtp, finish a game and finally jump into the suspected code, it would be easier if we have a standard test facility like cppunit in hand (and have proper test codes written).

I can add some hack on the code (a new "test" argument to autogtp for example), However in the long term, a unit test facility would be definitely required.

Shall we start working on creating a unit test project?

bood:
Attempt to fix #218 (Previous PR #441 closed due to merge issue)

Tested on `Mac` and `Ubuntu 16 + clang 4.0`. Still working on a Travis integration, but lack of supporting of ubuntu 16 really sucks, may end up with osx. For windows support, we need Appveyor integration, but I don't have a windows machine, so maybe someone else can work on it.

You probably need to run `git submodule update --init --recursive` first to pull the gtest submodule.

On Mac, just run
```
cmake CMakeLists.txt
make tests
./tests
```

For ubuntu, I cannot get it run with gcc, you probably need to specify the compiler with something like below:
```
sudo apt install clang-4.0 lldb-4.0
cmake -DCMAKE_CXX_COMPILER="clang++-4.0" CMakeLists.txt
```

gcc (tried both 5 and 6) is complaining some link error to vector for gtest library, maybe someone else could take a look @gcp @killerducky @sethtroisi ?


sethtroisi:
Worked out of the box for me on Ubuntu 17.10 with gcc-7 (7.2)

bood:
That's odd, could you double check with `cat CMakeFiles/tests.dir/link.txt` to confirm it's using gcc? CMake uses clang by default sometimes. I'm getting these linking errors:

```
gtest/googlemock/gtest/libgtest.a(gtest-all.cc.o): In function `testing::TestCase::TestCase(char const*, char const*, void (*)(), void (*)())':
gtest-all.cc:(.text+0x8882): undefined reference to `std::vector<int, std::allocator<int> >::~vector()'
gtest/googlemock/gtest/libgtest.a(gtest-all.cc.o): In function `testing::TestCase::~TestCase()':
gtest-all.cc:(.text+0x8929): undefined reference to `std::vector<int, std::allocator<int> >::~vector()'
gtest/googlemock/gtest/libgtest.a(gtest-all.cc.o): In function `testing::internal::UnitTestImpl::UnitTestImpl(testing::UnitTest*)':
gtest-all.cc:(.text+0xed70): undefined reference to `std::vector<int, std::allocator<int> >::~vector()'
gtest/googlemock/gtest/libgtest.a(gtest-all.cc.o): In function `testing::internal::UnitTestImpl::~UnitTestImpl()':
gtest-all.cc:(.text+0xef0f): undefined reference to `std::vector<int, std::allocator<int> >::~vector()'
collect2: error: ld returned 1 exit status
CMakeFiles/tests.dir/build.make:140: recipe for target 'tests' failed
make[2]: *** [tests] Error 1
CMakeFiles/Makefile2:106: recipe for target 'CMakeFiles/tests.dir/all' failed
make[1]: *** [CMakeFiles/tests.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
bood@bood-S2600CP:~/next$ cat CMakeFiles/tests.dir/link.txt
/usr/bin/g++-6   -Wall -Wextra -pipe -O3 -g -ffast-math -flto -march=native -std=c++14 -DNDEBUG  -flto -g CMakeFiles/tests.dir/src/tests/sample.cpp.o CMakeFiles/objs.dir/src/OpenCL.cpp.o CMakeFiles/objs.dir/src/Timing.cpp.o CMakeFiles/objs.dir/src/Network.cpp.o CMakeFiles/objs.dir/src/SMP.cpp.o CMakeFiles/objs.dir/src/FastState.cpp.o CMakeFiles/objs.dir/src/Utils.cpp.o CMakeFiles/objs.dir/src/TimeControl.cpp.o CMakeFiles/objs.dir/src/UCTNode.cpp.o CMakeFiles/objs.dir/src/Zobrist.cpp.o CMakeFiles/objs.dir/src/SGFTree.cpp.o CMakeFiles/objs.dir/src/SGFParser.cpp.o CMakeFiles/objs.dir/src/FullBoard.cpp.o CMakeFiles/objs.dir/src/GTP.cpp.o CMakeFiles/objs.dir/src/KoState.cpp.o CMakeFiles/objs.dir/src/TTable.cpp.o CMakeFiles/objs.dir/src/FastBoard.cpp.o CMakeFiles/objs.dir/src/Training.cpp.o CMakeFiles/objs.dir/src/UCTSearch.cpp.o CMakeFiles/objs.dir/src/Random.cpp.o CMakeFiles/objs.dir/src/GameState.cpp.o  -o tests -rdynamic -lboost_program_options -lopenblas -lOpenCL -lz gtest/googlemock/gtest/libgtest_main.a -lpthread gtest/googlemock/gtest/libgtest.a -lpthread
```

bood:
Travis integration was finally done using Docker

bood:
@gcp Any chance we can get this merged? So I can start to add some basic acceptance cases.

gcp:
>Note: Google Test is added as a submodule so `--recursive` is needed for
future clones.

Does this mean the README's should be updated to reflect this?

bood:
gtest module is not required unless you use CMake to build tests
There is no CMake related instructions in README now, I can add a section, what do you think?

gcp:
Sounds good.

I'm not sure if the current CMake setup works in MSVC/Windows. Actually I'm pretty sure it doesn't. But having tests - somewhere - is better than none so requiring CMake for them is fine for me.

bood:
Currently it does not work on Windows, but I will try to see if I can make
it work on windows too. CMake does support MSVC/Windows and Qt after all.

My ultimate plan is to make a cross-platform CMake for both autogtp and
leelaz.

Will submit an update to README soon.

bood:
@gcp You may want to update README again after merged and had your Travis account setup, to include the travis build status

README.md
```diff
@@ -114,6 +114,24 @@ source and remove the line that says "#define USE_OPENCL".
...
+    # Use stand alone directory to keep source dir clean
+    mkdir build && cd build
+    cmake ..
+    make leelaz
+    curl -O https://sjeng.org/zero/best_v1.txt.zip
+    unzip best_v1.txt.zip
+    ./leelaz --weights weights.txt
+
+    # If you want to run unit tests too
+    git submodule update --init --recursive
```
gcp: These instructions don't work. I think you now *require* the gtest stuff to be there before you can even run cmake.

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ubuntu:16.04
+
+# Install
+RUN apt-get -qq update
+RUN apt-get install -y curl
+RUN curl https://apt.llvm.org/llvm-snapshot.gpg.key|apt-key add -
+RUN apt-get install -y clang-4.0 lldb-4.0 cmake
```
gcp: Any reason why this uses clang-4.0?

The clang target in the original makefile is actually for 5.0: https://github.com/gcp/leela-zero/blob/7f9489ee1bc870ade3ee8c9db8a61d7977528304/src/Makefile#L19

README.md
bood: you are right, its now required to generate makefile, thanks for the correction

Dockerfile
bood: no particular reason, gcc works well in docker too

I will use the default gcc to save travis build time in next pr

killerducky:
@bood can you provide a non-trivial example test to show how this works?

bood:
@killerducky Are you asking how to write unit test code? I think it should be fairly easy if you follow the document of google test. Not sure what step you got stuck.

PR #444, opened 2017-12-20
Precompute batchnorm weights

Ttl:
Saves few sqrt calls, but no observable performance change in practice due to batchnorm being so fast compared to convolutions.

This should improve reproducibility as OpenCL sqrt is guaranteed to be accurate only within 3 ULP, but std::sqrt precision is 0.5 ULP.

### DIFF
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -109,6 +109,12 @@ void Network::benchmark(GameState * state, int iterations) {
              iterations, elapsed, (int)(iterations / elapsed));
 }
 
+void Network::process_bn_var(std::vector<float>& weights, const float epsilon) {
+    for(size_t i=0;i < weights.size(); i++) {
+        weights[i] = 1.0f / std::sqrt(weights[i] + epsilon);
+    }
+}
+
 void Network::initialize(void) {
 #ifdef USE_OPENCL
     myprintf("Initializing OpenCL\n");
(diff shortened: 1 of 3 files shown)


src/Network.cpp
```diff
@@ -109,6 +109,12 @@ void Network::benchmark(GameState * state, int iterations) {
              iterations, elapsed, (int)(iterations / elapsed));
 }
 
+void Network::process_bn_var(std::vector<float>& weights, const float epsilon) {
+    for(size_t i=0;i < weights.size(); i++) {
```
gcp: Should probably use a ranged based for or even replace it completely by a functional version.

src/Network.cpp
```diff
@@ -183,6 +189,7 @@ void Network::initialize(void) {
             } else if (linecount % 4 == 2) {
                 batchnorm_means.emplace_back(weights);
             } else if (linecount % 4 == 3) {
+                process_bn_var(weights);
                 batchnorm_variances.emplace_back(weights);
```
gcp: They're not variances any more now.

src/Network.cpp
```diff
@@ -335,18 +344,15 @@ template <size_t spatial_size>
 void batchnorm(size_t channels,
                std::vector<float>& data,
                const float* means,
-               const float* variances,
+               const float* scale_stddivs,
```
gcp: Name the parameter stddivs. (Which can then be made consistent with the global var name in Network) The scaling is done inside the function, it's not the stddiv that is scaled.

PR #451, opened 2017-12-21
AutoGTP: check for a minimum supported Qt5 version

akdtg:
Qt is needed for autogtp only.
This patch removes the check from leelaz's CMakeLists.txt and adds it to the autogtp/autogtp.pro.


marcocalignano:
Cool I didn't know you could do that in the .pro file.

PR #453, opened 2017-12-21
Encode resignation separately.

gcp:
Instead of handling a resignation as a bit of a pass with some special handling, just set a specific flag in the GameState object.

This should allow removing some hacks in AutoGTP where it has to fix up our SGF files.

@marcocalignano @sethtroisi can you review?

### DIFF
--- a/src/GameState.cpp
+++ b/src/GameState.cpp
@@ -167,7 +167,7 @@ void GameState::display_state() {
     m_timecontrol.display_times();
 }
 
-int GameState::has_resigned() {
+int GameState::has_resigned() const {
     return m_resigned;
 }
 
--- a/src/GameState.cpp
+++ b/src/GameState.cpp
@@ -41,8 +41,8 @@ void GameState::init_game(int size, float komi) {
     m_timecontrol.set_boardsize(board.get_boardsize());
     m_timecontrol.reset_clocks();
 
-    return;
-};
+    m_resigned = FastBoard::EMPTY;
+}
 
 void GameState::reset_game() {
     KoState::reset_game();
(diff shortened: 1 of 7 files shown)


src/GameState.cpp
```diff
@@ -171,6 +167,10 @@ void GameState::display_state() {
     m_timecontrol.display_times();
 }
 
+int GameState::has_resigned() {
```
marcocalignano: Here you forgot the `const`

src/SGFTree.cpp
```diff
@@ -449,9 +449,7 @@ std::string SGFTree::state_to_string(GameState& pstate, int compcolor) {
 
     while (state->forward_move()) {
         int move = state->get_last_move();
-        if (move == FastBoard::RESIGN) {
-            break;
-        }
+        assert(move != FastBoard::RESIGN) {
```
marcocalignano: Here I do not know what you want to do. Is assert meant as an if or you just forgot the semicolon?

marcocalignano:
So I checked it with leelaz at command line and it looks good. Now I am running autogtp to see if there are some incompatibilities.

marcocalignano:
can you check the sfg of the games
812f58cc0c2d62d52baf510b38efa58a46982d81a99cbebcceb9846cadf43561 
86ade8f5aa99bfac6ac388505a5f01e5dad7e16cbaae31b501a952f66f8fa8ae
on the server?. The first White resigned and the second Black resigned

marcocalignano:
But now I autogtp I get only resign as last move of a game  6 out of 6

marcocalignano:
Ok this commit is sound. I saw few games where both passed, so everything is working.

roy7:
812f58cc0c2d62d52baf510b38efa58a46982d81a99cbebcceb9846cadf43561
"score" : "B+Resign"

86ade8f5aa99bfac6ac388505a5f01e5dad7e16cbaae31b501a952f66f8fa8ae
"score" : "W+Resign"

RE[] matches on both

PR #464, opened 2017-12-22
autogtp: send moves count even for selfplayed games

marcocalignano:


roy7:
Ok I'll start saving these on the server side. Once this is merged and we have a new release and client version requirement, I could make it a mandatory field on server side (for error prevention).

PR #466, opened 2017-12-22
const GameState in Network.cpp

sethtroisi:
I've done some initial testing with comparing the result of
`echo "auto" | /usr/bin/time -v ./leelaz -t 1 -s 123 -w ../weights.txt -p 100 --noponder |& tee output.old`
on next and this branch but I'll follow up with more tests that involve calling undo / redo

I'm not in love with the new GameState constructor, I'm open to suggestions for redesign there (maybe another class or something?) that would avoid copying board_history then immediately clearing it.

I think this code is maybe 80% and expect that I'll need to make some changes but I wanted to start the code review process and see if you like the idea of this change before I go any further. 

Before: `Leela:  5000 evaluations in  4.88 seconds -> 1025 n/s`
After: `Leela:  5000 evaluations in  4.69 seconds -> 1065 n/s`


src/UCTSearch.cpp
```diff
@@ -39,9 +39,10 @@
 
 using namespace Utils;
 
-UCTSearch::UCTSearch(GameState & g)
+UCTSearch::UCTSearch(const GameState & g)
     : m_rootstate(g) {
     set_playout_limit(cfg_max_playouts);
+    m_rootstate.disable_history();
```
sethtroisi: This is the majority of the change, in the MCTS we don't need history.

We also save by not carrying around FullBoard which is much larger than the pair<bitset<361>,bitset<361>> that it get's converted to. 
Additionally we cache 14/18 of the input planes.

What's not yet perfect in this code is the fact that m_rootstate(g) full copies the history then immediately discards it :/

src/Network.cpp
```diff
@@ -629,50 +629,46 @@ void Network::show_heatmap(FastState * state, Netresult& result, bool topmoves)
...
+        white_to_move.set();
     }
 
-    // Now go back to present day
-    for (size_t h = 0; h < backtracks; h++) {
-        state->forward_move();
+    const auto moves = std::min<int>(state->get_movenum(), INPUT_MOVES);
+    for (auto h = 0; h < moves; h++) {
+        const auto& test = state->get_boardplanes(h);
+        planes[black_offset + h] = test.first;
```
sethtroisi: this performs a copy of the bitset and is avoidable by changing NNPlane to use pointers to bitsets but that's a future change

gcp:
I think I've said this before, but have you considered truncating the GameState object that is used in UCTSearch to the last 16 ply?

This is the kind of change where I'm not sure the speedup is worth the complexity.

sethtroisi:
Thanks for the initial opinion,
I like this approach because GameState can be based const to Network.gather_features which seems semantically more correct IMO. I'm wondering if I can achieve that using the current code and adding moves_ago(int).

I'm traveling the next couple of days but I'll collect some performance data and make another attempt limiting game history so we have more date to reflect on.

gcp:
My issue is that I don't think GameState should know about Neural Network planes. Caching them somehow in Network...OK...maybe indexed by hash.

sethtroisi:
I agree, let me try with the move_ago pattern and see if I can keep gather_features(const GameState&

sethtroisi:
I reverted much of this change.
New change is to pass const GameState to Network

Performance is slightly improved (but probably in the noise compared to the changes with Winograd / Caching...)

Before:  5000 evaluations in 26.24 seconds -> 190 n/s
After:  5000 evaluations in 25.12 seconds -> 199 n/s

src/Network.cpp
```diff
@@ -631,53 +631,51 @@ void Network::show_heatmap(FastState * state, Netresult& result, bool topmoves)
...
-                if (color != FastBoard::EMPTY) {
-                    if (color == to_move) {
-                        planes[our_offset + h][idx] = true;
-                    } else {
-                        planes[their_offset + h][idx] = true;
-                    }
-                }
-            }
-        }
-        if (!state->undo_move()) {
```
sethtroisi: I find this undo_move() call to be ugly as it prevents const (and performs non zero work)

src/Network.cpp
```diff
@@ -631,53 +631,51 @@ void Network::show_heatmap(FastState * state, Netresult& result, bool topmoves)
...
+                }
+            }
+            idx++;
+        }
+    }
+}
+
+void Network::gather_features(const GameState* state, NNPlanes & planes) {
+    planes.resize(INPUT_CHANNELS);
     BoardPlane& black_to_move  = planes[16];
```
gcp: This can be expressed in terms of INPUT_MOVES now.

sethtroisi:
Updated.

gcp:
I had to back this out as the engine crashed with it applied.

Issue #459, opened 2017-12-22
~28% of all the network evals are duplicates?


ywrt:
tl;dr: I think leela training games can easily be sped up by ~40%

Hacking in a duplicate tracker into Network::get_scored_moves() indicates that many calls to network to score the state have recently been evaluated.

I added something like: 
```
Network::get_scored_moves(...)
     [...]
     gather_features(state, planes);
 
+    std::vector<bool> bits;
+    for (const auto& p : planes)
+      for (unsigned int i = 0; i < p.size(); ++i)
+        bits.push_back(p[i]);
+    auto hash = std::hash<std::vector<bool>>{}(bits);
+    {
+      LOCK(tried_mutex, lock);
+      if (tried.count(hash)) ++tried_dup;
+      ++tried_total;
+      tried.insert(hash);
+    }
```

and after running a game with "-p 1600", it gave tried_dup=158745, tried_total=556768 for a duplicate rate of about 28.5%

As best as I can tell, most of these duplicates come from the node tree being expanded from scratch on every move.

For example, when UCTSearch::think() is called, the first thing it does is expand out the child nodes for the new root node. But that node was almost certainly already expanded for the previous move, so this becomes a duplicate call to Network::get_scored_moves().  etc, etc for a decent portion of the previous tree.

I suspect the right thing to do here is to enable re-using UCTSearch objects, and when a move it played it finds the appropriate child node, unlinks it (and it's children) from the existing tree, and the makes it the new root node.

bood:
Did #315 address this already?

ywrt:
I didn't see that. Which is silly, because I just reimplemented the same feature. 
https://github.com/gcp/leela-zero/compare/master...ywrt:reuse-nodes?expand=1
Eh. 

It turns out to be better, but not perfect:  It drops the duplicate calls from ~ 28% down to ~7%

gcp:
Consider rebasing to next, I can't merge any of that code as is. Given that there is still a discrepancy, maybe the Ttable should just store the network evaluations even if it has to be made a bit smaller?

ywrt:
I re-implemented this by just caching the network eval results. Based against next.

PR #471, opened 2017-12-22
Cache the network evaluation results. (~1.7x speedup)

ywrt:
Network::get_scored_moves() is often called with duplicate inputs due to both
search tree convergence, and portions of the search tree being in common between
moves.

1. Add a cache for the Network::Netresults
2. Remove TTable as it's no longer needed.
3. Print out cache stats on each turn.

Stats from from a random mid-game  (-t 1 -s 1 -p 1600 -m 30 -n -d  --noponder):
NNCache: 150545/296370 hits/lookups = 50.8% hitrate, 145825 inserts, 50000 size, 0 collisions


src/NNCache.cpp
```diff
@@ -0,0 +1,99 @@
...
+    for (const auto& p : features) {
+        hash = hash_combine(hash, p);
+    }
+    return hash;
+}
+
+// Return true if the feature planes are the same.
+static bool compare(const Network::NNPlanes& a, const Network::NNPlanes& b) {
+  if (a.size() != b.size()) return false;
+  for (size_t i = 0; i < a.size(); ++i) {
```
sethtroisi: Auto i

src/NNCache.cpp
```diff
@@ -0,0 +1,99 @@
...
+  }
+  return true;
+}
+
+
+const Network::Netresult* NNCache::lookup(const Network::NNPlanes& features) {
+  LOCK(m_mutex, lock);
+  ++m_lookups;
+
+  size_t hash = compute_hash(features);
```
sethtroisi: Auto hash

ywrt:
I think this resolves #459
I think it also obsoletes #315 (as there's little need to cache subtrees any more)

src/NNCache.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+template <class T>
+inline size_t hash_combine(size_t seed, const T& v)
+{
+      std::hash<T> hasher;
+      return seed ^ (hasher(v) + 0x9e3779b9 + (seed<<6) + (seed>>2));
+}
+
+static size_t compute_hash(const Network::NNPlanes& features) {
+    size_t hash = 0;
```
sethtroisi: Auto hash = size_t{0}

gcp:
Why would this obsolete the TTable? That still has the visit counts and evals etc.

src/NNCache.h
```diff
@@ -0,0 +1,72 @@
...
+
+    // Statistics
+    int m_hits{0};
+    int m_lookups{0};
+    int m_inserts{0};
+    int m_collisions{0};
+
+    struct Entry {
+        Entry(const Network::NNPlanes& f, const Network::Netresult& r) : features(f), result(r) {}
+        Network::NNPlanes features; // ~ 1KB
```
sethtroisi: Some food for thought, From my experience with the Chess engines most engines they don't store the position date in the TTable (this isn't directly compatible because the have really small TTable entries) as it's expense to store more bits). Given you only store 50,000 positions a hash collision on a 64 bit hash is fairly unlikely (I'll do the birthday paradox math later) (and it's not catastrophic if we return bad eval data) it might be faster and more space efficient to not store and compare each time.

src/NNCache.h
```diff
@@ -0,0 +1,72 @@
...
+        Network::Netresult result;  // ~ 3KB
+    };
+
+    // Map from hash to {features, result}
+    std::unordered_map<size_t, std::unique_ptr<const Entry>> m_cache;
+    // Order entries were added to the map.
+    std::deque<size_t> m_order;
```
sethtroisi: I think you might be able to use a normal queue

gcp:
Looks nice and clean otherwise. Update the copyright message to your own name. I'll do a full review later.

src/NNCache.cpp
gcp: I think the second loop can probably be replaced by something out of <algorithm>.

src/NNCache.h
gcp: The question is whether his hash has properties that are as good as typical Zobrist hashes. I don't think that's a given.

sethtroisi: If i read http://preshing.com/20110504/hash-collision-probabilities/ correctly,  it's 1/10 billion my own math to follow.

We have something slightly different.
Assume NNCache is 200K and full, for a 1000 move game where we do 100K playouts each move we'd have 1e8 lookups (assuming each position is unique AKA this never returns a hit) each with a 2^64/2e5 probability of collision => 2e13/2^64 = one in five million chance of collision (assuming hash is uniformly distributed) during the entire game

sethtroisi: @Gcp wish I was at home to evaluate and don't want to block the merge but always a thing we can investigate later!

gcp: >assuming hash is uniformly distributed

Right that's the problem here. Zobrist gives you that, but I'm not positive about this one. I guess the amount of collisions in a long game will tell?

src/NNCache.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+  m_cache.emplace(hash, std::make_unique<Entry>(features, result));
+  m_order.push_back(hash);
+  ++m_inserts;
+
+
+  // If the cache is too large, remove the oldest entry.
+  if (m_order.size() > m_size) {
+      m_cache.erase(m_order.front());
+      m_order.pop_front();
```
sethtroisi: More food for thought and later testing, age is probably an okay heuristic (particularly when playouts << cache size) but turnNumber might be even better (m_order would be converted to map<turn, queue> and you'd remove from the first not empty queue)

src/NNCache.cpp
ywrt: doh. std::vector already has an operator== overload.

src/NNCache.h
ywrt: The key (~1KB) is relatively small compared to the value (~3KB), so there's only small memory savings in dropping it entirely.

ywrt:
I think the TTable is obsolete because the visit counts and evals etc are rebuilt as part of rebuilding the tree.

My thinking was: Imagine a subtree from the previous move with visits + evals. That subtree was built from a number of calls to Network::get_scored_moves(). If the new tree is built from only that same Network calls in the subtree, then the vists + evals will be restored to what they were in the previous subtree.  Is that not right?

src/NNCache.h
ywrt: Amusingly, std::queue is normally implemented on top of std::deque.

gcp:
>are rebuilt as part of rebuilding the tree.

The whole idea is not to do this! If we know that after N visits we have a winrate X for a subtree due to the TTable, we will use the iterations/playouts on the current move to investigate the alternatives until UCT prefers that move again. But the extra iterations will be added onto the data in the TTable, we will not recreate it.

src/NNCache.cpp
ywrt: Agree, this is very sensitive to the playouts/move. I kept it simple for now, but having move epochs would be more robust.

src/NNCache.h
sethtroisi: :)

src/Makefile
```diff
@@ -49,7 +49,7 @@ sources = Network.cpp FullBoard.cpp KoState.cpp Training.cpp \
 	  TimeControl.cpp UCTSearch.cpp GameState.cpp Leela.cpp \
 	  SGFParser.cpp Timing.cpp Utils.cpp FastBoard.cpp \
 	  SGFTree.cpp Zobrist.cpp FastState.cpp GTP.cpp Random.cpp \
-	  SMP.cpp UCTNode.cpp OpenCL.cpp TTable.cpp
+	  SMP.cpp UCTNode.cpp OpenCL.cpp NNCache.cpp
```
sethtroisi: You need to revert the delete here too

ywrt:
Ok, I restored the TTable.

Thinking about this: If that's the desired behavior, shouldn't the nodes be sync'ed with TTable in UCTNode::create_children? With the current code, the previously known winrate for a node is only loaded after it's already been selected for expansion.

gcp:
Yes, it's possible to do that. It's a question of how much you want to be probing into the TTable and filling info into nodes that might never be selected.

src/NNCache.h
```diff
@@ -0,0 +1,72 @@
...
+    You should have received a copy of the GNU General Public License
+    along with Leela Zero.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef NNCACHE_H_INCLUDED
+#define NNCACHE_H_INCLUDED
+
+#include "config.h"
+
+#include <vector>
```
sethtroisi: I don't see vector used anywhere

ywrt:
After playing a full game with `-r 1 -t 1 -d -n -p 1600 -m 30 --noponder` and with the TTable restored:
`NNCache: 230002/552689 hits/lookups = 41.6% hitrate, 322687 inserts, 50000 size, 0 collisions`

src/NNCache.h
ywrt: thanks! fixed.

src/NNCache.h
sethtroisi: In addition to memory saving it makes cache lookup faster (in case of hit) because you don't have to perform the comparison

ywrt: Amused: I think you're reaching here ;) The comparison is much less than 1% of CPU.

FWIW: On my machine with a relatively fast GPU (Titan X), all the CPU operations take less than 10% of wall time. The actual CPU consumption of leelaz is 100%, but more than 90% of that is the terrible OpenCL busy-wait in clFinish().

As far as I can tell, it's basically useless to optimize anything on the CPU at the moment.
I recommend doing the simple reliable thing here and optimizing later if it proves to be worth it.

gcp:
2 questions:

After the maximum tree size is reached and we can't do any expansions, we evaluate random rotations of the node to refine the result a bit: https://github.com/gcp/leela-zero/blob/f8a1a5c5e6dcc07fb6119a214345d8fbf81d277a/src/UCTSearch.cpp#L68

I think this is now broken because only 1 rotation will be tried and always returned from the cache. A flag should be added to allow bypassing the cache.

Similarly, does "netbench" still do anything useful with this patch?

ywrt:
I amended get_scored_moves() to allow bypassing the cache, and changed Network::benchmark and UCTNode::get_eval to use that bypass.

isty2e:
I merged this branch with my local version and tested. For match games, the performance gain was around 13%. Presumably it is because two networks don't share the search tree and their policies are different? I haven't tested for self-plays, since autogtp is running matches all the time.

MartinDevelopment:
This gives me about a 30% speedup with CPU only on a Ryzen 1600X

isty2e:
For self-play games, the speedup was about 30–35%.

src/NNCache.cpp
```diff
@@ -0,0 +1,89 @@
...
+      return nullptr;
+  }
+
+  // Found it.
+  ++m_hits;
+  return &entry->result;
+}
+
+void NNCache::insert(const Network::NNPlanes& features, const Network::Netresult& result) {
+  LOCK(m_mutex, lock);
```
sethtroisi: trivial nit: there's a mix of 2, 4, and 6 space indents currently in this file. Most files in src use 4 space indent fairly consistently

src/NNCache.h
sethtroisi: Don't want to derail this pull request; I'm a huge fan and excited to see it going in.

I totally agree with you, the performance isn't big enough to warrant worrying about (at least right now). And it's a good self-consistency check / detector for bad hashing / overclocked computers / non-random bitflips.

Just adding some data because I was curious (and had free time),
https://gist.github.com/sethtroisi/76aeffe8b6b7a29ea4b3eabd49ca46c1

I tested hashing all `bitset<121>` with up to 5 bits set (197 million entries, 9 gigs of ram) for collisions. I didn't find any collisions, which is a good sign that gcc7, at least, uses a uniform hash (as is required by the [spec](http://en.cppreference.com/w/cpp/utility/hash)).

Inserting 197 million entries means 4e16 chances for collisions.
if `hasher<bitset<121>>` is uniform the chance of collision for a 64 bit hash is `4e16/2^64 ~= 2/1000 chance of collision`. if the hash was not uniform (even by a factor or 4) it would be much more likely to have see a collision.

I used the lower 32 bits from the hash and predictable see collisions after ~50-60 thousand insertions.

Additionally 4e16 collisions is greater than the number of comparisons that would be expected from generating 1 million self-play games (1600 node expansions * 65% cache miss * 400 moves/game * 50000 hashes to collide with * 1,000,000 self play games ~= 2e16 chances for collisions)

sethtroisi:
I did some analysis on the size of the cache

My methodology is to augment NNCache with a map inserted hash to insert number then on hit lookup how old the entry is. This allows me to estimate* hit rate based on different cache sizes.

[results (google sheets)](https://docs.google.com/spreadsheets/d/1KI6rRX_FAIHkq74p51weL2pRj2cMqlrZFCu6Ohh3kH8/edit#gid=474411313)

TL;DR the cache doesn't need to be larger than 10,000 with 1600 playouts.

----

code: https://github.com/sethtroisi/leela-zero/tree/o__cache_eval
diff + extra unrelated stuff: https://github.com/ywrt/leela-zero/compare/cache-nn...sethtroisi:o__cache_eval?expand=1#diff-62dbba49aec7fe1eecb18f44b50a2c08

\* as cache_size is reduced significantly hits become misses and potentially force entries out of the cache that later are marked as cache. For example entry is a hit with age = 16000, next search is hit with age = 10000, if cache_size is reduced to 10000 exactly then both would be misses as the first entry will cause eviction of the 2nd.

isty2e:
I think netbench command is broken with this PR. We probably need a workaround for that?

```
Leela: netbench
 1600 evaluations in  0.00 seconds -> 1264048 n/s
=
```

sethtroisi:
@isty2e  (on mobile) I believe this was already addressed, I see a skip_cache flag that's set true in the benchmark function. Can you pull and test again?

isty2e:
@sethtroisi That's the strange part, since my version is up-to-date. I'm not that accustomed to using git so the codes are merged in a strange way but I can confirm that this is the latest version of this PR.

ywrt:
There is a skip_cache flag, but it totally wasn't being passed correctly. Thanks for picking this up!
Now fixed.
```
Leela: netbench
 1600 evaluations in  0.78 seconds -> 2040 n/s
```

src/UCTNode.cpp
```diff
@@ -168,7 +168,7 @@ void UCTNode::kill_superkos(const KoState& state) {
 
 float UCTNode::eval_state(GameState& state) {
     auto raw_netlist = Network::get_scored_moves(
-        &state, Network::Ensemble::RANDOM_ROTATION);
+        &state, Network::Ensemble::RANDOM_ROTATION, 0, true);
```
oni-link: An assert in `get_scored_moves()` expects `-1` as value for `rotation` if `RANDOM_ROTATION` is used.

src/Network.cpp
```diff
@@ -98,7 +98,7 @@ void Network::benchmark(GameState * state, int iterations) {
         tg.add_task([iters_per_thread, state]() {
             GameState mystate = *state;
             for (int loop = 0; loop < iters_per_thread; loop++) {
-                auto vec = get_scored_moves(&mystate, Ensemble::RANDOM_ROTATION, true);
+                auto vec = get_scored_moves(&mystate, Ensemble::RANDOM_ROTATION, 0, true);
```
oni-link: In debug mode `rotation` argument is expected to be `-1`.

ywrt:
Sigh. Compiled with opt mode, asserts are disabled. Fixed. thanks.

src/NNCache.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "NNCache.h"
+
+#include <functional>
+
+#include "Utils.h"
+
+NNCache::NNCache(int size) : m_size(size) {}
+
+NNCache* NNCache::get_NNCache(void) {
+    static NNCache* cache = new NNCache();
```
gcp: Prefer a reference to a static here. (magic static)

src/NNCache.cpp
```diff
@@ -0,0 +1,88 @@
...
+    return cache;
+}
+
+template <class T>
+inline size_t hash_combine(size_t seed, const T& v)
```
gcp: nit: inconsistent braces

src/NNCache.cpp
```diff
@@ -0,0 +1,88 @@
...
+    // Found it.
+    ++m_hits;
+    return &entry->result;
+}
+
+void NNCache::insert(const Network::NNPlanes& features, const Network::Netresult& result) {
+    LOCK(m_mutex, lock);
+
+    size_t hash = compute_hash(features);
+    if (m_cache.count(hash)) return;  // Already in the cache.
```
gcp: Slight preference for find() != end(...) for existence checks.

src/UCTSearch.cpp
```diff
@@ -392,11 +396,14 @@ int UCTSearch::think(int color, passflag_t passflag) {
     Time elapsed;
     int elapsed_centis = Time::timediff_centis(start, elapsed);
     if (elapsed_centis > 0) {
-        myprintf("%d visits, %d nodes, %d playouts, %d n/s\n\n",
+        auto nn_hr2 = NNCache::get_NNCache()->hit_rate();
+        NNCache::get_NNCache()->dump_stats();
```
gcp: I'd propose to dump_stats when in DEBUG mode and keep the original output otherwise (without nncache stats)

src/NNCache.cpp
ywrt: fixed

src/NNCache.cpp
ywrt: fixed

src/NNCache.cpp
ywrt: fixed

ywrt:
also rebased onto next. no current conflicts.

gcp:
Merged in https://github.com/gcp/leela-zero/commit/26e82513ca5dad00920e79a2dd2ffb048583065c.

zediir:
That commit gives the following errors when I try to build in MSVC2017

>Severity	Code	Description	Project	File	Line	Suppression State
Error	C2511	'Network::Netresult Network::get_scored_moves(const GameState *,Network::Ensemble,int,bool)': overloaded member function not found in 'Network'	leela-zero	D:\My Folders\Documents\leela-zero-next\src\Network.cpp	481	
>Severity	Code	Description	Project	File	Line	Suppression State
Error	C2664	'void Network::gather_features(GameState *,Network::NNPlanes &)': cannot convert argument 1 from 'const GameState *' to 'GameState *'	leela-zero	D:\My Folders\Documents\leela-zero-next\src\Network.cpp	488	
>Severity	Code	Description	Project	File	Line	Suppression State
Error	C2664	'Network::Netresult Network::get_scored_moves_internal(GameState *,Network::NNPlanes &,int)': cannot convert argument 1 from 'const GameState *' to 'GameState *'	leela-zero	D:\My Folders\Documents\leela-zero-next\src\Network.cpp	499	
>Severity	Code	Description	Project	File	Line	Suppression State
Error	C2664	'Network::Netresult Network::get_scored_moves_internal(GameState *,Network::NNPlanes &,int)': cannot convert argument 1 from 'const GameState *' to 'GameState *'	leela-zero	D:\My Folders\Documents\leela-zero-next\src\Network.cpp	504	


When I reverted that commit the rest of next built properly.

gcp:
The MSVC project is probably missing NNCache.cpp.

zediir:
Yup. That's what it was. Added them and it's building now

PR #474, opened 2017-12-23
Fix broken USE_HALF compilation

ihavnoid:
The code is not being compatiable with the CPU implementation of OpenCL.  To make it somewhat compatiable I made changes to
minimize net_t usage on Network.cpp so that it does not further break things.  There is a slight performance overhead due to
additional copying operations but I think it is negligible.


gcp:
Much simpler fix in 41b99b3.

PR #481, opened 2017-12-23
Added weights file information to the GTP version command

OmnipotentEntity:
Minor tweak to give more information.  Motivating use case: KGS bots who automatically report version information on connection and other similar things.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -52,6 +52,7 @@ int cfg_lagbuffer_cs;
 int cfg_resignpct;
 int cfg_noise;
 int cfg_random_cnt;
+extern std::string cfg_weightsfile;
 std::uint64_t cfg_rng_seed;
 bool cfg_dumbpass;
 #ifdef USE_OPENCL
@@ -218,7 +219,9 @@ bool GTP::execute(GameState & game, std::string xinput) {
         gtp_printf(id, PROGRAM_NAME);
         return true;
     } else if (command == "version") {
-        gtp_printf(id, PROGRAM_VERSION);
+        size_t filename_start = cfg_weightsfile.find_last_of("/\\") + 1;
+        filename_start = filename_start % cfg_weightsfile.length();
+        gtp_printf(id, "%s with weightfile %s", PROGRAM_VERSION, cfg_weightsfile.substr(filename_start, 8).c_str());
         return true;
     } else if (command == "quit") {
         gtp_printf(id, "");


src/GTP.cpp
```diff
@@ -218,7 +219,9 @@ bool GTP::execute(GameState & game, std::string xinput) {
         gtp_printf(id, PROGRAM_NAME);
         return true;
     } else if (command == "version") {
-        gtp_printf(id, PROGRAM_VERSION);
+        size_t filename_start = cfg_weightsfile.find_last_of("/\\") + 1;
+        filename_start = filename_start % cfg_weightsfile.length();
+        gtp_printf(id, "%s with weightfile %s", PROGRAM_VERSION, cfg_weightsfile.substr(filename_start, 8).c_str());
```
sethtroisi: can you use basename(cfg_weightsfile)

src/GTP.cpp
```diff
@@ -218,7 +219,9 @@ bool GTP::execute(GameState & game, std::string xinput) {
         gtp_printf(id, PROGRAM_NAME);
         return true;
     } else if (command == "version") {
-        gtp_printf(id, PROGRAM_VERSION);
+        size_t filename_start = cfg_weightsfile.find_last_of("/\\") + 1;
+        filename_start = filename_start % cfg_weightsfile.length();
```
sethtroisi: can you use basename(cfg_weightsfile)?

OmnipotentEntity: basename is not a function in C++11.  It would require pulling in C++17 headers.

src/GTP.cpp
OmnipotentEntity: substr does not throw if you request a length longer than the remaining characters.  From cppreference.com: "Number of characters to include in the substring (if the string is shorter, as many characters as possible are used)."

gcp:
I'm not sure what this is trying to achieve. For online bots, you still have no idea about hardware, settings etc so I'm not sure if just knowing the weights file gains you much.

OmnipotentEntity:
I added the network weights file to the bots that I'm running because it was always the most requested piece of information from other users.  I put the network data in the game notes, but the game notes aren't visible by anyone after the game has already begun, but on the other hand the engine data that kgsGtp puts in the chat is visible.

src/GTP.cpp
sethtroisi: I apologize, I double checked some spec (wonder what I was looking at) because I had thought it did allow extra.

src/GTP.cpp
OmnipotentEntity: Correction, it's not part of C++17 either, it's actually part of boost::filesystem.

PR #482, opened 2017-12-23
Minor tidyup: Switch to using the std mutex everywhere.

ywrt:
Previously, there was a mix of std::mutex (ThreadPool, Utils) and SMP::Mutex (TTable, UCTNode).


sethtroisi:
Looks good is general and I'm a fan of unifying code style.

I'm unsure if the change to UCTNode destructor is safe (or can the entire destructor be removed?)

ywrt:
There's never a need to lock in either the constructor or destructor to access non-static member variables.

By definition, constructors are single threaded (not possible to have multiple threads constructing the same object).
By necessity, destructors must be single threaded: If there are multiple threads destroying the same object then the same area of memory will be freed twice(!). Else if thread A is using an object at the same time as thread B is destructing it, then it's a use-after-free race. Either way, the program behavior is undefined (aka it's a broken program).

OmnipotentEntity:
Moreover, it seems like the dtor's only job is to clear a vector of std::unique_ptr, this happens when the vector's dtor is called.  So the dtor seems to be superfluous.

gcp:
2 things:

1) The LOCK macro must stay. The comment explains what it's for.
2) There is a lock in the UCTNode destructor because we need to ensure that other threads are not still busy traversing the m_children array. It has nothing to do with destructing an object twice.

ywrt:
1. Could you expand on what the LOCK() macro is supposed to do?
Note that std::mutex doesn't have copy or move constructors, so it should be just about impossible to get a temporary. Similarly, std::lock_guard<> doesn't have a copy constructor.

2. Indeed, it's never useful to lock in a destructor. Among other items, the destructor will destroy the mutex, meaning that if any other thread blocked on the mutex, it will now be using a destroyed mutex.

gcp:
Hmm, indeed I can't convince std::lock_guard<> to make a temporary, due to the restricted interface. I'll pull this.

PR #486, opened 2017-12-24
autogtp: reduce verbosity in console output

marcocalignano:
Merry Christmas to everyone that celebrate!


marcocalignano:
I added the winner to a self played game as requested on #492

roy7:
Thanks!

Issue #484, opened 2017-12-24
Is Leelaz trained off all the the games, or just the more recent ones?


ywrt:
I'm assuming it's trained off the more recent ones? If so, how many are used?

As an aside: are the scripts used to drive http://zero.sjeng.org/ checked in anywhere? I suspect that would answer all of my questions :)

OmnipotentEntity:
250k games are used in training typically.

ywrt:
I'm assuming that the script used for training isn't training/tf/parse.py, but something else?
(Because parse.py is CPU bound parsing the training files, and not GPU bound actually training the network)

gcp:
It is parse.py.

The network is fairly small and TensorFlow/python is rather slow so it ends up largely CPU limited indeed.

gcp:
>As an aside: are the scripts used to drive http://zero.sjeng.org/ checked in anywhere? I suspect that would answer all of my questions :)

FWIW the server is @roy7's work and he was interested in open sourcing it, but I think he hasn't gotten around to cleaning up the code yet. It was originally just a few lines of node.js that shoved the submitted data into mongodb, but due to the distributed testing it has gotten more complicated.

The code that dumps the training data is here: https://pastebin.mozilla.org/9075155

Obviously it's useless without the full DB which I won't publish right now because it has IP addresses etc. I might anonymize it after the project.

ywrt:
Few more questions...

1. I'm assuming that training continues from the previous network? That is: At some interval: New chunks are created from games that have arrived, some old chunks are removed and then training resumes?

Or alternatively: training is stopped, all chunks are removed, the most recent 250k games are dumped, and training resumes?

2. What policy loss is currently seen in training? 
(My from-scratch training is currently showing `step 157600, policy=2.90912 mse=0.155379 reg=0.181329 total=3.24583 (1320.76 pos/s`)

3. Why is the batch size set to 256? As far as I can tell, this is too small for the learning rate (as indicated by saturation in the policy loss)?
I suspect the batch size should be increased to at least 1024 (or alternatively: reduce the learning rate, but larger batch is preferable). FWIW: The AlphaZero paper mentions a batch size of 4096. `Training proceeded for 700,000 steps (mini-batches of size 4,096)`

roy7:
You're correct on #1, if I understand what you are asking. Training doesn't go on forever though.

[...]

PR #493, opened 2017-12-24
Speed up record processing in tensorflow training.

ywrt:
Currently, training is cpu-bound loading move records in the worker threads. 

Changes:
1. Uses numpy to significantly speed up applying the symmetries.
2. Don't apply symmetries to planes 16 and 17 (both are uniform so the symmetry is a no-op)
3. Move the NaN filter into a numpy vector op.

On my machine with 3 worker threads:
Before:  ~520 records/second
After : ~1360 records/second


ljn917:
Why not plane 16?

gcp:
Programmers count from 0 so yes.

gcp:
Am I correct that this changes the data format being passed between threads from python list of lists to numpy arrays?
Could be a big part of the speedup, as in my experience the code is serially limited and more data processing threads won't help.

ljn917:
Thanks. I was confused by the conventions.

ywrt:
Apologies, I re-worked this a little and it's faster again.  More importantly, it now comes with a test! :)

With 3 workers threads:
Base: ~ 520 records/second
After: ~ 6700 records/second

It now fully saturates my GPU with ~ 50% cpu load.
```
2017-12-26 12:16:33.882987: I tensorflow/core/kernels/shuffle_dataset_op.cc:110] Filling up shuffle buffer (this may take a while): 47268 of 65536
2017-12-26 12:16:38.660416: I tensorflow/core/kernels/shuffle_dataset_op.cc:121] Shuffle buffer filled.
step 100, policy=6.10998 mse=0.304389 reg=0.563884 total=6.97826 (0 pos/s)
step 200, policy=6.08004 mse=0.298817 reg=0.563286 total=6.94214 (1389.11 pos/s)
step 300, policy=6.05026 mse=0.292761 reg=0.562289 total=6.90531 (1327.17 pos/s)
step 400, policy=6.02196 mse=0.285393 reg=0.561013 total=6.86837 (1316.13 pos/s)
step 500, policy=5.98251 mse=0.276803 reg=0.559583 total=6.8189 (1328.04 pos/s)
step 600, policy=5.91035 mse=0.267624 reg=0.558132 total=6.7361 (1324.22 pos/s)
step 700, policy=5.8299 mse=0.257874 reg=0.556702 total=6.64448 (1324.44 pos/s)
```

And yes, there were a number of independent bottlenecks. Fixing each of these just uncovered the next one.

1. apply_symmetry() was very slow due to processing each element in python. I replaced these with numpy gathers. And then replaced the individual plane gathers with a single wide gather.
2. Actually, pretty much anything that touches an individual element is slow in python. I replaced all of these with numpy vector ops (np.any(), np.isnan() etc)
3. multiprocessing.Queue() is slow. It takes a shared lock per record, so it drives extensive context-switching. I changed this to use multiple Pipe()s instead
4. The pickled data was huge and the sheer volume of data that needed to be moved between threads was a bottleneck. I changed to passing byte strings of uint8s instead of lists of floats.
5. The implicit conversion from python lists of lists of floats (returned by ParseChunks) to tensors was very cpu expensive. This one was particularly bad because it didn't scale (single-threaded due to python) and ate CPU in the main tensorflow thread. I changed to using tr.train.Example() and using the tensorflow graph to decode the protobuf and constructors the tensors. Something like 1/9th the CPU cost(!)

It could still go faster by moving more processing into the tensorflow graph, but given that it now fully saturates my GTX Titan X with much CPU to spare I don't think there's much point.

gcp:
Amazing work.

roy7:
I go from about 473 pos/sec pretty steady, to jumping around 1400-2200 pos/sec. Wow. This is on my Ryzen 1700 + 1080 Ti. The variation in speed is because RoyalLeela is playing a game so it isn't getting all of the GPU. I'd say I'm probably a 473 -> 2200 upgrade if RL was turned off?

And no more long shuffle buffer setup phase!

evanroberts85:
That is some speed up, is it possible to produce a validation dataset to check training losses against, or is there a reason this is not needed?

roy7:
@ywrt is a hero.

ywrt:
Good to hear it goes faster for you. Could you confirm that your GPU is saturated? i.e. GPU utilization stays over 95%?

And roy7, while I have you attention, I had a few more questions over in #484 ... :)

roy7:
I'm over 95% non stop according to nvidia-smi.

ywrt:
@evanroberts85 This change just modifies record parsing and decoding into tensors. There's a test to verify that going from lists -> text -> protobuf -> tensors -> lists ends back up where it started. So in this case I don't think a validation dataset would add much.

evanroberts85:
@ywrt The validation dataset is to see if the change is loss values for the training data mirrors the change in loss for data not used for training. It is a common way to check for overfitting. So from the 250k window, you might take a random 1%  to use for validating. 

Admitedly slightly off topic from your changes in this issue, but as you wrote a test for the tensors I thought it would be worth bringing up

ywrt:
@evanroberts85 gcp said elsewhere that training using 250k games and runs for 128k steps.
Each step is a mini-batch of 512 positions. 250k games * 150 moves-per-game / 128k batches / 512 moves-per-batch means that each move is expected to be seen by ~ twice.

The currently configured validation dataset is drawn from the main dataset with replacement. i.e. there's no current hard partition. But given that every move is only looked at ~twice, the error versus a fully distinct validation set is going to be small.

evanroberts85:
When I looked through the code I saw no validation being used, certainly it
is not included in the console printout.

You can have poor generalisation/ovrerfitting with data only seen once. But
there are other things I take issue with your observation. While each
position is technically unique, moves x and x+1 in the same came are up to
99% the same. Also for each training session only a small percentage of the
data is new, as we use a 250k window. By the time a game position has left
the training window it will have been used for training around 10-30 times.

gcp:
It should use a separate dataset for real validation but that simply isn't implemented yet and not all that important for our setup where we validate in match games.

PR #501, opened 2017-12-26
Use windows average instead of exponential average for TF statistics.

ywrt:
Change the avg_policy_loss et al to be average only over the step just
completed, rather than being an exponential average over all steps.
This is a more accurate estimate of the current losses.


ywrt:
(Took me a while to work out why I got a sharp improvement in the loss every time I stopped and resumed the training! :)

roy7:
Hah that always bugged me. Thanks!

PR #513, opened 2017-12-28
convert centis in Timing to ms

sethtroisi:
Smoked tested with
Unlimited
`echo -e "time_settings 0 1 0\nauto" | ./leelaz -w ../weights.txt`
Byo stones
`echo -e "time_settings 120 60 5\nauto" | ./leelaz -w ../weights.txt`
KGS style Fischer GO
`echo -e "kgs-time_settings byoyomi 20 6 4\nauto" | ./leelaz -w ../weights.txt`
lagbuffer
`echo -e "kgs-time_settings byoyomi 20 6 4\nauto" | ./leelaz --lagbuffer 700 -w ../weights.txt`

The `--lagbuffer` parameter used to accept centiseconds which we/you might want to change,
These are some of the options,
* keeping parameter as centiseconds
* changing parameter to ms, (what I coded)\*
* changing parameter name to lagbuffer_cs (or lagbuffer_ms)

\*One downside is anyone passing lagbuffer will now be passing a value ten times smaller :/

Infinite time setting, 1 month, overflowed the ms counter (max value 560 hours ~=25 days) and was reduced to 1 week.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -48,7 +48,7 @@ using namespace Utils;
 bool cfg_allow_pondering;
 int cfg_num_threads;
 int cfg_max_playouts;
-int cfg_lagbuffer_cs;
+int cfg_lagbuffer_ms;
 int cfg_resignpct;
 int cfg_noise;
 int cfg_random_cnt;
@@ -69,7 +69,7 @@ void GTP::setup_default_parameters() {
     cfg_allow_pondering = true;
     cfg_num_threads = std::max(1, std::min(SMP::get_num_cpus(), MAX_CPUS));
     cfg_max_playouts = std::numeric_limits<decltype(cfg_max_playouts)>::max();
-    cfg_lagbuffer_cs = 100;
+    cfg_lagbuffer_ms = 1000;
 #ifdef USE_OPENCL
     cfg_gpus = { };
     cfg_rowtiles = 5;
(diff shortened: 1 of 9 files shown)


gcp:
Would need to change the name of lagbuffer, don't like silently changing the meaning.

1 week is probably too little for infinite time - even if due to the hard tree size limit we have you're not likely to run into that problem for now.

I'm not sure "just" changing it from centi to millis is worth it...I understand that they're more common but for an engine like this centis works just as well. This may be more worthwhile if the stuff in TimeControl.h is changed to use chrono units. 

If this helps further progresss towards the latter (and I think it does due to easier conversion?) I'll pull it though.

sethtroisi:
All timing has already been changed over to chrono, but I hadn't planned on changing to chono types (i'm not sure that will end up being any cleaner).

I updated `lagbuffer` parameter to `lagbuffer_ms` unfortunately boost::program_options allows partial parameter names so `--lagb 100` or `--lagbufffer 100` both work and people might not notice that the parameter value is now 1/10th

1 week is the infinite time **per move**, correspondence games are 1 move / {1,3,5} days so 1 week seems like a reasonable amount of time

I like ms slightly better (more precision, more of a standard in timing) but I understand not wanting to change current behavior. If you don't want to change the current behavior I understand and we'll just close this.

roy7:
Can you keep old --lagbuffer and add a new --lagbuffer_ms? So you can use either.

gcp:
Going to close this as (for now) there are downsides and the upside isn't so clear.

PR #515, opened 2017-12-29
Change visits from int to uint64

GeneralZero:
When running the program for a long time I got "assert(best != nullptr);" for uct_select_child. It took a while to pin point the error but what happened is the number of visits went over the max 32bit int value. 

This creates a large negative value for for the selection value.

This edit change the visits in the UCTNode.* and the TTable.* files.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -247,7 +247,7 @@ void UCTNode::randomize_first_proportionally() {
     auto accum_vector = std::vector<uint32>{};
 
     auto child = m_firstchild;
-    auto accum = uint32{0};
+    auto accum = uint64{0};
     while (child != nullptr) {
         accum += child->get_visits();
         accum_vector.emplace_back(accum);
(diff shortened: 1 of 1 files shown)


sethtroisi:
all pull requests are merged into 'next' branch.

in 'next' typedef for uint64 was removed so it will need to be replaced with 'std::uint64_t'

gcp:
This code doesn't work because of things like:

"auto pick = Random::get_Rng().randuint32()"

Additionally a sanity check for computations that are cast to float is needed, and/or accumulating visits (much of those should already be using double, but maybe not all).

gcp:
To fix that random call you can probably use the C++ standard library to get a correctly generated 64-bit sized int and use the fact that Random confirms to the UniformRandomBitGenerator interface (as IIRC dirichlet_noise already does).

gcp:
I'll close this now. Feel free to reopen if you get a more complete fix.

Issue #498, opened 2017-12-26
Implement F(2x2, 3x3) Winograd convolution


Ttl:
I implemented Winograd convolution to learn some OpenCL. It should work pretty well in theory as the runtime is dominated by the batched GEMM and transformations are only about 12% of the runtime. Batched GEMM I'm using is from CLBlast library which should be pretty fast.

On my slow laptop with Intel GPU it is about 40% faster, but on my desktop computer with 1050 Ti Winograd convolution is 5% to 50% slower depending on the network size.

If anyone wants to look at the code it can be found here: https://github.com/Ttl/leela-zero/tree/winograd

Also just a note that when GEMM is first called it will compile it causing the runtime to be longer. In benchmarks at least one NN evaluation should be run before netbench.

gcp:
>On my slow laptop with Intel GPU it is about 40% faster, but on my desktop computer with 1050 Ti Winograd convolution is 5% to 50% slower depending on the network size.

The speed advantage on the Intel GPU is because my OpenCL direct convolution kernel is hampered by the fact that the execution units on Intel GPUs can't do effective SIMT and hence are presented as vector based rather than scalar based as on AMD and NVIDIA architectures. So it's more of a SIMD than a SIMT architecture, and it ends up not being able to use more than 1 ALU lane. CLBlast's GEMM kernels don't have this problem.

When you say transformations are only 12% of the run-time, was that measured on the Intel, the 1050 Ti, or both?

Ttl:
Transformations are about 12% on both devices.

gcp:
Ok. Consider doing a CLBlast tuning run with the exact matrix sizes you require?

How hard would it be to translate this back to CPU code? Could be a pretty large speedup on those.

Ttl:
I'll test the tuning tomorrow. The matrix sizes are not nice powers of two so it might hit some edge case.

CPU implementation shouldn't be too hard. I think lot of code could be copy pasted.

Ttl:
I did the tuning but it didn't change the results at all. I investigated a little bit more and it seems that for small matrices CLBlast GEMM spends more time in checking the inputs and copying the arguments to device than actually doing the matrix product. According to profiler runKernel takes only 20% of the execution time with 64 channel network. Biggest overhead seems to be creating new buffers for offsets, alphas and betas and transferring them to the device on each call.

I think I need to copy the OpenCL GEMM kernel from CLBlast and call it directly for this to be faster.

[...]

Issue #458, opened 2017-12-22
Bad games being uploaded?


Hersmunch:
Found a game uploaded that ends without passing or resignation. http://zero.sjeng.org/view/901a20869e379a162556b5965f498cef9f0c4575f2bc8391c563e9dd977fd03f?viewer=wgo
It is a short game that ends after a black move and reports a white win so perhaps two passes occurred but weren't put in the sgf?

````
(;GM[1]FF[4]RU[Chinese]DT[2017-12-21]SZ[19]KM[7.5]PB[Leela Zero 0.9 ../leela]PW[Human]RE[W+6.5]
--
  |  
  | ;B[jb];W[qq];B[iq];W[rd];B[nb];W[dd];B[mq];W[eq];B[ph];W[ge]
  | ;B[lf];W[db];B[pm];W[rp];B[ln];W[in];B[ig];W[qi];B[el];W[rb]
  | ;B[qp];W[oq];B[hf];W[oh];B[rq];W[pp];B[hp];W[fg];B[dg];W[pf]
  | ;B[ff];W[eg];B[go])
````

The player name doesn't show the hash... hmm
Wait I found another.
http://zero.sjeng.org/view/433aa775f5b4578aba79a571adf24aeea591143d289f89af6be6d2cfa9e149d0?viewer=wgo
and more:
http://zero.sjeng.org/view/377cfae402a169137ffecb3892ea086626692a8efa611d7b4d54023610139fa5?viewer=wgo
http://zero.sjeng.org/view/9543a673320074b86b2aa502560474801eed5dbb3d097cb7bc6fb629daa2f105?viewer=wgo

Is someone uploading bad games?

kimitsu:
Definitely. 2421 games per hour with 216 clients seem like a higher-than-usual number, and you do hit these sgfs quite often (>50%) on the web page as last uploaded self-play, as of now. The training data will get spammed by those. Hope @gcp can filter that out...

barrtgt:
according to https://docs.google.com/spreadsheets/d/1QhYWYQvXBhI5eHJm45fvf97CaV2rRmgrATwVgBFGKoU/htmlview#
we should expect about 1200 games per hour

roy7:
Since the hash they are playing with is "../leela" I'd like to request a feature that the player name actually only show the first 8 characters of the hash itself, not the path to the weights file. Since "../leela" is useless to know what weights file they are using. :) @marcocalignano

killerducky:
[...]

Issue #375, opened 2017-12-13
Heatmap not deterministic on Mac with Intel GPU


jkiliani:
After doing some tests @killerducky encouraged me to, I found out that leelaz heatmap command is not repeatable on my machine. 
[heatmap3.txt](https://github.com/gcp/leela-zero/files/1556300/heatmap3.txt)

I did some testing, repeatedly calling the same heatmap from the starting position, using the supervised network. Here's the summary: Out of 43 identical function calls, 33 were correct, 10 were scrambled. The scrambled calls actually were identical between each other, just incorrect.

I opened a new issue for this since this looks like a major bug, most likely with OpenCL on some architectures.

isty2e:
Wow, this looks like a serious one, the heatmap became a total mess. What was the options for leelaz?

killerducky:
See also #359

jkiliani:
Shouldn't heatmap depend not at all on options? Anyway I used options as suggested by @killerducky in the thread about short games on AMD GPUs. I can look up the specifics in a while, using my phone atm.

isty2e:
Yes, what I was wondering is what options make this happen and how to reproduce it.

jkiliani:
I think it's likely to not be reproducible on every system. Since heatmap shows the raw neural network evaluations, I can only speculate that some hardware/software configurations mess up the NN calculation itself under certain conditions. I hope others here come up with more concrete theories, or ways to test which clients are affected.

odeint:
All those heatmap calls are just in the same Leela console, right? Without restarting or changing options in between?

jkiliani:
Correct, same console, no restarts, identical function call.

marcocalignano:
We are compiling our leelaz with the option -ffast-math. I think this makes the games not reproducible.

> -ffast-math
> This option allows GCC to violate some ANSI or IEEE rules and/or specifications in the interest of optimizing code for speed. For example, it allows the compiler to assume arguments to the sqrt function are non-negative numbers and that no floating-point values are NaNs.
> This option should never be turned on by any `-O' option since it can result in incorrect output for programs which depend on an exact implementation of IEEE or ANSI rules/specifications for math functions.

jkiliani:
@marcocalignano I just tried omitting -ffast-math and -O3, either, in the makefile. Neither removed the occurrences of the corrupted heatmaps.

evanroberts85:
Could it have something to do hardware memory errors, that could cause bits
[...]

PR #523, opened 2017-12-30
F(2x2, 3x3) Winograd convolution for CPU and GPU

Ttl:
See #498 for previous discussion. Winograd convolution is a little more inaccurate than direct convolution and with supervised network the difference was sometimes big enough that the OpenCL self-check failed. With CPU Winograd convolution check passed always in my testing. 5x64 network didn't have this problem.

CPU performance doesn't seem to be that bad with small networks. It currently uses only one core and for autogtp several games can be run at the same time. For example a high-end 8 core CPU with 8 simultaneous games should be about equivalent to one low-end GPU.

Benchmarks with "-t 1":
```
CPU

5x256:
winograd: 1600 evaluations in 113.42 seconds -> 14 n/s
direct: 1600 evaluations in 168.83 seconds -> 9 n/s

6x128:
winograd: 1600 evaluations in 35.78 seconds -> 44 n/s
direct: 1600 evaluations in 55.86 seconds -> 28 n/s

5x64:
winograd: 1600 evaluations in  9.91 seconds -> 161 n/s
direct: 1600 evaluations in 14.83 seconds -> 107 n/s

GPU

5x256:
winograd: 1600 evaluations in  7.96 seconds -> 200 n/s
direct: 1600 evaluations in 16.65 seconds -> 96 n/s

6x128:
winograd: 1600 evaluations in  3.19 seconds -> 501 n/s
direct: 1600 evaluations in  5.52 seconds -> 289 n/s

5x64:
winograd: 1600 evaluations in  1.30 seconds -> 1227 n/s
direct: 1600 evaluations in  1.75 seconds -> 914 n/s
```

There are some tuner parameters in config.h for SGEMM that affect the performance. I'm not sure if the current tuner parameters are good for every GPU and I would be interested in performance tests with other GPUs. Tuners need to be tuned with [CLBlast](https://github.com/CNugteren/CLBlast) tuners. Instructions are in config.h. Even if you are not able or want to compile CLBlast performance comparison to the current next branch would be interesting. Start leelaz with "-t 1" option and use "netbench" for benchmarks.

It should be possible to increase the performance even more with some tweaks in the transformation kernels. Some OpenCL optimizations that I can think of are:

* Cache input in local memory in input transformation. Currently every input value is read 4 times.
* Fused output + batchnorm + relu + input transformation kernel to be used between the residual blocks. This seems complicated and might not be worth the trouble yet with this low number of blocks.
* ~~Use CLBlast indirect sgemm with pre and post transformations included in Winograd in and out transformations.~~ Done

USE_HALF is currently broken with this code.


isty2e:
This somehow fails to run, failing to compile OpenCL kernels.

```
Device ID:     2
Device name:   AMD Radeon Pro 560 Compute Engine
Device type:   GPU
Device vendor: AMD
Device driver: 1.2 (Nov  9 2017 18:48:40)
Device speed:  907 MHz
Device cores:  16 CU
Device score:  1112
Selected platform: Apple
Selected device: AMD Radeon Pro 560 Compute Engine
with OpenCL 1.2 capability
Error building kernels: <program source>:196:36: error: expected expression
        for (int i = 0; i < 4; i++) {
                                   ^
<program source>:205:17: error: use of undeclared identifier 'pred'
            if (pred[i]) {
                ^
```

This must be out_transform_fused_bn in OpenCL.cpp, but I can't figure out why this happened, since the source code seems just fine for me...

Ttl:
There was an invisible unicode character in the source. Can you try again with the latest commit?

isty2e:
It is now running fine. A simple benchmark result:

```
5x64:
winograd:  1600 evaluations in 48.56 seconds -> 32 n/s
direct: 1600 evaluations in  3.11 seconds -> 514 n/s
```

So apparently the tuning parameters are not universal. After tuning:

`winograd: 1600 evaluations in  3.30 seconds -> 485 n/s`

FWIW, the optimal tuning parameters were "KWID=8 MDIMAD=32 MDIMCD=8 NDIMBD=16 NDIMCD=8 PADA=1 PADB=0 PRECISION=32 VWMD=1 VWND=2 WGD=32" for me.

isty2e:
By the way, given that the parameters should be tuned, what happens to the distributed windows binary? Could it be more desirable to store the parameters in a local storage? Maybe it can be useful for multiple GPU systems with different graphic cards.

Ttl:
I think that a tuner needs to be implemented in leelaz that finds the optimal parameters and stores them on disk.

Other problem is that direct convolution is faster for you with 64 channel network. I don't really want to support both of them because CPU needs to use the same algorithm for the OpenCL self-check to work. Hopefully with some optimizations the Winograd convolution is faster for all nets so that direct convolution doesn't need to be used.

How the performance changes for you if you replace the OpenCL compilation args in OpenCL.cpp line 762 with "-cl-mad-enable -cl-fast-relaxed-math -cl-no-signed-zeros -cl-denorms-are-zero -DUSE_STAGGERED_INDICES=1"?

isty2e:
Using staggered indices, I experienced the speed dropping by 10% or so. 

By the way, the clBlast tuner seems to optimize parameters for several GEMM routines. If we restrict this to the set of matmuls needed for 3x3 convolution, we might get better parameters?

Ttl:
The command line to use with tuners is: "./clblast_tuner_xgemm_direct -m 128 -n 128 -k 128 -alpha 1 -beta 0". You can also try with "-m 64" and "-k 64". These settings should be closest to the ones leelaz needs.

isty2e:
Those parameters are obtained the very command, as written in config.h. I wonder why m, n, k are 128 though, could you explain a bit?

Ttl:
Winograd convolution calculates the convolution by rewriting the convolution as 16 multiplications of big matrices.

See https://arxiv.org/abs/1509.09308 and https://colfaxresearch.com/falcon-library/ for more information.

m is number of outputs (=filters in this case), n is number of tiles (100 with 19x19 array) and k is number of filters. CLblast tuner has issues with non-power of two sizes so number of tiles is rounded up. Using 128 for m and k gave better results than 64 even with 64 channel network for me.

isty2e:
I just wondered why m and k are 128, not 64 :) But indeed, m, k=128 yielded better results. I have been playing with parameters a bit, and my conclusions are:

1. Due to the randomness in parameters search, the parameters are somewhat random.
2. This shouldn't matter much if the results are consistent, but even if the flops doesn't fluctuate much, netbench results do.

Ttl:
I replaced the SGEMM OpenCL kernel with faster one used in CLBlast `BatchedGemmIndirect` function. On my computer this is good amount of faster:


```
5x256:
winograd: 1600 evaluations in  6.45 seconds -> 248 n/s
direct: 1600 evaluations in 16.44 seconds -> 97 n/s

6x128:
winograd: 1600 evaluations in  2.66 seconds -> 601 n/s
direct: 1600 evaluations in  5.45 seconds -> 293 n/s

5x64:
winograd: 1600 evaluations in  1.11 seconds -> 1444 n/s
direct: 1600 evaluations in  1.69 seconds -> 949 n/s
```

@isty2e Can you try the new code on your computer? Note that you need to use different tuner this time: "./clblast_tuner_xgemm -m 128 -n 128 -k 128 -alpha 1 -beta 0".

Other testers are also appreciated.

isty2e:
```
5x64:
winograd (XGEMM-direct): 1600 evaluations in  3.30 seconds -> 485 n/s
winograd (XGEMM-indirect): 1600 evaluations in  3.12 seconds -> 512 n/s
direct: 1600 evaluations in  3.11 seconds -> 514 n/s

6x128:
winograd (XGEMM-direct): 1600 evaluations in  6.59 seconds -> 242 n/s
winograd (XGEMM-indirect): 1600 evaluations in  6.48 seconds -> 246 n/s
direct: 1600 evaluations in  7.47 seconds -> 214 n/s
```

So this is more or less faster than direct XGEMM. Again, the result from the tuner was not very stable so I had to cherry pick the best performing one.

MartinDevelopment:
Is there a way to combine this with https://github.com/gcp/leela-zero/pull/471 ? If that were possible it would be amazing. From this alone I am getting a 30% speed increase over the normal next branch win CPU only.

Ttl:
Yes, both can be used at the same time. I tested it before and there aren't even any conflicts in the merge. You can merge the branches manually locally if you want to test it right now.

With both changes especially the supervised network is multiple times faster than before.

isty2e:
@MartinDevelopment They should be totally compatible after resolving conflicts if any. But I am not sure if this will be pulled without writing a decent tuner or at least providing an easy way to tune especially for distributed windows binary.

@Ttl The instability of the clBlast tuner and a low correlation between tuned clBlast performance and leelaz performance is potentially problematic, I suppose. In the long term, wouldn't it be better to write one or at least put some modifications on the clBlast tuner?

Ttl:
@isty2e For release it's necessary to write a tuner into leelaz, but I haven't had the motivation to start writing it yet. I don't mind if someone else wants to write it.

MartinDevelopment:
Could anyone tell me how to merge 2 branches from different people? I can't find a working way of doing it.

MartinDevelopment:
It seems like they work almost perfect together. The only problem is that I don't see the NNCache output in the console after it played a move.

bood:
Unfortunately I didn't see any improvement on my test machine.
I'm using RX 480 card with `-g 6`, it's both around `540 ms/move` before and after apply this PR.

Besides, looks like the parameters given by `./clblast_tuner_xgemm -m 128 -n 128 -k 128 -alpha 1 -beta 0` performs worse than the one provided in the code.

alreadydone:
On Tesla K20 with `-g 1`:
current master branch: ~3200 ms/move, 62-64% GPU usage
with cache-nn: ~2500 ms/move, 48-62% GPU usage
with cache-nn & winograd: ~1900 ms/move, 32-47% GPU usage

On another machine with K20c with cache-nn & winograd:
1 instance     1632ms/move    51%
2 instances     905ms/move    87% 
3 instances     818ms/move    95%
4 instances     765ms/move    95% 
5 instances     741ms/move    95%

jkiliani:
You could try also pulling #526 and #546 and see how your benchmarks look then...

gcp:
@Ttl "Biggest overhead seems to be creating new buffers for offsets, alphas and betas and transferring them to the device on each call."

1) It may be worthwhile to raise this issue with the clBlast dev and explain the performance problem the current API poses, and see if a solution can be found that does not require us to duplicate/fork part of clBlast.

2) How usable are the default clBlast GEMM tunings for us? Even if not optimal, starting from or being able to fall back to its database might be preferable than trying to construct our own.

Ttl:
I have already implemented almost working tuner. I just need to add checking of the output against reference and writing/reading results to disk. Tuning is much simpler when only sgemm needs to be tuned and the current implementation takes about 1 second on my GPU to find similar parameters that CLBlast tuner finds.

The default tuning parameters are not very good. Performance with them is worse than with the previous code and according to the tests above the parameters that were optimal for my GPU didn't work for others. Tuning on the end user computer is in my opinion necessary because there are going to be many GPUs that are not in the CLBlast database.

There are other issues with using CLBlast library with the newest faster sgemm kernel. CLBlast library is currently hard coded to use the slower direct kernel and this faster kernel can't be used with the library methods. The faster kernel also requires the matrices to be zero-padded to certain dimensions that depend on the tuning parameters. If the matrices are not right size CLBlast copies them and adds the necessary padding. Copying would be too slow so the matrix dimensions must be correct before calling the method. I didn't find any method for getting the tuning parameters from the CLBlast library so getting the best performance with the library seems impossible at the moment.

Ttl:
I just finished a simple SGEMM tuner. Not the best code, but it works. This should be ready for testing and review now.

isty2e:
This somehow does not run appropriately for me:

```
...
Device ID:     2
Device name:   AMD Radeon Pro 560 Compute Engine
Device type:   GPU
Device vendor: AMD
Device driver: 1.2 (Nov  9 2017 18:48:40)
Device speed:  907 MHz
Device cores:  16 CU
Device score:  1112
Selected platform: Apple
Selected device: AMD Radeon Pro 560 Compute Engine
with OpenCL 1.2 capability

Started OpenCL SGEMM tuner
Found 578 valid configurations
 -DKWG=32 -DKWI=2 -DMDIMA=8 -DMDIMC=8 -DMWG=16 -DNDIMB=8 -DNDIMC=8 -DNWG=16 -DSA=0 -DSB=0 -DSTRM=0 -DSTRN=0 -DVWM=1 -DVWN=1 : 0.2416 ms
 -DKWG=32 -DKWI=2 -DMDIMA=8 -DMDIMC=8 -DMWG=32 -DNDIMB=8 -DNDIMC=8 -DNWG=16 -DSA=0 -DSB=0 -DSTRM=0 -DSTRN=0 -DVWM=1 -DVWN=1 : 0.1695 ms
 -DKWG=32 -DKWI=2 -DMDIMA=8 -DMDIMC=8 -DMWG=16 -DNDIMB=8 -DNDIMC=8 -DNWG=32 -DSA=0 -DSB=0 -DSTRM=0 -DSTRN=0 -DVWM=1 -DVWN=1 : 0.1692 ms
 -DKWG=32 -DKWI=2 -DMDIMA=8 -DMDIMC=8 -DMWG=32 -DNDIMB=8 -DNDIMC=8 -DNWG=32 -DSA=0 -DSB=0 -DSTRM=0 -DSTRN=0 -DVWM=1 -DVWN=1 : 0.1250 ms
 -DKWG=32 -DKWI=2 -DMDIMA=16 -DMDIMC=16 -DMWG=64 -DNDIMB=16 -DNDIMC=16 -DNWG=64 -DSA=0 -DSB=0 -DSTRM=0 -DSTRN=0 -DVWM=1 -DVWN=1 : 0.1240 ms
libc++abi.dylib: terminating with uncaught exception of type cl::Error: clEnqueueNDRangeKernel
zsh: abort      ./leelaz -t 1 -w
```

Ttl:
The latest commit should hopefully fix it.

isty2e:
```
5x64:
direct: 1600 evaluations in  3.11 seconds -> 514 n/s
winograd (XGEMM-direct): 1600 evaluations in  3.30 seconds -> 485 n/s
winograd (XGEMM-indirect): 1600 evaluations in  3.12 seconds -> 512 n/s
winograd-tuned: 1600 evaluations in  3.04 seconds -> 526 n/s

6x128:
direct: 1600 evaluations in  7.47 seconds -> 214 n/s
winograd (XGEMM-direct): 1600 evaluations in  6.59 seconds -> 242 n/s
winograd (XGEMM-indirect): 1600 evaluations in  6.48 seconds -> 246 n/s
winograd-tuned: 1600 evaluations in  5.74 seconds -> 278 n/s
```

Apparently the tuner is working well and the results are better now.

isty2e:
I gave it a try with tuning with a larger matrix, but after a series of failures (probably cl::Error: clEnqueueNDRangeKernel), the "Failed to find a working configuration." message is printed out.

bood:
This implementation seems to be breaking the heatmap according to @alreadydone in #458, I think a further investigation needs to be done.

isty2e:
That is strange. I get exactly the same heatmap with this. Probably the OpenCL driver is broken for @alreadydone or something? And here is a probably good news: #375 does not happen with this Winograd transformation.

alreadydone:
@isty2e Notice that the output from the CPU version (compiled by setting USE_OPENCL undefined) gives the same output as GPU. Maybe this problem only occurs when cache-nn and winograd are used together? Can you test with [this](https://github.com/alreadydone/leela-zero)?

isty2e:
@alreadydone I already has both merged in the version I am running for test, so that should be not it. I wonder if you witnessed my comment edited: I first tested with 5x64 and it was fine, but with the 6x128 network (weights.txt), I also get exactly the same heatmap you mentioned, which is different from that obtained with the direct convolution.

Ttl:
Winograd convolution has bigger rounding errors than direct convolution. It might not give the exact same result, but the difference should not be big. That's why Winograd convolution was also needed on CPU so that the OpenCL self-check passes.

@isty2e I guess the maximum error is exceeded. Can you try replacing cl_args on OpenCL.cpp line 46 with empty string and trying if it passes then. If it still doesn't work try increasing MAX_ERROR in Tuner.cpp line 47. The current error limit is pretty strict.

alreadydone:
@isty2e I tested the 5x64 network 55bf55dc (current best), and indeed all platforms and versions I mentioned in #458 and they all give the same result as follows:
>     Leela: heatmap
>       0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
>       0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
>       0   0   8   4   1   0   0   0   0   0   0   0   0   0   1   3  11   0   0
>       0   0   4 169   0   1   0   0   0   0   0   0   0   1   0 176   4   1   0
>       0   0   1   0   0   0   0   0   0   0   0   0   0   0   0   0   1   0   0
>       0   0   0   1   0   0   0   0   0   0   0   0   0   0   0   1   0   0   0
>       0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
>       0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
>       0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
>       0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
>       0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
>       0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
>       0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
>       0   0   0   1   0   0   0   0   0   0   0   0   0   0   0   1   0   0   0
>       0   0   1   0   0   0   0   0   0   0   0   0   0   0   0   0   2   0   0
>       0   0   3 178   0   1   0   0   0   0   0   0   0   1   0 192   3   1   0
>       0   0   9   4   1   0   0   0   0   0   0   0   0   0   1   3  10   0   0
>       0   0   0   1   0   0   0   0   0   0   0   0   0   0   0   1   0   0   0
>       0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
>     pass: 0
>     winrate: 0.514386

So this problem indeed seems only pertaining to larger networks, on which winograd results in bigger errors, which if I understand correctly is what @Ttl suggests above.

isty2e:
@Ttl Increasing the MAX_ERROR worked, with the heatmap preserved.

It is expected that this implementation can lead to different results from that of a direct one, but I can hardly imagine why the error is negligible for 5x64 (as tested by me and @alreadydone) and suddenly increases for 6x128.

isty2e:
So this has something to do with the number of channels. I made a 5x128 and a 6x64 network and tested. For the 6x64 network, the error was not visible, but for the 5x128 one, the difference was quite significant.

isty2e:
I ran a benchmark for Intel HD Graphic 630/Mac for fun:

```
5x64:
direct: 1600 evaluations in 88.04 seconds -> 18 n/s
winograd: 1600 evaluations in  4.58 seconds -> 349 n/s
```

After a few netbenches, the direct convolution failed the self-check and terminated, while winograd one did not, which also gives with a reasonable speed. Of course that speed difference between Intel and AMD chips is more than it should be I believe, somehow Intel graphics driver for mac should not be working well with the previous method apparently.

And for me, the parameters tuned with m=k=64 worked slightly better than those from 128. I hope as many people as possible to test and report the results here.

Ttl:
Here's some data about Winograd convolution accuracy with different amount of channels. Data is normally distributed with mean 0 and std_dev 1. Reference convolution is naive implementation on CPU using double precision.

```
Convolution size, maximum diff, average diff
64x64x19x19: 3.43323e-05, 5.30173e-06
128x128x19x19: 8.39233e-05, 1.0371e-05
256x256x19x19: 1.79291e-04, 2.0342e-05
```

Because Winograd calculates one 2x2 tile at a time one input value can affect the other nearby output value precision. The error is bigger when values have very different magnitudes.

It seems that 6x128 network output is sensitive to errors. 5x256 network that is still almost random that I used for benchmarks doesn't have issue with accuracy even though accuracy in convolution is even worse.

isty2e:
@Ttl The 5x128 network I used for benchmark was a simple truncation of the supervised 6x128 network, and it showed some accuracy issues. Maybe the error in heatmaps can be larger if the output values are condensed?

isty2e:
Anyway, if the non-negligible error is unavoidable, the question is that whether this should be pulled or not. This can give a lot of boost in speed especially for larger networks which is to be presumably used in the next run, but involves some errors. In terms of consistency, self-play games will not be affected in general, but the possible discrepancy between the TF-trained model and the actually running one can be a thing. @gcp I would like to know what your opinion is.

gcp:
>the possible discrepancy between the TF-trained model and the actually running one can be a thing

Well in theory TF is using cuDNN which under the hood is likely using Winograd. But for them it wasn't trivial either (https://github.com/tensorflow/tensorflow/issues/9339).

My main problem with this pull request is that we end up duplicating large parts of clBlast. On the other hand, being able to tune to the users' GPU on startup is pretty darned awesome.

isty2e:
AFAIK the Winograd convolution is currently available in TF by setting TF_ENABLE_WINOGRAD_NONFUSED environment variable to 1, but I am not convinced that it would lead to the same result (though theoretically it should). Maybe worth testing?

And I am almost ignorant at managing projects like this, so I wonder what would be the problem if we clBlast code. It is under Apache license, so shouldn't there be any problem as far as license is concerned?

roy7:
I no nothing about any of these libraries, but when I see GCP mention "we end up duplicating large parts of clBlast", it does make me wonder why not just include clBlast in the project if it's useful and has a compatible open source license. :)

roy7:
@Ttl Ahh ok. :) I've not paid too much attention to discussions like this because it's just way over my pay grade.

gcp:
>It is under Apache license, so shouldn't there be any problem as far as license is concerned?

It's not the license, but if clBlast makes improvements to their code (performance, compatibility, whatever) we won't get them unless we translate the changes back.

Ttl:
I have filed an issue with CLBlast (https://github.com/CNugteren/CLBlast/issues/237).

isty2e:
I assume it will definitely take some time to resolve, especially if it is related to the interface. On the other hand, I would really like this to be pulled as soon as possible, especially with this easy-to-use tuner. What would be the remaining tasks?

1. Determining m and k: As written above, m=k=64 led to better results than m=k=128 for me. Also, this could be determined after reading the network file so that it can be changed based on the number of filters in the network?
2. Checking consistency with TF+Winograd.
3. Having more people to test this.

gcp:
> I would really like this to be pulled as soon as possible

It's a massive code change that introduces difficult to quantity functional differences. I like the speedups as much as anyone but don't count on this getting merged any time soon. It'll take a while to review.

isty2e:
Of course, I am well aware of it. I hope this to be merged for 1.1 (if the versioning principle does not change) release though.

gcp:
@Ttl I saw your comments about GemmStridedBatched in clBlast. How does it relate to this pull request?

For example, does

>The faster kernel also requires the matrices to be zero-padded to certain dimensions that depend on the tuning parameters.

Still block you from reaching comparable performance with clBlast?

gcp:
Current version merged in winograd branch, will do cleanup work there.

Ttl:
Implementation using CLBlast is still quite bit slower than this one even with the new `GemmStridedBatched` function.

```
next:
 1600 evaluations in  1.65 seconds -> 971 n/s

CLBlast StridedGemmBatched:
 1600 evaluations in  1.55 seconds -> 1029 n/s

Copied StridedGemmBatched kernel + hard coded arguments:
 1600 evaluations in  1.46 seconds -> 1097 n/s

This pull request:
 1600 evaluations in  1.13 seconds -> 1422 n/s
```
Library overhead seems to be about 10% with 64 channel network compared directly calling the kernel and hard coding the constant arguments.

src/Network.cpp
```diff
@@ -378,12 +643,19 @@ void Network::forward_cpu(std::vector<float>& input,
     // Input convolution
     constexpr int width = 19;
     constexpr int height = 19;
+    constexpr int tiles = (width + 1) * (height + 1) / 4;
     // Calculate output channels
     const auto output_channels = conv_biases[0].size();
+    //Assumes that residual blocks are identical and have same
+    //number of inputs and outputs
```
gcp: I think that's a requirement for it to work as you need to add the input to the second layer.

src/Network.cpp
```diff
@@ -378,12 +643,19 @@ void Network::forward_cpu(std::vector<float>& input,
...
+
+    const auto input_channels = output_channels;
     auto conv_out = std::vector<float>(output_channels * width * height);
 
-    convolve<3>(output_channels, input,
-                conv_weights[0], conv_biases[0], conv_out);
+    auto V = std::vector<float>(16*input_channels*tiles);
+    auto M = std::vector<float>(16*output_channels*tiles);
+
+    winograd_convolve3(output_channels, input, conv_weights[0], V, M, conv_out);
```
gcp: Might as well remove the conv_biases[] as this point.

src/OpenCL.h
```diff
@@ -135,14 +141,24 @@ class OpenCL_Network {
...
 private:
+    void tune_sgemm(void);
+    void process_tuners(std::string tuners);
+
     cl::Program m_program;
+    std::string m_cl_args;
 
+    size_t m_sgemm_mwg, m_sgemm_nwg, m_sgemm_kwg;
+    size_t m_sgemm_vwm, m_sgemm_vwn;
+    size_t m_sgemm_mdimc, m_sgemm_ndimc;
```
gcp: Maybe make a struct for these?

src/Tuner.cpp
```diff
@@ -0,0 +1,424 @@
...
+#include <Accelerate/Accelerate.h>
+#endif
+#ifdef USE_MKL
+#include <mkl.h>
+#endif
+#ifdef USE_OPENBLAS
+#include <cblas.h>
+#endif
+
+#define TUNER_FILE_LOCAL "leelaz_tuners"
```
gcp: const(expr) auto = std::string

src/Tuner.cpp
```diff
@@ -0,0 +1,424 @@
...
+#define MAX_ERROR 1e-4
```
gcp: constexpr auto

src/Tuner.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+using namespace Utils;
+
+static void sgemmBatched_ref(const std::vector<float>& a, const std::vector<float>& b,
```
gcp: This is winograd_sgemm from Network.cpp?

src/Network.cpp
```diff
@@ -267,6 +365,173 @@ void Network::initialize(void) {
 }
 
 #ifdef USE_BLAS
+void winograd_transform_in(const std::vector<float>& in, std::vector<float>& V,
```
gcp: static

src/Tuner.cpp
```diff
@@ -0,0 +1,424 @@
...
+                    &a[offset_u], m,
+                    &b[offset_v], n,
+                    0.0f,
+                    &c[offset_m], n);
+    }
+}
+
+
+bool IsMultiple(const size_t a, const size_t b) {
+    return ((a/b)*b == a);
```
gcp: Isn't this just "% == 0"?

src/Tuner.cpp
```diff
@@ -0,0 +1,424 @@
...
+        return false;
+    }
+    if (p["SA"] != p["SB"]) {
+        return false;
+    }
+    return true;
+}
+
+
+Parameters Tuner::get_parameters_by_int(const std::vector<Configurations> opts, const int n) {
```
gcp: Take opts by ref?

src/Tuner.cpp
```diff
@@ -0,0 +1,424 @@
...
+            auto s = o.first;
+            auto v = o.second[j % choices[c]];
+            j /= choices[c];
+            param[s] = v;
+        }
+
+        return param;
+}
+
+std::string Tuner::parameters_to_string(const Parameters p) {
```
gcp: const ref

src/Tuner.cpp
```diff
@@ -0,0 +1,424 @@
...
+    std::string s;
+    for (auto const& x : p) {
+        s += " -D" + x.first + "=" + std::to_string(x.second);
+    }
+    return s;
+}
+
+size_t next_power_of_two(const size_t x) {
```
gcp: static

src/Tuner.cpp
```diff
@@ -0,0 +1,424 @@
...
+    return 2 << (size_t)(std::ceil(std::log2(x)) - 1);
+}
+
+static void sgemm_generate_data(std::vector<float> &x, const int m, const int n,
+        const int batch_size, const int m_ceil, const int n_ceil) {
+    for (auto batch = 0; batch < batch_size; batch++) {
+        for (auto i = 0; i < n_ceil; i++) {
+            if (i < n) {
+                for (auto j = 0; j < m; j++) {
+                    x[batch*n_ceil*m_ceil + i*m_ceil + j] = 0.01f*(((i ^ j) + batch - 50) % 100);
```
gcp: Won't i ^ j overflow? Why not random floats?

src/Tuner.cpp
```diff
@@ -0,0 +1,424 @@
...
+                sum += (r - y) * (r - y);
+            }
+        }
+    }
+    return sum/(m*n);
+}
+
+std::string Tuner::tune_sgemm(const int m, const int n, const int k, const int batch_size, const int runs) {
+
+    std::vector<Configurations> opts = {
```
gcp: Ah, the how much tuning dilemma. I think you already pruned the possibilities from CLBlast here, right?

I'd propose this:
a) Add an --full-tuning or --exhaustive-tuning flag, which tries all sensible combinations (including the ones already pruned here).
b) Run it on a few systems you have access to. Include the results in the source.
c) By default, we only test the exact combinations from (b).
d) If exhaustive-tuning is requested, and we get an improvement of more than 5% over something from (b), mark that in the tuned output.
e) Ask people to send in the output from (d). Add to source if >5% better.

This should allow getting to within say 5% of optimal performance with maybe 20-30 attempts, covering all common GPU. People who want the last 5% can use the flag.

src/Tuner.cpp
```diff
@@ -0,0 +1,424 @@
...
+                                  (size_t)batch_size};
+
+        auto sum = 0.0;
+        auto max_error = 0.0f;
+        for (auto r = 0; r < runs; r++) {
+
+            try {
+                queue.enqueueNDRangeKernel(sgemm_kernel, cl::NullRange,
+                                            size_sgemm, local_sgemm,
+                                            nullptr, &event);
```
gcp: You don't need the event here.

src/Tuner.cpp
```diff
@@ -0,0 +1,424 @@
...
+        }
+        if (max_error < MAX_ERROR && (best_time == 0 || sum < best_time)) {
+            myprintf("%s : %.4f ms\n", defines.c_str(), 1e-6*sum/runs);
+            best_time = sum;
+            best_params = defines;
+        }
+
+    }
+    if (best_time == 0) {
+        myprintf("Failed to find a working configuration.\nCheck your OpenCL drivers.\n");
```
gcp: Make this a printf to make it more likely it gets shown to the user.

src/Tuner.cpp
```diff
@@ -0,0 +1,424 @@
...
+        std::exit(-1);
```
gcp: Current code does a throw in these cirumstances.

src/Tuner.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include "config.h"
+#include <vector>
+#include <map>
+#include <string>
+
+using Configurations = std::pair<std::string, std::vector<size_t>>;
+using Parameters = std::map<std::string, size_t>;
+
+#define TUNER_VERSION 0
```
gcp: constexpr in the class

src/clblast_level3/xgemm_part4.opencl
```diff
@@ -0,0 +1,140 @@
...
+  #elif SB == 1
+    XgemmBody(kSizeN, kSizeN, kSizeK, agm, bgm, cgm, alpha, beta, blm);
+  #else
+    XgemmBody(kSizeN, kSizeN, kSizeK, agm, bgm, cgm, alpha, beta);
+  #endif
+}
+
+// =================================================================================================
+// If not using a triangular version, include the regular kernel
+#else
```
gcp: We could strip the above, and stuff referring to complex ops from the other files, I guess. I think that if we're going to pull in this part of CLBlast, it's worthwhile. Of course this can be avoided if we can just add it as a dependency.

src/OpenCL.cpp
```diff
@@ -52,285 +56,212 @@ static std::string sourceCode_config = R"(
...
-        const int o = output;
-        const float bias = vload_net_t(o, biases);
+        const float mean = vload_net_t(k, means);
+        const float scale_stddiv = vload_net_t(k, stddivs);
 
-        float sum = bias;
-        for (int c = 0; c < channels; c++) {
-            sum += vload_net_t((c * boardsize + b) * outputs + o, in);
+        for (int i = 0; i < 4; i++) {
+            o[i] = scale_stddiv * (o[i] - mean);
```
gcp: I think this can be merged with the next loop.

src/OpenCL.cpp
```diff
@@ -507,107 +449,119 @@ void OpenCL_Network::forward(const std::vector<net_t>& input,
...
-        channelShift = 1;
+    try {
+        in_transform_kernel.setArg(0, bufferInOut);
+        in_transform_kernel.setArg(1, bufferV);
+        in_transform_kernel.setArg(2, channels);
+        in_transform_kernel.setArg(3, k_ceil);
+        in_transform_kernel.setArg(4, n_ceil);
+
+        queue.enqueueNDRangeKernel(in_transform_kernel, cl::NullRange,
+                                   cl::NDRange(wgs, channels));
```
gcp: I think this introduces a problem that if channels is not a power of 2, you require OpenCL 2.0. (The current code works with e.g. 192 or 384)

In general for squeezing performance out of OpenCL you're going to want to control the workgroup size and 2d blocking yourself.

src/OpenCL.cpp
```diff
@@ -507,107 +449,119 @@ void OpenCL_Network::forward(const std::vector<net_t>& input,
...
+            if (bufferResidual) {
+                out_transform_bn_kernel.setArg(5, *bufferResidual);
+            } else {
+                out_transform_bn_kernel.setArg(5, nullptr);
+            }
+            out_transform_bn_kernel.setArg(6, (*bn_weights)[0]);
+            out_transform_bn_kernel.setArg(7, (*bn_weights)[1]);
+
+            queue.enqueueNDRangeKernel(out_transform_bn_kernel, cl::NullRange,
+                                       cl::NDRange(outputs, wgs));
```
gcp: Same remark wrt outputs and power of 2.

src/OpenCL.cpp
```diff
@@ -664,6 +618,97 @@ static std::string trim(std::string trim_me) {
...
+        if (!vwn) {
+            std::cerr << " VWN";
+        }
+        std::cerr << std::endl;
+        std::exit(-1);
+    }
+}
+
+std::vector<size_t> OpenCL::get_sgemm_tuners(void) {
+    auto tuners = std::vector<size_t>(7);
```
gcp: Just emplace instead of preallocating this?

src/OpenCL.cpp
```diff
@@ -772,19 +818,29 @@ void OpenCL::initialize(void) {
...
+                                + sourceCode_sgemm);
     } catch (const cl::Error &e) {
         myprintf("Error getting kernels: %s: %d", e.what(), e.err());
         throw;
     }
+
+    m_cl_args = cl_args;
+
+    auto t = Tuner();
+    auto sgemm_tuners = t.load_sgemm_tuners(128, 100, 128, 16);
```
gcp: Magic constants? Also you're tuning 128x6 here instead of the actual network size?

src/Tuner.cpp
Ttl: event is used for timing the kernel.

src/Tuner.cpp
Ttl: Isn't ^ xor operation for ints? I can't use the builtin random generator here, because it would mess the reproducibility of the games. This is just something pseudo-random data with uniform like distribution.

gcp: Ugh yes of course I'm not sure what I was smoking here.

You can just instantiate a Random() instance with a fixed seed here.

src/Tuner.cpp
Ttl: CLBlast has two tuning methods for sgemm. Reduced set of arguments that are tried exhaustively and full set that are tried randomly. This is the reduced set.

I agree that your changes would be good, but I don't have the motivation to implement them. I just wanted to learn OpenCL when I started this pull request and this change goes too far. I'll leave the tuner improvements for someone else.

isty2e: The implement tuner did not consume too much time, so is there any reason for configurations not to remain as is?

gcp: It doesn't take too much time because it only tries a limited subset, leaving some performance on the table. I'm thinking of ways to recover that without making the first startup take 15 mins for everyone.

src/OpenCL.cpp
Ttl: This works with non-power of two arguments for me and I don't think I'm using OpenCL 2.0.

src/Tuner.cpp
gcp: The exhaustive tuner could likely also be much faster by making it use a genetic algorithm, or CEM or whatever optimization method, rather than blindly trying all combinations. But realistically there's maybe 20 common models of GPU, so if you have the tuning for most of those, it could finish near instantly.

src/OpenCL.cpp
gcp: Hmm, I guess it's OK because you're not specifying a local range, so it will find a divisor of your global one.

src/Tuner.cpp
marcocalignano: This is faster.
```
size_t r = 1;
while (x >>= 1) {
    r <<= 1;
}
return  r << 1;
```

Armavica: These two implementations differ on powers of 2, though. The original leaves `x` unchanged whereas this one doubles the number. I am not sure if this matters.

marcocalignano: Of course x cannot be const and then is a local variable so it will be change inside this function only. and why do you say the results differ?

Armavica: [This](https://paste.fulltxt.net/3,F=uf) prints [that](https://paste.fulltxt.net/xXM@O38_7d9) on my machine.

marcocalignano: You are right try this: 

size_t next_power_of_two_loop(size_t x) {
    size_t r = 1;
    r <<= ( x % 2);
    while (x >>= 1) {
        r <<= 1;
    }
    return r ;
}

marcocalignano: And this on top fixes the equation for x = 1.

Armavica: This does not seem to work either (for even numbers that are not powers of 2)… Something like
```C
x--;
x |= x >> 1;
x |= x >> 2;
x |= x >> 4;
x |= x >> 8;
x |= x >> 16;
x |= x >> 32;
return x + 1;
```
should be okay, assuming `size_t` is 64 bits.

Ttl: Speed isn't really that important here. I'm leaving the current implementation since it works and is more readable than alternatives.

marcocalignano: This should work but I need to see if it is better that the original (except the erron when x is 1)
```
size_t next_power_of_two_loop(const size_t x) {
    size_t r = 1;
    size_t n = x;
    while (n >>= 1) {
        r <<= 1;
    }
    r <<= (x != r)  ;
    return r;
}
```

Ttl:
I hope I got the every change request. I left the data generation in tuner as is, because it wasn't immediately clear how to get random floats in range [-1, 1] from the random generator. The exact data doesn't matter anyway and the current one should work fine.

I did try setting the transform kernel local sizes before, but leaving it for the driver worked best. Maybe someone who knows OpenCL can find some better local sizes.

I'll leave the tuner improvements for someone else.

Also tuner should probably write the best tuners to some other place than the current directory. When leelaz is packaged and installed for example in "/usr/bin" it can't write anything there. It probably should write them to for example in "~/.config/leelaz/tuners" and somewhere similar with other operating systems. This isn't a problem with autogtp because leelaz is in the same directory and user should always have access there.

src/OpenCL.cpp
```diff
@@ -403,36 +344,48 @@ void OpenCL_Network::forward(const std::vector<net_t>& input,
...
         opencl_thread_data.m_inBuffer = cl::Buffer(
             CL_MEM_READ_WRITE, alloc_inSize);
         opencl_thread_data.m_tmpBuffer = cl::Buffer(
             CL_MEM_READ_WRITE, alloc_inSize);
         opencl_thread_data.m_residualBuffer = cl::Buffer(
             CL_MEM_READ_WRITE, alloc_inSize);
-        opencl_thread_data.m_mergeBuffer = cl::Buffer(
-            CL_MEM_READ_WRITE | CL_MEM_HOST_NO_ACCESS, alloc_mergeSize);
+        opencl_thread_data.m_VBuffer = cl::Buffer(
+            CL_MEM_READ_WRITE | CL_MEM_HOST_NO_ACCESS | CL_MEM_COPY_HOST_PTR, alloc_vm_size, v_zeros.data(), nullptr);
```
gcp: Does this actually need to be zeroed? It feels weird because it's not going to be used and filled up. Has to do with padding?

gcp:
I cleaned up everything to my liking and fixed an issue I found on macOS. (Trivia: on macOS the Intel HD 4000 now outperforms my GTX650M, even though the tuner suggests the latter does the kernels much faster. Thermal throttling maybe?)

Just want to do a strength check to see the effect of accuracy loss (if any) and I'll merge this to next.

On my GTX1060 I seem to get a good boost from exhaustive tuning, so I'd still want to add that.

gcp:
On CPU, winograd_transform_in/out are very visible in profiles. Looking at the assembler, it seems to compiler isn't able to vectorize them, I guess due to all the reordering going on. So I guess on CPU more gains should be possible.

src/OpenCL.cpp
Ttl: This is for zero padding. It's zeroed at the beginning once and the padding is not touched during the operations.

Ttl:
CPU Winograd transformations were literally copy pasted from the OpenCL kernels. I just added loops around them. I haven't spent any effort in trying to optimize them.

Optimizing the transformations on GPU should also give good speed up especially for networks with small amount of channels. Currently transformation take about 50% of the runtime for 64 channel network on my computer.

Adding local memory to input transformation to avoid loading every value 4 times from the global memory would be one optimization. I tried to implement it, but didn't get the indexing to work correctly.

gcp:
I have some work in progress on an exhaustive tuning option, but I think I found a bug in the existing tuning code.

The OpenCL buffers are allocated according to:
https://github.com/gcp/leela-zero/blob/898fa8453e7fdbed74b5a680bd1d8e5cae283f5c/src/Tuner.cpp#L215

Which is based on sizes like "batch_size * next_power_of_two(k * m)" that I admittedly don't entirely understand at this point.

However, the memory required is potentially quite different:
https://github.com/gcp/leela-zero/blob/898fa8453e7fdbed74b5a680bd1d8e5cae283f5c/src/Tuner.cpp#L266

Note that those bounds depend on things like the selected MWG & NWG. And those are the actual bounds used in the OpenCL call. 

What if those are bigger (by more than 1 order of 2) than the network size being tuned?

I think that if you try to run the tuner with a small network, it will crash.

Ttl:
When `MWG` is less than `m` the maximum value for `m_ceil` is `next_power_of_two(m)`. However when `MWG` exceeds `m` it can go up to `next_power_of_two(MWG)`.

I think it will allocate too small buffer when` MWG` exceeds `next_power_of_two(m)` and same with `n` and `k`. This doesn't happen with 64 channel network and this tuner because maximum `MWG` is 64. Exhaustive tuner tries with `MWG` up to 128 which can cause crash.

But is there really a point in using that large `MWG`? At that point when the limit is exceeded there is more zero padding than actual data. I don't think that is going to be faster than just using smaller `MWG` with half the zero padding. Adding a check that `MWG` is less than `next_power_of_two(m)` and `NWG` is less than `next_power_of_two(n)` should guard against it and prune obviously too large values.

Now that I look at the code I think that `next_power_of_two(k * m)` should be replaced with `next_power_of_two(k) * next_power_of_two(m)` to be safe in case the tuner is called with non-power of two `m` or `k`. `next_power_of_two(k)` should also be at minimum the maximum value of `KWG`.

zediir:
This probably needs some kind of a message in autogtp to the user if the first time they run leelaz and the tuning is through autogtp or the user might think the program is stuck and terminate it before the tuner has finished.

marcocalignano:
I am sorry but I didn't get what autogtp should do.

gcp:
@marcocalignano In this new branch the first run of Leela Zero will tune the parameters to the machine used on start. This can take a while and there will be no move output while it happens. So it would be good if autogtp displayed a message.

Right now Leela will output 
Started OpenCL SGEMM tuner.
to stderr when this starts.

marcocalignano:
```
When I try to build leelaz with cmake I have the following error: 
CMake Error at CMakeLists.txt:20 (add_subdirectory):
  The source directory

    /home/marcuzzo/workspace/leela-zero/gtest

  does not contain a CMakeLists.txt file.

```

gcp:
@marcocalignano check out the README, you need to pull in the gtest submodule.

marcocalignano:
@gcp autogtp starts leelaz with the option -q (quiet) with this option I do not see ant output. Maybe you need to change the code from myprintf to printf like for the OpenCl driver error?

gcp:
Hmm, but that would break the GTP compliance, so I can't do that. (It's ok to do that on a fatal error, but this isn't an error)

gcp:
(Also note the code we're talking about is in the winograd branch, not the next branch)

zediir:
Should autogtp just assume leelaz will run tuners if it does not find the leelaz_tuners file?

gcp:
That requires autogtp to know about that file though, which will be annoying if it's not in the current dir. (See remark above about linux packaging etc)

marcocalignano:
@gcp you couls print that string as a GTP comment. (https://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html#SECTION00039000000000000000)

gcp:
@marcocalignano Ok done!

marcocalignano:
Where did you commit it?

gcp:
It's in winograd now.

marcocalignano:
Is there also a comment when it is finished tuning. If not I need that to know when to sent the first GTP command.

gcp:
You can send commands while it's tuning. It just won't reply till it's done.

gcp:
Tuning on a laptop may be tricky. My Macbook's performance varies by a factor 2, I guess due to thermal throttling.

marcocalignano:
GTP print the tuning comment now see #636

zediir:
Intel HD Graphics 4000 @ 1200MHz

next-branch ~8000 ms/move -> winograd with quick-tuner ~5000 ms/move -> winograd with full-tuner ~4100 ms/move

Ttl:
When autogtp is started with "-g" bigger than one all leelaz instances will start tuning at the same time. It finds the same tuning values for me as one instance, but there might be some potential issues with it and at least it's a waste of resources.

It might be a good idea to do tuning using a single instance at the start. I think starting a single leelaz instance, sending "quit" to it and waiting it to exit would work.

gcp:
@marcocalignano I guess in general it may be a good idea to stagger startup of leelaz instances when -g is used. i.e. not launch new ones until the old ones have responded to GTP commands.

gcp:
Rebased to next.

I think that there's still going to be a minor bug if you use exhaustive tuning with multiple network sizes. All but the last size tested will be removed if you force retuning.

marcocalignano:
@gcp what about a command line option like --tune that makes the tuning and exit so I can run that and then start with all threads?

gcp:
Added --tune-only.

marcocalignano:
Does leelaz need to be tuned every time a new network is downloaded?

jkiliani:
With my MacBook Pro (Intel Iris 6100), the tuner finds multiple configurations acceptable. However, the automatically picked one later fails the self-check. Can there be an option for the tuner to do a stricter self-check with multiple tries?

isty2e:
@jkiliani It could be a problem with the OS X itself. If a configuration passes the test during tuning, its error cannot be larger than the preset bound, but the error detected with the self-check is usually much larger than that; I once saw an error of 7000% or something.

jkiliani:
The problem with my system in the past was that a single call of heatmap did not reliably detect it. If the tuner self-test works by comparing the error of a single neural net eval with the OpenBLAS output, it may stay within the error bound and pass the test. Later, this same configuration may fail during play in a repeat call.

killerducky:
@jkiliani your system has the OpenCL driver bug right? Winograd is not going to solve or workaround that bug. Only using CPU (avoiding OpenCL) will be a safe workaround.

jkiliani:
@killerducky That's a pity, but thanks for your input. I'll continue with the CPU then.

zediir:
@marcocalignano leelaz saves the tuning information to leelaz_opencl_tuning file after tuning.

 Will it tune again if it encounters one of the 6x128 matches when running autogtp?

isty2e:
Currently the tuner seems to write the only set of tuning parameters which fits the network size, unlike the previous version. Wouldn't it be better to store all the tuned parameters for several m and k values, as well as those for different GPUs, as in the previous version of the tuner? Tuning again every time whenever the network size changes does not seem very efficient.

zediir:
I'm getting 

>Error	C2398	Element '3': conversion from 'const int' to 'cl::size_type' requires a narrowing conversion	leela-zero	D:\My Folders\Documents\leela-zero-next\src\OpenCL.cpp	514	

when I try to build with MSVC.

marcocalignano:
So to make the tuning I need any 5x64 network right?
But then, autogtp create **one** tuning parameters file for all the threads, and if a new network (with a different structure) comes in, leelaz will **overwrite** this file with the new tuning (braking it fir all the other thread still working with the old structure network.

zediir:
It will append it to the file right? At least when I ran with -g2 it wrote the tuning parameters twice to the file.

gcp:
The file is read once at startup so overwriting sounds like it wouldn't be an issue.

It seems using comments is going to break other GTP clients (it's breaking ringmaster). We'll have to find something else.

gcp:
>Currently the tuner seems to write the only set of tuning parameters which fits the network size, unlike the previous version. Wouldn't it be better to store all the tuned parameters for several m and k values, as well as those for different GPUs, as in the previous version of the tuner?

"Yes."

marcocalignano:
@gcp that means then with the option --tune-only you wouldn't need to give a network on the command line but the turner does the tuning for "all" networks. 
And since autogtp will run leelaz with  --tune-only before starting the threads the GTP comment about tuning is not needed anymore because the tuning file is already there when the threads start.
I think this llok like a good plan.

Ttl:
It seems that tuning for different network sizes are not saved anymore. Previously different tuners were stored for 64 and 128 channel networks.

gcp:
>that means then with the option --tune-only you wouldn't need to give a network on the command line but the turner does the tuning for "all" networks.

The program is currently flexible in the network setup it supports so tuning "all" networks is not possible. I don't like this.

>It seems that tuning for different network sizes are not saved anymore. Previously different tuners were stored for 64 and 128 channel networks.

I tried to support multiple GPU and overwriting the existing tuning with --full-tuner, but it doesn't work correctly yet with multiple network configs.

gcp:
```
leelaz-46ff v leelaz-46ff-wn (359/1000 games)
board size: 19   komi: 7.5
leelaz-46ff       172 47.91%       84  46.67%     88  49.16%    177.35
leelaz-46ff-wn    187 52.09%       91  50.84%     96  53.33%    154.18
                                   175 48.75%     184 51.25%
```

Absolutely no ill effects seen from any loss in accuracy. (The heatmaps seem very close anyway)

gcp:
Do we want to include some of the logic from here?
https://github.com/CNugteren/CLBlast/blob/b35e3d1e5326cdc257daa170eb243800616cfc26/src/utilities/compile.cpp#L37

Ttl:
I had staggered offsets for AMD before, but removed it because @isty2e said it was slower. Staggered offsets were added to CLBlast, because there were some issues with AMD cards without it on large matrices (https://github.com/CNugteren/CLBlast/issues/53). I don't think we need it.

I didn't think the performance on ARM GPUs was too important. Is anyone using leelaz on ARM?

Inline keyword didn't have any effect on performance for me. Since it was the last compilation option left I took it out too.

gcp:
What about USE_CL_MAD?

I have an A10-7860K, I guess I can run some tests on that.

gcp:
Guess that's redundant due to -cl-mad-enable in the global options.

gcp:
Made the tuner fixes and merged to next.

roy7:
If the auto detected config used uses a config # from the 5128 that are valid from --full-tuner, should the message "Loaded existing SGEMM tuning." state the config # it is using?

I'm running a --full-tuner now for the heck of it, but I don't think I'll have any way to know if the best one it finds is actually the same one it was using before or not.

gcp:
Store the tuning file and compare?

--full-tuner actually runs a large random subset from all possible tunings (there's like 250k). So you can run it multiple times if you want to try to squeeze out another 0.1% (and pick a worse tuning due to Boost clocks :-)

It's a very good candidate for using a smart global optimization algorithm (I like Cross Entropy Method for this kind of thing). But everything on it's time...

roy7:
No worries then. When I saw "Will try 5128 valid configurations." I thought to myself that was the total of all there was. I didn't realize it was a tiny random sample of the whole space. My leelaz_opencl_tuning was:

> 0;XgemmBatched;64;100;64;16; -DKWG=32 -DKWI=2 -DMDIMA=16 -DMDIMC=16 -DMWG=64 -DNDIMB=8 -DNDIMC=8 -DNWG=16 -DSA=1 -DSB=1 -DSTRM=0 -DSTRN=0 -DVWM=4 -DVWN=2;OpenCL: NVIDIA Corporation GeForce GTX 1080 Ti @ 1683MHz

So we'll see what the final result of my random exploration brings me:

(4148/5128) KWG=16 KWI=8 MDIMA=8 MDIMC=8 MWG=32 NDIMB=16 NDIMC=16 NWG=64 SA=1 SB=1 STRM=0 STRN=0 VWM=4 VWN=4 0.0107 ms (1228.2 GFLOPS)
Wavefront/Warp size: 32
Max workgroup size: 1024
Max workgroup dimensions: 1024 1024 64

> 0;XgemmBatched;64;100;64;16; -DKWG=16 -DKWI=8 -DMDIMA=8 -DMDIMC=8 -DMWG=32 -DNDIMB=16 -DNDIMC=16 -DNWG=64 -DSA=1 -DSB=1 -DSTRM=0 -DSTRN=0 -DVWM=4 -DVWN=4;OpenCL: NVIDIA Corporation GeForce GTX 1080 Ti @ 1683MHz

So it found some changed settings, but I don't know the GFLOPS value of the previously tried one. (Does a --full-tune include the existing settings in the list of random new settings it'll try, so you never get worse if you tune multiple times?)

gcp:
>Should a --full-tune include the existing settings in the list of random new settings it'll try, so you never get worse if you tune multiple times?

You can easily get worse even if it did that. Modern cards have boost clocks and what not, so these measurements aren't 100% stable.

marcocalignano:
In autogtp I want to run --tune-only before I start the threads with the game, so each thread can use the tuning without loosing time retuning. My problem is that I have to give leelaz a network file like that: `leelaz --tune-only -w 55d235`. Can we change the --tune-only so that it take a parmeter like 5x64 to tell which network struct has to tune?

roy7:
Maybe just check/download best network before the --tune-only step? Then it's already downloaded before the real threads launch.

marcocalignano:
@roy7 IIf I do this I have to start a thread (each thread ask for work and download the needed networks) sisnce the weight are not important for the tuning but only the structure why not use a parameter?

roy7:
Will need you need an endpoint on the server to ask for the proper structure in case the structure changes over time? I don't know how else you'd know what structure to ask it to tune for, if you aren't using the weight file.

roy7:
(4090/5243) KWG=16 KWI=8 MDIMA=16 MDIMC=16 MWG=32 NDIMB=32 NDIMC=8 NWG=32 SA=1 SB=1 STRM=0 STRN=0 VWM=2 VWN=1 0.0106 ms (1235.6 GFLOPS)
Wavefront/Warp size: 32
Max workgroup size: 1024
Max workgroup dimensions: 1024 1024 64
BLAS Core: ZEN

I'll leave it at that since it was almost identical to the last run. :) 

> 0;XgemmBatched;64;100;64;16; -DKWG=16 -DKWI=8 -DMDIMA=16 -DMDIMC=16 -DMWG=32 -DNDIMB=32 -DNDIMC=8 -DNWG=32 -DSA=1 -DSB=1 -DSTRM=0 -DSTRN=0 -DVWM=2 -DVWN=1;OpenCL: NVIDIA Corporation GeForce GTX 1080 Ti @ 1683MHz

bood:
is it required to run the tuner on an idle card or it is fine even having
other leelaz running self plays? @gcp

2018年1月16日 04:16，"Gian-Carlo Pascutto" <notifications@github.com>写道：

gcp:
@bood You want the card to be idle as it's doing a speed measurement.

gcp:
@marcocalignano This would require a bunch of stuff to be suddenly aware of what the network structure is. I think it's better to use the actual weights file.

marcocalignano:
@gcp Have a look at my PR, that, if correct, would solve my problem.

marcocalignano:
@gcp the next improvement would be that, if a new network comes along and a new tuning is done, the old one does not get throw away but just one line is added in the tuning file. That would save a lot in games between two different networks.

marcocalignano:
In this way autogtp would run leelaz --tune-only for each GPU and for 5x64 and 6x128 network at the beginning, save the file and use it for all the self played games and matches.

gcp:
>if a new network comes along and a new tuning is done, the old one does not get throw away but just one line is added in the tuning file

I already fixed that.

gcp:
>Have a look at my PR, that, if correct, would solve my problem.

But how are you going to know what amount of channels to tune without reading in the network first? If you want the server to tell it to you, now parsing the file is the servers's problem, *AND* you could have asked just as well asked the server for the weights file instead of its structure....

I don't like this at all.

marcocalignano:
autogtp would do the tuning for both multiple network (5x64, 6x128) at the beginning.

gcp:
>autogtp would do the tuning for both multiple network (5x64, 6x128) at the beginning.

As said, I don't like this at all compared to using the actual network that you know is going to be used, but I gather from your resistance and PR #655 that this is rather complicated in current AutoGTP.

I think #655 can be simplified a lot so let me give that a try.

marcocalignano:
I guess I can ask the nest network to the self playing only URL. @gcp let me try it.

marcocalignano:
Ok @gcp look at PR #657

Alderi-Tokori:
Tested the winograd version with a 6x128 network for 9x9 boards and got
`leelaz v leelaz-wn (400/400 games)
board size: 9   komi: 5.5
leelaz       199 49.75%       111 55.50%     88  44.00%     88.68
leelaz-wn    201 50.25%       112 56.00%     89  44.50%     73.16
                              223 55.75%     177 44.25%`

PR #526, opened 2017-12-31
Batching NN evaluation + Multi-GPU

ihavnoid:
- Use float2/float4/float8 to concurrently evaluate 2/4/8 forward networks with a single OpenCL invocation
- To do this, two additional worker threads are spawned, detects the queue (which keeps the required evaluations) and uses either the 1-net/2-net/4-net/8-net variation depending on the queue depth. 
- Autodetects whether the OpenCL can/cannot handle each variation, and drops the variations that doesn't work.
- Average 33% more evaluations/sec on both GTX1080 and GTX1050 observed when 8 threads are used (which results in most evaluations being the 4-net variation)


gcp:
This absolutely needs a strength measurement. More evaluations per second are worthless if this makes the search vastly less efficient, which is exactly what one would expect from batching.

ihavnoid:
I would be interested to do one myself - what would be the easiest way to do so?  Is there some variation of autogtp that will handle this, or do I have to come up with my own script doind this?

gcp:
AutoGTP doesn't easily allow multiple binaries I think, so I use "gomill - ringmaster" which is a Python package.

roy7:
On the /next branch you can use the /validation folder instead of /autogtp to do your own offline strength testing.

gcp:
@roy7 as said above this doesn't really work for different executables (or parameters aside from the network)

roy7:
Yeah I just ended up asking since this was referenced in this reddit thread:

https://www.reddit.com/r/cbaduk/comments/7n4j8j/1_day_1451_games_7kwh_050/

About getting games played as fast as possible. :)

ihavnoid:
Got gomill - ringmaster running on my machine with constraints on 10 seconds per move.  3 players:
- leelaz default 2 threads
- leelaz default 8 threads
- leelaz batched mode 8 threads

ihavnoid:
24 hours in, here is the current status.  On the same machine with a six-core i7-8700, two sets of tournaments are running, one with GTX1080 and 10 seconds/move, another with GTX1050 and 11 seconds/move.  None of the cases were core-bound - it seems that it it is being GPU-bound at this point.

GTX1080
{code}
leelaz-8t-batch v leelaz-2t-base (21/100 games)
board size: 19   komi: 7.5
                  wins              black        white       avg cpu
leelaz-8t-batch     17 80.95%       8 72.73%     9  90.00%   4074.72
leelaz-2t-base       4 19.05%       1 10.00%     3  27.27%   2158.48
                                    9 42.86%     12 57.14%

leelaz-8t-batch v leelaz-8t-base (20/100 games)
board size: 19   komi: 7.5
leelaz-8t-batch     17 85.00%       9  90.00%     8 80.00%   4033.87
leelaz-8t-base       3 15.00%       2  20.00%     1 10.00%   2323.20
                                    11 55.00%     9 45.00%
{code}

GTX1050
{code}
leelaz-8t-batch v leelaz-2t-base (15/100 games)
board size: 19   komi: 7.5
leelaz-8t-batch      7 46.67%       5  62.50%     2 28.57%   3969.65
leelaz-2t-base       8 53.33%       5  71.43%     3 37.50%   2971.33
                                    10 66.67%     5 33.33%

leelaz-8t-batch v leelaz-8t-base (15/100 games)
board size: 19   komi: 7.5
leelaz-8t-batch     11 73.33%       5 62.50%     6 85.71%   4335.06
leelaz-8t-base       4 26.67%       1 14.29%     3 37.50%   5335.29
                                    6 40.00%     9 60.00%
{code}

FYI, GTX1080 seems to have 3x evaluation throughput compared to GTX1050.

It seems that for quick evaluations, less threads is better, but as we add more effort/time, higher parallelism starts to pay off.

I will add a patch which makes this a command-line option, so that we can still use the previous implementation.  Will leave the ringmaster running for another day.

ihavnoid:
By the way, I do have a feeling that this is a bit better than I was expecting.  Is there anybody who is willing to verify what I am seeing here?  e.g., longer games (1 minute per move) will be interesting.

killerducky:
Can you post your ringmaster cfg file?

ihavnoid:
Here you go.  Can you see if what I did makes sense?

[playoff.ctl.txt](https://github.com/gcp/leela-zero/files/1596882/playoff.ctl.txt)

killerducky:
Looks good. You mentioned it is GPU bound. So my interpretation of the log results for the fast GPU case: the 8t-batch version is able to use the GPU more efficiently, therefore uses more CPU time as well keeping up (~4000). The 2t-base already uses the GPU 100%, so the 8t-base version doesn't gain anything, it uses ~2000 CPU in both cases. 

```
leelaz-8t-batch 17 80.95% 8 72.73% 9 90.00% 4074.72
leelaz-2t-base 4 19.05% 1 10.00% 3 27.27% 2158.48

leelaz-8t-batch 17 85.00% 9 90.00% 8 80.00% 4033.87
leelaz-8t-base 3 15.00% 2 20.00% 1 10.00% 2323.20
```

To me this means it doesn't hurt much in cases where the user has a slow GPU, but when they have a fast GPU it helps a lot.

ihavnoid:
48 hrs update.  I am killing this evaluation as I need to run something else.

GTX 1080, 10 seconds
```
leelaz-8t-batch v leelaz-2t-base (39/100 games)
board size: 19   komi: 7.5
leelaz-8t-batch     31 79.49%       13 65.00%     18 94.74%   4288.47
leelaz-2t-base       8 20.51%       1   5.26%     7  35.00%   2268.53
                                    14 35.90%     25 64.10%

leelaz-8t-batch v leelaz-8t-base (38/100 games)
board size: 19   komi: 7.5
leelaz-8t-batch     34 89.47%       18 94.74%     16 84.21%   4069.59
leelaz-8t-base       4 10.53%       3  15.79%     1   5.26%   2344.19
                                    21 55.26%     17 44.74%
```

GTX 1050, 11 seconds per move
```
leelaz-8t-batch v leelaz-2t-base (31/100 games)
board size: 19   komi: 7.5
leelaz-8t-batch     14 45.16%       8  50.00%     6  40.00%   3909.21
leelaz-2t-base      17 54.84%       9  60.00%     8  50.00%   2915.60
                                    17 54.84%     14 45.16%

leelaz-8t-batch v leelaz-8t-base (31/100 games)
board size: 19   komi: 7.5
leelaz-8t-batch     22 70.97%       11 68.75%     11 73.33%   4027.12
leelaz-8t-base       9 29.03%       4  26.67%     5  31.25%   4918.58
                                    15 48.39%     16 51.61%
```

Next thing - I figured out that I broke USE_HALF, so I am going to fix it.

gcp:
One of your results show 8t-batch being weaker than 2t-base, which is what I feared.

However, the GTX 1080 results are very suprising. You said earlier you measured this to be a 33% speedup. A 33% speedup can't cause a 80% winrate? Something seems off.

ihavnoid:
@gcp  - yes I am also surprised and have a feeling that something unusual is happening here.  That's why I would like to have some other people verify this.

One possible thing is that I was running both comparisons on the same machine (this is a i7-8700 with two GPUs), and if there is a thread scheduling problem because of those two runs interfering each other, there can be some unknown factors there, but I don't have a good theory why the 2t-base case may suffer more.  I would like to run it on a separate machine with no other jobs running, but I don't have a machine that has no other tasks running as of now.

MartinDevelopment:
It could be because you used too few games and it just got lucky.

ihavnoid:
One possible theory I have is that this might be a consequence of using identical networks.  Whoever sees a bit beyond the other gets a definite advantage.  A car analogy would be racing two completely identical cars with completely identical drivers, except one car is 1km/h faster than the other.  The faster car will always win by a very slim margin.

In this case, the only way to test this is to have a 3rd AI which is completely different, and if there are any win/lose rate difference between the two.  This isn't easy either because to have this meaningful the 3rd AI shouldn't be too strong not too weak.

Maybe I am hypothesizing too much.  At least I made this fix command-line configurable, though.

ihavnoid:
I ended up adding more and more commits to this branch.  Some highlights:
- back-merged to the latest 'next' branch
- Multi-GPU support - I am running a GTX 1080 and GTX1050 at the same time
- A bit more tweaks to avoid the 'clFinish busy loop' as much as possible
- Batching is now purely opportunistic - if there are two or more threads waiting for a NN evaluation, it does the batched mode.  If not, it does not wait for other threads to add more evaluations.

Running a set of evaluations in 10 second/move settings.  Keep in mind that there were a lot of updates on the next branch, including the NNCache which moved the overall picture quite a bit.

```
56/120 games played

                         A   B   C   D
A leelaz-8t-batch           3-7 4-6 5-4
B leelaz-12t-batch-2gpu 7-3     7-2 4-5
C leelaz-2t-base        6-4 2-7     5-4
D leelaz-8t-base        4-5 5-4 4-5
```

To run multi-gpu, put multiple --gpu command line options to pick all the GPUs you want to add, e.g., 
`./leelaz --gpu 1 --gpu 2 [your usual command line]`

gcp:
> Whoever sees a bit beyond the other gets a definite advantage.

FWIW I don't think this holds given that you saw 33% on both chips yet a win on one and a loss on the other.

I will almost certainly pull #523 instead. But the idea to opportunistically batch might still be applicable there.

ihavnoid:
Yes I think I will also try to apply batching on top of #523 at some point.  There is no reason that the batching idea will not work with #523 (or maybe, it will make the whole batching idea obsolete, hard to say), nor I am not in a hurry to get this merged anywhere - to me this is just a learning experience. :)

Upon analyzing what's going on, I think there are many weird situations regarding the OpenCL implementation, e.g., queue.finish() seems to be a busy loop and having too many net evaluations piling up will result in CPU cycles wasted for nothing, hence making any rational comparison even more complicated.

I think I will be leaving this branch running for a while on my PC (just to gather some more confidence on the result) and then probably once #523 is good to be merged I will try to create a batching implementation of it.

gcp:
> queue.finish() seems to be a busy loop and having too many net evaluations piling up will result in CPU cycles wasted for nothing

I think this is true. In the original Leela, I spent a lot of time trying to make network evaluation asynchronous, including working through piles of bugs in the OpenCL implementations, with the idea it could run Monte Carlo simulations on the CPU while the network was being evaluated on the GPU. I never managed to get a benefit, and analyzing seemed to indicate the OpenCL implementations were just pegging the CPU and thus slowing down what should have been an asynchronous request. Sad times. Leela (and Leela Zero) are fully synchronous for this reason.

gcp:
Won't be merged as is, as per discussion above.

ihavnoid:
Final update of the 120 games between various comparisons:

```
120/120 games played

                         A     B    C     D
A leelaz-8t-batch             8-12 10-10 10-10
B leelaz-12t-batch-2gpu 12-8       15-5  12-8
C leelaz-2t-base        10-10 5-15       9-11
D leelaz-8t-base        10-10 8-12 11-9
```

For now, It looks like (with all the latest changes from v0.10) basically the single GPU versions are similar regardless of thread count or batching/non-batching.  The only thing that shows a definite advantage is when there are additional GPUs - this is an additional GTX 1050 on top of the usual GTX 1080 I was using.

Hard to explain why I am getting this, though...

PR #537, opened 2018-01-01
autogtp: leelaz command line option output

marcocalignano:


marcocalignano:
The second commit output is something like this: 

`30 game(s) (29 self played and 1 matches) played in 2 hours and 50 minutes = 5 minutes and 41 seconds per game, 948 ms/move, last game took 43 minutes and 3 seconds.
`

gcp:
Not sure about this. Comparing speeds and so on is much easier if everything is in seconds?

marcocalignano:
That is true.

marcocalignano:
Ok so this is what you want.

Issue #538, opened 2018-01-01
Reusing the tree overwhelms the Dirichlet noise


killerducky:
I saw some self-play games were this is causing a problem. During a forced sequence of plays the tree is reused and the PV can grow very large. During a ladder it grew so large that it reused 12,257 playouts, and did an additional 1600 at the root for a total of 13,857. Only the additional 1600 playouts will have Dirichlet noise, making it impossible for the noise to change the move selected. 

The peak of the problem happened in the middle of a ladder that doesn't work. After the end of the ladder was reached, continued play was still affected in a very bad way, Leela misses capturing the tail of the ladder for several moves. It's fine that the network doesn't see it deep in the tree, but at the root the purpose of the noise is for Leela to be able to correct itself. In this case it's not able to correct itself until the size of the PV shrinks so that the Dirichlet noise is not overwhelmed.

Full log file here: https://pastebin.com/hNB3eMtZ

First you can see X has failed to connect the tail of the long ladder at C16. The policy has low prior for O capturing at C16. But at the bottom you can see there are 5049 playouts, most of which came from before this node was root, so even if the noise hit C16 hard, it will be ignored.

```
Passes: 0            Black (X) Prisoners: 2
White (O) to move    White (O) Prisoners: 0

   a b c d e f g h j k l m n o p q r s t 
19 . . . . . . . . . . . . . . . . . . . 19
18 . . . . . . . . . . . . . . . O X O . 18
17 . X X O O . . . . . . . . . X X O . . 17
16 X . . X X O . . . + . . . . X O(X). . 16
15 . X X O X X O . . . . . . . . O . . . 15
14 . . . O O X X O . . . . . . . . . . . 14
13 . . X . . O X X O . . . . . . . . . . 13
12 . . . . . . O X X O . . . . . . . . . 12
11 . . . . . . . O X X O . . . . . . . . 11
10 . . . + . . . . O X X O . . . + . . . 10
 9 . . . . . . . . . O X X O . . . . . .  9
 8 . . . . . . . . . . O X X O . . . . .  8
 7 . . . . . . . . . . . O X X O . . . .  7
 6 . . . . . . . . . . . . O X O . . . .  6
 5 . . . . . . . . . . . . . O O . . . .  5
 4 . . . X . . . . . + . . . . . X . . .  4
 3 . . . . . . . . . . . . . . . . . . .  3
 2 . . . . . . . . . . . . . . . . . . .  2
[...]

PR #546, opened 2018-01-03
Change definition of playout_limit_reached

killerducky:
See issue #538. Instead of using m_playouts, use m_root.get_visits to decide when we have reached the playout limit.

BEFORE: 1147 ms/move
AFTER: 602 ms/move
1.9X faster

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -297,7 +297,7 @@ bool UCTSearch::is_running() const {
 }
 
 bool UCTSearch::playout_limit_reached() const {
-    return m_playouts >= m_maxplayouts;
+    return m_root.get_visits() >= m_maxplayouts;
 }
 
 void UCTWorker::operator()() {


killerducky:
I starting out focused on the impact of tree reuse and Dirichlet noise. But I think on a more fundamental level this pull is the right thing to do. It makes no sense to spend extra time searching a position during a forced sequence. You should either a) pocket the 1.9X savings and run more self-play games. Or b) increase the playouts from 1600 to something higher. Use the extra playouts on positions that are not obvious forced sequences. To use those playouts on forced sequences is the opposite of good time management.

I vote option (a), pocket the extra time.

MartinDevelopment:
How does this affect the strength of the network at 1600 playouts?

killerducky:
I'm running a tournament with gomill ringmaster now. For self-play there is a 1.9X speedup, but for two separate engines there is only a 1.35X speedup due to less tree reuse. So both the speed difference and the Elo difference will be larger than what I report below in the self-play case.

Option A: Pocket the speed increase, accept a loss of strength. For matches we gain 1.35X in speed but lose 76 Elo.
```
lzbase-1600 v lzmaxp-1600 (28/1000 games)
board size: 19   komi: 7.5
lzbase-1600     17 60.71%       7  50.00%     10 71.43%    348.37
lzmaxp-1600     11 39.29%       4  28.57%     7  50.00%    258.22
                                11 39.29%     17 60.71%
```

Option B: Increase playouts so the average number of playouts over the game is actually the same. Games will take the same amount of time, but we gain 159 Elo by distributing the extra playouts evenly over all moves instead of mostly in forced sequences. In the match above I measured that the old code with -p 1600 was actually doing 2278 playouts on average due to tree reuse. For the match below I set the new maxp code to use -p 2200 to be conservative. You can see by the avg cpu column they use about the same amount of CPU time.
```
lzbase-1600 v lzmaxp-2200 (28/1000 games)
board size: 19   komi: 7.5
lzbase-1600      8 28.57%       3  21.43%     5  35.71%    363.33
lzmaxp-2200     20 71.43%       9  64.29%     11 78.57%    366.84
                                12 42.86%     16 57.14%
```
[maxplayouts.ctl.py.txt](https://github.com/gcp/leela-zero/files/1600549/maxplayouts.ctl.py.txt)

MartinDevelopment:
Now the question is if it is worth it to lose that amount of Elo for such a large speed increase.

jkiliani:
That's pretty big... hope this gets pulled soon, along with other promising speed-ups.

roy7:
I'm kinda learning towards preferring lzmaxp-2200 (smarter use of the same CPU we already spend, giving us stronger training data) over lzmaxp-1600 (faster results, more training games, but weaker strength games).

But over time maybe they converge to the same eventual target.

killerducky:
Final results, I'm stopping the tournament now. A) take the speed gain, lose 80 Elo. B) keep the same CPU/time usage, gain 75 Elo. I kinda regret framing this as A/B because you could choose any amount of playouts really. Unless someone does some serious cost/benefit analysis it makes sense to me to stick to 1600 since that's what AGZ did.

```
lzbase-1600 v lzmaxp-1600 (62/1000 games)
board size: 19   komi: 7.5
lzbase-1600     38 61.29%       19 61.29%     19 61.29%    344.56
lzmaxp-1600     24 38.71%       12 38.71%     12 38.71%    255.34
                                31 50.00%     31 50.00%
```

```
lzbase-1600 v lzmaxp-2200 (61/1000 games)
board size: 19   komi: 7.5
lzbase-1600     24 39.34%       15 48.39%     9  30.00%    361.19
lzmaxp-2200     37 60.66%       21 70.00%     16 51.61%    363.81
                                36 59.02%     25 40.98%
```

roy7:
We may need to invoke the power of @gcp to make a command decision for us.

If we move to option A, we'll be putting weaker data into the training database than we have been.  But we'll generate a lot more example games. And we stick with the 1600 magic number.

billyswong:
Do we know if AGZ do the visit caching and/or capping too? If they don't do visit caching, then the 1600 magic number can be said to be "designed" for non-caching / caching+capping playing strength, not the current strength.

There is of course still an option C) take smaller speed gain, and keep roughly the same elo.

gcp:
Yes, running the engine with less playouts than intended makes it a lot faster and weaker. No, we don't want that at all. This isn't a speedup or a speed gain, at all. It makes nothing faster. That's a serious misrepresentation.

What this pull request does is add time management code to the engine: don't search so deeply if we already have a lot of search on the current position. The test result above gives some indication this time management may be effective.

Changing the meaning of the -p option in this manner is not something I'd want to pull, though. -p says that the engine is allowed to do x playouts every move. It looks like you're really wanting to add a new "-v" engine option and are arguing that -v 2200 is a more effectively use of computing resources than -p 1600 because it plays stronger in the same amount of time.

jkiliani:
The original reason for the pull request was the observation that Dirichlet noise gets suppressed by high numbers of carried over playouts. If this effects of using visit count instead of playout count is confirmed, we should change to visit count for the reason of fixing blind spots irrespective of the speed-up/strength gain that can be derived from it.

I think there's a high likelihood someone (or quite a number of people) are now running autogtp with this pull, or some other speed-ups. The rate of self play roughly doubled this morning.

gcp:
>I think there's a high likelihood someone (or quite a number of people) are now running autogtp with this pull, or some other speed-ups. The rate of self play roughly doubled this morning.

That's terrible :(

gcp:
Has anyone verified that the probability map constructed in the learning data still has a sane/expected shape with this patch enabled? It seems that if the search aborts almost immediately because a move that was expected with high probability was played, the constructed map will be the one corresponding to only a few hundred playouts, rather than the expected 1600+?

gcp:
>That's terrible :(

To be clear, the NN caching patch, and the Winograd patch are fine: they just compute the right thing faster. But this one fundamentally changes what is being computed on the clients.

gcp:
I'm going to close this pull request to make it very clear this change won't be pulled and shouldn't be used as-is.

Adding a -v option and using a target of 2200 instead of 1600 might be usable for a next run:
https://github.com/gcp/leela-zero/pull/546#issuecomment-355174499

IFF we can verify that the probability map that is constructed for the learner is still OK.

gcp:
https://github.com/gcp/leela-zero/blob/19a9129dc2e3d529ba09a27a8770447122cb0e32/src/Training.cpp#L119

The fact that this counts the actual values in the children should stop some of the worst things that can go wrong, but note that 1600 or 2200 counted visits at the root position does not necessarily imply we will find back those numbers in all the leaves, for example in case of a TT overwrite.

killerducky:
Ok I'll look into the impact of caching and the shape of the training data.

roy7:
@gcp should @killerducky update this PR to make a new -v visit count cap option, instead of changing the playouts cap current -p behavior? So at least it's in the code for future runs if we want to use it, or for use by individuals outside of autogtp? I was thinking it might be nice to run my bot on OGS with a visit cap instead of a playout cap (effectively make better use of the time it does think).

Once we're sure it's safe to use of course.

PR #559, opened 2018-01-04
Add NNCache to msvc builds

zediir:


zediir:
Was from the wrong local branch

Issue #330, opened 2017-12-10
fp16 training


ihavnoid:
I am trying to test how good (or bad) it is if we move the training using fp16.  Quick testing using supervised learning seems to show that my GPU (GTX 1080 non-TI) is running out of memory when I tried to train using the full network (20 deep, 256 planes).  Moving to fp16 at least did seem to help by reducing the memory footprint of various data structures (still a smaller batch was needed), and even on the reduced network, the reduced bandwidth requirement seemed to help training performance.

(BTW, tensorflow seemed to keep the in-memory structure at fp16 but still do compute on fp32 - not sure how it was done on my GTX1080)

Questions:
- Is it expected that GPU memory capacity is going to be a bottleneck?  If so, any hint on where it would be good to start?
- Is there where I can download the raw training data rather than the sgf files?

Ttl:
Raw training data can be found at #167.

billyswong:
Most consumer graphic cards are very weak in fp16 compute, and can be even slower than fp32. Thus the tensorflow hybrid design.

gcp:
I tried fp16 training on a V100 but I had problems with the converege of the value network, even after re-scaling the backprop gradient (see NVIDIA's papers on why that's needed).

ihavnoid:
Was the policy head the only problem?  What I am trying to do is to cast everything back to fp32 on the 'dual' heads while leaving the residual tower and those parameters at fp16.

ihavnoid:
By the way, is this the nvidia paper you are referring to?

https://devblogs.nvidia.com/parallelforall/mixed-precision-training-deep-neural-networks/

gcp:
Yes, that and the slides to go with it.

I think what you're trying is a good idea (fp16 res tower, fp32 heads).

ihavnoid:
Training the ad8a1d02 dataset for 24 hours, and this is the current status:

`step 132000, policy loss=5.11683 mse=0.0368261 reg=0.285763 (226.404 pos/s)`
`step 132000, training accuracy=16.0938%, mse=0.0652886`

Accuracy seems to be moving around 15~20% since the last 12 hours.  Is this normal?

gcp:
Yes, that sounds about right. The MSE is very low, you're probably overfitting.

ihavnoid:
Okay I thought we should be hitting around 50% accuracy but maybe that's only for the supervised learning data.

[...]

PR #566, opened 2018-01-05
Training pipeline changed to support fp16 based training

ihavnoid:
See issue #330 for relevant discussion.   Reworked patch.

See dtype on main function for changing to float16 training.  Performance difference was not tested as I don't have a decent GPU with good fp16 support, though.


ihavnoid:
One thing to beware is that this cannot use the old model data dumps (even if we run fp32) because there is a change of variable scope naming.  Probably we won't want to merge as-is, but this is to figure out what else we want to do (or not).

gcp:
At some point I want to fix the weight format (dump the redundant convolution bias weights, and add a bias/translate layer after the batch-norm one). I guess at that point this could be pulled, but not during a run as that would cause quite some headache.

gcp:
Going to close this as it would need to be rebased to all recent training pipeline changes.

gcp:
>One thing to beware is that this cannot use the old model data dumps (even if we run fp32) because there is a change of variable scope naming.

FWIW, as long as net_to_model works, I believe that's a non-issue.

PR #567, opened 2018-01-05
UCTNode size reduction optimization

ihavnoid:
- Changed order of declaration of instance variables for better packing
- Removed mutex and used a shared mutex (40 byte per mutex!)
- 112B -> 56B on a 64-bit Linux gcc compilation

Next obvious target is the std::vector (28 bytes) but probably not worth the additional hassle.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -44,6 +44,8 @@
 
 using namespace Utils;
 
+std::mutex UCTNode::s_nodemutex[UCTNode::MUTEX_TABLE_SIZE];
+
 UCTNode::UCTNode(int vertex, float score, float init_eval)
     : m_move(vertex), m_score(score), m_init_eval(init_eval) {
 }
@@ -61,7 +63,7 @@ bool UCTNode::create_children(std::atomic<int> & nodecount,
     }
     {
       // acquire the lock
-      std::lock_guard<std::mutex> lock(m_nodemutex);
+      std::lock_guard<std::mutex> lock(get_mutex());
       // no successors in final state
       if (state.get_passes() >= 2) {
           return false;
@@ -123,7 +125,7 @@ void UCTNode::link_nodelist(std::atomic<int> & nodecount,
     // Use best to worst order, so highest go first
     std::stable_sort(rbegin(nodelist), rend(nodelist));
 
-    std::lock_guard<std::mutex> lock(m_nodemutex);
+    std::lock_guard<std::mutex> lock(get_mutex());
 
     for (const auto& node : nodelist) {
         m_children.emplace_back(
@@ -311,7 +313,7 @@ UCTNode* UCTNode::uct_select_child(int color) {
     UCTNode* best = nullptr;
     auto best_value = -1000.0f;
 
-    std::lock_guard<std::mutex> lock(m_nodemutex);
+    std::lock_guard<std::mutex> lock(get_mutex());
 
     // Count parentvisits.
     // We do this manually to avoid issues with transpositions.
@@ -370,13 +372,13 @@ private:
 };
 
 void UCTNode::sort_root_children(int color) {
-    std::lock_guard<std::mutex> lock(m_nodemutex);
+    std::lock_guard<std::mutex> lock(get_mutex());
     std::stable_sort(begin(m_children), end(m_children), NodeComp(color));
     std::reverse(begin(m_children), end(m_children));
 }
 
 UCTNode& UCTNode::get_best_root_child(int color) {
-    std::lock_guard<std::mutex> lock(m_nodemutex);
+    std::lock_guard<std::mutex> lock(get_mutex());
     assert(!m_children.empty());
 
     return *(std::max_element(begin(m_children), end(m_children),
--- a/src/UCTNode.h
+++ b/src/UCTNode.h
@@ -74,29 +74,44 @@ public:
     UCTNode& get_best_root_child(int color);
 
 private:
+    // BEWARE : this instance is very size-sensitive as we generate tens of millions of these.
+    // Think twice before adding a field here, and see how you can reduce instance footprint.
+
+    // instead of placing mutex instances per instance (40 bytes on Linux 64-bit),
+    // we will create a pool of mutex instances and then use mutexes based on the
+    // hash of the this pointer.  This will create some false locks but it should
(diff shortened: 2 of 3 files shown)


sethtroisi:
:)

ihavnoid:
Hmm okay it seems that we took out the atomic for no apparent reason than tidying things up.  In that case I agree that we should return to atomics.

However, reordering variables to densely pack the memory layout will still save another 10~20 bytes on sizeof(UCTNode).  std::atomic<bool> seems to be only a single byte so we will also get sizeof(UCTNode) == 56 with clever variable reordering, though.

gcp:
I'd could still consider taking that part - though it's a smaller savings that makes some parts of the code arguably less logical.

gcp:
I'd could still consider taking that part - though it's a smaller savings that makes some parts of the code arguably less logical.

ywrt:
gcp, can I suggest taking this patch anyway? I think the hashed mutexs make a lot of sense.
1. They pull the mutex out of node anyway. even at 8 bytes that's still better.
2. They open the way to UCTNode being copyable, which allows flattening the vector, which becomes both faster and smaller.

I'll sketch something up to show what I mean

gcp:
Having copyable or movable UCTNodes would indeed allow for simplifications, but not a fan of hashed mutexes at all.

gcp:
Actually UCTNodes won't be copyable/movable unless there's no atomics in use, which this doesn't bring us much closer to achieving.

ihavnoid:
I also thought of flattening the vectors simply because it's additional bytes everywhere, but if we really want to do that I think the only way to make it work is to have a global lock which isn't acceptable.  If one node is being moved while another node is using the child, or even the child's child, everything falls apart.  The 'this' pointer will point to nowhere.

The hashed mutex uses the this pointer for the hash function, so moving objects around will break the locking mechanism, too.  It just assumes the object is not movable.

sethtroisi:
When leelaz is run with a single thread is the mutex even needed? Would this pull request better facilitate wrapping the mutex in `if (cfg_threads > 1) {`

ywrt:
I was going to strip all the atomics out instead.

Note that the taking a lock is about the same cost as a seq consistent atomic write.
At the moment ~ every node visit does an atomic write to virtual loss twice plus an atomic write to visits, plus an atomic write to black evals. 

So it's equiv in cost to take a lock, get the current stats, release the lock, do the decent, take the lock again, update stats and release the lock.

They're both 4 locked operations.

My plan is something like:
1. Remove m_virtual_loss from UCTNode and put it into a side hash instead (because it's only used when the node is actually being descended through which very few nodes are).
2. Remove m_valid, and instead actually remove the nodes that are invalid.
3. Change m_children to be std::unique_ptr<std::vector<UCTNode>> which removes one pointer per node (plus whatever overhead malloc is introducing).
4. Remove m_has_children (because it's equiv to m_children != nullptr)
5. Remove m_expanding and put it into a side hash instead (because there's only one node per thread where it can be set).
6. As an aside, turn m_children into a heap, so that uct_select_child() can always just use the first node.

That reduces UCTNode to:

std::unique_ptr<std::vector<UCTNode>> m_children;  // (8 bytes)
double m_blackevals; // (8 bytes)
int m_visits; // (4 bytes)
float m_score; // (4 bytes)
float m_init_eval; // (4 bytes)
int m_move; // ( 4 bytes)

for a total of 32 bytes per node. aka ~ 1.3GiB for a 40 million node search tree.

Possible future reductions if we're very keen:
1. Move m_visits + m_blackevals out to a side hash. Vast majority of nodes have a zero visit count.
2. m_init_eval can be moved out with some work.

That would get it down to 16 bytes: 8 bytes for child ptr, 4 bytes for score and 4 bytes for move.

ywrt:
Hmm. Note that I think about it:

m_visits, m_blackevals, and (effectively) m_init_eval are only used if the the node has children (aka the node has been visited).

So a structure like:
```
struct UCTNode {
  std::unique_ptr<UCTVisited> m_full;
  int m_move;
  float m_score;
};  // 16 bytes

struct UCTVisited {
  uint64_t m_moves;
  double m_blackevals;
  float m_init_eval;
  std::vector<UCTNode> m_children;
};
```

Which would be an average of 16.08 bytes per node (given that the typical node has (say) 200 children, less than 1 node in 200 is expanded).

But we can actually do better again:

```
struct UCTNode {
  std::vector<std::pair<int, float>> m_children;  // entry per move returned by network.
  std::vector<UCTNode*> m_expanded;  // entry per expanded move.
  float m_child_eval;  // this is the init_eval for the CHILD nodes, not this node.
  uint64 m_visits;
  /// etc
};
```

Which gets us down to ~8.04 bytes per 'node' as originally conceived of. In actual practice there's now just < 0.5% of the original quantity of UCTNode's allocated.

Obv, the API for UCTNode changes somewhat.

In principle, we can get smaller by relying on the statistics of node expansion. e.g. if most nodes have more than 180 legal moves, then std::vector<float>(361) is better than std::vector<std::pair<int, float>>(). But that's only true in the early game, so it seems dubious.

Similarly, we could quantize the score returned by the network and pack it with the move to get down to ~4 bytes per 'node', but we pay the tradeoff of lower accuracy.

gcp:
In general I'm not interested in changes that reduce the size of UCTNode at the cost of adding complexity.

Making nodes copyable/movable would simplify things, so that's a trickier case. But your reasoning there about the locking assumes no contention at all, making it IMHO rather handy-wavy. Locks are a bit easier to reason about than atomics, so I'll give you that. But to make the UCTNodes movable, you'd need to move all the atomics/mutexes out. And if you make the nodes movable, your hashed mutexes get tricky?

In general I think "try to squeeze more UCTNodes into a given amount of RAM" is the wrong approach. You already realize in your second post that most nodes just don't ever have any information added. It may make a lot more sense to track the hash of the positions/node more closely and instead of using a dynamically constructed tree, shove all information into a fixed-size hashtable, simply overwriting the node with the least amount of visits in case you get a collision.

gcp:
If I understand your second suggestion, you wouldn't allocate any UCTNode, but just store the policy network output, and store nodes that got visited at least once. That sounds like it may have an acceptable ratio of memory saving vs added complexity.

Maybe instead of an pair<int, float>, it should be tuple<int, float, UCTNode*>?

PR #571, opened 2018-01-05
Resize cache if max_playout is set

sethtroisi:
Testing is documented [here (Google Sheets)](https://docs.google.com/spreadsheets/d/1KI6rRX_FAIHkq74p51weL2pRj2cMqlrZFCu6Ohh3kH8/edit#gid=785169179)

For people who use high playouts memory is already REALLY high (several gigs, thought I hope pull/567 helps with that) so the extra cache is unnoticed by them.

For people running autogtp performance will be the same (with size = 15,000 it's identical because no cache hit happens with age > 15,000) but 70 megs lighter. This allows for 40% more instances of leelaz to be run which is useful for people using cuDNN and large batch_sizes.

---
Spot testing

First 38 moves are the same when testing with
`(yes "go" | head -n60) | /usr/bin/time -v ./leelaz -t1 -s120 -w ../weights.txt -p 1000 --noponder |& tee smaller_cache`

Memory is reduced by ~70 megs

```
< 	System time (seconds): 10.08
< 	Maximum resident set size (kbytes): 198600
---
> 	System time (seconds): 10.61
> 	Maximum resident set size (kbytes): 271700
```

### DIFF
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -32,6 +32,7 @@
 #include "GTP.h"
 #include "GameState.h"
 #include "Network.h"
+#include "NNCache.h"
 #include "Random.h"
 #include "ThreadPool.h"
 #include "Utils.h"
@@ -264,6 +265,15 @@ int main (int argc, char *argv[]) {
 
     // Initialize network
     Network::initialize();
+
+    // Initialize cache if playouts is set
+    if (cfg_max_playouts) {
+        // cache hits are generally from last several moves so setting cache
+        // size based on playouts increases the hit rate while balancing memory
+        // usage for low playout instances. 100'000 cache entries is ~500 MB
+        auto max_size = std::min(100'000, std::max(6'000, 3*cfg_max_playouts));
+        NNCache::get_NNCache()->resize(max_size);
+    }
 
     auto maingame = std::make_unique<GameState>();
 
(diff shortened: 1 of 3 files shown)


sethtroisi:
@ywrt I suggesting changing the cache size to 3 * max_playouts, any input on lower, upper bound?

src/Leela.cpp
```diff
@@ -265,6 +266,15 @@ int main (int argc, char *argv[]) {
...
 
+    // Initialize cache if playouts is set
+    if (cfg_max_playouts) {
+        // cache hits are generally from last several moves so setting cache
+        // size based on playouts increases the hit rate while balancing memory
+        // usage for low playout instances. 100'000 cache entries is ~500 MB
+        auto max_size = std::min(100'000, std::max(6'000, 3*cfg_max_playouts));
+        NNCache::get_NNCache()->resize(max_size);
+    }
+
```
gcp: I prefer making a function inside NNCache that takes cfg_max_playouts as the parameter and calculates an appropriate size for itself.

gcp: That is, the NNCache specific logic for figuring out the size should be in NNCache.cpp, not in Leela.cpp.

sethtroisi: Updated, I also moved all of the global object initialization (threadpool, rng, zobrist, nncache) into a method.

sethtroisi:
Addressed comment

PR #574, opened 2018-01-06
Reduce the memory consumption of nodes.

ywrt:
Primarily by not expanding child nodes until they're actually visited.
This splits m_children into m_child_scores and m_expanded.

The average cost of a 'node' after this change is a fraction over 8 bytes.
(Somewhat squishy because 'node' doesn't really mean the same thing after this change).

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -49,9 +49,6 @@ UCTNode::UCTNode(int vertex, float score, float init_eval)
 }
 
 UCTNode::~UCTNode() {
-    LOCK(get_mutex(), lock);
-    // Empty the children array while the lock is held
-    m_children.clear();
 }
 
 bool UCTNode::first_visit() const {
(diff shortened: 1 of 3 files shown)


ywrt:
This is mostly a draft at this point, looking for comments on the approach.

ywrt:
(Ref #567)

src/UCTNode.h
```diff
@@ -92,6 +99,7 @@ class UCTNode {
     // UCT eval
     float m_score;
     float m_init_eval;
```
sethtroisi: init_eval is no longer used.

gcp:
Seems ok.

src/UCTNode.cpp
```diff
@@ -49,9 +49,6 @@ UCTNode::UCTNode(int vertex, float score, float init_eval)
 }
 
 UCTNode::~UCTNode() {
```
killerducky: I'm a C++ noob, is there a reason not to delete the destructor function completely now that it's empty?

gcp: It's already like that in "next" actually.

~UCTNode() = default;

src/UCTNode.cpp
```diff
@@ -135,35 +132,36 @@ void UCTNode::link_nodelist(std::atomic<int> & nodecount,
...
             mystate.play_move(move);
 
             if (mystate.superko()) {
-                // Don't delete nodes for now, just mark them invalid.
-                child->invalidate();
+                // Skip invalid moves.
+                continue;
             }
         }
+        good_moves.push_back(child);
```
oni-link: Why not reorder the children inplace (`std::partition`) and use `erase` afterwards? Could save some copying and entries are only ordered in `m_expanded`.

killerducky:
Pinging @ywrt. Will you be able to work on this soon? I have an idea that would work better if this was finished.

roy7:
Hopefully @ywrt is still around. :)

sethtroisi:
I'm familiar with most of this code and can take over if ywrt doesn't show back up.

sethtroisi:
There's quite a bit of work that needs to be done to verify this change doesn't break KOs / narrow trees / dumppass.

I've started a pass trying to merge this change with next (primarily tree reuse), see https://github.com/sethtroisi/leela-zero/tree/v__small_node

killerducky:
@sethtroisi @ywrt the idea I was working on that would rely on this pull didn't work out. I did a quick resurrection of TTable that reuses the entire tree, but I wasn't able to get any Elo improvement out of it. Seems just relying on NNCache for transpositions is good enough, and it's much simpler.

This pull would still be nice to reduce memory usage, but at least for me I don't see any near term changes that depend on it.

sethtroisi:
I'm almost done with the change. I couldn't find clean solutions with the existing UCTNode so I'm cleaning up some of it [first](https://github.com/gcp/leela-zero/pull/799)

WhiteHalmos:
I was thinking of something similar but a different approach (rather than separating out scores and children, replace child array with an edge array that contains {move, score, child_node}).

PR #575, opened 2018-01-06
Fix assert in GameState

killerducky:
This assert fires when you run these commands:
```
play b q4
undo
genmove b
```

@sethtroisi do you see any other problems? From a quick test seems the rest of the code works.


sethtroisi:
Good catch, 

This is supposed to be handled by the [resize on line 111](https://github.com/gcp/leela-zero/blob/master/src/GameState.cpp#L111) but that happens after move is generated!

you could change the assert to <=

killerducky:
I tested `<=` -- it works. Rebased to that instead.

PR #577, opened 2018-01-06
delete more dead code

sethtroisi:


gcp:
I think that if you delete is_held in the SMP.cpp, then you can replace it by an atomic flag instead of an atomic bool.

PR #578, opened 2018-01-06
change m_lastmove to int

sethtroisi:
This compliments #577 and cleans up m_lastmove from which we only use the latest value.


gcp:
Needs a rebase due to redundancy with your other pull request.

sethtroisi:
rebased

PR #580, opened 2018-01-06
UCTNode field reorganization to optimize memory layout (attempt 2)

ihavnoid:
We had the discussion on how to reduce memory footprint and have many good ideas, but this is a simple stopgap change which is better than nothing until we have something better.

(prior discussion : https://github.com/gcp/leela-zero/pull/567)

72 bytes -> 56 bytes, this is 2 cache lines vs. 1 cache line in many CPUs.

### DIFF
--- a/src/UCTNode.h
+++ b/src/UCTNode.h
@@ -79,16 +79,11 @@ private:
     void link_nodelist(std::atomic<int>& nodecount,
                        std::vector<Network::scored_node>& nodelist,
                        float init_eval);
-
-    // Tree data
-    std::atomic<bool> m_has_children{false};
-    std::vector<node_ptr_t> m_children;
-
     // Move
-    int m_move;
+    int16_t m_move;
     // UCT
+    std::atomic<int16_t> m_virtual_loss{0};
     std::atomic<int> m_visits{0};
-    std::atomic<int> m_virtual_loss{0};
     // UCT eval
     float m_score;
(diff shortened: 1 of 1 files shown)


src/UCTNode.h
```diff
@@ -79,16 +79,11 @@ class UCTNode {
...
                        std::vector<Network::scored_node>& nodelist,
                        float init_eval);
-
-    // Tree data
-    std::atomic<bool> m_has_children{false};
-    std::vector<node_ptr_t> m_children;
-
     // Move
-    int m_move;
+    int16_t m_move;
```
gcp: Needs a std:: prefix as it's not guaranteed to be in the global namespace. Also perhaps the move type should be a typedef if we do this.

src/UCTNode.h
```diff
@@ -79,16 +79,11 @@ class UCTNode {
...
     // UCT
+    std::atomic<int16_t> m_virtual_loss{0};
```
gcp: Same remark. Note that this could be an (u)int8_t too.

ihavnoid:
Pushed a new commit addressing concerns.  I did leave the m_virtual_loss at uint16_t simply because further reducing it won't save any memory footprint at this point.

gcp:
I guess in theory you could have 128 CPUs * 3 > 256 virtual losses anyway.

PR #581, opened 2018-01-06
After search, sync all TTable entries in root node

killerducky:
Because of the interaction of a new search and the TTable, the policy prior can be distorted. If the new search doesn't do any additional searches to a node that is in the TTable, we don't pick up those playouts. I found cases were nodes with as many as 487 playouts in a TTable entry were left out. In another case 87 different moves were left out, although most of them were very low playouts. 

See below a couple examples of the dump_stats output before and after this change. And here is the full [diff.txt](https://github.com/gcp/leela-zero/files/1609157/diff.txt)

This is the most conservative fix for the problem, the search and gameplay is completely unaffected, only the dump_stats output and dump_training output is changed. I spot checked the dump_training output, it looks ok.

I think this sort of interaction between TTable and the search will distort the shape of the search throughout the tree. After the first search in a node, it will suddenly have many playouts for that one node. Q, puct, and U are supposed to control the narrowness of the search in a smooth way but that will be distorted by this. I think most of the problems would go away if we implement a real tree reuse.

There are many moving parts here so it's complicated. @gcp I'm looking for some input on what to do about these issues.

```
rm before1.log; printf "auto\nprintsgf before.sgf\ndump_training b train_before.txt" | ./leelaz -t 1 -p 1600 --noponder -r 1 -n -d -w ~/networks/58da61769a5fb12ff555ee927f16b50d4c24de73a1170a85df1b0d2d706bdef3 -s 3 -l before1.log

rm after1.log; printf "auto\nprintsgf after.sgf\ndump_training b train_after.txt" | ./leelaz -t 1 -p 1600 --noponder -r 1 -n -d -w ~/networks/58da61769a5fb12ff555ee927f16b50d4c24de73a1170a85df1b0d2d706bdef3 -s 3 -l after1.log

diff --side-by-side -W 88 before1.log after1.log

 G10 ->    4847 (V: 45.13%) (N: 21.92%)          G10 ->    4847 (V: 45.13%) (N: 21.92%)
 M12 ->     640 (V: 43.88%) (N:  9.10%)          M12 ->     640 (V: 43.88%) (N:  9.10%)
  R2 ->     632 (V: 44.59%) (N:  2.44%)           R2 ->     632 (V: 44.59%) (N:  2.44%)
                                           >     G11 ->     487 (V: 44.83%) (N:  0.33%)
 H12 ->     196 (V: 43.04%) (N:  4.66%)          H12 ->     196 (V: 43.04%) (N:  4.66%)
 B17 ->     113 (V: 41.31%) (N:  4.66%)          B17 ->     113 (V: 41.31%) (N:  4.66%)
 L12 ->      82 (V: 44.22%) (N:  0.77%)          L12 ->      82 (V: 44.22%) (N:  0.77%)
 H11 ->      80 (V: 43.34%) (N:  1.44%)          H11 ->      80 (V: 43.34%) (N:  1.44%)
 J13 ->      64 (V: 35.93%) (N:  8.23%)          J13 ->      64 (V: 35.93%) (N:  8.23%)
  P8 ->      58 (V: 43.85%) (N:  0.77%)           P8 ->      58 (V: 43.85%) (N:  0.77%)
  S4 ->      56 (V: 44.06%) (N:  0.48%)           S4 ->      56 (V: 44.06%) (N:  0.48%)
  Q2 ->      37 (V: 41.67%) (N:  1.56%)           Q2 ->      37 (V: 41.67%) (N:  1.56%)
 N12 ->      36 (V: 34.05%) (N:  4.63%)          N12 ->      36 (V: 34.05%) (N:  4.63%)
  C5 ->      32 (V: 43.28%) (N:  0.50%)           C5 ->      32 (V: 43.28%) (N:  0.50%)
  B6 ->      31 (V: 41.22%) (N:  1.79%)           B6 ->      31 (V: 41.22%) (N:  1.79%)
 O14 ->      22 (V: 31.27%) (N:  3.47%)          O14 ->      22 (V: 31.27%) (N:  3.47%)
  F2 ->      21 (V: 42.99%) (N:  0.51%)           F2 ->      21 (V: 42.99%) (N:  0.51%)
  B4 ->      20 (V: 41.13%) (N:  1.10%)           B4 ->      20 (V: 41.13%) (N:  1.10%)
  C8 ->      18 (V: 40.48%) (N:  0.92%)           C8 ->      18 (V: 40.48%) (N:  0.92%)
 A13 ->      18 (V: 34.26%) (N:  2.86%)          A13 ->      18 (V: 34.26%) (N:  2.86%)
 M17 ->      17 (V: 36.94%) (N:  1.77%)          M17 ->      17 (V: 36.94%) (N:  1.77%)
  E7 ->      16 (V: 33.01%) (N:  2.80%)           E7 ->      16 (V: 33.01%) (N:  2.80%)
 E14 ->      15 (V: 39.52%) (N:  1.28%)          E14 ->      15 (V: 39.52%) (N:  1.28%)
 N14 ->      14 (V: 38.87%) (N:  1.12%)          N14 ->      14 (V: 38.87%) (N:  1.12%)
  G2 ->      14 (V: 37.33%) (N:  1.63%)           G2 ->      14 (V: 37.33%) (N:  1.63%)
 J12 ->      13 (V: 41.21%) (N:  0.69%)          J12 ->      13 (V: 41.21%) (N:  0.69%)
 K12 ->      13 (V: 40.83%) (N:  0.59%)          K12 ->      13 (V: 40.83%) (N:  0.59%)
 F13 ->      11 (V: 41.41%) (N:  0.47%)          F13 ->      11 (V: 41.41%) (N:  0.47%)
 A18 ->      11 (V: 37.43%) (N:  1.24%)          A18 ->      11 (V: 37.43%) (N:  1.24%)
  E8 ->      10 (V: 38.35%) (N:  0.77%)           E8 ->      10 (V: 38.35%) (N:  0.77%)
 L14 ->      10 (V: 37.06%) (N:  0.77%)          L14 ->      10 (V: 37.06%) (N:  0.77%)
                                           >     H10 ->       9 (V: 41.90%) (N:  0.29%)
  S5 ->       9 (V: 40.11%) (N:  0.44%)           S5 ->       9 (V: 40.11%) (N:  0.44%)
 C13 ->       9 (V: 38.04%) (N:  0.63%)          C13 ->       9 (V: 38.04%) (N:  0.63%)
 M18 ->       4 (V: 38.12%) (N:  0.49%)          M18 ->       4 (V: 38.12%) (N:  0.49%)
 S10 ->       4 (V: 30.80%) (N:  0.97%)          S10 ->       4 (V: 30.80%) (N:  0.97%)
  L4 ->       4 (V: 25.57%) (N:  1.10%)           L4 ->       4 (V: 25.57%) (N:  1.10%)
                                           >     S19 ->       2 (V: 31.00%) (N:  0.29%)
 L19 ->       2 (V: 27.86%) (N:  0.55%)          L19 ->       2 (V: 27.86%) (N:  0.55%)
7685 visits, 364118 nodes, 1600 playouts   |    7685 visits, 364118 nodes, 1600 playouts

  J3 ->    3832 (V: 62.50%) (N: 30.17%)           J3 ->    3832 (V: 62.50%) (N: 30.17%)
  H5 ->      61 (V: 40.80%) (N: 22.36%)           H5 ->      61 (V: 40.80%) (N: 22.36%)
  H6 ->      58 (V: 54.67%) (N:  5.48%)           H6 ->      58 (V: 54.67%) (N:  5.48%)
 R19 ->      19 (V: 48.24%) (N:  5.25%)          R19 ->      19 (V: 48.24%) (N:  5.25%)
 R10 ->      13 (V: 29.14%) (N:  8.44%)          R10 ->      13 (V: 29.14%) (N:  8.44%)
 O18 ->      11 (V: 56.52%) (N:  0.36%)          O18 ->      11 (V: 56.52%) (N:  0.36%)
 O14 ->      11 (V: 51.52%) (N:  1.37%)          O14 ->      11 (V: 51.52%) (N:  1.37%)
  A2 ->      10 (V: 58.88%) (N:  0.28%)           A2 ->      10 (V: 58.88%) (N:  0.28%)
 N15 ->      10 (V: 55.97%) (N:  0.29%)          N15 ->      10 (V: 55.97%) (N:  0.29%)
  P2 ->      10 (V: 55.77%) (N:  0.35%)           P2 ->      10 (V: 55.77%) (N:  0.35%)
 Q11 ->      10 (V: 52.98%) (N:  1.15%)          Q11 ->      10 (V: 52.98%) (N:  1.15%)
 E15 ->       8 (V: 55.92%) (N:  0.46%)          E15 ->       8 (V: 55.92%) (N:  0.46%)
 S17 ->       7 (V: 44.19%) (N:  1.40%)          S17 ->       7 (V: 44.19%) (N:  1.40%)
  Q4 ->       7 (V: 21.37%) (N:  5.59%)           Q4 ->       7 (V: 21.37%) (N:  5.59%)
 J12 ->       6 (V: 57.06%) (N:  0.14%)          J12 ->       6 (V: 57.06%) (N:  0.14%)
  O8 ->       6 (V: 54.86%) (N:  0.18%)           O8 ->       6 (V: 54.86%) (N:  0.18%)
 S19 ->       6 (V: 53.28%) (N:  0.29%)          S19 ->       6 (V: 53.28%) (N:  0.29%)
  F5 ->       6 (V: 46.72%) (N:  0.38%)           F5 ->       6 (V: 46.72%) (N:  0.38%)
  J9 ->       6 (V: 40.00%) (N:  0.46%)           J9 ->       6 (V: 40.00%) (N:  0.46%)
 G11 ->       6 (V: 37.80%) (N:  1.31%)          G11 ->       6 (V: 37.80%) (N:  1.31%)
 G12 ->       6 (V: 34.39%) (N:  0.29%)          G12 ->       6 (V: 34.39%) (N:  0.29%)
  O3 ->       6 (V: 32.99%) (N:  0.30%)           O3 ->       6 (V: 32.99%) (N:  0.30%)
 O19 ->       6 (V: 32.39%) (N:  0.41%)          O19 ->       6 (V: 32.39%) (N:  0.41%)
  G9 ->       6 (V: 29.59%) (N:  0.38%)           G9 ->       6 (V: 29.59%) (N:  0.38%)
  A7 ->       6 (V: 21.19%) (N:  0.79%)           A7 ->       6 (V: 21.19%) (N:  0.79%)
  Q1 ->       5 (V: 56.43%) (N:  0.08%)           Q1 ->       5 (V: 56.43%) (N:  0.08%)
  E8 ->       5 (V: 55.99%) (N:  0.21%)           E8 ->       5 (V: 55.99%) (N:  0.21%)
  D4 ->       5 (V: 33.93%) (N:  0.26%)           D4 ->       5 (V: 33.93%) (N:  0.26%)
  G4 ->       5 (V: 32.36%) (N:  0.25%)           G4 ->       5 (V: 32.36%) (N:  0.25%)
 H19 ->       5 (V: 31.75%) (N:  0.28%)          H19 ->       5 (V: 31.75%) (N:  0.28%)
 F19 ->       5 (V: 30.82%) (N:  0.17%)          F19 ->       5 (V: 30.82%) (N:  0.17%)
 A17 ->       5 (V: 24.96%) (N:  0.24%)          A17 ->       5 (V: 24.96%) (N:  0.24%)
  S2 ->       4 (V: 52.17%) (N:  0.10%)           S2 ->       4 (V: 52.17%) (N:  0.10%)
  B6 ->       4 (V: 47.54%) (N:  0.13%)           B6 ->       4 (V: 47.54%) (N:  0.13%)
  K5 ->       4 (V: 41.32%) (N:  0.13%)           K5 ->       4 (V: 41.32%) (N:  0.13%)
  E5 ->       4 (V: 41.01%) (N:  0.11%)           E5 ->       4 (V: 41.01%) (N:  0.11%)
  F8 ->       4 (V: 39.41%) (N:  0.10%)           F8 ->       4 (V: 39.41%) (N:  0.10%)
 P12 ->       4 (V: 36.16%) (N:  0.16%)          P12 ->       4 (V: 36.16%) (N:  0.16%)
  K2 ->       4 (V: 31.78%) (N:  1.40%)           K2 ->       4 (V: 31.78%) (N:  1.40%)
 B15 ->       4 (V: 31.16%) (N:  0.11%)          B15 ->       4 (V: 31.16%) (N:  0.11%)
 Q12 ->       4 (V: 28.40%) (N:  0.15%)          Q12 ->       4 (V: 28.40%) (N:  0.15%)
 P19 ->       4 (V: 27.65%) (N:  0.14%)          P19 ->       4 (V: 27.65%) (N:  0.14%)
                                           >     O13 ->       3 (V: 51.52%) (N:  0.07%)
                                           >      E2 ->       3 (V: 47.04%) (N:  0.04%)
 T11 ->       3 (V: 45.69%) (N:  0.14%)          T11 ->       3 (V: 45.69%) (N:  0.14%)
  K6 ->       3 (V: 31.19%) (N:  0.57%)           K6 ->       3 (V: 31.19%) (N:  0.57%)
  A5 ->       3 (V: 29.13%) (N:  1.57%)           A5 ->       3 (V: 29.13%) (N:  1.57%)
                                           >      F1 ->       2 (V: 42.83%) (N:  0.03%)
 G13 ->       2 (V: 41.77%) (N:  0.12%)          G13 ->       2 (V: 41.77%) (N:  0.12%)
  Q3 ->       2 (V: 40.68%) (N:  0.58%)           Q3 ->       2 (V: 40.68%) (N:  0.58%)
                                           >     T12 ->       2 (V: 37.46%) (N:  0.02%)
                                           >      S1 ->       2 (V: 37.26%) (N:  0.02%)
  Q7 ->       2 (V: 34.02%) (N:  0.59%)           Q7 ->       2 (V: 34.02%) (N:  0.59%)
  S6 ->       2 (V: 30.63%) (N:  0.13%)           S6 ->       2 (V: 30.63%) (N:  0.13%)
  T9 ->       2 (V: 27.26%) (N:  1.18%)           T9 ->       2 (V: 27.26%) (N:  1.18%)
                                           >     C19 ->       2 (V: 27.12%) (N:  0.07%)
                                           >     D19 ->       2 (V: 25.26%) (N:  0.07%)
  A6 ->       2 (V: 18.16%) (N:  0.26%)           A6 ->       2 (V: 18.16%) (N:  0.26%)
 H11 ->       2 (V: 10.36%) (N:  0.13%)          H11 ->       2 (V: 10.36%) (N:  0.13%)
                                           >     T15 ->       1 (V: 58.10%) (N:  0.03%)
                                           >      L1 ->       1 (V: 56.13%) (N:  0.02%)
                                           >      T4 ->       1 (V: 51.38%) (N:  0.03%)
                                           >      H1 ->       1 (V: 48.50%) (N:  0.05%)
                                           >      A3 ->       1 (V: 48.49%) (N:  0.02%)
<74 move moves in after1.log but not in before 1.log>
4332 visits, 220085 nodes, 1600 playouts        4332 visits, 220085 nodes, 1600 playouts
```

### DIFF
--- a/src/GameState.cpp
+++ b/src/GameState.cpp
@@ -336,6 +336,6 @@ void GameState::place_free_handicap(int stones) {
 
 const FullBoard& GameState::get_past_board(int moves_ago) const {
     assert(moves_ago >= 0 && (unsigned)moves_ago <= m_movenum);
-    assert(m_movenum + 1 == game_history.size());
+    assert(m_movenum + 1 <= game_history.size());
     return game_history[m_movenum - moves_ago]->board;
 }
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -44,6 +44,18 @@ UCTSearch::UCTSearch(GameState & g)
     set_playout_limit(cfg_max_playouts);
 }
 
+// Iterate through all children and sync their TTable results.
+void UCTSearch::ttable_sync_all_children(GameState & currstate, UCTNode* const node) {
+    const auto komi = currstate.get_komi();
+    for (auto& child : node->get_children()) {
+        auto move = child->get_move();
+        currstate.play_move(move);
+        const auto hash = currstate.board.get_hash();
+        TTable::get_TT()->sync(hash, komi, child.get());
+        currstate.undo_move();
+    }
+}
+
 SearchResult UCTSearch::play_simulation(GameState & currstate, UCTNode* const node) {
     const auto color = currstate.get_to_move();
     const auto hash = currstate.board.get_hash();
@@ -386,6 +398,10 @@ int UCTSearch::think(int color, passflag_t passflag) {
     // display search info
     myprintf("\n");
 
+    // Conservative way to pick up playouts from the TTable
+    // that we haven't already. This method does not change
+    // the way the search works.
+    ttable_sync_all_children(m_rootstate, &m_root);
     dump_stats(m_rootstate, m_root);
     Training::record(m_rootstate, m_root);
 
--- a/src/UCTSearch.h
+++ b/src/UCTSearch.h
@@ -88,6 +88,7 @@ private:
     std::string get_pv(KoState& state, UCTNode& parent);
     void dump_analysis(int playouts);
     int get_best_move(passflag_t passflag);
+    void ttable_sync_all_children(GameState & currstate, UCTNode* const node);
 
     GameState & m_rootstate;
     UCTNode m_root{FastBoard::PASS, 0.0f, 0.5f};


gcp:
What's the goal of this? The effect I can see is that moves with a low policy prior that were previously discovered to be reasonable will be sure to be re-discovered. Fine, but do we really care about this? Why would this be a problem?

I don't see why "the policy prior can be distorted" or "Q, puct, and U are supposed to control the narrowness of the search in a smooth way but that will be distorted by this".

gcp:
I'm aware of the "problem" with carried over search results meaning that it could be effectively impossible for a new best move to overtake the current one. But that's only a "problem" in as that it's possible to find a better allocation for search timing, which you've already demonstrated (-v 2200 vs -p 1600).

The gain of this pull request on the other hand, I'm at a loss?

killerducky:
Take the example of `G11 ->     487 (V: 44.83%) (N:  0.33%)`. It had a prior of 0.33%. But it has the 2nd best V: 44.59, only G10 had a higher one with 45.13%. But the current code will give the target policy of 0. It will give many other moves with much worse Values a much higher target policy prior. For example it will give J13 a policy target of 64 even though the search quickly realized J13 is very bad (V: 35.93%). Why do we want to train the network to give J13 a higher policy when the value is 9% lower and it received over 400 fewer visits overall than G11?

Another angle: I think the results of the search should be independent of what is in the cache (outside of randomness effects). Taking the G11 example again: the final policy target of a move that received 487 out of 7685 visits should not depend on if those visits happened before a node was root or after.

Another angle: This is information we spent CPU time computing, and the CPU cost to retrieve it is very small. Why throw away that information?

amj:
@killerducky when you say "the policy prior can be distorted" i assume you mean that the problem is that the final outputs of the probabilities -- "pi" in the paper, i.e. the targets for training -- are distorted.

So the problem would be distorting the pi-values that are used as the training target for the policy net.  That'd be a big deal.

If the ttable is changing the output of the search (i.e., the probabilities-pi, not the final move), then it's gonna change the training.

Is that what you meant, @killerducky ?

roy7:
I've noticed before that moves that showed up in prior search but aren't re-visited at all in a later search don't appear in the summary printout after a move, but the visits are counted in the total visits count at the bottom of the summary printout. I thought that was just a logging oddity with no actual effect otherwise. Is the training data stored and send into the database behaving the same way? Nodes that aren't re-visited aren't just hidden in the console output, but are also hidden in the training data weights?

For example with playouts limited to 1, we can see something like this:

> genmove w
> Thinking at most 50.0 seconds...
> NN eval=0.041853
>   L7 ->       1 (V: 16.93%) (N: 32.61%) PV: L7 K9
>   K9 ->       1 (V: 15.87%) (N:  8.11%) PV: K9 H11
> 2 visits, 825 nodes, 2 playouts, 100 n/s
> = L7
> genmove b
> Thinking at most 50.0 seconds...
> NN eval=0.888761
>   K6 ->       1 (V: 93.50%) (N: 14.17%) PV: K6 L6
>   G7 ->       1 (V: 93.44%) (N: 16.88%) PV: G7 F8
> 3 visits, 821 nodes, 2 playouts, 66 n/s
> = K6

where K9 must be the hidden third visit in the "3 visits".

killerducky:
@amj yes thanks for the clarification. I'm talking about the AGZ paper `probability pi` that for LZ is written out by dump_training. In debug_stats, "N" is the `probability p` that the current network outputs, plus the Dirichlet noise. 

@roy7 I think that's right. For the second move K9 will not be recorded in the dump_training for the `probability pi`. Instead the code in dump_training adds up the visit count of each child
https://github.com/gcp/leela-zero/blob/next/src/Training.cpp#L119

And outputs child->get_visits/sum_visits:
https://github.com/gcp/leela-zero/blob/next/src/Training.cpp#L135

In your example K6 and G6 will get 0.5, and everything else including K9 gets 0.

roy7:
I don't know LZ internals so please humor on this, but I have a question on how the tree data is reused on later steps. Say we have a situation where we have moves A B, followed by 2 choices, call them C with prior 99.9% and D with .1%, and all other priors are 0%. So search will heavily lean at first to A B C.

Say we go 50K search that's almost all in A B C and looking good, when we discover a trap down that path where we're going to lose. We also discover A B D doesn't lose at all. So the V for A B C drops low and the V for A B D turns out really high.

Search continues another 50K steps, so we have two moves A B C and A B D. Just to make up sample figures, say it is like after A B:

C -> 50001 (V: 5%) (N: 99.9%)
D -> 49999(V: 95%) (N: 0.1%)

And we move to A. Opponent moves to B. Now we are in a situation where we have 2 moves in our heatmap, C with 99.9% and D with 0.1%. Is search starting at this point, with our 100K visits cached from the prior turn, going to start heavy with D (high V% and equal visits but low prior N%), or heavy with C (low V% but equal visits and high N% prior).

In other words, when we start our search on our second move, are we only using the heatmap (priors) at that moment in time and not accessing the TTable for move D until we happen to have 1 search go there naturally and then it'll "discover" all of the data in the TTable waiting for it? Or is the TTable knowledge on C and D immediately available at the start of the search of move 2, so really all search will pour into D and C will mostly be ignored?

I'm just nervous our D result we found on move 1 is "forgotten" until we stumble back into it and we'll start move 2 based on the C prior.

ywrt:
Absent the jitter due to the random rotation/reflection of the board before being fed into the network, the non-root nodes of the tree search should be deterministic. Which is to say if that a TTable entry is created at depth N, it should be reused for depth N-1.

However, the root node is different due to the addition of Dirichlet noise. That would result in some nodes no longer being expanded (and other nodes being expanded instead).

Which seems to match what you see? some nodes (G11) are missed, and other nodes (H10, S19) have been added instead?

roy7:
@killerducky I'm not sure if this is working as expected, but I checked out your ttable_sync branch and compiled a fresh leelaz from your fork. Doing my 1 playout test still hides a visit. From game start:

> NN eval=0.471891
> 
>  Q16 ->       1 (V: 53.73%) (N: 20.85%) PV: Q16 Q4
>  D16 ->       1 (V: 53.18%) (N: 20.55%) PV: D16 D4
> 2 visits, 1084 nodes, 2 playouts, 33 n/s
> 
> = Q16
> 
> genmove w
> Thinking at most 50.0 seconds...
> NN eval=0.462715
> 
>   Q4 ->       1 (V: 53.60%) (N: 26.90%) PV: Q4 D16
>  D16 ->       1 (V: 52.16%) (N: 26.68%) PV: D16 D4
> 2 visits, 1081 nodes, 2 playouts, 100 n/s
> 
> = Q4
> 
> genmove b
> Thinking at most 50.0 seconds...
> NN eval=0.464028
> 
>   D4 ->       1 (V: 54.79%) (N: 42.69%) PV: D4 D16
>  D16 ->       1 (V: 54.48%) (N: 43.65%) PV: D16 D4
> 3 visits, 1078 nodes, 2 playouts, 100 n/s
> 
> = D4

gcp:
>However, the root node is different due to the addition of Dirichlet noise. That would result in some nodes no longer being expanded (and other nodes being expanded instead).

I agree this is what likely happens. Some minor skew due to Dirichlet or rotations can cause another node to get a bit more of the exploration target, which means we will never care about expanding the node that is seen as missing here. I don't see why we would care about this.

>Another angle: This is information we spent CPU time computing, and the CPU cost to retrieve it is very small. Why throw away that information?

You can extend this argument to the entire tree and do TTable probes in every node on create_children. Maybe that makes the program stronger and maybe it does not. That would make more sense to test than a hack at the root only.

>If the ttable is changing the output of the search (i.e., the probabilities-pi, not the final move), then it's gonna change the training.

The WHOLE IDEA of recovering previous search outputs is to use them and change the output. That's the EXPECTED result. Claiming that this is a "distortion of the search probabilities" is missing the point by a few miles.

amj:
Sorry, maybe i misunderstand killerducky's report: I took it to mean that all the visits of the children of the root node are *not* being correctly counted when saving the probabilities.  

I.e., the search outputs two things: 
 - a move
 - a set of visit counts of child nodes. 

If it's getting the visit counts of child nodes wrong through the ttable evicting nodes or something, then something's wrong.  If it's not then it's fine.  I took @killerducky 's comments to mean that the visit counts aren't correct -- e.g., where did G11, the 4th best move go?  It's just vanished?

It's probably fine, because after the 30th move the temperature infinitesimal turns it into a onehot anyway, but it seems like it could have some weird side effects.  

Sorry for the drive by comments.  Good luck!

killerducky:
@gcp ok how about instead of this approach I do a different pull re-implementing my previous idea, but with a new command line switch --visits? I think you were open to merging that idea.

Assuming I work on that, I want to double check that you agree with this algorithm, I will use the -v 2200 version you preferred in my example:

- Previous search did 2200 playouts, of which 2000 are for D4, so that move is picked.
- Search for the position after D4 is played will do 200 additional playouts, making the total visit count 2200 for this root node.
- Before calling dump_training, I will do the same thing as this pull, probe the entire TTable for the root only, to ensure all 2200 of those visits are used for the output of dump_training.

Does this sound good?

roy7:
@amj 

> If it's getting the visit counts of child nodes wrong through the ttable evicting nodes or something, then something's wrong. 

I believe that is indeed the concern. dump_stats and dump_training output the same information. In the original example, G11's visits were not appearing in the dump. It just vanished. Without that in the training output, the NN would learn G11 as a 0 prior instead of a 487/7685 prior. Making it even less likely to find in the future, although it's the 2nd strongest move that was found.

@gcp 

> The WHOLE IDEA of recovering previous search outputs (with TTable or tree reuse) is to use them and change the output.

I feel like you and @killerducky are in perfect agreement here. But in the example, G11 is not being recovered and used. It is being lost and wrong information sent to the NN to learn (dump_training 0 prior instead of 487/7685 prior). I wonder if dump_training even scales the targets to add to 1 (if that is needed), since the values printing don't sum to the visits count at the bottom.

Apologies if I'm just plain wrong here, but for the TTable info to appear in dump_stats and dump_training, one new visit this move has to be generated to G11 for the prior search to load in, is that correct? If so, that one extra search would result in a prior of 488/7686 instead of 0. That seems like a huge swing for a single visit (.000 -> .063).

roy7:
@killerducky If you didn't spot my earlier comment, is it an issue that using your ttable_sync branch I was still able to come up with a dump that only contained 2 visits even though the summary stats at end said 3 visits? I cloned your fork and changed to that branch to test it.

gcp:
>I do a different pull re-implementing my previous idea, but with a new command line switch --visits? I think you were open to merging that idea.

Yes, I think it would be good default for a next run. You've proven it's superior for the same computing effort.

I think it might also be interesting to extend what you tried here to every node created in create_children, though there's a speed impact. That is, make the logic here apply to every node added to the tree, i.e. do the TTable probe *before* they get selected.

killerducky:
@roy7 I saw that, I fixed a couple display errors in dump_stats and will make a pull request for those. After those fixes there is still some discrepancy, but it's due to the way visits to a node are counted. I think the first time the root node is visited that is included in the visit count. But it doesn't create any children to list in the dump_stats output -- the first visit to a node stops there, doesn't descend to any children. It doesn't happen on the first few moves, but later the root visit count will always be slightly larger than the sum of the child counts. As @gcp pointed out there is code in dump_training to account for this difference.

@gcp ok I will work on that. I did try the idea of extending this concept recursively to every child before the search, and I've started a tournament with that change. So far there is a very small improvement but I don't have enough results yet to be sure.

roy7:
Even if @killerducky's tournament doesn't show a big change in a single head to head tournament, I think the issue is it will provide stronger training long-term since currently we're sometimes losing information on low probability moves that have good exploration data being lost.

killerducky:
#590 is a better solution than this one.

A better solution for #315 would also help.

roy7:
@killerducky Since #590 didn't end up addressing this since the children sync was removed, does that make #632 as the better solution now?

killerducky:
Yes #632 will fix this. See also #645.

PR #586, opened 2018-01-07
Removes a rare possible race condition on NNCache

ihavnoid:
I am not sure if this is a real bug or not, so I am posting a pull request anyway.  Please have a look and close if this seems to be a non-issue.

The concern here is that the prior lookup() does things this way:
1) acquire lock
2) search cache and find entry
3) return method, and release lock
4) create a copy, and return (this happens on Network::get_scored_moves)

If the entry gets deleted on insert() between 3) and 4) we can potentially segfault or have a corrupted table because the pointer is going to be deleted.

This patch tries to plug the hole by copying Netresult inside the lookup() method while holding the lock.

### DIFF
--- a/src/NNCache.cpp
+++ b/src/NNCache.cpp
@@ -43,26 +43,28 @@ static size_t compute_hash(const Network::NNPlanes& features) {
     return hash;
 }
 
-const Network::Netresult* NNCache::lookup(const Network::NNPlanes& features) {
+bool NNCache::lookup(const Network::NNPlanes& features, Network::Netresult & result) {
+    auto hash = compute_hash(features);
+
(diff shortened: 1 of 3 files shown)


SimonX200:
Not sure, but for me it looks like that this change will reduce performance by duplicating a vector.

using scored_node = std::pair<float, int>;
using Netresult = std::pair<std::vector<scored_node>, float>;

Have you checked how the performance changed?

gcp:
1. Worrying about the performance impact is completely and utterly pointless if the original code is broken in the first place.
2. The copy is already happening outside the function, this change just moves it under the lock.

PR #589, opened 2018-01-07
Minor fixes for dump_stats and get_pv.

killerducky:
Three minor fixes, found while testing very low playouts (`-t 1 -p 2 --noponder`).

- The PV was sometimes outputting an extra random move. 
- movecount>2 check causes it to output an extra random move when there is only 1 move searched. I think the !node->get_visits check is sufficient? Or is there some other corner case I didn't consider?
- the elapsed_centis check may not be needed at all, because the code below adds 1 to avoid the possibility of divide by zero. I left it in but changed it to check for it being less than negative one, although that should be impossible. The code as is suppresses the output for the testing I was doing since it ran fast enough to make elapsed_centis=0.

@roy7 this fixes a few issues you mentioned.

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -108,7 +108,7 @@ void UCTSearch::dump_stats(KoState & state, UCTNode & parent) {
     const int color = state.get_to_move();
 
     // sort children, put best move on top
-    parent.sort_root_children(color);
+    parent.sort_children(color);
 
 
     if (parent.get_first_child()->first_visit()) {
@@ -117,7 +117,9 @@ void UCTSearch::dump_stats(KoState & state, UCTNode & parent) {
 
     int movecount = 0;
     for (const auto& node : parent.get_children()) {
-        if (!node->get_visits()) break;
+        // Always display at least two moves. In the case there is
+        // only one move searched the user could get an idea why.
+        if (++movecount > 2 && !node->get_visits()) break;
 
         std::string tmp = state.move_to_text(node->get_move());
         std::string pvstring(tmp);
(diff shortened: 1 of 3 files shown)


killerducky:
I just amended this with another minor fix to do sort_root_children on parent instead of m_root. We could also just remove those two arguments since right now dump_stats is only ever called on the root node.

BTW when I rewrote history by `git commit --amend` and `git push -f` it wasn't logged in any way here. I wanted to keep the commits clean but the silent update is a bit scary...

src/UCTSearch.cpp
```diff
@@ -108,7 +108,7 @@ void UCTSearch::dump_stats(KoState & state, UCTNode & parent) {
     const int color = state.get_to_move();
 
     // sort children, put best move on top
-    m_root.sort_root_children(color);
+    parent.sort_root_children(color);
```
gcp: I think sort_root_children is now safe to use anywhere, so it should probably just be renamed.

gcp:
>Or is there some other corner case I didn't consider?

I did this to make it clear the display is not truncated but there were simply no other moves searched, and to show the "second best" after what the engine considers forced. Thought it would be useful for analysis.

src/UCTSearch.cpp
```diff
@@ -117,7 +117,7 @@ void UCTSearch::dump_stats(KoState & state, UCTNode & parent) {
 
     int movecount = 0;
     for (const auto& node : parent.get_children()) {
-        if (++movecount > 2 && !node->get_visits()) break;
```
gcp: See comment on why this behaves like it does.

killerducky: Ah yes good idea, in fact I'm tempted to just always display the first move with 0 visits. What do you think? For now I'll just put it back the way it was.

src/UCTSearch.cpp
killerducky: renamed.

Issue #576, opened 2018-01-06
TTable::sync in a ko situation


mkw1234:
Following two positions (case 1 and case 2) have equal hash values but do not have same positional values.  However, Leelaz MCTS might treat these two positions as equal when a node is synced with another node in the TTable of equal hash value.  Shouldn't the nodes not the be synced in this case?

![image](https://user-images.githubusercontent.com/35014179/34636960-68ceb2a4-f2ef-11e7-9636-f13e72cfb221.png)
**Case 1**

![image](https://user-images.githubusercontent.com/35014179/34636966-74076652-f2ef-11e7-9199-b330c92db8f5.png)
**Case 2**

gcp:
Shouldn't the prisoners by unequal if it is a ko?

mkw1234:
No, not necessarily.  In two positions shown above there is 1 black stone prisoner.  Same hash values.  The difference is in case 1, black can take the ko, but in case 2 black can not take the ko.

ThePokerDude:
>  Shouldn't the prisoners be unequal if it is a ko? They are included in the hash. <

Not necessarily. For examlpe: 
1) (move number for example 24 (e6f6 already played))  last move = e4 wins opponent's stone -> ko 
2) (move number 22 - e6f6 not played yet) last moves: e4 wins opponent's stone, e6 f6 -> no ko

joseph-ireland:
This can happen in a triple ko too. The only fully correct solution I can see is to enumerate the superko-illegal points for each position in advance and include them in the hash.

mkw1234:
In addition to ko issue, I think there might be a general problem with the syncing concept.  In baduk,  correct timing of moves is very important.  If the program plays a good sequence of moves to reach a good position, NN should be rewarded.  If the program plays bad sequence of moves but reaches same position by accident, NN should not be rewarded for that.  But by syncing a node to another node based on matching board configuration, NN might be rewarded for a bad sequence of moves and develop a bad habit.

gcp:
Your last comment makes no sense. The evaluation fed back to the root is the one from the leaves. If it comes from a bad sequence, that means the opponent could have forced an inferior branch earlier up the tree and have stopped you from reaching it.

gcp:
>The difference is in case 1, black can take the e4 stone, but in case 2 black can not take the e4 stone.

[...]

PR #590, opened 2018-01-07
Add new option --visits

killerducky:
When this option is used, include past search results and stop the search when the sum of the children's visits reaches --visits.

I'm re-running the tournament to ensure we still get an Elo improvement on same CPU usage. 

I made it so ttable_sync_all_children is only called when the --visits argument is used. This way the behavior when using --playouts doesn't change. I also verified the code in Training::record that does sum_visits always gets exactly --visits.

### DIFF
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -63,8 +63,7 @@ static void parse_commandline(int argc, char *argv[], bool & gtp_mode) {
                        "Weaken engine by limiting the number of playouts. "
                        "Requires --noponder.")
         ("visits,v", po::value<int>(),
-                       "Weaken engine by limiting the number of visits. "
-                       "Requires --noponder.")
+                       "Weaken engine by limiting the number of visits. ")
         ("lagbuffer,b", po::value<int>()->default_value(cfg_lagbuffer_cs),
                         "Safety margin for time usage in centiseconds.")
         ("resignpct,r", po::value<int>()->default_value(cfg_resignpct),
@@ -201,12 +200,6 @@ static void parse_commandline(int argc, char *argv[], bool & gtp_mode) {
 
     if (vm.count("visits")) {
         cfg_max_visits = vm["visits"].as<int>();
-        if (!vm.count("noponder")) {
-            myprintf("Nonsensical options: Visits are restricted but "
-                     "thinking on the opponent's time is still allowed. "
-                     "Add --noponder if you want a weakened engine.\n");
-            exit(EXIT_FAILURE);
-        }
     }
 
     if (vm.count("resignpct")) {
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -62,6 +62,9 @@ static void parse_commandline(int argc, char *argv[], bool & gtp_mode) {
         ("playouts,p", po::value<int>(),
                        "Weaken engine by limiting the number of playouts. "
                        "Requires --noponder.")
+        ("visits,v", po::value<int>(),
+                       "Weaken engine by limiting the number of visits. "
+                       "Requires --noponder.")
         ("lagbuffer,b", po::value<int>()->default_value(cfg_lagbuffer_cs),
                         "Safety margin for time usage in centiseconds.")
         ("resignpct,r", po::value<int>()->default_value(cfg_resignpct),
@@ -190,6 +193,16 @@ static void parse_commandline(int argc, char *argv[], bool & gtp_mode) {
         cfg_max_playouts = vm["playouts"].as<int>();
         if (!vm.count("noponder")) {
             myprintf("Nonsensical options: Playouts are restricted but "
+                     "thinking on the opponent's time is still allowed. "
+                     "Add --noponder if you want a weakened engine.\n");
+            exit(EXIT_FAILURE);
+        }
+    }
+
+    if (vm.count("visits")) {
+        cfg_max_visits = vm["visits"].as<int>();
+        if (!vm.count("noponder")) {
+            myprintf("Nonsensical options: Visits are restricted but "
                      "thinking on the opponent's time is still allowed. "
                      "Add --noponder if you want a weakened engine.\n");
             exit(EXIT_FAILURE);
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -117,7 +117,7 @@ void UCTSearch::dump_stats(KoState & state, UCTNode & parent) {
 
     int movecount = 0;
     for (const auto& node : parent.get_children()) {
-        if (++movecount > 2 && !node->get_visits()) break;
+        if (!node->get_visits()) break;
 
         std::string tmp = state.move_to_text(node->get_move());
         std::string pvstring(tmp);
@@ -266,6 +266,9 @@ std::string UCTSearch::get_pv(KoState & state, UCTNode& parent) {
     }
 
     auto& best_child = parent.get_best_root_child(state.get_to_move());
+    if (best_child.first_visit()) {
+        return std::string();
+    }
     auto best_move = best_child.get_move();
     auto res = state.move_to_text(best_move);
 
@@ -391,7 +394,7 @@ int UCTSearch::think(int color, passflag_t passflag) {
 
     Time elapsed;
     int elapsed_centis = Time::timediff_centis(start, elapsed);
-    if (elapsed_centis > 0) {
+    if (elapsed_centis+1 > 0) {
         myprintf("%d visits, %d nodes, %d playouts, %d n/s\n\n",
                  m_root.get_visits(),
                  static_cast<int>(m_nodes),
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -118,7 +118,7 @@ void UCTSearch::dump_stats(KoState & state, UCTNode & parent) {
 
     int movecount = 0;
     for (const auto& node : parent.get_children()) {
-        if (!node->get_visits()) break;
+        if (++movecount > 2 && !node->get_visits()) break;
 
         std::string tmp = state.move_to_text(node->get_move());
         std::string pvstring(tmp);
@@ -267,9 +267,6 @@ std::string UCTSearch::get_pv(KoState & state, UCTNode& parent) {
     }
 
     auto& best_child = parent.get_best_root_child(state.get_to_move());
-    if (best_child.first_visit()) {
-        return std::string();
-    }
     auto best_move = best_child.get_move();
     auto res = state.move_to_text(best_move);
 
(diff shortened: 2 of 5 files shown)


gcp:
Please pull the ttable_sync_children change out of this change.

killerducky:
Ok then how should I measure visits? root.get_visits? Or redo the sum_visits calculation on each playout?

killerducky:
Ok I removed it and replaced it with root.get_visits.

killerducky:
https://github.com/gcp/leela-zero/pull/546#issuecomment-355256661
> Has anyone verified that the probability map constructed in the learning data still has a sane/expected shape with this patch enabled? It seems that if the search aborts almost immediately because a move that was expected with high probability was played, the constructed map will be the one corresponding to only a few hundred playouts, rather than the expected 1600+?

Locally I added a debug output and ran a game with -v 2200. Snippet and link to full logs below. Most cases are close to 2200 but some are lower. Presumably the missing playouts are hidden in the TTable for moves that received 0 playouts in the new search.

[maxv.log](https://github.com/gcp/leela-zero/files/1610152/maxv.log)
grep sum_visits maxv.log | sort -r > visits.log
[visits.log](https://github.com/gcp/leela-zero/files/1610151/visits.log)

```
sum_visits=2491 root_visits=2200
sum_visits=2367 root_visits=2200
sum_visits=2366 root_visits=2200
sum_visits=2341 root_visits=2200
sum_visits=2325 root_visits=2200
<cut>
sum_visits=1872 root_visits=2200
sum_visits=1865 root_visits=2200
sum_visits=1806 root_visits=2200
sum_visits=1676 root_visits=2200
sum_visits=1652 root_visits=2200

```

gcp:
I think that's OK (those are moves which got very low priors *this* move).

I am doing some testing and the benefit of the TTable is right now rather low, I guess partly because of NNCache, partly because of #576, and the general issue that TTable entries are only added after the node has been selected once.

That makes me think it's worth testing the ttable_sync_children on all moves.

If I would remove the TTable, and we have no sub-tree reuse, the improvement from visits=2200 instead of playouts=1600 would also disappear, right? So that may be an argument for keeping it.

killerducky:
Yes I think this pull depends on the TTable right now. But if you add real tree reuse at the same time you remove TTable this code should still work. As you pointed out, recursively calling ttable_sync is just another way to do tree reuse.

I'm running a tournament testing both versions of this pull, I will report the results later.

ywrt:
This seems to be equivalent to counting NNCache misses?
i.e. Don't use the TTable. Instead,  run until there's been 1600 new network evaluations (aka, 1600 NNCache misses).

In principle, this should give the same result the sub-tree reuse (mod dirichlet noise)?

And it would get rid of the TTable which I personally find difficult to accurately reason about! ;)

killerducky:
The results are worse than before, in #546 it was 37-24 (61%) in favor of maxv-v-2200. But now it's only 43-41(51%) for lzmaxv-v-2200 and worse at 36-48 (43%) for lzmaxv-ttab-v-2200. Also according to avg cpu 2200 visits is using more cpu than lzbase 1600 playouts so if I lowered the visits it might lose the 51% advantage it was showing. 

I'll do some investigating later to see if I messed something up. As long as it's not worse my opinion is this is still useful because the output of dump_training will have a more consistent sum_visits. But that's just an opinion, now for the evidence (or lack of it):

Naming convention:
```
lzbase-p-1600 - base with -p 1600
lzmaxv-v-2200 - this pull without ttable sync, with -v 2200
lzmaxv-ttab-v-2200 - this pull but including the ttable sync, with -v 2200
```
And results:
```
lzbase-p-1600 v lzmaxv-v-2200 (84/1000 games)
board size: 19   komi: 7.5
lzbase-p-1600     41 48.81%       22 52.38%     19 45.24%    278.69
lzmaxv-v-2200     43 51.19%       23 54.76%     20 47.62%    291.26
                                  45 53.57%     39 46.43%

lzbase-p-1600 v lzmaxv-ttab-v-2200 (84/1000 games)
board size: 19   komi: 7.5
lzbase-p-1600          47 55.95%       20 47.62%     27 64.29%    286.60
lzmaxv-ttab-v-2200     37 44.05%       15 35.71%     22 52.38%    298.05
                                       35 41.67%     49 58.33%

lzmaxv-ttab-v-2200 v lzmaxv-v-2200 (84/1000 games)
board size: 19   komi: 7.5
lzmaxv-ttab-v-2200     48 57.14%       26 61.90%     22 52.38%    299.14
lzmaxv-v-2200          36 42.86%       20 47.62%     16 38.10%    300.38
                                       46 54.76%     38 45.24%
```

billyswong:
@killerducky Is avg cpu strictly proportional to ms/move?

killerducky:
See below the relevant gomill source code line, looks like unix user time + system time. I have no idea if I can trust that but it seemed to be reasonable.
https://github.com/mattheww/gomill/blob/master/gomill/gtp_controller.py#L1132

gcp:
>I'll do some investigating later to see if I messed something up.

Probably it's simply statistical variance, your original results weren't statistically significant.

gcp:
Even if it turns out there's no benefit from this for self-play, I'm still interested in pulling it because it presents a way to play at a fixed level even with pondering enabled, which some people have been asking for.

src/Leela.cpp
```diff
@@ -62,6 +62,9 @@ static void parse_commandline(int argc, char *argv[], bool & gtp_mode) {
         ("playouts,p", po::value<int>(),
                        "Weaken engine by limiting the number of playouts. "
                        "Requires --noponder.")
+        ("visits,v", po::value<int>(),
+                       "Weaken engine by limiting the number of visits. "
+                       "Requires --noponder.")
```
gcp: This is not true :-)

src/Leela.cpp
```diff
@@ -196,6 +199,16 @@ static void parse_commandline(int argc, char *argv[], bool & gtp_mode) {
         }
     }
 
+    if (vm.count("visits")) {
+        cfg_max_visits = vm["visits"].as<int>();
+        if (!vm.count("noponder")) {
```
gcp: Again, not true and not necessary.

src/UCTSearch.cpp
```diff
@@ -117,7 +118,7 @@ void UCTSearch::dump_stats(KoState & state, UCTNode & parent) {
 
     int movecount = 0;
     for (const auto& node : parent.get_children()) {
-        if (++movecount > 2 && !node->get_visits()) break;
+        if (!node->get_visits()) break;
```
gcp: These are unrelated changes from another branch.

src/UCTSearch.cpp
```diff
@@ -266,6 +267,9 @@ std::string UCTSearch::get_pv(KoState & state, UCTNode& parent) {
     }
 
     auto& best_child = parent.get_best_root_child(state.get_to_move());
+    if (best_child.first_visit()) {
```
gcp: Same.

jkiliani:
> Probably it's simply statistical variance, your original results weren't statistically significant.

@gcp, @marcocalignano This seems to be a key problem with many proposed pull requests or parameter tuning. Is there any chance to use autogtp in a way similar to Fishtest, i.e. to run distributed SPRT tests not just of the same binary but different neural nets, but also same net but different binaries (or binary run with different parameters)? As it is, it looks very difficult for contributing developers to always prove the statistical significance of proposed code changes, especially if a lot of tuning parameters are involved.

gcp:
A validate with the possibility to specify different binaries would be useful, yes. (And the upper SPRT bound in it should probably be configurable, i.e. code changes don't need [0, 35] but more like [0, 5] or [-1, 3] as per fishtest)

gcp:
Running *distributed* tests for code changes isn't reasonably possible though. Fishtest has a whole bunch of infrastructure to do code compiles, and user accounts so trusted users can authorize tests (so people don't submit trojans...). Out of scope for us.

jkiliani:
I can see the problem with automatically compiling binaries for testing, but what if proposed code changes could be implemented in such a way that they could be optionally called by a runtime parameter? In that way, such a commit could possibly be merged to next branch, tested by /next clients, and reverted if it turned it to be useless.

gcp:
>In that way, such a commit could possibly be merged to next branch, tested by /next clients, and reverted if it turned it to be useless.

Strong no to this idea (merging bad changes and backing them out).

If a code change has a big positive impact, it should be possible to demonstrate that it works with a local test, as SPRT will terminate quickly.

Strength scaling and tuning are not things that need to be done regularly, so there's no pressing need to write a shitton of code to support it in a distributed manner. But if you want to do all that, be my guest.

jkiliani:
It was only an idea to get around the issues preventing us from having a Fishtest analogon. If you say it's not feasible I'll take your word for it.

marcocalignano:
I think the different binary part is doable, and we could use the validation QT project for that. I did already some changes to run the SPRT against Pachi and GnuGo adding some option to choose different binaries shouldn't be that difficult.

marcocalignano:
So if someone want to test the different binary validation program, can try it on my repository in the branch /bin_option. 
An example of the command line is this: 

`./validation -n 097deeb9 -n 0c09415 -b ./leelaz1 -o " -g -p 2200 --noponder -r 1 -w "  `

this will start the binary `./leelaz1` with the net 097deeb9 and the option ` -g -p 2200 --noponder -r 1 -w ` and will play against `./leelaz` with net 0c09415 and default options (`-g  -p 1600 --noponder -t 1 -q -d -r 0 -w`.
After the engines start and before the games starts each game will prompt the command line for each game for verification.

marcocalignano:
I added also an option to set the SPRT boundaries from command line.

roy7:
What's the status on the ttable_sync_children front? Since both https://github.com/gcp/leela-zero/pull/315 and https://github.com/gcp/leela-zero/pull/581 have been closed.

I just want to make sure somewhere, somehow, the issue with data in the TTable being lost if it wasn't visited at least once in a later step goes way. :) I can't stand the fact in the example in the other thread, LZ found the answer it needed, but then 'forgot' about it and dump_training pushed that move towards 0 when it should be raising the prior instead.

marcocalignano:
@gdp 
-Does autogtp need to support this option, i.e. will the server send this option in the json at work request?
-Should the branch /bin_option in my repository also be merged in next?

gcp:
@marcocalignano It is not clear whether it is beneficial yet. But it would not hurt, it's an alternative to playouts.

What does bin_option do?

marcocalignano:
You can specify different leelaz binary, different option and different SPRT threshold  for the validation program.

roy7:
I'm assuming my comments/results on https://github.com/gcp/leela-zero/issues/651 are because ttable_sync_all_children wasn't applied in this specific pull request, but may be at some time in the future?

Issue #592, opened 2018-01-07
PRNG is not properly seeded with probability 1/2^64.


thynson:
As a [reference implementation of xoroshiro+128](http://xoroshiro.di.unimi.it/xoroshiro128plus.c) said:
```
The state must be seeded so that it is not everywhere zero. If you have
   a 64-bit seed, we suggest to seed a splitmix64 generator and use its
   output to fill s.
```
While in `Random.cpp`:
```cpp
void Random::seedrandom(uint64 seed) {
    // Magic values from Pierre L'Ecuyer,
    // "Tables of Linear Congruental Generators of different sizes and
    // good lattice structure"
    m_s[0] = (1181783497276652981ULL * seed);
    m_s[1] = (1181783497276652981ULL * m_s[0]);
}
```
Althrough magic value here is multiply to the seed, it doesn't help if the seed is `0`, and then the random engine will produce nothing but 0. Note that the constructor do check if it's seeded by 0, and will use thread id instead, but `Leela.cpp` call `Random::seedrandom` directly. 

Through from the way how the the random engine is initialized, it's very unlikely be seeded with 0. Unless the program is compiled on a system that doesn't have an working random device with a buggy `std::high_resolution_clock`.  

Don't know if this bug links to any junk game issue, but we should get rid of such risk anyway.

gcp:
> except on any buggy system that doesn't have an working random device as well as a buggy std::high_resolution_clock

They would still have to return sequences that when mixed together work out to 0. I don't see how this can happen with a probability of more than 1/2^64, even for broken implementations(!). This is the same probability as the server seeding with "0".

We can "fix" this but the issue is purely theoretical.

thynson:
> They would still have to return sequences that when mixed together work out to 0.

I checked Leela.cpp again and found `Random::get_Rng` is returning an thread local instance seed by thread id, by pass 0 to the constructor, while `init_global_objects` call `Random::seedrandom` directly with `cfg_rng_seed` as its argument, so P(cfg_rng_seed = 0) is not likely as small as 1/2^64.

And I'm raising another question here: Does thread pool need to seed its random engine by itself?

gcp:
>while P(cfg_rng_seed = 0) is likely not as small as 1/2^64

Why?

As said above, even for broken implementations, they would have to be broken in such a way that the mixing of them ends up at 0. I'd say the probability is likely *smaller* than 1/2^64 there.

gcp:
[...]

PR #594, opened 2018-01-07
Fix Random::seedrandom to initilize xoroshiro128+ prng properly

thynson:
Fix for #592. Throgh it's very unlikely to happen, in a probability of around 1/2^64.

### DIFF
--- a/src/Random.cpp
+++ b/src/Random.cpp
@@ -26,6 +26,13 @@
 #include "GTP.h"
 #include "Utils.h"
 
+static std::uint64_t splitmix64(std::uint64_t z) {
+    z += 0x9e3779b97f4a7c15;
+    z = (z ^ (z >> 30)) * 0xbf58476d1ce4e5b9;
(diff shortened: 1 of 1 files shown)


src/Random.cpp
```diff
@@ -69,11 +69,19 @@ std::uint32_t Random::randuint32() {
...
-    // "Tables of Linear Congruental Generators of different sizes and
-    // good lattice structure"
-    m_s[0] = (1181783497276652981ULL * seed);
-    m_s[1] = (1181783497276652981ULL * m_s[0]);
+
+    // Initialize state of xoroshiro128+ by transform the seed
+    // with splitmix64 algorithm.
+    // As suggested by http://xoroshiro.di.unimi.it/xoroshiro128plus.c
+
+    seed += 0x9e3779b97f4a7c15;
```
gcp: Pull out the splitmix64 call rather than duplicating the code and constants twice.

Or just add a m_s[1] |= 1;

thynson:
OK, amended.

PR #595, opened 2018-01-07
move rotate_nn_idx in a table

marcocalignano:
This is a small optimization but could be useful for CPU only leelaz.


src/Network.cpp
```diff
@@ -529,7 +740,7 @@ Network::Netresult Network::get_scored_moves_internal(
     for (int c = 0; c < INPUT_CHANNELS; ++c) {
         for (int h = 0; h < height; ++h) {
             for (int w = 0; w < width; ++w) {
-                auto rot_idx = rotate_nn_idx(h * 19 + w, rotation);
+                auto rot_idx = rotate_nn_idx_table[rotation][h * 19 + w];
```
sethtroisi: According to Intel vtune the program spends a surprising amount of time here (3% overall, 10% of CPU time note these have big error bars). I tried several refactors of rotate_nn_idx (using std::div, passing in h, w) and none of them speed it up.

If I find time I'll open vtune again and test this.

marcocalignano: Please let me know the result.

gcp:
I won't pull this as is anyway. If this is actually beneficial then construct the table at startup, don't include it in the source.

Issue #602, opened 2018-01-08
build tips for macos using macports instead of brew


mjhsieh:
It would be nice if the document can mention the example for building leelaz on macos using macports boost. Thank you!

PR #603, opened 2018-01-08
Example for building with macports

mjhsieh:
See https://github.com/gcp/leela-zero/issues/602 feedbacks are welcome.


gcp:
Not sure if we want to have instructions for every possible Linux distro or mac packager in the README.

gcp:
Should be rebase against the next branch, not master.

mjhsieh:
Sorry, I will do that next time.

Issue #582, opened 2018-01-06
Feature request: Allowing playout cap without --noponder


roy7:
I discussed this with GCP ages ago but it didn't make sense to sort out making this work for Leela since it's such a niche feature, that would only maybe benefit bot operators (or heck, just me :) ). 

Currently if you use --playouts, you also need to use --noponder. I'd like to request one of the volunteers out there working on LZ sort out a way to remove that requirement.

The idea is that after making a move, some playouts (or visits if a visit cap is added) will be pruned because of the paths explored but not taken. Thus whatever cap we just hit before moving, we're now under that cap and have some more work we could do while the opponent is thinking. Once we're back up to the cap, pondering stops and we just wait for the opponent's move.

Where this has come up before for me is running bots on OGS with various time settings and configurations. For ranked games for instance, with RoyalLeela, I run no playout cap and allow pondering. For unranked games and correspondence games I turn off pondering and have playout caps in place. The reason is because, unlike KGS where I believe bots only play 1 game at a time, OGS bots by default can play an unlimited number of games at once. Each game is fighting for the same GPU, so the more games playing at once, the weaker the bot will become.

If I want the bot to benefit from opponent time, I can't do a playout cap, and so the bot may think for long periods of time if the game setting is something like byoyomi 8h + 5m x 5. That's a long time for the GPU to be tied up on one game, while many other games might also be running. So a playout cap is needed... except with a playout cap and --noponder I can't start thinking until the opponent moves, which takes more time for the bot to make its move. The strength end result is the same (X playouts/visits are done before a move is made), but the user experience for the human opponent is better if the bot moves faster (sometimes all of the pondering work will be directly in the path the human went, so much of the playout/visit cap is still valid and new work done after opponent's move is very brief).

[...]

PR #609, opened 2018-01-08
Feature #582

Cabu:
Implement Feature request 582: Allowing playout cap without --noponder.
Thinking + Pondering are now limited to max playout.


jkiliani:
One general rule is that @gcp merges all pull requests against /next, not against /master. Also, I recommend always giving a pull request a meaningful name including the essence of what it does. You can still reference the issue it originated from in the description.

PR #610, opened 2018-01-08
Feature #582

Cabu:
Implement Feature request 582: Allowing playout cap without --noponder.
Thinking + Pondering are now limited to max playout.


gcp:
I will take visit counts instead, which do what the requester achieved in a manner that is consistent with pondering.

Issue #622, opened 2018-01-09
LeelaZ resign too 'easily' while playing handicap games


Cabu:
LeelaZ resign too 'easily' while playing handicap games (I tested for >=5 stones) leaving no time to let the weaker player (me ~10kyu) to make my first big mistake or enough little mistake to nullify the stones advantage.
Could it be feasible to offset the winrate with a function of the handicap and the number of played move?

killerducky:
Try "--resign 1" to lower the threshold from the default 10% to 1%. There are probably some code improvements we could do but that should get you going right away.

gcp:
Resigning is already capped to not happen early in the game. As said above, when playing with handicap you probably want a much lower threshold than the default.

gcp:
Maybe if no resignation threshold is specified, the engine could divide 10% by the number of handicap stones or something.

Cabu:
I like that idea. I though about something more complex where the threshold get back up as the game move to the end. But simply dividing by the number of stone is easy and should work.

sethtroisi:
Code would go here if you want to write a pull request https://github.com/gcp/leela-zero/blob/fbade9fcc8cb4746f9111213c1df1c7840136501/src/UCTSearch.cpp#L253

something like 

```
auto default_threshold = (float)cfg_resignpct / 100.0f;
auto handicap_threshold = resign_threshold / (1 + m_rootstate.get_handicap());
// Blend default_threshold with reduced handicap_threshold based on number of moves into game
auto blend_ratio = max<float>(1.0, ((float) m_rootstate.get_movenum()) / (0.6 * m_rootstate.board.get_boardsize() * m_rootstate.board.get_boardsize()));
auto resign_threshold = handicap_threshold + (default_threshold - handicap_threshold) * blend_ratio;
```

gcp:
I don't want to override the behavior of the user specifies the resign percentage, as that would be even more confusing. Alternatively, we could change the description in the options to make it clear it is a kind of "base" percentage.

sethtroisi:
What do you suggest for checking if the user has specified a resign percentage?

I can test if cfg_resignpct == 10 but this is duplicating the default value (10).
I can add another global default_resignpct.
or something else ...

gcp:
Default it to -1 and check for that? Maybe a bit ugly, but I don't immediately know a better way.

>Also should resign rate be reduced when Leelaz is playing for the side who received the handicap?

No. (And I guess not the opposite either!)

[...]

PR #626, opened 2018-01-10
new resign logic for handicap threshold

sethtroisi:
New resign logic discussed in https://github.com/gcp/leela-zero/issues/622

Leelaz doesn't resign before move 90 (92?).
For games with handicap reduce handicap rate even more for moves 90 to ~215.

[Old graph of resign pct vs move number](
https://docs.google.com/spreadsheets/d/1AIGYObGCGWadcdWwN1ceMuPd_qu1hyQd2tD8QzSsGV8/edit#gid=1843999952)

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -77,6 +77,7 @@ void GTP::setup_default_parameters() {
 #endif
     cfg_puct = 0.85f;
     cfg_softmax_temp = 1.0f;
+    // Keep in sync with UCTSearch::should_resign
     cfg_resignpct = 10;
     cfg_noise = false;
     cfg_random_cnt = 0;
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -77,8 +77,8 @@ void GTP::setup_default_parameters() {
 #endif
     cfg_puct = 0.85f;
     cfg_softmax_temp = 1.0f;
-    // Keep in sync with UCTSearch::should_resign
-    cfg_resignpct = 10;
+    // see UCTSearch::should_resign
+    cfg_resignpct = -1;
     cfg_noise = false;
     cfg_random_cnt = 0;
     cfg_dumbpass = false;
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -250,12 +250,13 @@ int UCTSearch::get_best_move(passflag_t passflag) {
                                  * m_rootstate.board.get_boardsize();
             auto move_threshold = board_squares / 4;
             auto resign_threshold = 0.01 * cfg_resignpct;
+            // TODO add something to disable if -r is user set.
             if (m_rootstate.get_handicap() > 0) {
                 auto handicap_resign_threshold =
-                    resign_threshold / std::sqrt(1 + m_rootstate.get_handicap());
-                // blend the thresholds for the first ~180 moves.
+                    resign_threshold / (1 + m_rootstate.get_handicap());
+                // blend the thresholds for the first ~215 moves.
                 auto blend_ratio = std::max(1.0, m_rootstate.get_movenum() /
-                                                 (0.5 * board_squares));
+                                                 (0.6 * board_squares));
                 resign_threshold = handicap_resign_threshold +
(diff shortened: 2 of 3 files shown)


sethtroisi:
Refactored resign code to a method so it can be spaced out a little more.

This logic is only used if cfg_resignpct is not set via the commandline by duplicating the constant (10)
I added a comment in GTP.cpp to keep the two in sync.

sethtroisi:
I tested a couple of games and see this working (as well as discovering another issue)

"0.073 (0.025 * 0.646 + 0.100) vs 0.050, 140
Eval (0.050) looks bad. Resigning."

evanroberts85:
I do not like this solution, Leela should still be able to resign mid game even in a handicap game. Instead would it not be better to lower the resign threshold for handicap games rather than increase the minimum number of moves before resigning.

gcp:
@evanroberts85 Isn't that exactly what this does?

src/UCTSearch.cpp
```diff
@@ -137,6 +137,54 @@ void UCTSearch::dump_stats(KoState & state, UCTNode & parent) {
...
+        // low visits
+        return false;
+    }
+
+    const size_t board_squares = m_rootstate.board.get_boardsize()
+                               * m_rootstate.board.get_boardsize();
+    const auto move_threshold = board_squares / 4;
+    const auto movenum = m_rootstate.get_movenum();
+    if (movenum <= move_threshold) {
+        // to early in game to resign
```
gcp: too early

src/GTP.cpp
```diff
@@ -77,6 +77,7 @@ void GTP::setup_default_parameters() {
 #endif
     cfg_puct = 0.85f;
     cfg_softmax_temp = 1.0f;
+    // Keep in sync with UCTSearch::should_resign
```
gcp: I don't like this, easy to mess up. Making cfg_resignpct = -1 be something like "use default logic", and then putting 10 as a default in that logic, is also not perfect but seems preferable.

src/UCTSearch.cpp
```diff
@@ -137,6 +137,54 @@ void UCTSearch::dump_stats(KoState & state, UCTNode & parent) {
...
+
+    const auto color = m_rootstate.board.get_to_move();
+    const auto best_eval = m_root.get_first_child()->get_eval(color);
+    auto resign_threshold = 0.01 * cfg_resignpct;
+    if (best_eval > resign_threshold) {
+        // eval > cfg_resign
+        return false;
+    }
+
+    const auto is_default_cfg_resign = (cfg_resignpct == 10);
```
gcp: With the above this would then just be a check for < 0, but be moved above the previous block, which would then use 10 as the default.

src/UCTSearch.cpp
```diff
@@ -239,21 +287,11 @@ int UCTSearch::get_best_move(passflag_t passflag) {
...
-                                * m_rootstate.board.get_boardsize()) / 4;
-            // bad score and visited enough
-            if (bestscore < ((float)cfg_resignpct / 100.0f)
-                && visits > 500
-                && m_rootstate.m_movenum > movetresh) {
-                myprintf("Score looks bad. Resigning.\n");
-                bestmove = FastBoard::RESIGN;
-            }
+        if (should_resign(passflag)) {
+            myprintf("Eval (%.3f) looks bad. Resigning.\n", bestscore);
```
gcp: Scores are winrates, so convert to percentage and add a %%.

src/GTP.cpp
gcp: Too bad std::optional is C++17. But boost has a version too, and it's header only, so that would be usable.

sethtroisi: Logic now lives in UCTSearch

src/UCTSearch.cpp
sethtroisi: I don't agree that it's a winrate (in the same way a p value != probability in statistics) but will fix

sethtroisi:
Updated,

"dump_debug" from training prints cfg_resignpct which will now be -1 if not specified, is this okay?

PR #629, opened 2018-01-10
autogtp: adding time x game statistic to tell the server how fast we are

marcocalignano:


marcocalignano:
This PR is for people who want to test, but most of all if for @roy7 to tell us if it a feasible way to solve the #545 Issue.

gcp:
So I shouldn't pull this into next yet?

marcocalignano:
I would like to hear from @roy7 first if we need it or if he want to solve this on the server side only.

roy7:
I'm pretty worn out today, have plans tonight, and I'm not sure on the best way to handle it. So I'm tempted to just get the server into github this weekend and see if anyone volunteers to take the client allocation logic across the finish line with a pull request. ;)

marcocalignano:
@roy7 no problem I can wait!

marcocalignano:
Closing this because it need rebasing.

roy7:
Turns out I'm sick with something, sorry for delays. I'll look more tomorrow if I'm up and about.

marcocalignano:
@roy7 I hope you are better. Do you want me to let a client run with this patch so you can try experiment on this or we won't go down this path?

roy7:
Sorry, been behind on looking at this. Things seem to be working so I just let myself get distracted elsewhere. If you want to feed the param in I can log it in my log files to check it out. It doesn't hurt to send the data into the server even if we didn't end up using it in this way. If nothing we could collect some stats on client speeds.

marcocalignano:
@roy7 So now you should have two threads one fast and one slow, sending you the statistics. I was without internet for almost the whole day do it will send (or already did) all the games from today.
So do not panic ;)

marcocalignano:
>  It doesn't hurt to send the data into the server even if we didn't end up using it in this way. If nothing we could collect some stats on client speeds.

@roy7 so do you want me to do the PR anyway?

PR #630, opened 2018-01-11
Added validation of fixed/free handicap input

sethtroisi:
"place_free_handicap" without specifying a number of stones results in leelaz adding an arbitrary number of handicap stones.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -529,8 +529,8 @@ bool GTP::execute(GameState & game, std::string xinput) {
         cmdstream >> tmp;   // eat fixed_handicap
         cmdstream >> stones;
 
-        if (game.set_fixed_handicap(stones)) {
-            std::string stonestring = game.board.get_stone_list();
+        if (!cmdstream.fail() && game.set_fixed_handicap(stones)) {
+            auto stonestring = game.board.get_stone_list();
             gtp_printf(id, "%s", stonestring.c_str());
         } else {
             gtp_fail_printf(id, "Not a valid number of handicap stones");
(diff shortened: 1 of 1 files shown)


src/GTP.cpp
```diff
@@ -544,10 +544,13 @@ bool GTP::execute(GameState & game, std::string xinput) {
         cmdstream >> tmp;   // eat place_free_handicap
         cmdstream >> stones;
 
-        game.place_free_handicap(stones);
-
-        std::string stonestring = game.board.get_stone_list();
-        gtp_printf(id, "%s", stonestring.c_str());
+        if (!cmdstream.fail()) {
+            game.place_free_handicap(stones);
+            std::string stonestring = game.board.get_stone_list();
```
gcp: nit: prefer auto

src/GTP.cpp
```diff
@@ -544,10 +544,13 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
+            gtp_printf(id, "%s", stonestring.c_str());
+        } else {
+            gtp_fail_printf(id, "Not a valid number of handicap stones");
```
gcp: I think this is the wrong error message in this case. "Not a valid list of vertices" or something similar?

sethtroisi: If you look five lines above you'll see it reads an integer "stones" from cmdstream, so I think this is the right error message.

Also see https://www.gnu.org/software/gnugo/gnugo_19.html
"Fails:     invalid number of stones"

sethtroisi:
updated to using auto

PR #632, opened 2018-01-12
Implement tree reuse.

killerducky:
Overview of implementation: UCTSearch holds a static copy of m_root and m_rootstate to save the tree. When UCTSearch::think is called, it checks m_root and it's children to see if the tree can be reused. Otherwise create a new tree. When UCTSeach::think picks a move, update m_root to that child so the tree is immediately pruned. This should cover the most common use cases of autoplay and match play.

I didn't test ponder yet, I think it's probably broken right now. Also I'm still learning the std::move stuff, hope I didn't butcher that too badly. I just want some early feedback, let me know if this is a good direction.

Related closed pulls: #315 and #581.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -67,8 +67,6 @@ FILE* cfg_logfile_handle;
 bool cfg_quiet;
 std::string cfg_options_str;
 
-std::unique_ptr<UCTSearch> search = nullptr;
-
 void GTP::setup_default_parameters() {
     cfg_allow_pondering = true;
     cfg_num_threads = std::max(1, std::min(SMP::get_num_cpus(), MAX_CPUS));
@@ -163,11 +161,7 @@ std::string GTP::get_life_list(const GameState & game, bool live) {
 
 bool GTP::execute(GameState & game, std::string xinput) {
     std::string input;
-    // This is a way to make sure e.g. cfg_max_playouts is
-    // initialized before we construct the search object.
-    if (search == nullptr) {
-        search = std::make_unique<UCTSearch>();
-    }
+    static UCTSearch search;
 
     bool transform_lowercase = true;
 
@@ -281,6 +275,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
     } else if (command.find("clear_board") == 0) {
         Training::clear_training();
         game.reset_game();
+        search.set_gamestate(game);
         gtp_printf(id, "");
         return true;
     } else if (command.find("komi") == 0) {
@@ -348,10 +343,8 @@ bool GTP::execute(GameState & game, std::string xinput) {
             }
             // start thinking
             {
-                search->set_gamestate(game);
-
                 game.set_to_move(who);
-                int move = search->think(who);
+                int move = search.think(who, game);
                 game.play_move(move);
 
                 std::string vertex = game.move_to_text(move);
@@ -360,8 +353,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
             if (cfg_allow_pondering) {
                 // now start pondering
                 if (!game.has_resigned()) {
-                    search->set_gamestate(game);
-                    search->ponder();
+                    search.ponder(game);
                 }
             }
         } else {
@@ -387,10 +379,8 @@ bool GTP::execute(GameState & game, std::string xinput) {
             }
             game.set_passes(0);
             {
-                search->set_gamestate(game);
-
                 game.set_to_move(who);
-                int move = search->think(who, UCTSearch::NOPASS);
+                int move = search.think(who, game, UCTSearch::NOPASS);
                 game.play_move(move);
 
                 std::string vertex = game.move_to_text(move);
@@ -399,8 +389,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
             if (cfg_allow_pondering) {
                 // now start pondering
                 if (!game.has_resigned()) {
-                    search->set_gamestate(game);
-                    search->ponder();
+                    search.ponder(game);
                 }
             }
         } else {
@@ -483,8 +472,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
                 // KGS sends this after our move
                 // now start pondering
                 if (!game.has_resigned()) {
-                    search->set_gamestate(game);
-                    search->ponder();
+                    search.ponder(game);
                 }
             }
         } else {
@@ -493,9 +481,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
         return true;
     } else if (command.find("auto") == 0) {
         do {
-            search->set_gamestate(game);
-
-            int move = search->think(game.get_to_move(), UCTSearch::NORMAL);
+            int move = search.think(game.get_to_move(), game, UCTSearch::NORMAL);
             game.play_move(move);
             game.display_state();
 
@@ -503,9 +489,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
 
         return true;
     } else if (command.find("go") == 0) {
-        search->set_gamestate(game);
-
-        int move = search->think(game.get_to_move());
+        int move = search.think(game.get_to_move(), game);
         game.play_move(move);
 
         std::string vertex = game.move_to_text(move);
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -161,7 +161,7 @@ std::string GTP::get_life_list(const GameState & game, bool live) {
 
 bool GTP::execute(GameState & game, std::string xinput) {
     std::string input;
-    static UCTSearch search;
+    static auto search = std::make_unique<UCTSearch>();
 
     bool transform_lowercase = true;
 
@@ -275,7 +275,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
     } else if (command.find("clear_board") == 0) {
         Training::clear_training();
         game.reset_game();
-        search.set_gamestate(game);
+        std::make_unique<UCTSearch>().swap(search);
         gtp_printf(id, "");
         return true;
     } else if (command.find("komi") == 0) {
@@ -344,7 +344,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
             // start thinking
             {
                 game.set_to_move(who);
-                int move = search.think(who, game);
+                int move = search->think(who, game);
                 game.play_move(move);
 
                 std::string vertex = game.move_to_text(move);
@@ -353,7 +353,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
             if (cfg_allow_pondering) {
                 // now start pondering
                 if (!game.has_resigned()) {
-                    search.ponder(game);
+                    search->ponder(game);
                 }
             }
         } else {
@@ -380,7 +380,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
             game.set_passes(0);
             {
                 game.set_to_move(who);
-                int move = search.think(who, game, UCTSearch::NOPASS);
+                int move = search->think(who, game, UCTSearch::NOPASS);
                 game.play_move(move);
 
                 std::string vertex = game.move_to_text(move);
@@ -389,7 +389,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
             if (cfg_allow_pondering) {
                 // now start pondering
                 if (!game.has_resigned()) {
-                    search.ponder(game);
+                    search->ponder(game);
                 }
             }
         } else {
@@ -472,7 +472,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
                 // KGS sends this after our move
                 // now start pondering
                 if (!game.has_resigned()) {
-                    search.ponder(game);
+                    search->ponder(game);
                 }
             }
         } else {
@@ -481,7 +481,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
         return true;
     } else if (command.find("auto") == 0) {
         do {
-            int move = search.think(game.get_to_move(), game, UCTSearch::NORMAL);
+            int move = search->think(game.get_to_move(), game, UCTSearch::NORMAL);
             game.play_move(move);
             game.display_state();
 
@@ -489,7 +489,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
 
         return true;
     } else if (command.find("go") == 0) {
-        int move = search.think(game.get_to_move(), game);
+        int move = search->think(game.get_to_move(), game);
         game.play_move(move);
 
         std::string vertex = game.move_to_text(move);
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -67,6 +67,8 @@ FILE* cfg_logfile_handle;
 bool cfg_quiet;
 std::string cfg_options_str;
 
+std::unique_ptr<UCTSearch> search = nullptr;
+
 void GTP::setup_default_parameters() {
     cfg_allow_pondering = true;
     cfg_num_threads = std::max(1, std::min(SMP::get_num_cpus(), MAX_CPUS));
@@ -161,6 +163,11 @@ std::string GTP::get_life_list(const GameState & game, bool live) {
 
 bool GTP::execute(GameState & game, std::string xinput) {
     std::string input;
+    // This is a way to make sure e.g. cfg_max_playouts is
+    // initialized before we construct the search object.
+    if (search == nullptr) {
+        search = std::make_unique<UCTSearch>();
+    }
 
     bool transform_lowercase = true;
 
@@ -341,7 +348,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
             }
             // start thinking
             {
-                auto search = std::make_unique<UCTSearch>(game);
+                search->set_gamestate(game);
 
                 game.set_to_move(who);
                 int move = search->think(who);
@@ -353,7 +360,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
             if (cfg_allow_pondering) {
                 // now start pondering
                 if (!game.has_resigned()) {
-                    auto search = std::make_unique<UCTSearch>(game);
+                    search->set_gamestate(game);
                     search->ponder();
                 }
             }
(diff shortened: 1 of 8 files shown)


src/UCTSearch.h
```diff
@@ -89,8 +89,8 @@ class UCTSearch {
     void dump_analysis(int playouts);
     int get_best_move(passflag_t passflag);
 
-    GameState & m_rootstate;
-    UCTNode m_root{FastBoard::PASS, 0.0f, 0.5f};
+    static std::unique_ptr<GameState> m_rootstate;
```
gcp: GameState is copied in by UCTSearch, based on the state of the UCTSearch's controller. So this should be a simple member. From looking through the code I see no clear reason why you need this to change (neither to a pointer nor to a global).

src/UCTSearch.cpp
```diff
@@ -39,8 +39,23 @@
 
 using namespace Utils;
 
-UCTSearch::UCTSearch(GameState & g)
-    : m_rootstate(g) {
+std::unique_ptr<UCTNode> UCTSearch::m_root = nullptr;
```
gcp: This is a bit ugly, and would prevent using multiple UCTSearches. If we want to reuse information without globals, the owner of UCTSearch can also do that, and that is effectively GTP.cpp's execute().

I would suggest to try the following instead: make think() return the UCTNode for the best move. execute() can call get_move on that to get the move as before. It can also keep a local std::unique_ptr var to point to the result of the last search, and then pass that into the UCTSearch constructor (may be null).

You can then do a check if g.get_last_move() == last_search_node_my_new_root.get_move() and if so (maybe) check hashes to make sure. Checking hashes won't be needed if execute() properly clears it's saved UCTNode on new games etc.

src/UCTNode.cpp
```diff
@@ -394,10 +394,30 @@ UCTNode* UCTNode::get_first_child() const {
     return m_children.front().get();
 }
 
+UCTNode::node_ptr_t UCTNode::move_first_child() {
```
gcp: It seems this function is not used.

src/UCTSearch.cpp
```diff
@@ -239,24 +254,29 @@ int UCTSearch::get_best_move(passflag_t passflag) {
...
-                && m_rootstate.m_movenum > movetresh) {
+                && m_rootstate->m_movenum > movetresh) {
                 myprintf("Score looks bad. Resigning.\n");
                 bestmove = FastBoard::RESIGN;
             }
         }
     }
 
+    // TODO: Instead of searching for the child we could keep track of it.
+    // This is just an easy way to be safe.
```
gcp: At this point you know you want the UCTNode out of m_children that has get_move() == bestmove, so all the object copying and hash lookups aren't needed?

src/UCTSearch.h
```diff
@@ -89,8 +89,8 @@ class UCTSearch {
...
+    static std::unique_ptr<UCTNode> m_root;
```
gcp: Remove static. You can't avoid making this a pointer but there's no need for it to be a global.

gcp:
I think this is the right direction. It can be simplified quite a bit, but in order to have saner ownership I think GTP::execute needs to know about saving the last search results, which will complicate it slightly again.

killerducky:
About ownership, how about a similar idea but GTP holds a UCTSearch object instead of constructing a new one every time. UCTSearch m_root will be a normal member not a global static, and UCTSearch remains in charge of m_root which is like a cache of it's internal search state. GTP doesn't need to know anything about UCTNode. If GTP ever does multiple UCTSearches he can still do it by holding multiple of those objects.

killerducky:
I found a problem, m_nodes accounting is broken. A couple options:
A) Traverse the entire tree and count m_nodes.
B) Use some sort of object pool for nodes that provides a current count of allocated nodes. Maybe an object pool would provide a performance benefit also? Some quick googles and it doesn't look like boost object pool is suitable.

Any ideas?

src/UCTNode.cpp
killerducky: removed.

src/UCTNode.cpp
```diff
@@ -394,10 +394,30 @@ UCTNode* UCTNode::get_first_child() const {
...
+    }
+    auto child = std::move(m_children.front());
+    return child;
+}
+
 const std::vector<UCTNode::node_ptr_t>& UCTNode::get_children() const {
     return m_children;
 }
 
+UCTNode::node_ptr_t UCTNode::find_and_take_ownership(GameState& g_new, GameState& g_curr) {
```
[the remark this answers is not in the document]
killerducky: fixed.

src/UCTNode.cpp
```diff
@@ -394,10 +394,30 @@ UCTNode* UCTNode::get_first_child() const {
...
+    for (auto& child : m_children) {
+        auto move = child->get_move();
+        g_curr.play_move(move);
```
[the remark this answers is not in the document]
killerducky: Changed to check child->get_move first, but still do the full check next to make sure it works in all cases.

src/UCTSearch.cpp
killerducky: See main thread comments for discussion.

src/UCTSearch.cpp
killerducky: Created a new version of find_and_take_ownership in UCTNode to handle this one. Let me know if you're ok with overloading the same function name with different arguments.

src/UCTSearch.h
killerducky: It's needed for the conservative checks I do in find_and_take_ownership were I pass GTP's gamestate (g_new) and the Search's saved games state (g_curr).

src/UCTSearch.h
killerducky: Done.

marcocalignano:
@killerducky You could have a `m_nodeCount` in the `UTCNode` that hold the total. you need only to calculate that number in the `create_children` function.

killerducky:
@gcp first I suggest we work on #645 first because some details of this pull depend on exactly how the hashes work.

I've come to realize how leaving m_rootstate as a reference works. I still prefer GTP holding a Search object than having it also manage a UCTNode. But maybe change it so GTP informs Search every time the GameState changes and Search can update it's own m_root accordingly. The current solution is nice too though since GTP doesn't have to manage as many details about the Search object. I don't really know which is best, they both have good and bad. What's your opinion?

gcp:
>I still prefer GTP holding a Search object than having it also manage a UCTNode.

Yeah that's fine. You can null it out in key places (like on reset_board) where it seems sensible to free the tree.

>But maybe change it so GTP informs Search every time the GameState changes and Search can update it's own m_root accordingly.

Seems more error prone than detecting it automagically.

src/GTP.cpp
```diff
@@ -67,6 +67,8 @@ FILE* cfg_logfile_handle;
 bool cfg_quiet;
 std::string cfg_options_str;
 
+std::unique_ptr<UCTSearch> search = nullptr;
```
gcp: Use uniform initialization: std::unique_ptr<UCTSearch> search{nullptr};

This can also be a static variable inside GTP::execute, which reflects ownership better, and which means it will only get its constructor called when cfg_* is already set up.

src/GTP.cpp
```diff
@@ -161,6 +163,11 @@ std::string GTP::get_life_list(const GameState & game, bool live) {
 
 bool GTP::execute(GameState & game, std::string xinput) {
     std::string input;
+    // This is a way to make sure e.g. cfg_max_playouts is
+    // initialized before we construct the search object.
```
gcp: I think that logic can go to the top of think(). (Or if you follow the previous suggestion, the problem disappears)

src/GTP.cpp
```diff
@@ -353,7 +360,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
             if (cfg_allow_pondering) {
                 // now start pondering
                 if (!game.has_resigned()) {
-                    auto search = std::make_unique<UCTSearch>(game);
+                    search->set_gamestate(game);
```
gcp: Does it not make more sense to have think() and ponder() take the gamestate?

src/UCTNode.cpp
```diff
@@ -398,6 +398,50 @@ const std::vector<UCTNode::node_ptr_t>& UCTNode::get_children() const {
     return m_children;
 }
 
+int UCTNode::count_nodes() const {
```
gcp: Make this return a size_t.

src/UCTNode.cpp
```diff
@@ -398,6 +398,50 @@ const std::vector<UCTNode::node_ptr_t>& UCTNode::get_children() const {
...
+    auto nodecount = 0;
```
gcp: = size_t{0};

src/UCTNode.cpp
```diff
@@ -398,6 +398,50 @@ const std::vector<UCTNode::node_ptr_t>& UCTNode::get_children() const {
...
+        }
+    }
+    return nodecount;
+}
+
+// Use this version if you know the child is directly under the parent.
+UCTNode::node_ptr_t UCTNode::find_and_take_ownership(const int move) {
+    if (m_has_children) {
+        // TODO: I couldn't get the find_if version to work.
+        // auto& child = std::find_if(begin(m_children), end(m_children),
```
gcp: find_if probably returns an iterator (behaving more like a ptr, not a ref)

src/UCTNode.cpp
```diff
@@ -398,6 +398,50 @@ const std::vector<UCTNode::node_ptr_t>& UCTNode::get_children() const {
...
+        //     [&move](auto &child) { return child->get_move == move; });
```
gcp: child->get_move()

src/UCTNode.cpp
```diff
@@ -398,6 +398,50 @@ const std::vector<UCTNode::node_ptr_t>& UCTNode::get_children() const {
...
+        for (auto& child : m_children) {
+            if (child->get_move() == move) {
+                return std::move(child);
+            }
+        }
+    }
+    return nullptr;
+}
+
+// Use this version if the child could be anywhere.
```
gcp: Not sure I understand in which case this one is useful? I think it's more that if you don't have the move? But get_last_move() will have it...

src/UCTSearch.cpp
```diff
@@ -39,12 +39,31 @@
...
 
+void UCTSearch::set_gamestate(GameState & g) {
+    // Definition of m_playouts is playouts from a certain GameState.
+    // So reset this count now.
+    m_playouts = 0;
+    if (m_root == nullptr) {
+        m_root = std::make_unique<UCTNode>(FastBoard::PASS, 0.0f, 0.5f);
+    }
+    if (m_rootstate == nullptr) {
+        m_rootstate = std::make_unique<GameState>(g);
```
gcp: Doesn't need to be a ptr.

src/GTP.cpp
killerducky: fixed

src/GTP.cpp
killerducky: fixed

src/GTP.cpp
killerducky: fixed.

src/UCTNode.cpp
killerducky: fixed

src/UCTNode.cpp
killerducky: fixed

src/UCTNode.cpp
killerducky: Gave up on this, just left it as a loop.

src/UCTNode.cpp
killerducky: GTP calls this one and I check the hashes to guarantee the entire position is match not just the last move. This covers cases where GTP gets multiple play commands followed by a genmove, without requiring GTP to do extra bookkeeping.

The other version should only be called by UCTSearch::think. In this case we can guarantee the child is a directly under the parent. We also don't have two separate GameStates to do a hash compare anyways.

src/Training.cpp
```diff
@@ -105,10 +105,6 @@ void Training::record(GameState& state, UCTNode& root) {
     step.planes = Network::NNPlanes{};
     Network::gather_features(&state, step.planes);
 
-    auto result =
-        Network::get_scored_moves(&state, Network::Ensemble::DIRECT, 0);
```
gcp: Not sure this belongs in this patch?

src/GTP.cpp
```diff
@@ -161,6 +161,7 @@ std::string GTP::get_life_list(const GameState & game, bool live) {
 
 bool GTP::execute(GameState & game, std::string xinput) {
     std::string input;
+    static UCTSearch search;
```
gcp: You could make this a unique_ptr<UCTSearch>. The reason for doing so would be that it avoids putting a potentially very big object on the stack, and it reduces the changes you need to make in GTP::execute as well.

(This has become relevant since it now carries a full copy of GameState, rather than just a reference)

src/UCTSearch.cpp
```diff
@@ -39,10 +39,29 @@
...
 
-UCTSearch::UCTSearch(GameState & g)
-    : m_rootstate(g) {
+UCTSearch::UCTSearch() {
     set_playout_limit(cfg_max_playouts);
     set_visit_limit(cfg_max_visits);
+    m_root = std::make_unique<UCTNode>(FastBoard::PASS, 0.0f, 0.5f);
+}
+
+void UCTSearch::set_gamestate(GameState & g) {
```
gcp: The argument here can be const.

src/UCTSearch.cpp
```diff
@@ -39,10 +39,29 @@
...
+    // Definition of m_playouts is playouts from a certain GameState.
+    // So reset this count now.
+    m_playouts = 0;
+    if (m_rootstate.get_komi() != g.get_komi()) {
+        m_root = std::make_unique<UCTNode>(FastBoard::PASS, 0.0f, 0.5f);
+        m_rootstate = g;
```
gcp: Is there ever a case where you wouldn't want to do m_rootstate = g in this function? I think it can go above the ifs.

Maybe if get_hash() != get_hash()? Then I'd pull that case out to remove the code duplication.

src/GTP.cpp
killerducky: done.

src/Training.cpp
killerducky: Reverted. I'll work on that later.

src/UCTSearch.cpp
killerducky: I refactored the code and added a comment to be clear. It avoids doing the `m_rootstate=g` copy when there is a match at the root, which happens every move in self play games.

BTW I think komi should be included in m_hash, that would get rid of this special case and also the one in NNCache. Save that change for later though. For now it doesn't matter since we only support komi=7.5.

killerducky: After a couple iterations I have a much cleaner way of doing this.

killerducky:
Ok I think I should take a break from fiddling with the division of work between set_gamestate and find_and_take_ownership. I guess it's good enough now, let me know if you see a better way.

Maybe a better name for the function is get_new_root.

gcp:
Got this merged and cleaned up, running test match now.

gcp:
One effect that is obvious is that this makes 1600 playouts a fair bit stronger, at the cost of being a bit slower.

Before, we would spend some amount of playouts on re-expanding the tree. With NNCache this was fast, but the playouts were still spent. (TTable compensates a bit for it, but not entirely)

With this patch, we don't spend any playouts re-expanding, we immediately deepen the tree. But the new positions are, of course, for the most part not in the NNCache.

It might be good to measure the NNCache impact now, it's possible that it can be made a lot smaller.

jkiliani:
I guess up until now NNCache had a net effect very similar to using visit count instead of playout count, and it was especially noticeable that it could wrap up no-resign games a lot quicker with the cache. It would be a pity if it can't do this anymore... any chance of changing to using -visits along with the merge of this?

gcp:
>any chance of changing to using -visits along with the merge of this?

It needs testing.

gcp:
>NNCache had a net effect very similar to using visit count instead of playout count

NNCache had no effect at all. It's 100% a speedup with no impact on explored nodes.

jkiliani:
What is the current state of knowledge on https://github.com/gcp/leela-zero/issues/538? Will this implementation of tree reuse still reduce the effectiveness of Dirichlet noise?

gcp:
```
leelaz-c83e-ntr v leelaz-c83e-tru (75/1000 games)
board size: 19   komi: 7.5
leelaz-c83e-ntr     24 32.00%       15 39.47%     9  24.32%    187.51
leelaz-c83e-tru     51 68.00%       28 75.68%     23 60.53%    242.83
                                    43 57.33%     32 42.67%
```

Merged.

gcp:
>What is the current state of knowledge on #538? Will this implementation of tree reuse still reduce the effectiveness of Dirichlet noise?

Unclear. The "new" root will be noised after being reused, so it certainly won't make anything worse. Probably better behaved than the TTable in that regard.

roy7:
> One effect that is obvious is that this makes 1600 playouts a fair bit stronger, at the cost of being a bit slower.

I mentioned this to Alderi since he runs the 9x9 project and the minusgo bot on OGS (LZ 9x9 bot). He said the change made his generation speed about half what it used to be, so I imagine in 9x9 he had a lot of duplicate past history being hidden in the TTable before. He still makes enough games/day to be okay, hopefully his data quality/sterength will be higher now though.

killerducky:
I guess extra CPU time should be free as long as we are still GPU limited. The ringmaster avg cpu stat is only skewed because it was run with `-t 1`. I think it would equalize if ran with -t 2 or larger to saturate the GPU. I'll try testing this later. 

Similar for autogtp which runs LZ with `-t 1` but if you use -g 2 or larger the CPU penalty of this change should be completely hidden by GPU.

For 9x9 maybe it's CPU limited since the network is so much faster.

roy7:
@killerducky Is it possible tree re-use could be pulling an old TimeControl out of a branch and replacing a newer time control that has been set since? Per https://github.com/gcp/leela-zero/issues/690#issuecomment-359194241.

PR #645, opened 2018-01-14
Fixes to m_hash and m_ko_hash.

killerducky:
Fixes issue #576.

Definition of m_komove changed slightly to be only set to the vertex if that vertex is actually a ko. That part used to be lazy calculated in is_move_legal. I think that extra code could be removed now from is_move_legal but I need to check weird cases like same color gets to genmove commands. Maybe we don't care about that.

Reproduce bug:
`rm master.log; cat testgtp.txt | $lzmaster -w latest.txt -p 10 --noponder -l master.log
`
Verify fix:
`rm test.log; cat testgtp.txt | ./leelaz -w latest.txt -p 10 --noponder -l test.log
`

[testgtp.txt](https://github.com/gcp/leela-zero/files/1629027/testgtp.txt)
[master.log](https://github.com/gcp/leela-zero/files/1629028/master.log)
[test.log](https://github.com/gcp/leela-zero/files/1629029/test.log)

I will to add a few more tests with undos and passes later.

Note different hashes based on ko status. Also note Ko-Hash is purely positional, so these positions have the same Ko-Hash. Superkos will be detected by going through the stack of them. 
```
   a b c d e f g h j k l m n o p q r s t
19 . . . . . . . . . . . . . . . . . . . 19
18 . . . . . . . . . . . . . . . . . . . 18
17 . . . . . . . . . . . . . . . . . . . 17
16 . . . + . . . . . + . . . . . + . . . 16
15 . . . . . . . . . . . . . . . . . . . 15
14 . . . . . . . . . . . . . . . . . . . 14
13 . . . . . . . . . . . . . . . . . . . 13
12 . . . . . . . . . . . . . . . . . . . 12
11 . . . . . . . . . . . . . . . . . . . 11
10 . . . + . . . . . + . . . . . + . . . 10
 9 . . . . . . . . . . . . . . . . . . .  9
 8 . . . . . . . . . . . . . . . . . . .  8
 7 . . . . . . . . . . . . . . . . . . .  7
 6 . . . . X O . . . . . . . . . . . . .  6
 5 . . . . X O . . . . . . . . . . . . .  5
 4 . . . X(O). O . . + . . . . . + . . .  4
 3 . . . . X O . . . . . . . . . . . . .  3
 2 . . . . . . . . . . . . . . . . . . .  2
 1 . . . . . . . . . . . . . . . . . . .  1
   a b c d e f g h j k l m n o p q r s t

Hash: 58C219E7A22CBE0D Ko-Hash: CF5EC0B8BB82BF52

   a b c d e f g h j k l m n o p q r s t
19 . . . . . . . . . . . . . . . . . . . 19
18 . . . . . . . . . . . . . . . . . . . 18
17 . . . . . . . . . . . . . . . . . . . 17
16 . . . + . . . . . + . . . . . + . . . 16
15 . . . . . . . . . . . . . . . . . . . 15
14 . . . . . . . . . . . . . . . . . . . 14
13 . . . . . . . . . . . . . . . . . . . 13
12 . . . . . . . . . . . . . . . . . . . 12
11 . . . . . . . . . . . . . . . . . . . 11
10 . . . + . . . . . + . . . . . + . . . 10
 9 . . . . . . . . . . . . . . . . . . .  9
 8 . . . . . . . . . . . . . . . . . . .  8
 7 . . . . . . . . . . . . . . . . . . .  7
 6 . . . . X(O). . . . . . . . . . . . .  6
 5 . . . . X O . . . . . . . . . . . . .  5
 4 . . . X O . O . . + . . . . . + . . .  4
 3 . . . . X O . . . . . . . . . . . . .  3
 2 . . . . . . . . . . . . . . . . . . .  2
 1 . . . . . . . . . . . . . . . . . . .  1
   a b c d e f g h j k l m n o p q r s t

Hash: 38EAE2514E7FA02E Ko-Hash: CF5EC0B8BB82BF52
```

This example shows why the zobrish hash isn't a simple "ko last turn" but specifies which intersection the ko is on. It matters which ko is active.
```
   a b c d e f g h j k l m n o p q r s t
19 O X . . . . . . . . . . . . . . . . . 19
18 . . . . . . . . . . . . . . . . . . . 18
17 . . . . . . . . . . . . . . . . . . . 17
16 . . . + . . . . . + . . . . . + . . . 16
15 . . . . . . . . . . . . . . . . . . . 15
14 . . . . . . . . . . . . . . . . . . . 14
13 . . . . . . . . . . . . . . . . . . . 13
12 . . . . . . . . . . . . . . . . . . . 12
11 . . . . . . . . . . . . . . . . . . . 11
10 . . . + . . . . . + . . . . . + . . . 10
 9 . . . . . . . . . . . . . . . . . . .  9
 8 . . . . . . . . . . . . . . . . . . .  8
 7 . . . . . . . . . . . . . . . . . . .  7
 6 . . . . . . . . . . . . . . . . . . .  6
 5 . . X . . . . . . . . . . . . . X . .  5
 4 . X(O)X . . . . . + . . . . . X O X .  4
 3 . O . O . . . . . . . . . . . O . O .  3
 2 . . O . . . . . . . . . . . . . O . .  2
 1 . . . . . . . . . . . . . . . . . . .  1
   a b c d e f g h j k l m n o p q r s t

Hash: 5BF1BD611EA9A560 Ko-Hash: 5DEAB48D9EB780B7

   a b c d e f g h j k l m n o p q r s t
19 O X . . . . . . . . . . . . . . . . . 19
18 . . . . . . . . . . . . . . . . . . . 18
17 . . . . . . . . . . . . . . . . . . . 17
16 . . . + . . . . . + . . . . . + . . . 16
15 . . . . . . . . . . . . . . . . . . . 15
14 . . . . . . . . . . . . . . . . . . . 14
13 . . . . . . . . . . . . . . . . . . . 13
12 . . . . . . . . . . . . . . . . . . . 12
11 . . . . . . . . . . . . . . . . . . . 11
10 . . . + . . . . . + . . . . . + . . . 10
 9 . . . . . . . . . . . . . . . . . . .  9
 8 . . . . . . . . . . . . . . . . . . .  8
 7 . . . . . . . . . . . . . . . . . . .  7
 6 . . . . . . . . . . . . . . . . . . .  6
 5 . . X . . . . . . . . . . . . . X . .  5
 4 . X O X . . . . . + . . . . . X(O)X .  4
 3 . O . O . . . . . . . . . . . O . O .  3
 2 . . O . . . . . . . . . . . . . O . .  2
 1 . . . . . . . . . . . . . . . . . . .  1
   a b c d e f g h j k l m n o p q r s t

Hash: F43208F94077881F Ko-Hash: 5DEAB48D9EB780B7

```

### DIFF
--- a/src/FastState.cpp
+++ b/src/FastState.cpp
@@ -34,7 +34,7 @@ void FastState::init_game(int size, float komi) {
 
     m_movenum = 0;
 
-    m_komove = 0;
+    m_komove = FastBoard::PASS;
     m_lastmove = 0;
     m_komi = komi;
     m_handicap = 0;
@@ -53,7 +53,7 @@ void FastState::reset_game(void) {
     m_movenum = 0;
     m_passes = 0;
     m_handicap = 0;
-    m_komove = 0;
+    m_komove = FastBoard::PASS;
     m_lastmove = 0;
 }
 
@@ -74,6 +74,10 @@ void FastState::play_pass(int color) {
 
     m_lastmove = FastBoard::PASS;
 
+    board.m_hash ^= Zobrist::zobrist_ko[m_komove];
+    m_komove = FastBoard::PASS;
+    board.m_hash ^= Zobrist::zobrist_ko[m_komove];
+
     if (board.m_tomove == color) {
         board.m_hash ^= 0xABCDABCDABCDABCDULL;
     }
@@ -90,9 +94,16 @@ void FastState::play_move(int vertex) {
 
 void FastState::play_move(int color, int vertex) {
     if (vertex != FastBoard::PASS) {
+        board.m_hash ^= Zobrist::zobrist_ko[m_komove];
         int kosq = board.update_board(color, vertex);
+        if ( board.get_square(kosq) == FastBoard::EMPTY &&
+            !board.is_suicide(kosq, !color)) {
+            m_komove = kosq;
+        } else {
+            m_komove = FastBoard::PASS;
+        }
+        board.m_hash ^= Zobrist::zobrist_ko[m_komove];
 
-        m_komove = kosq;
         m_lastmove = vertex;
         m_movenum++;
 
--- a/src/FastState.cpp
+++ b/src/FastState.cpp
@@ -95,13 +95,7 @@ void FastState::play_move(int vertex) {
 void FastState::play_move(int color, int vertex) {
     if (vertex != FastBoard::PASS) {
         board.m_hash ^= Zobrist::zobrist_ko[m_komove];
-        int kosq = board.update_board(color, vertex);
-        if ( board.get_square(kosq) == FastBoard::EMPTY &&
-            !board.is_suicide(kosq, !color)) {
-            m_komove = kosq;
-        } else {
-            m_komove = FastBoard::PASS;
-        }
+        m_komove = board.update_board(color, vertex);
         board.m_hash ^= Zobrist::zobrist_ko[m_komove];
 
         m_lastmove = vertex;
--- a/src/FastState.cpp
+++ b/src/FastState.cpp
@@ -148,7 +148,7 @@ int FastState::get_to_move() const {
 }
 
 void FastState::set_to_move(int tom) {
-    board.m_tomove = tom;
+    board.set_to_move(tom);
 }
 
 void FastState::display_state() {
--- a/src/FullBoard.cpp
+++ b/src/FullBoard.cpp
@@ -84,7 +84,7 @@ std::uint64_t FullBoard::calc_hash(int komove) {
     res ^= Zobrist::zobrist_pris[1][m_prisoners[1]];
 
     if (m_tomove == BLACK) {
-        res ^= 0xABCDABCDABCDABCDULL;
+        res ^= Zobrist::zobrist_blacktomove;
     }
 
     m_hash ^= Zobrist::zobrist_ko[komove];
@@ -100,6 +100,13 @@ std::uint64_t FullBoard::get_hash(void) const {
 
 std::uint64_t FullBoard::get_ko_hash(void) const {
     return m_ko_hash;
+}
+
+void FullBoard::set_to_move(int tomove) {
+    if (m_tomove != tomove) {
+        m_hash ^= Zobrist::zobrist_blacktomove;
+    }
+    FastBoard::set_to_move(tomove);
 }
 
 int FullBoard::update_board(const int color, const int i) {
--- a/src/FullBoard.cpp
+++ b/src/FullBoard.cpp
@@ -70,7 +70,7 @@ std::uint64_t FullBoard::calc_ko_hash(void) {
     return res;
 }
 
-std::uint64_t FullBoard::calc_hash(void) {
+std::uint64_t FullBoard::calc_hash(int komove) {
     auto res = std::uint64_t{0x1234567887654321ULL};
 
     for (int i = 0; i < m_maxsq; i++) {
@@ -86,6 +86,8 @@ std::uint64_t FullBoard::calc_hash(void) {
     if (m_tomove == BLACK) {
         res ^= 0xABCDABCDABCDABCDULL;
     }
+
+    m_hash ^= Zobrist::zobrist_ko[komove];
 
     m_hash = res;
 
@@ -179,6 +181,6 @@ void FullBoard::display_board(int lastmove) {
 void FullBoard::reset_board(int size) {
     FastBoard::reset_board(size);
 
-    calc_hash();
+    calc_hash(FastBoard::PASS);
     calc_ko_hash();
 }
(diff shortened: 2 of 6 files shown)


killerducky:
Created some more test cases in testgtp.txt:
https://senseis.xmp.net/?SendingTwoReturningOne
And a basic triple-ko.

These tests found some more problems, which I have fixed. I verified the hashes look correct but I still haven't actually verified UCTNode::kill_superkos works.

I also slipped in a small change in GTP.cpp to handle # comments. Before it only worked if there was a lone # by itself.

[testgtp.txt](https://github.com/gcp/leela-zero/files/1629945/testgtp.txt)
[test.log](https://github.com/gcp/leela-zero/files/1629946/test.log)

killerducky:
BTW note that m_hash and m_ko_hash are identical now except one includes side-to-move. For m_hash which is used by TTable this is required. For m_ko_hash which is super-ko, it depends which flavor of super-ko we want. Tromp-Taylor is positional super ko so does not include site-to-move. It's also most restrictive, and therefore safest -- we restrict ourselves to only moves that are legal in all super-ko rules.

killerducky:
Ok after actually testing super-ko I realized I got confused along the way, m_ko_hash never had prisoners or passes in it, only m_hash did.

I need to add back at least passes to m_hash. I'm not sure about prisoners, is there a reason to include them?

Clearly I've gotten myself confused on this so please bear with me while I get straight which parts of this change are good and which ones are messed up.

gcp:
>I'm not sure about prisoners, is there a reason to include them?

They're included in the score for Japanese rules, so that seemed sensible. (m_hash is mostly there for rulesets with Situtional Superko) In terms of TTable hashing, different prisoners indicate different move histories.

killerducky:
Ok then I think this is ready.

src/FastState.cpp
```diff
@@ -90,9 +94,16 @@ void FastState::play_move(int vertex) {
 
 void FastState::play_move(int color, int vertex) {
     if (vertex != FastBoard::PASS) {
+        board.m_hash ^= Zobrist::zobrist_ko[m_komove];
         int kosq = board.update_board(color, vertex);
+        if ( board.get_square(kosq) == FastBoard::EMPTY &&
+            !board.is_suicide(kosq, !color)) {
```
gcp: Nice fix, but maybe this logic belongs in update_board? It can be its responsibility to not return fake ko squares.

src/FullBoard.cpp
```diff
@@ -179,6 +188,6 @@ void FullBoard::display_board(int lastmove) {
 void FullBoard::reset_board(int size) {
     FastBoard::reset_board(size);
 
-    calc_hash();
+    calc_hash(FastBoard::PASS);
```
gcp: Default the argument to calc_hash as FastBoard::PASS, I think that's cleaner.

src/Zobrist.cpp
```diff
@@ -32,6 +33,11 @@ void Zobrist::init_zobrist(Random& rng) {
         }
     }
 
+    for (int j = 0; j < FastBoard::MAXSQ; j++) {
+        Zobrist::zobrist_ko[j]  = ((std::uint64_t)rng.randuint32()) << 32;
+        Zobrist::zobrist_ko[j] ^= (std::uint64_t)rng.randuint32();
```
gcp: At some point Random() should expose randuint64(no arg) and this can all be cleaned up using std::for_each or whatever.

src/FullBoard.cpp
```diff
@@ -84,9 +84,11 @@ std::uint64_t FullBoard::calc_hash(void) {
     res ^= Zobrist::zobrist_pris[1][m_prisoners[1]];
 
     if (m_tomove == BLACK) {
-        res ^= 0xABCDABCDABCDABCDULL;
+        res ^= Zobrist::zobrist_blacktomove;
     }
 
+    m_hash ^= Zobrist::zobrist_ko[komove];
```
oni-link: Should this be `res`? The next statement is `m_hash = res;`.

src/FastState.cpp
```diff
@@ -74,6 +74,10 @@ void FastState::play_pass(int color) {
 
     m_lastmove = FastBoard::PASS;
 
+    board.m_hash ^= Zobrist::zobrist_ko[m_komove];
+    m_komove = FastBoard::PASS;
+    board.m_hash ^= Zobrist::zobrist_ko[m_komove];
```
oni-link: Here we access field (size_t)(-1), because FastBoard::PASS is `-1`.

gcp:
Thanks for pointing that out. You are right of course, the code is buggy.

gcp:
Fixed in https://github.com/gcp/leela-zero/commit/25e14a2b8e2dc1d2da9458f197e538011a5dbce1.

killerducky:
Thanks @oni-link!

Sometime I'll have a look at how to write a gtest that would catch that bug.

gcp:
I'm writing a gtest for hashing right now.

PR #652, opened 2018-01-15
replaced #include <lib.h> with equivilant <clib>

sethtroisi:
`grep -rh '^#include <c[a-z]*>' *.cpp *.h | sed -e 's/.*<c//' -e s'/.$//' | sort -u | xargs -I{} sh -c 'echo "$(grep -riho "include <c{}>" *.cpp *.h | wc -w) c{} vs $(grep -riho "include <{}.h>" *.cpp *.h | wc -w) {}.h"' | sort -rn`

14 cassert vs 12 assert.h
12 cstdint vs 2 stdint.h
6 cstdlib vs 0 stdlib.h
6 cmath vs 0 math.h
6 chrono vs 0 hrono.h
6 cctype vs 0 ctype.h
4 cstdio vs 6 stdio.h
4 climits vs 0 limits.h
2 ctime vs 0 time.h
2 cstddef vs 12 stddef.h

afterwards I found one extra instance (of stdarg.h) with
`grep -rh '^#include .*\.h>' *.cpp *.h`

If you prefer the .h version I can convert all existing <cLIB> version to <LIB.h> I just prefer consistence

### DIFF
--- a/src/FastBoard.cpp
+++ b/src/FastBoard.cpp
@@ -18,7 +18,7 @@
 
 #include "FastBoard.h"
 
-#include <assert.h>
+#include <cassert>
 #include <array>
 #include <iostream>
 #include <queue>
--- a/src/FastState.h
+++ b/src/FastState.h
@@ -19,7 +19,7 @@
 #ifndef FASTSTATE_H_INCLUDED
 #define FASTSTATE_H_INCLUDED
 
-#include <stddef.h>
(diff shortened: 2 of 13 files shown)


gcp:
Surprised you didn't have to add more std:: prefixes to stuff. Maybe a build on macOS will point out any missing ones :-)

PR #654, opened 2018-01-15
autogtp: allow leelaz three numbers version

marcocalignano:


roy7:
@marcocalignano This broke compilation of the validation too.

> Validation.cpp:22:42: error: conflicting declaration âusing VersionTuple = class std::tuple<int, int>â
>  using VersionTuple = std::tuple<int, int>;
>                                           ^
> In file included from Validation.h:28:0,
>                  from Validation.cpp:19:
> ../autogtp/Game.h:25:47: note: previous declaration as âusing VersionTuple = class std::tuple<int, int, int>â
>

marcocalignano:
Will be solve with #661. Thanks @roy7

PR #656, opened 2018-01-16
Add autogtp to Travis

bood:
As you wish. Also changed to use gcc instead of clang.

Will try migrating to CMake system later when I have time (will be another PR).


gcp:
Thanks a lot!

PR #657, opened 2018-01-16
autogtp: make tuning before starting threads

marcocalignano:
I guess this you will like more. The tuning process is also verbose.

### DIFF
--- a/autogtp/Management.cpp
+++ b/autogtp/Management.cpp
@@ -57,8 +57,36 @@ Management::Management(const int gpus,
     m_fallBack(Order::Error) {
 }
 
+void Management::runTuningProcess(const QString &tuneCmdLine) {
+    QTextStream(stdout) << tuneCmdLine << endl;
+    QProcess tuneProcess;
+    tuneProcess.start(tuneCmdLine);
+    tuneProcess.waitForStarted(-1);
+    while(tuneProcess.state() == QProcess::Running) {
+        tuneProcess.waitForReadyRead(1000);
+        QTextStream(stdout) << tuneProcess.readAllStandardError();
+    }
+    tuneProcess.waitForFinished(-1);
+}
+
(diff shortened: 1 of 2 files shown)


autogtp/Management.cpp
```diff
@@ -57,8 +57,36 @@ Management::Management(const int gpus,
...
+        tuneProcess.waitForReadyRead(1000);
+        QTextStream(stdout) << tuneProcess.readAllStandardError();
+    }
+    tuneProcess.waitForFinished(-1);
+}
+
 void Management::giveAssignments() {
     sendAllGames();
+
+    //Make the OpenCl tunig before starting the threads
```
gcp: nit: typo "tuning"

autogtp/Management.cpp
```diff
@@ -57,8 +57,36 @@ Management::Management(const int gpus,
...
+    QTextStream(stdout) << "Start tuning process please wait ..." << endl;
```
gcp: "Starting tuning process, please wait..."

autogtp/Management.cpp
```diff
@@ -57,8 +57,36 @@ Management::Management(const int gpus,
     m_fallBack(Order::Error) {
 }
 
+void Management::runTuningProcess(const QString &tuneCmdLine) {
+    QTextStream(stdout) << tuneCmdLine << endl;
+    QProcess tuneProcess;
+    tuneProcess.start(tuneCmdLine);
```
gcp: If the user is running a CPU only client, it will reject --tune-only, so you need to handle this case and just bail out.

gcp:
Thanks!

autogtp/Management.cpp
marcocalignano: But the process will run, say something like "no OpenCl capable Device found" and exit. The code should not have any problem with this.

PR #669, opened 2018-01-18
OpenCl Network some math optimizations

marcocalignano:
with these optimizations my system is I get 15 ms/move. I guesse the same optimization can be done the CPU netwotk and probably would yield more performance gain but I have no way to test the CPU code. (or have I?).

### DIFF
--- a/src/FastBoard.cpp
+++ b/src/FastBoard.cpp
@@ -281,14 +281,7 @@ void FastBoard::display_board(int lastmove) {
     int boardsize = get_boardsize();
 
     myprintf("\n   ");
-    for (int i = 0; i < boardsize; i++) {
-        if (i < 25) {
-            myprintf("%c ", (('a' + i < 'i') ? 'a' + i : 'a' + i + 1));
-        } else {
-            myprintf("%c ", (('A' + (i-25) < 'I') ? 'A' + (i-25) : 'A' + (i-25) + 1));
-        }
-    }
-    myprintf("\n");
+    print_columns();
     for (int j = boardsize-1; j >= 0; j--) {
         myprintf("%2d", j+1);
         if (lastmove == get_vertex(0, j))
@@ -312,14 +305,20 @@ void FastBoard::display_board(int lastmove) {
         myprintf("%2d\n", j+1);
     }
     myprintf("   ");
-    for (int i = 0; i < boardsize; i++) {
-         if (i < 25) {
+    print_columns();
+    myprintf("\n");
+}
+
+void FastBoard::print_columns() {
+    for (int i = 0; i < get_boardsize(); i++) {
+        if (i < 25) {
             myprintf("%c ", (('a' + i < 'i') ? 'a' + i : 'a' + i + 1));
-        } else {
-            myprintf("%c ", (('A' + (i-25) < 'I') ? 'A' + (i-25) : 'A' + (i-25) + 1));
+        }
+        else {
+            myprintf("%c ", (('A' + (i - 25) < 'I') ? 'A' + (i - 25) : 'A' + (i - 25) + 1));
         }
     }
-    myprintf("\n\n");
+    myprintf("\n");
 }
 
 void FastBoard::merge_strings(const int ip, const int aip) {
--- a/src/FastBoard.h
+++ b/src/FastBoard.h
@@ -136,6 +136,7 @@ protected:
     void merge_strings(const int ip, const int aip);
     void add_neighbour(const int i, const int color);
     void remove_neighbour(const int i, const int color);
+    void print_columns();
 };
 
 #endif


killerducky:
15ms/move really? What was before?

gcp:
>I have no way to test the CPU code. 

It always runs for self-check. You can just disable USE_OPENCL in config.h

src/OpenCL.cpp
```diff
@@ -148,11 +156,12 @@ __kernel void out_transform(__global float *M, __global net_t *Y,
...
 
     if (k < K && block < P) {
+        int kHW = k * HW;
         int b = block_y * WTILES + block_x;
         float temp_m[16];
-        for (int xi = 0; xi < 4; xi++) {
+        for (int xi = 0; xi < 16; xi+=4) {
             for (int nu = 0; nu < 4; nu++) {
-                temp_m[xi*4 + nu] = M[xi*(4*Kpad*Ppad) + nu*(Kpad*Ppad)+ b*Kpad + k];
+                temp_m[xi + nu] = M[(xi + nu)*(KPpad)+ b*Kpad + k];
```
WhiteHalmos: b*Kpad might be worth extracting as it doesn't change, same for the loop below. Another thing is extracting xi * KPpad to the outer loop, and just adding KPpad each inner loop to replace nu*KPpad with simple addition, up to you.

Awesome changes by the way, nice to optimize away expensive operations.

Ttl:
This gives a 4% speed increase in netbench with 64 channel network on my computer.

marcocalignano:
@Ttl I know it is not much but still!

roy7:
Alteri might eventually submit a PR replacing all of the use of '19' and related numbers with a set of defines, so it's easier for him to track the master repo while also running in 9x9 instead of 19x19. Does the optimizer already realize W and H are never changing though I wonder, or would making them defines so the literal 19s go into the loop be any faster? I assume g++ -O3 can figure out variables that can't change from ones that might.

gcp:
>I assume g++ -O3

This is OpenCL code so we don't know what compiler is actually processing it. (Odds are it's LLVM derived, I guess)

marcocalignano:
I ma testing now the suggestion from @WhiteHalmos

marcocalignano:
Oh sorry the 15ms / move is the improvement. I go from 545 to 530.

Ttl:
One more change in the same style would be removing the T2 array in in_transform and instead assigning the T1 calculations directly to V.

marcocalignano:
@Ttl I was thinking of that but that saves little and I think I have a better operation for the all stuff (If I make it work!)

marcocalignano:
So 516 ms/move so a 29 ms gain that is good I think!

src/OpenCL.cpp
```diff
@@ -148,11 +157,14 @@ __kernel void out_transform(__global float *M, __global net_t *Y,
...
         int b = block_y * WTILES + block_x;
+        int bKpad = b*Kpad;
         float temp_m[16];
-        for (int xi = 0; xi < 4; xi++) {
-            for (int nu = 0; nu < 4; nu++) {
-                temp_m[xi*4 + nu] = M[xi*(4*Kpad*Ppad) + nu*(Kpad*Ppad)+ b*Kpad + k];
+        for (int xi = 0; xi < 16; xi+=4) {
+            int xiKPpad = xi * KPpad;
+            for (int nu = 0, nuKPpad = 0; nu < 4; nu++, nuKPpad += KPpad) {
+                temp_m[xi + nu] = M[xiKPpad + nuKPpad + bKpad + k];
```
Ttl: This can be replaced with a single loop.

```
int offset = b*Kpad + k;
for (int xn = 0; xn < 16; xn++) {
    temp_m[xn] = M[xn*KPpad + offset];
}
```

marcocalignano: But this is worst for the multiplication, then use that same trick like nuKPpad.
```
int offset = b*Kpad + k;
int xnKPpad = 0;
for (int xn = 0, xnKPpad = 0; xn < 16; xn++, xnKPpad += KPpad) {
    temp_m[xn] = M[xnKPpad + offset];
}
```

marcocalignano: Testing now!

Ttl: I think you could initialize xnKPad with offset to get rid of the addition in array access.

WhiteHalmos: Wow, nice job guys!

marcocalignano:
All the suggestion are in, I guess the check for the OpenCl driver check also for the correctness of this patch, right?

src/OpenCL.cpp
Hersmunch: How about
```
for (int xn = 0, xnKPpad = b*Kpad + k; xn < 16; xn++, xnKPpad += KPpad) {
    temp_m[xn] = M[xnKPpad];
}
```

src/OpenCL.cpp
```diff
@@ -125,7 +132,7 @@ __kernel void in_transform(__global net_t *in, __global float *V,
 
         for (int i = 0; i < 4; i++) {
             for (int j = 0; j < 4; j++) {
-                V[(i*4 + j)*Cpad*Ppad + offset] = T2[i][j];
+                V[(i*4 + j)*CPpad + offset] = T2[i][j];
```
Hersmunch: Similarly
```
for (int i = 0, v_idx = offset; i < 4; i++, v_idx += CPpad) {
    for (int j = 0; j < 4; j++) {
        V[v_idx] = T2[i][j];
    }
}
```

Hersmunch: Similarly
```
for (int i = 0, v_idx = offset; i < 4; i++) {
    for (int j = 0; j < 4; j++, v_idx += CPpad) {
        V[v_idx] = T2[i][j];
    }
}
```

src/OpenCL.cpp
```diff
@@ -172,14 +180,15 @@ __kernel void out_transform(__global float *M, __global net_t *Y,
                     temp_m[2*4 + 1] + temp_m[2*4 + 2] + temp_m[2*4 + 3] -
                     temp_m[3*4 + 1] + temp_m[3*4 + 2] + temp_m[3*4 + 3];
```
Hersmunch: Would calculating these only if they are needed, within the following ifs save anything?

src/OpenCL.cpp
marcocalignano: This one above is wrong: 
For each `j` you have the same index `v_idx`.

marcocalignano: Like I said I have something else in store for that part of code.

src/OpenCL.cpp
marcocalignano: We can do better here, but to late for I go to bed. look at it tomorrow.

gcp: I already tested that a while ago and I saw no gain.

On GPUs you're often forced to run all code in all ifs, even if some get discarded/masked, so these kind of tricks almost never works.

src/OpenCL.cpp
Hersmunch: Ah sorry. I missed that. I’ll try harder to read previous comments in future. Good luck with your idea!

marcocalignano:
@gcp So you deemed this PR useless?

WhiteHalmos:
I think he was commenting that optimizing with ifs won't matter. Other optimizations here are legit.

marcocalignano:
But they didn't make it in the release.

marcocalignano:
I have other ideas (a couple turn out not to be good), but I'd like to know if it is worth or not, @Ttl was saying that the improvement is not that big but I think could be really substantial for slow GPU, for CPU when I implement it and when we move to the 6x128 network.

roy7:
A small improvement is better than no improvement. And since everyone here is volunteers, it isn't like you're wasting company time/money on meaningless work. ;) If someone enjoys optimizing things here and there for small gains, more power to them. It all adds up in the end.

gcp:
>But they didn't make it in the release.

I had the impression you were still working on this, and I needed to stop pulling in changes at some point.

gcp:
>I think he was commenting that optimizing with ifs won't matter. Other optimizations here are legit.

Right. For example, you can optimize the for loop into a "non-edge" case and an "edge" case, and this will also give like 1-2%. But moving things in and out of if's usually makes no difference on a GPU, as long as the entire workgroup has to run both paths (divergence).

marcocalignano:
I close this because I will complete and rebase this PR to next.

PR #681, opened 2018-01-19
Next 679 tiny refactor bb

barrybecker4:
I made a tiny change in FastBoard::display_board to avoid the duplicating the code which prints the columns. I purposely made just a tiny change so that I can get comfortable with the process, and it will be more likely to be accepted before lots of other changes conflict and complex rebasing becomes necessary.


gcp:
I think you made a mistake with the base branch or something, as you can see from the list of changes in here.

barrybecker4:
OK. I'll try again. I did have some trouble merging from next into my fork.

Issue #679, opened 2018-01-19
tiny FastBoard refactor to avoid duplicate code in display_board


barrybecker4:
The code that prints the letters at the top and bottom of the board is duplicated.
I realize that this is a trivial thing, but I just want to see if I can create a PR and get it accepted (my last attempt at FastBoard changes got lost among many other modifications). If this goes through successfully, perhaps I will try more ambitious changes.

barrybecker4:
Closing. The change was merged.

PR #682, opened 2018-01-19
fix for #679 tiny FastBoad refactor to avoid duplicate code in display_board

barrybecker4:
Second attempt. Rather than try to fix my original repo, I just forked again and redid my change from scratch. Hope it can be merged this time.


barrybecker4:
Thanks! I will close the issue.

Issue #689, opened 2018-01-20
SGEMM optimizer


Cabu:
The old optimizer tested more than 5500 combinaisons of parameters to tune the sgemm. I got 2220 ms/move with this 'old' one. The new one just tested 578 combinaisons and run at 2550 ms/move...

alreadydone:
I guess you can just use your old configuration? (Just copy `leelaz_opencl_tuning` to the directory containing `autogtp` and `leelaz`.) If you prefer running tuning again, you can run `leelaz --full-tuner -w` with the weight file of a network of the proper size (5x64 or 6x128 for now).

gcp:
Use "--full-tuner" to get a tuning run trying more parameters.

Cabu:
Haa... I got tricked by the meaning of --full-tuner vs --tuner-only. Because when I use --full-tuner at the end of the tunning process, leela then start its command line interpreter, but with --tuner-only it quit right after the tuning process. But I didn't realized that with --tuner-only, the tunning was less thorough :( Perhaps the tunning process should be the same with the 2 options, or the second one could be called --quick-tuner?

marcocalignano:
Should I put --full-tuner --tuner-only  also in autogtp, I saw that with both option it stops after tuning.

PR #692, opened 2018-01-20
More tuner options

Cabu:
Add all the combinations for the tuner:
- full
- full & quit
- quick
- quick & quit


Cabu:
Patch to make tuner option more clear and solve issue "SGEMM optimizer #689"

Cabu:
Missed that autogp used the special case for tune-only. I prepare a new version.

PR #694, opened 2018-01-20
More tuning choices

Cabu:
Add all the combinations for the tuner:
    --full-tuner: Try harder to find an optimal OpenCL tuning.
    --full-tuner-only: Try harder to find an optimal OpenCL tuning and then exit.
    --quick-tuner: Tune OpenCL.
    --quick-tuner-only: Tune OpenCL only and then exit.
    --tune-only: Tune OpenCL only if needed and then exit.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -59,7 +59,8 @@ bool cfg_dumbpass;
 #ifdef USE_OPENCL
 std::vector<int> cfg_gpus;
 bool cfg_sgemm_exhaustive;
-bool cfg_tune_only;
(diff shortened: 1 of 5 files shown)


gcp:
What's the point of this?

Cabu:
This is actually the case. --tune-only just do a quick tuning and quit lella after doing so. --full-tuner do a thorough tuning and stay on the leela command interpreter. There is no way to do a full and quit of a quick and stay. This patch comme with these possibilities.

zediir:
Just use both commands. --tune-only --full-tuner

Issue #697, opened 2018-01-20
Improve error messages in GTP


barrybecker4:
For many GTP commands, if the user types the wrong thing, it just says "syntax not understood". 
This does not give the user much guidance on what they need to type so that their input will be accepted. There were a few times that I had to go look at the code to figure out what needed to be entered. Most users will not do that. I would like to add a few clarifying words to the error messages as suggested below.
 
- boardsize ->
  "unacceptable size" -> "%d is an unacceptable size.\nOnly a size of 19 is currently supported."
- genmove ->
  "syntax error" -> "You need to specify the color for whom to generate the move.\nFor example, 'genmove white'."
- kgs-genmove_cleanup ->
 "syntax error" -> "You need to specify the color to cleanup.\nFor example, 'kgs-genmove_cleanup white'."
- time-left ->
   "Color in time adjust not understood.\n" -> "Color $s in time adjust not understood.\nAcceptable values are 'b' or 'w'.", color
  "syntax not understood" -> "Syntax not understood. Expected parameters are color time stones"  
- fixed-handicap, place_free_handicap -> 
   "Not a valid number of handicap stones" -> "%d is not a valid number of handicap stones.\nUse a number that is in the range 2 - 9".
- kgs-time_settings ->
   "syntax not understood" -> "%s is not an expected KGS time setting. Try one of 'none', 'absolute', 'canadian', or 'byoyomi'."
- dump_training  
   "syntax not understood" -> "Unexpected winner color: %s.\nSpecify either 'b' or 'w'."
   
Also this block: 
```
     if (!cmdstream.fail()) {
        gtp_printf(id, "");
     } else {
        gtp_fail_printf(id, "syntax not understood");
     }
```
is repeated many times. I will make it a method.
   I'll create a PR shortly.

gcp:
I closed that PR because:

a) GTP is meant to be used between programs, it's not a human UI. If a human uses it, he's likely to be the one developing the engine or the GUI...
b) Some of the changes actually violated the GTP spec.

>There were a few times that I had to go look at the code to figure out what needed to be entered

You should look at the GTP spec.

>Most users will not do that.

I surely hope most people aren't trying to "speak" GTP!

barrybecker4:
I suppose the main use case is machines sending commands, but I thought the improved messages would have been useful for developers, like me, who were trying to test things out from the command line.

wctgit:
[...]

PR #699, opened 2018-01-20
#697 Enhance error message text in GTP.

barrybecker4:
Give the user more feedback if they incorrectly type a GTP command.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -265,7 +265,8 @@ bool GTP::execute(GameState & game, std::string xinput) {
 
         if (!cmdstream.fail()) {
             if (tmp != 19) {
-                gtp_fail_printf(id, "unacceptable size");
+                gtp_fail_printf(id, 
+                    "%d is an unacceptable size.\nOnly a size of 19 is currently supported.", tmp);
             } else {
                 float old_komi = game.get_komi();
                 Training::clear_training();
@@ -273,7 +274,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
                 gtp_printf(id, "");
             }
         } else {
-            gtp_fail_printf(id, "syntax not understood");
+            gtp_fail_printf(id, "Syntax not understood");
         }
 
         return true;
@@ -298,7 +299,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
             }
             gtp_printf(id, "");
         } else {
-            gtp_fail_printf(id, "syntax not understood");
+            gtp_fail_printf(id, "Syntax not understood");
         }
 
         return true;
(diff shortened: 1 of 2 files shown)


src/GTP.cpp
```diff
@@ -265,15 +265,16 @@ bool GTP::execute(GameState & game, std::string xinput) {
 
         if (!cmdstream.fail()) {
             if (tmp != 19) {
-                gtp_fail_printf(id, "unacceptable size");
+                gtp_fail_printf(id,
```
gcp: This violates the GTP protocol, see: 
https://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html#sec:list-of-all-commands

fails | Syntax error. If the engine cannot handle the new                 size, fails with the error message "unacceptable size".

It's the job of the GTP interface to translate these error messages to something more user friendly.

Using newlines in the responses seems to me like you risk breaking some GTP interfaces that don't expect this. I would avoid it if possible.

src/GTP.cpp
```diff
@@ -704,18 +717,17 @@ bool GTP::execute(GameState & game, std::string xinput) {
         } else if (winner_color == "b" || winner_color == "black") {
             who_won = FullBoard::BLACK;
         } else {
-            gtp_fail_printf(id, "syntax not understood");
+            gtp_fail_printf(id, 
+                "Unexpected winner color: %s.\n"
```
gcp: These commands are only produced by automated tools like GTP, so there's no point to this.

src/GTP.cpp
```diff
@@ -547,7 +558,9 @@ bool GTP::execute(GameState & game, std::string xinput) {
             auto stonestring = game.board.get_stone_list();
             gtp_printf(id, "%s", stonestring.c_str());
         } else {
-            gtp_fail_printf(id, "Not a valid number of handicap stones");
+            gtp_fail_printf(id, 
+                "%d is not a valid number of handicap stones.\n"
+                "Use a number that is in the range 2 - 9", stones);
```
gcp: Free handicap allows more than 9 stones. Also if you get into this condition then the option was missing, so "stones" won't have a valid value.

src/GTP.cpp
```diff
@@ -465,7 +470,9 @@ bool GTP::execute(GameState & game, std::string xinput) {
             } else if (color == "b" || color == "black") {
                 icolor = FastBoard::BLACK;
             } else {
-                gtp_fail_printf(id, "Color in time adjust not understood.\n");
+                gtp_fail_printf(id, 
+                    "Color $s in time adjust not understood.\n"
```
gcp: The command is time_left, not time adjust.

gcp:
The user isn't supposed to type GTP commands. I'm not sure there is much point to this.

gcp:
I'll close this because it's just very unlikely I'll end up merging it. GTP is meant to be used by developers that are working on some kind of interface. I am not sure these error messages help much there.

barrybecker4:
OK. I didn't realize that the spec specified the actual error message content and the the GTP interface was supposed to take the GTP error message and translate it into something else that was more user friendly.

PR #704, opened 2018-01-21
net2net

Ttl:
Implementation of net2net (https://arxiv.org/abs/1511.05641) for leelaz network.

Adding blocks works, but widening has still some bugs and the output of the widened network isn't exactly equal to the old one. It's still pretty close and it should be a better initialization than random even now.

Just putting this here in case anyone else was working on this. I'll investigate the widening later.

Example usage:

`python net2net.py 1 64 <network>` adds 1 block and 64 filters to the network. Output is written to `<network>_net2net`.

### DIFF
--- a/training/tf/net2net.py
+++ b/training/tf/net2net.py
@@ -45,15 +45,15 @@ def read_net(filename):
         return blocks, channels, weights
 
 def conv_bn_wider(weights, next_weights, inputs, channels,
-                  new_channels, noise_std=0, next_conv=True,
-                  verify=False):
-    """new_channels = number of channels to add"""
+                  new_channels, noise_std=0, last_block=False,
+                  rand=None, verify=False):
 
     if new_channels == 0:
         return weights, next_weights
 
-    rand = list(range(channels))
-    rand.extend(np.random.randint(0, channels, new_channels))
+    if rand == None:
+        rand = list(range(channels))
+        rand.extend(np.random.randint(0, channels, new_channels))
     rep_factor = np.bincount(rand)
 
     #Widen the current layer
@@ -62,7 +62,7 @@ def conv_bn_wider(weights, next_weights, inputs, channels,
     w_bn_vars = np.array(weights[3])[rand]
 
     #Widen the next layer inputs
-    if next_conv:
+    if not last_block:
         w_filter = 3
     else:
         w_filter = 1
@@ -112,7 +112,7 @@ if __name__ == "__main__":
 
     args = parser.parse_args()
     new_blocks = args.blocks
-    new_filters = args.filters
+    new_channels = args.filters
     net_filename = args.network
     noise_std = args.noise
     verify = args.verify
@@ -127,7 +127,7 @@ if __name__ == "__main__":
     if new_blocks < 0:
         raise ValueError("Blocks must be non-negative")
 
-    if new_filters < 0:
+    if new_channels < 0:
         raise ValueError("Filters must be non-negative")
 
     input_planes = 18
@@ -162,8 +162,14 @@ if __name__ == "__main__":
     #Version
     out_file.write('1\n')
 
+    #Making widening choice deterministic allows residual connection to be left
+    #as identity map. If the choice is not deterministic then the output of the
+    #widened network doesn't match the original one.
+    rand = list(range(channels))
+    rand.extend(np.random.randint(0, channels, new_channels))
+
     #Input
-    w_wider, conv_next = conv_bn_wider(w_input, [w_convs[0][0]], input_planes, channels, new_filters, noise_std, verify=verify)
+    w_wider, conv_next = conv_bn_wider(w_input, [w_convs[0][0]], input_planes, channels, new_channels, noise_std, rand=rand, verify=verify)
     w_convs[0][0] = conv_next[0]
 
     write_layer(w_wider, out_file)
@@ -171,12 +177,12 @@ if __name__ == "__main__":
     for e, w in enumerate(w_convs[:-1]):
         if e % 2 == 0:
             print("Processing block", 1 + e//2)
-        w_wider, conv_next = conv_bn_wider(w, [w_convs[e+1][0]], channels + new_filters, channels, new_filters, noise_std, verify=verify)
+        w_wider, conv_next = conv_bn_wider(w, [w_convs[e+1][0]], channels + new_channels, channels, new_channels, noise_std, rand=rand, verify=verify)
         w_convs[e+1][0] = conv_next[0]
         write_layer(w_wider, out_file)
 
     #The last block is special case because of policy and value heads
-    w_wider, w_next = conv_bn_wider(w_convs[-1], [w_pol[0], w_val[0]], channels + new_filters, channels, new_filters, noise_std, next_conv=False, verify=verify)
+    w_wider, w_next = conv_bn_wider(w_convs[-1], [w_pol[0], w_val[0]], channels + new_channels, channels, new_channels, noise_std, last_block=True, rand=rand, verify=verify)
     w_pol[0] = w_next[0]
     w_val[0] = w_next[1]
 
--- a/training/tf/net2net.py
+++ b/training/tf/net2net.py
@@ -1,4 +1,21 @@
 #!/usr/bin/env python3
+#
+#    This file is part of Leela Zero.
+#    Copyright (C) 2017 Henrik Forsten
+#
+#    Leela Zero is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation, either version 3 of the License, or
+#    (at your option) any later version.
+#
+#    Leela Zero is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with Leela Zero.  If not, see <http://www.gnu.org/licenses/>.
+
 import argparse
 import os
 import numpy as np
--- a/training/tf/net2net.py
+++ b/training/tf/net2net.py
@@ -5,7 +5,7 @@ import numpy as np
 import scipy.signal as signal
 from copy import deepcopy
 
-def convolve(w, x):
+def convolve(w, x , bn=None, bn_epsilon=1e-5):
     x_ch, x_w, x_h = x.shape
     outputs, inputs, _, __ = w.shape
     assert x_ch == inputs
@@ -13,6 +13,14 @@ def convolve(w, x):
     for o in range(outputs):
         for c in range(inputs):
             res[o,:,:] += signal.correlate2d(x[c,:,:], w[o,c,:,:], mode='same')
+    if bn == None:
+        return res
+    bn_means = bn[0]
+    bn_vars = bn[1]
+    for o in range(outputs):
+        scale = 1.0 / np.sqrt(bn_epsilon + bn_vars[o])
+        v = scale * (res[o,:,:] - bn_means[o])
+        res[o,:,:] = np.maximum(v, 0)
     return res
 
 def read_net(filename):
@@ -72,9 +80,9 @@ def conv_bn_wider(weights, next_weights, inputs, channels,
 
     if noise_std == 0 and verify:
         x = np.random.random((inputs, 19, 19))
-        old1 = convolve(np.array(weights[0]).reshape(channels, inputs, 3, 3), x)
+        old1 = convolve(np.array(weights[0]).reshape(channels, inputs, 3, 3), x, bn=[weights[2], weights[3]])
         old2 = convolve(np.array(next_weights[0]).reshape(-1, channels, w_filter, w_filter), old1)
-        new1 = convolve(np.array(w_conv_new).reshape(channels + new_channels, inputs, 3, 3), x)
+        new1 = convolve(np.array(w_conv_new).reshape(channels + new_channels, inputs, 3, 3), x, bn=[w_bn_means, w_bn_vars])
         new2 = convolve(np.array(next_weights_new[0]).reshape(-1, channels + new_channels, w_filter, w_filter), new1)
         assert (np.abs(old2 - new2) < 1e-6).all()
 
--- a/training/tf/net2net.py
+++ b/training/tf/net2net.py
@@ -124,7 +124,7 @@ if __name__ == "__main__":
     parser.add_argument("blocks", help="Residual blocks to add", type=int)
     parser.add_argument("filters", help="Filters to add", type=int)
     parser.add_argument("network", help="Input network", type=str)
-    parser.add_argument("--noise", nargs='?', help="Standard deviation of noise to add to new filters/blocks. Default: 1e-5", default=1e-5, type=float)
+    parser.add_argument("--noise", nargs='?', help="Standard deviation of noise to add to new filters/blocks. Default: 5e-3", default=5e-3, type=float)
     parser.add_argument("--verify", help="Verify that output matches", default=False, action='store_true')
 
     args = parser.parse_args()
--- a/training/tf/net2net.py
+++ b/training/tf/net2net.py
@@ -66,7 +66,7 @@ def conv_bn_wider(weights, next_weights, inputs, channels,
 
         for i in range(len(rand)):
             noise = 0
-            if i > channels:
+            if i >= channels:
                 noise = np.random.normal(0, noise_std)
             next_weights_new[j][:, i, :, :] *= (1.0 + noise)/rep_factor[rand[i]]
 
--- a/training/tf/net2net.py
+++ b/training/tf/net2net.py
@@ -1,0 +1,180 @@
+#!/usr/bin/env python3
+import argparse
+import os
+import numpy as np
+import scipy.signal as signal
+from copy import deepcopy
+
+def convolve(w, x):
+    x_ch, x_w, x_h = x.shape
+    outputs, inputs, _, __ = w.shape
+    assert x_ch == inputs
+    res = np.zeros((outputs, x_w, x_h))
+    for o in range(outputs):
+        for c in range(inputs):
+            res[o,:,:] += signal.correlate2d(x[c,:,:], w[o,c,:,:], mode='same')
+    return res
+
+def read_net(filename):
+    with open(filename, 'r') as f:
+        weights = []
+        for e, line in enumerate(f):
+            if e == 0:
+                print("Version", line.strip())
+                if line != '1\n':
+                    raise ValueError("Unknown version {}".format(line.strip()))
+            else:
+                weights.append(list(map(float, line.split(' '))))
+            if e == 2:
+                channels = len(line.split(' '))
+                print("Channels", channels)
+        blocks = e - (4 + 14)
+        if blocks % 8 != 0:
+            raise ValueError("Inconsistent number of weights in the file")
+        blocks //= 8
+        print("Blocks", blocks)
+
+        return blocks, channels, weights
+
+def conv_bn_wider(weights, next_weights, inputs, channels,
+                  new_channels, noise_std=0, next_conv=True,
+                  verify=False):
+    """new_channels = number of channels to add"""
+
+    if new_channels == 0:
+        return weights, next_weights
+
+    rand = range(channels)
+    rand.extend(np.random.randint(0, channels, new_channels))
+    rep_factor = np.bincount(rand)
+
+    #Widen the current layer
+    w_conv_new = np.array(weights[0]).reshape(channels, inputs, 3, 3)[rand, :, :, :]
+    w_bn_means = np.array(weights[2])[rand]
+    w_bn_vars = np.array(weights[3])[rand]
+
+    #Widen the next layer inputs
+    if next_conv:
+        w_filter = 3
+    else:
+        w_filter = 1
+
+    next_weights_new = []
+    for j in range(len(next_weights)):
+        n = np.array(next_weights[j]).reshape(-1, channels, w_filter, w_filter)
+        next_weights_new.append(n[:, rand, :, :])
+
+        for i in range(len(rand)):
+            noise = 0
+            if i > channels:
+                noise = np.random.normal(0, noise_std)
+            next_weights_new[j][:, i, :, :] *= (1.0 + noise)/rep_factor[rand[i]]
+
+    if noise_std == 0 and verify:
+        x = np.random.random((inputs, 19, 19))
+        old1 = convolve(np.array(weights[0]).reshape(channels, inputs, 3, 3), x)
+        old2 = convolve(np.array(next_weights[0]).reshape(-1, channels, w_filter, w_filter), old1)
+        new1 = convolve(np.array(w_conv_new).reshape(channels + new_channels, inputs, 3, 3), x)
+        new2 = convolve(np.array(next_weights_new[0]).reshape(-1, channels + new_channels, w_filter, w_filter), new1)
+        assert (np.abs(old2 - new2) < 1e-6).all()
+
+    w_conv_new = w_conv_new.flatten()
+    for j in range(len(next_weights)):
+        next_weights_new[j] = next_weights_new[j].flatten()
+
+    #Biases are always zero
+    bias = np.zeros(channels + new_channels)
+
+    w_new = [w_conv_new, bias, w_bn_means, w_bn_vars]
+
+    return w_new, next_weights_new
+
+def write_layer(weights, out_file):
+    for w in weights:
+        out_file.write(' '.join(map(str,w)) + '\n')
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description='Add filters/blocks to existing network such that the output is preserved.')
+    parser.add_argument("blocks", help="Residual blocks to add", type=int)
+    parser.add_argument("filters", help="Filters to add", type=int)
+    parser.add_argument("network", help="Input network", type=str)
+    parser.add_argument("--noise", nargs='?', help="Standard deviation of noise to add to new filters/blocks. Default: 1e-5", default=1e-5, type=float)
+    parser.add_argument("--verify", help="Verify that output matches", default=False, action='store_true')
+
+    args = parser.parse_args()
+    new_blocks = args.blocks
+    new_filters = args.filters
+    net_filename = args.network
+    noise_std = args.noise
+    verify = args.verify
+
+    if verify and noise_std != 0:
+        raise ValueError("Noise must be zero if verify is enabled.")
+
+    base, ext = os.path.splitext(net_filename)
+    output_filename = base + "_net2net" + ext
+    blocks, channels, weights = read_net(net_filename)
+
+    if new_blocks < 0:
+        raise ValueError("Blocks must be non-negative")
+
+    if new_filters < 0:
+        raise ValueError("Filters must be non-negative")
+
+    input_planes = 18
+
+    #Input convolution, bias, batch norm means, batch norm variances
+    w_input = weights[:4]
+
+    #Residual block convolution + batch norm
+    w_convs = []
+    for b in range(2*blocks):
+        w_convs.append(weights[4 + b*4: 4 + (b+1)*4])
+
+    i = ((b+1)*4) + 4
+    w_pol = weights[i:i+6]
+    w_val = weights[i+6:]
+
+    if new_blocks > 0:
+        #New blocks must have zero output due to the residual connection
+        new_block_conv = np.random.normal(0, noise_std, 9*(channels)**2)
+        new_block_bias = np.zeros(channels)
+        new_block_bn_mean = new_block_bias.copy()
+        new_block_bn_variances = np.ones(channels)
+        new_block = [new_block_conv, new_block_bias, new_block_bn_mean, new_block_bn_variances]
+
+        for i in range(2*new_blocks):
+            w_convs.append(deepcopy(new_block))
+
+        blocks += new_blocks
+
+    out_file = open(output_filename, 'w')
+
+    #Version
+    out_file.write('1\n')
+
+    #Input
+    w_wider, conv_next = conv_bn_wider(w_input, [w_convs[0][0]], input_planes, channels, new_filters, noise_std, verify=verify)
+    w_convs[0][0] = conv_next[0]
+
+    write_layer(w_wider, out_file)
+
+    for e, w in enumerate(w_convs[:-1]):
+        if e % 2 == 0:
+            print("Processing block", 1 + e//2)
+        w_wider, conv_next = conv_bn_wider(w, [w_convs[e+1][0]], channels + new_filters, channels, new_filters, noise_std, verify=verify)
+        w_convs[e+1][0] = conv_next[0]
+        write_layer(w_wider, out_file)
+
+    #The last block is special case because of policy and value heads
+    w_wider, w_next = conv_bn_wider(w_convs[-1], [w_pol[0], w_val[0]], channels + new_filters, channels, new_filters, noise_std, next_conv=False, verify=verify)
+    w_pol[0] = w_next[0]
+    w_val[0] = w_next[1]
+
+    write_layer(w_wider, out_file)
+
+    write_layer(w_pol, out_file)
+    write_layer(w_val, out_file)
+
+    out_file.close()
--- a/training/tf/net2net.py
+++ b/training/tf/net2net.py
@@ -63,7 +63,7 @@ def read_net(filename):
 
 def conv_bn_wider(weights, next_weights, inputs, channels,
                   new_channels, noise_std=0, last_block=False,
-                  rand=None, verify=False):
+                  rand=None, dir_alpha=None, verify=False):
 
     if new_channels == 0:
         return weights, next_weights
@@ -72,6 +72,24 @@ def conv_bn_wider(weights, next_weights, inputs, channels,
         rand = list(range(channels))
         rand.extend(np.random.randint(0, channels, new_channels))
     rep_factor = np.bincount(rand)
+
+    factor = np.zeros(len(rand))
+
+    #In the net2net paper every input weight was weighted equally,
+    #but in general we can have unequal division of the weights
+    if dir_alpha == None:
+        #Equal division
+        for i in range(len(rand)):
+            factor[i] = 1.0/rep_factor[rand[i]]
+    else:
+        #Unequal input weighting determined by dirichlet distribution
+        for i in range(channels):
+            x = np.random.dirichlet([dir_alpha]*rep_factor[i])
+            e = 0
+            for j in range(channels + new_channels):
+                if rand[j] == i:
+                    factor[j] = x[e]
+                    e += 1
 
     #Widen the current layer
     w_conv_new = np.array(weights[0]).reshape(channels, inputs, 3, 3)[rand, :, :, :]
@@ -93,7 +111,7 @@ def conv_bn_wider(weights, next_weights, inputs, channels,
             noise = 0
             if i >= channels:
                 noise = np.random.normal(0, noise_std)
-            next_weights_new[j][:, i, :, :] *= (1.0 + noise)/rep_factor[rand[i]]
+            next_weights_new[j][:, i, :, :] *= (1.0 + noise)*factor[i]
 
     if noise_std == 0 and verify:
         x = np.random.random((inputs, 19, 19))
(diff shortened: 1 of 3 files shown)


training/tf/net2net.py
```diff
@@ -0,0 +1,180 @@
...
+def conv_bn_wider(weights, next_weights, inputs, channels,
+                  new_channels, noise_std=0, next_conv=True,
+                  verify=False):
+    """new_channels = number of channels to add"""
+
+    if new_channels == 0:
+        return weights, next_weights
+
+    rand = range(channels)
+    rand.extend(np.random.randint(0, channels, new_channels))
```
isty2e: In python 3, ranges cannot be directly extended:
```
AttributeError: 'range' object has no attribute 'extend'
```
Changing `rand = range(channels)` to `rand = list(range(channels))`, it works fine for me.

isty2e:
I tested with the latest best net, adding a block and 128 channels, and it does not undergo any verification problem. In "the output of the widened network isn't exactly equal to the old one", do you mean the deviation under 1e-6?

Ttl:
Verification in the program works. However if you try the widened net in leelaz it doesn't give the same heatmap.

isty2e:
Do you have any slightest clue what is the problem here? Having read the code, I still don't see why the heatmap and NN evals are different.

Ttl:
As you can see the included verification code passes. I have no idea why the heatmaps are different.

The verification is calculated without batch norms and residual connections so maybe something related to those? Although I believe I'm processing the batch norm weights correctly and the residual connection shouldn't affect the widening. Or maybe there's a bug with the text processing somewhere?

isty2e:
As far as I have tested, text processing and I/O seem okay, which was tested by adding 0 block and 0 channel, and removing `if new_channels == 0` in `conv_bn_wider()`. Also after adding blocks and filters, the first values of weights were appropriately devided by `rep_factor`, and I don't think there is anything wrong with the batchnorm layer too. Can there anything with the last layers?

training/tf/net2net.py
```diff
@@ -0,0 +1,180 @@
...
+        w_filter = 1
+
+    next_weights_new = []
+    for j in range(len(next_weights)):
+        n = np.array(next_weights[j]).reshape(-1, channels, w_filter, w_filter)
+        next_weights_new.append(n[:, rand, :, :])
+
+        for i in range(len(rand)):
+            noise = 0
+            if i > channels:
```
Hersmunch: Should this be >= ?

Hersmunch:
Does the order of deepening and widening matter? Has the order of deepen then widen been chosen for a particular reason?
From what I read I don't think it matters but perhaps it is cheaper/simpler to widen then deepen since the deepening appears to be cheaper. I think it also would mean that noise won't be applied twice to the new filters in the new blocks.

training/tf/net2net.py
Ttl: Good catch. It should be >=.

Ttl:
It was easier to do it this way since widening also modifies the next layer inputs. New channels in the new blocks will have more noise, but I'm not worried about that.

bood:
A noob question: if this works correctly, are we expecting a 6b network
with identical outputs as the original 5b network? Any drawbacks compared
to supervised 6b version currently used?

Ttl:
The modified network should have identical outputs to the old one. It's meant to be used as initialization for the training. Training should converge much quicker when output is already almost correct compared to starting from completely random network.

Ttl:
Training of the 6 block network was started from random initialization. There aren't other alternatives right now. Don't confuse random weight initialization with self-play of random network.

This doesn't solve whatever problem is causing the value network to be overfit. Network initialized with this method still needs to be trained and it would end up being overfit in the same way.

bood:
Thanks for clarify. I get your point about 6b net being trained from
random. My questions is actually about the converted 6b net from 5b net.
I thought the output 6b network could be used for self-play too, no?

The first 6b net playing self-play (ed0/7fd) is supervised training from
last 5b net (c83), but ed0/7fd appears to be weak in its value network
because of overfitting issue (I'm not so unsure though).

My question is if net2net is able to create an identical 6b network from
c83, we can use it to start self-play and it seems it will have a better
value network than ed0/7fd thus save us a lot of time.

Correct me if anything I said is not true.

roy7:
I get @bood's point and I assume he's correct. If we extend the best 5x64 into a 6x128 in a way that has the same results for all inputs, it's simply larger, it should basically just keep going from where 5x64 is currently without the artifacts a supervised training 6x128 has had.

If nothing else, if we can extend the network successfully, training could be tried with both 6x128 networks to see which can promote to a better new network sooner. And it'd also be a success to build on later for 7x128 or 6x256 or whatever once the technique is proven.

isty2e:
My opinion is that it is unclear if there is any overfitting in the current 6-block networks since there is no evidence to support it, and if any, it can be cured by the self-play reinforcement learning. At any rate, this provides a safe and fast connection from a small net to a larger one, hence the bootstrapping is much easier and reliable. Of course, the first training cycle after bootstrapping probably will take more time compared with a regular cycle.

jkiliani:
Has any play testing been done yet how the enlarged net does compared to the base version when matched against each other?

bood:
I think this still has some implementation problem, heatmap of the
generated 6b network is not the same as the original 5b network.

isty2e:
Yes, due to the issue @bood mentioned it is not complete yet. Adding blocks does not matter, but upon addition of filters the heatmap and eval are somehow distorted, so it is not the time for actual match-tests yet. Like I said, I could not find why that happened, so if anyone is interested, please take a look and figure out why.

bood:
> My opinion is that it is unclear if there is any overfitting in the current 6-block networks since there is no evidence to support it, and if any, it can be cured by the self-play reinforcement learning. At any rate, this provides a safe and fast connection from a small net to a larger one, hence the bootstrapping is much easier and reliable. Of course, the first training cycle after bootstrapping probably will take more time compared with a regular cycle.

Yes, it's maybe not due to overfitting, I don't what it should be called, but it seems that supervise trained 6b would lose some information. e.g. c83 knows quite a bit of ladder but ed0/7fd seems to be weak on ladder since c83 plays little ladder so ed0/7fd has nothing to learn from. Just like the weights.txt we get from human games.

On the other hand, if we have an identical 6b from 5b, we even don't have any bootstrapping problem! we can continue to include those self-play games generated from c83 in the training window, since in theory this 6b would just generate the same games as c83.

isty2e:
@bood Due to the difference in the training set, the skills they have learned should be different, so it will be eventually handled once we are on track again. And even using this net2net method, practically a larger net is expected to behave differently, since we have to add random noise to break the symmetry in weights to avoid some neurons to be exactly identical. Thus there should be some training before replacing the best net. 

Still, this can save a lot of time, since when we start with a larger net, we do not have to start from a random network and train it with all the games played historically, but instead start with a net2net version and train with a given training window.

Mardak:
Does it make sense to try training a deeper net first? E.g., 5x64 -> 10x64?

```
$ ./net2net.py 5 0 c83e1b6e0ffbf8e684f2d8f6261853f14c553b29ee0e70ff6c34e87d28009c43
Version 1
Channels 64
Blocks 5
Processing block 1
Processing block 2
Processing block 3
Processing block 4
Processing block 5
Processing block 6
Processing block 7
Processing block 8
Processing block 9
Processing block 10

$ ./leelaz -w c83e1b6e0ffbf8e684f2d8f6261853f14c553b29ee0e70ff6c34e87d28009c43_net2net
…
Detecting residual layers...v1...64 channels...10 blocks.
…
Leela: heatmap
  0   0   1   1   1   1   1   0   0   0   0   0   0   0   0   0   1   1   1 
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0   0 
  0   1   3   5   1   0   1   1   1   1   1   1   1   0   1   7   3   1   0 
  0   1   5 144   3   1   1   1   1   1   1   1   1   0   2 133   6   1   0 
  0   1   1   3   1   1   1   1   1   1   1   1   1   1   1   3   1   1   0 
  0   1   0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0 
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0 
  0   0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0   0 
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0 
  0   0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0   0 
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0 
  0   0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0 
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0 
  0   1   0   0   1   1   1   1   1   1   1   1   1   1   1   0   0   1   0 
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   2   1   1   0 
  0   1   5 134   2   0   1   1   1   1   1   1   1   0   2 127   5   1   0 
  0   1   3   6   1   0   0   1   1   1   1   1   1   0   1   5   3   1   0 
  0   1   1   1   1   1   1   0   1   0   1   0   1   1   1   1   1   1   0 
  0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0 
pass: 0
winrate: 0.440024
```

Ttl:
The non-matching output is caused by the residual connection. The first convolution has some channels duplicated and the next layer's inputs for the duplicated channels are divided by the duplication factor causing the output to be the same as before. Residual connection however messes this up since it doesn't include the division by the duplication factor.

To confirm it I commented the residual connection out in the leelaz source code and the output matches.

However I'm not sure what would be the best way to fix it. This might be the closest it gets.

roy7:
I'll ping @paengs in case he has any ideas on why the new expanded network isn't identical input/output to the prior network as a starting point after the discussion here. :) He offered to help in the other original thread where Net2Net came up. (Thanks!)

isty2e:
Oh, the skip connection. [This paper](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/11/subnet_morph_v3.9.1_submitted.pdf) seems relevant; I will read it thoroughly when I have some spare time.

roy7:
Ok @Ttl I'm back up to speed for training. If someone can sort out a way to fix the residual layer when adding filters that'd be awesome. In the mean time I might try training some larger 64 filter networks to see what happens. First I'm going to validate 46ff 5x64 vs 46ff 6x64 for my peace of mind that they are roughly the same strength.

alreadydone:
I think the problem is due to randomness of the function g (see original paper). If you have a layer giving output [ax+by,cx+dy] when given input [x,y], then after widening the input, one possibility is that it gives output [ax+by,cx+dy] when given input [x,y,x,y]. But after widening the output, the output may well become [ax+by,cx+dy,cx+dy,ax+by] (even [ax+by,cx+dy,cx+dy,cx+dy] is possible). When the skip connection is added, the layer transforms [x,y,x,y] into [x+ax+by,y+cx+dy,x+cx+dy,y+ax+by] which is clearly not what we want: we want duplicated [x+ax+by,y+cx+dy].

My solution to this is to make g deterministic: we can do this since we are exactly doubling the number of filters. Just defining g(i)=i for 1≤i≤4 and g(i)=i-64 for 65≤i≤128 should work.

Mardak:
Just a quick sanity check with 1 playout of `c83e1b6e` deepened to 20x64 vs the original:

```
206 wins, 194 losses
400 games played.
Status: 0 LLR -0.824481 Lower Bound -2.94444 Upper Bound 2.94444
```

roy7:
I'm doing my own check right now although I'm currently at 128-101 for 5x64 vs 6x64 net2net with 50 playouts.

Ttl:
@alreadydone Nice work. Making rand deterministic fixes the output. Every channel doesn't need to be doubled, it works in all cases. When rand is deterministic the output head input convolution input modification using the same rand fixes the output.

I modified a 2 block 8 channel to 12 block 256 channel network and the output still matched. Looks like it's working now.

bood:
Tests on c83 too, heatmap looks very similar now, maybe just due to some precision loss?

```
c83
...
Leela: heatmap
  0   0   1   1   1   1   1   0   0   0   0   0   0   0   0   0   1   1   1
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0   0
  0   1   3   5   1   0   1   1   1   1   1   1   1   0   1   7   3   1   0
  0   1   5 144   3   1   1   1   1   1   1   1   1   0   2 133   6   1   0
  0   1   1   3   1   1   1   1   1   1   1   1   1   1   1   3   1   1   0
  0   1   0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0
  0   0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0   0
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0
  0   0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0   0
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0
  0   0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0
  0   1   0   0   1   1   1   1   1   1   1   1   1   1   1   0   0   1   0
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   2   1   1   0
  0   1   5 134   2   0   1   1   1   1   1   1   1   0   2 127   5   1   0
  0   1   3   6   1   0   0   1   1   1   1   1   1   0   1   5   3   1   0
  0   1   1   1   1   1   1   0   1   0   1   0   1   1   1   1   1   1   0
  0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
pass: 0
winrate: 0.440025

6x128 from c83
...
Leela: heatmap
  0   0   1   1   1   1   1   0   0   0   0   0   0   0   0   0   1   1   1
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0   0
  0   1   3   6   1   1   1   1   1   1   1   1   1   0   1   7   3   1   0
  0   1   6 146   3   1   1   1   1   1   1   1   1   0   2 132   6   1   0
  0   1   1   3   1   1   1   1   1   1   1   1   1   1   1   3   1   1   0
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0
  0   0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0   0
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0
  0   0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0   0
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0
  0   0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   1   0
  0   1   0   0   1   1   1   1   1   1   1   1   1   1   1   0   0   1   0
  0   1   1   1   1   1   1   1   1   1   1   1   1   1   1   2   1   1   0
  0   1   4 131   2   1   1   1   1   1   1   1   1   0   2 125   5   1   0
  0   1   3   6   1   0   1   1   1   1   1   1   1   0   1   5   3   1   0
  0   1   1   1   1   1   1   0   1   0   1   0   1   1   1   1   1   1   0
  0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
pass: 0
winrate: 0.438191
```

Let's do some tests now! @roy7

Ttl:
If you add "--noise 0" option the heatmap should be identical. Don't use noiseless network for training though.

I think that the first convolution in a block can be widened randomly. I'll add that in afternoon when I'm done with work. Then this should be ready to be merged.

roy7:
I wasted 4 hours tonight, and still going, in an upgrade nightmare because apparently Manjaro-stable pushed out CUDA 9.1 which requires a newer nvidia driver than is under the LTS track, but Manjaro-stable only uses LTS nvidia drivers. So it isn't possible with an updated system to use CUDA. Since I build my own tensorflow (for optimizations) it's like ~30 minutes to rebuild on top of the other issues.

Almost finally there. I hoped to train and test things all evening until I discovered CUDA broken lol. At least killerducky has his training setup working now too. I was going to try training a 10x64 overnight. Maybe I should wait and try an expanded 6x128 as a comparison vs the bootstrapped 6x128?

roy7:
Ok I got system working again. Converted the 10x64 to a model for training with net_to_model, fired off

 ./parse.py train_46ff4b94 leelaz-model-0

and we're on our way:

step 1000, policy=3.45318 mse=0.230687 reg=2.53024 total=6.90617 (0 pos/s)
step 2000, policy=3.19963 mse=0.196101 reg=2.34277 total=6.32681 (2422.65 pos/s)
step 3000, policy=3.16869 mse=0.172517 reg=2.14376 total=6.00252 (2904.45 pos/s)
step 4000, policy=3.13173 mse=0.15754 reg=1.96568 total=5.72757 (2905.81 pos/s)

I'm using only train_46ff4b94 data, with an expanded 10x64 model from 46ff4b94. I'll let it run over night and tomorrow we'll see if it's any better (or can even tie) the existing model. I set RESIDUAL_FILTERS = 64
and RESIDUAL_BLOCKS = 10 but made no other changes to learning rate, batch sizes, etc.

bood:
ok, but why not just expanding c83 to 6x128? since we're doing 6x128 right now, a good comparison can be made.

alreadydone:
> I think that the first convolution in a block can be widened randomly. 

@Ttl Nice observation; I think so. Another possible randomization is that instead of 
U<sup>(i+1)</sup><sub>j,h</sub>=1/|{x|g(x)=g(j)}| W<sup>(i+1)</sup><sub>g(j),h</sub>
we may replace 1/|{x|g(x)=g(j)}| by any function w(j) such that summation of w(x) over all x satisfying g(x)=g(j) yields 1. In other words,
Σ<sub>x such that g(x)=g(j)</sub>w(x)=1 for all j.
This corresponds to replacing f/2, f/2 in Figure 2 by say f/3, 2f/3.
(Algorithm 1 is very confusing, containing typos like 1/c_j which should be 1/c_k; the main text is clear though.)

Because there are so many possible randomizations, we may use any randomized network for self-play, and the generated games can be used to train all these randomized networks; we'll see which fare best.

roy7:
The c83 data wasn't listed as available for download. So I took the most recent batch of 250K+ games to save effort. :) Should I cancel and do 6x128 instead of 10x64 now that you guys fixed the expansion?

bood:
I think you should 😄  result of 6x128 seems to be more interesting to check right now @roy7

roy7:
Ok ok. :)

After 8000 steps my 10x64 tried a -p 1 faceoff vs the original 46ff. Original wins 27-1. Stomping.

roy7:
No joy. I can't run the net_to_model on the 6x128.

```
[roy@ryzen-one tf]$  python net2net.py 1 64 46ff
Version 1
Channels 64
Blocks 5
Processing block 1
Processing block 2
Processing block 3
Processing block 4
Processing block 5
Processing block 6
[roy@ryzen-one tf]$ python net_to_model.py 46ff_net2net
Version 1
Channels 128
Blocks 6.0
2018-01-23 02:03:22.082229: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:895] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero
2018-01-23 02:03:22.082737: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1105] Found device 0 with properties:
name: GeForce GTX 1080 Ti major: 6 minor: 1 memoryClockRate(GHz): 1.683
pciBusID: 0000:09:00.0
totalMemory: 10.91GiB freeMemory: 10.12GiB
2018-01-23 02:03:22.082756: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1195] Creating TensorFlow device (/device:GPU:0) -> (device: 0, name: GeForce GTX 1080 Ti, pci bus id: 0000:09:00.0, compute capability: 6.1)
WARNING:tensorflow:From /home/roy/code/leela-zero/training/tf/tfprocess.py:67: softmax_cross_entropy_with_logits (from tensorflow.python.ops.nn_ops) is deprecated and will be removed in a future version.
Instructions for updating:

Future major versions of TensorFlow will allow gradients to flow
into the labels input on backprop by default.

See tf.nn.softmax_cross_entropy_with_logits_v2.

  File "net_to_model.py", line 33, in <module>
    tfprocess.replace_weights(weights)
  File "/home/roy/code/leela-zero/training/tf/tfprocess.py", line 128, in replace_weights
    new_weight = tf.constant(new_weights[e], shape=shape)
  File "/usr/lib/python3.6/site-packages/tensorflow/python/framework/constant_op.py", line 212, in constant
    value, dtype=dtype, shape=shape, verify_shape=verify_shape))
  File "/usr/lib/python3.6/site-packages/tensorflow/python/framework/tensor_util.py", line 468, in make_tensor_proto
    (shape_size, nparray.size))
ValueError: Too many elements provided. Needed at most 10368, but received 20736
```

roy7:
I wonder if I'm curious enough to compare heatmaps between the 10x64 and 5x64. I wonder what it is about those 8000 steps that make it incapable of winning a game any more.

isty2e:
@roy7 My understanding is that though Net2Net provides a better initialization, it still requires some amount of training to work properly. Weights in the converged network should be quite different from the initialized values. We might need 0.5M steps or so?

roy7:
Perhaps so. But since the expanded/deeper network is the same input/output as the prior network, I figured any further training would simply make it more accurate. But of course we don't see that normally anyway. It just is counter intuitive to me as I tend to think that more training = more accurate neural networks.

alreadydone:
Why not train c83e1b6e converted to 6x128 using the games generated by itself? If migration is to work we should expect this to produce a better 6x128 network, though it failed to produce a better 5x64 network.

roy7:
I didn't see those games available for download in the archive web site/post, is why. I took the newest games available.

isty2e:
By the way, what was the loss function component values for the original network? Also, the larger the network, the more vulnerable it is under overfitting, I wonder if there is a overfitting issue.

roy7:
What does a training step in LZ context actually mean? I'm assuming it isn't showing 1 example pattern to the network and making tiny changes. Since there's no way 8000 example board states out of 250K games could overfit.

isty2e:
A training step would correspond to a minibatch. Still I am not sure if it is due to overfitting; There can be other reasons, of course.

Ttl:
@roy7 I think your issue with net_to_model was that tfprocess had different blocks and channels than the network you were trying to import.

Ttl:
Added the random replication of channels to first convolution in block and unequal input weight division suggested by @alreadydone . Output still matches with the new options.

This should be now ready to be used and can be merged.

roy7:
Ohh ok thanks. I was thinking tfprocess was only used by parse not net_to_model. Will try again.

roy7:
Yes that fixed it. Perhaps when a model is provided for input (net_to_model and parse) the # of filters and blocks could be auto detected?

Ttl:
I just added a commit that will make the error message more clear in #732

roy7:
Interesting. I didn't pull in @ttl's latest changes but my 5x64 beats 10x64 with a sprt pass. 

I'll git pull and start again to be sure they are identical. Or should they not be?

roy7:
-p 1 is maybe too unreliable, thought maybe I'd get lucky. ;) This is 5x64 vs 6x128 with no training, just net2net upgrade.

```
46ff     v 46ff_net ( 339 games)
              wins        black       white
46ff      163 48.08%   86 48.04%   77 48.13%
46ff_net  176 51.92%   93 51.96%   83 51.88%
                      179 52.80%  160 47.20%
```

I guess I should man up and run -p 1600 for a true confirmation the nets play identically.

isty2e:
@roy7 176-163 is not statistically significant to prove that the first is worse. Beside that the winrate is in 46.6% - 57.2% range with the 95% confidence interval, the output that "The first net is worse than the second" is in fact very misleading, since you are running SPRT(0, 35). If A is better than B, then B should be worse than A: It is an antisymmetric relation. You will notice while 163-176 fails the SPRT(0, 35), 176-163 does not pass, since the SPRT(0, 35) is not antisymmetric at all. I suggest you to use anti-symmetric ones such as SPRT(-13, 13) to figure out which is better.

roy7:
That's more like it. -p 50 validation, 5x64 vs 6x128 net2net, 92-91 so far.

roy7:
@isty2e Ahh thank you. I tend to forget the "is worse" message is talking about the 55% 35 ELO test, Not actually "worse" head to head. Maybe I'll change validation to use -13,13 like you say.

isty2e:
@roy7 It is not even actually testing if its winrate is 55% or higher for 400 games, the number 35 in the SPRT is pretty arbitrary here. Anyway, that is beyond the scope of this PR.

roy7:
Yes no worries. Thanks for your feedback. This is just a one time thing to assure myself net2net was working as expected.

roy7:
Final results of -p 50 5x64 vs 6x128 net2net when I stopped it, 265 wins, 246 losses.

alreadydone:
Forgive my obsession in math but I came up with another randomization. I suspect that the existing randomizations already make the enlarged networks trainable though, **even without noise** (all weights should already be distinct, though I don't know whether this suffices), and we probably don't need to widen anymore unless we transition from 128 filters to 196 or 256.

The idea is that w(j) can be chosen independently for each output channel h and each spot in the 3x3 convolutional kernel. I didn't point out that w(j) can also be allowed to depend on h, though this is a trivial observation and maybe already implemented? Even negative values of w(j) can be allowed, though I don't know whether it's advantageous to keep it close to 1/c_j so as not to affect the variance of the weights.

Since the layers are 3x3 convolutional, the q inputs are effectively 9q inputs, and we label the r-th entry in the j-th input channel by {j,r} (r=1,2,...9). Then for any function w(j,r,h) satisfying
Σ<sub>x such that g(x)=g(j)</sub>w(x,r,h)=1 for all j,r,h,
we can let
U<sup>(i+1)</sup><sub>{j,r},h</sub>=w(j,r,h)W<sup>(i+1)</sup><sub>{g(j),r},h</sub>
(cf. page 4 of paper)

(Just to clarify, the skip connection does not need to be divided by anything. The input/output of each layer of the enlarged network is supposed to be the same as the original network, only replicated into (different numbers of) copies.)

alreadydone:
What I said above is not completely correct; it only works if we are widening but not deepening. If we are only deepening but not widening, it seems inevitable to add noise to get rid of the completely 0 weights in the added block. But if we deepen and then widen, then the randomizations above apply, and we can expect a trainable enlarged network without noise (hence with identical output). The formula needs to be modified though.

First of all, to ensure that we get rid of all of the zeros, g(j) must hit each number in 1,2,..64 at least twice, so we must use g(j)=j for 1≤j≤64 and g(j)=j-64 for 65≤i≤128 (or something equivalent) not only for the inputs/outputs of the blocks but also for the intermediate between the two convolutions. To ensure that all weights are distinct, we pick distinct numbers w(j,r,h) for 1≤j≤64, 1≤r≤9 and 1≤h≤(number of output channels at that step). We then simply take
U<sup>(i+1)</sup><sub>{j,r},h</sub>=w(j,r,h)
U<sup>(i+1)</sup><sub>{j+64,r},h</sub>=-w(j,r,h)
These should ensure that the added layer(s) have distinct weight.

 (If noise is to be added, I think the new block should be added at the bottom (closest to the final output layer) to minimize accumulated errors.)

Ttl:
@alreadydone Interesting, but I think the current implementation is good enough as is. I don't think I'm going to implement it, but if you want you can code it yourself and submit a pull request.

alreadydone:
After the 6x128 network stalls and we use net2net to migrate to 10x128, I wonder whether it's a good idea to freeze the old blocks in training. This seems to be a common technique in transfer learning and a good way to make sure the new network doesn't forget too much when subjected to training even with a high learning rate; the feature extracted by the 6-block network should be useful to extract further features anyway. (I would add the 4 new blocks near the output and only allow weights in these blocks and in the policy/value heads to vary.) Also this helps to reduce the training time. When 10x128 stalls we could unfreeze to squeeze out the last bit of potential.

I wonder whether this strategy works for transferring knowledge between different komi and board sizes. It seems we could just copy all the weights in the residual blocks, freeze all (except maybe the last few blocks), and only train the fully-connected layers from random initialization; the quality of the self-play games should improve faster/more steadily than if the blocks are not frozen. If anyone tries this please let me know about the result.

Reference: https://www.analyticsvidhya.com/blog/2017/06/transfer-learning-the-art-of-fine-tuning-a-pre-trained-model/ Section "Ways to Fine tune the model"

Ttl:
I tried training a 8x128 network using net2net before and it seemed to work just fine without freezing weights. After 351k steps using 128 batch size I got the following result in validation:
```
101 wins, 75 losses
176 games played.
Status: 0 LLR 1.73266 Lower Bound -2.94444 Upper Bound 2.94444
```
Not statistically significant yet, but seems to be doing well considering this was the first network I tested and normally finding a better network seems to require testing many networks.

The previous issues with bootstrapping were probably caused by the shuffling bug.

bood:
@Ttl Are you using 5x64 network c83 self-plays to do the training? What is the learning rate are you using? The net2net paper seems to suggest using a 1/10 learning rate for the new generated network?

I agree that we should do some testing on net2net before 6x128 network stalls.

Ttl:
I started with 63498 network. Learning rate was 0.005 for 150k steps and after that I dropped it to 0.0005.

bood:
I see. But for batch size 128, shouldn't it be using 0.0025 and 0.00025 to match the current setting 512/0.005?

Ttl:
Yes, if you want to match to the current parameters.

bood:
Also tried net2net from 5x64 to 6x128. Trained with c83 self-plays.

Parameters:
Batch size: 128, Learning rate: 0.0025 for the first 152k moves 0.00025 for the rest

Current Results:
`./validation -g 1 -n ~/5b/c83e1b6e0ffbf8e684f2d8f6261853f14c553b29ee0e70ff6c34e87d28009c43 -n /root/6b/leelaz-model-352000.txt -k 352k | tee 352k.log`
312k.log:16 wins, 38 losses
352k.log:2 wins, 14 losses

Looks promising too.

alreadydone:
I read about CNN backpropagation and it seems that freezing 6 blocks of weights can indeed cut training time by 40%: there are roughly the same number of weights in a fc layer as in a convolutional layer (361x2x362 in policy head vs. 128x128x9), but the amount of computation in convolutional layers need to be multiplied by 361x361, so it dominates. Hope I didn't get it wrong... Training may not be the bottleneck though.

bood:
More match results about the 6x128 leelaz-model-352000.txt I trained:

`47 wins, 12 loses` (data from last post included)

Also, I tested some ladder scenarios with the tool mentioned in #905 , it shows this 6x128 performs better compared to ed0 which is trained by SL, still worse than c83 though (leelaz v0.12 1600 playouts):

c83|ed0|352k
---|---|---
51/70|13/70|23/70

alreadydone:
Is GPU memory what is limiting us from using the same batch size (2048) as in AGZ paper?  If so, [this work](https://github.com/openai/gradient-checkpointing) which trades off memory usage with time may be useful. Implementation in Tensorflow as well. A graph near the end shows the effect for 1-20 blocks with batch size 1280.
Eventually we'll switch to larger networks and self-play games will be generated more slowly, so a slight increase in training time seems acceptable. Shuffling may need to be further optimized to work with larger batch size, though.

alreadydone:
@Ttl I guess net2net doesn't currently support adding input planes. Would you mind adding in this functionality? I am considering adding a liberty plane to pass to the NN the number of liberties of the group containing each stone, and this also seems to be [Minigo's plan](https://www.reddit.com/r/cbaduk/comments/86dvib/minigo_will_be_at_the_nvidia_conference_in_san/dw6721s/). This helps bridge the gap between the local nature of convolutional filters and the global/long-range aspects of Go. When combined with tree search and 3x3 filters, this will allow the network to deal with (self-)atari, illegal moves, life-and-death, and ladder problems more efficiently (using less of its capacity, which is especially crucial for a small-sized network). From what I heard people expect LZ to easily beat top pros and rise to the tier of DeepZenGo and Fine Art once these problems are fixed. I think there's enough evidence that it's much more efficient to count liberties with an optimized algorithm than forcing the network to grasp it. Hopefully introducing this plane will boost the training accuracy and then the strength of the network through self-play.

@gcp Having become the strongest engine on everyone's call, nowadays Leela Zero is not just pure (reproducing) scientific research, but also a widely used tool for analyzing and learning Go; as you said, we are pushing the limit. What will benefit the most people is a network that is simultaneously strong and efficient (in computing resources), friendly to the environment, to weaker machines, and to mobile devices. Some people may still want to see the limit a Zero bot can eventually reach with a full-sized 20x256 or 40x256 network (expected to [read all meaningful ladders correctly](https://www.reddit.com/r/MachineLearning/comments/76xjb5/ama_we_are_david_silver_and_julian_schrittwieser/dolmac1/) with ? playouts). However, AFAIK there's currently no plan to switch to 20x256 because it's too slow, and even the strong 20x256 network 93229e still shows [ladder problems](https://github.com/gcp/leela-zero/issues/1113#issuecomment-376613086). I do have some [new ideas](https://www.reddit.com/r/baduk/comments/86a6qo/would_you_write_a_go_game_bot_if_it_was_earning/dweerkv/) to help bring in more computing power, though it will take some time and momentum to start running.

If most people stick to the Zero approach and don't agree to use a liberty-enhanced network for self-play, being able to add the plane through net2net at least allows interested people to train such a network themselves. Since the new input plane is not included in current training data, the liberties would better be calculated during training; hopefully some additions to the training code will also allow people to experiment with other input features.

Footnote: I don't see any point using separate planes for 1,2,3,... liberties as in the first AG paper; liberties are additive when you connect two groups (for this to work better, put the number of empty adjacent points for an empty point in the input plane), and AlphaZero also used only one plane for move count.

gcp:
You know you don't need net2net support for that right, and in fact using net2net for that seems extremely bad: quite a bit of the existing net is going to be trained to count liberties, which you're now making redundant. What would be the point of retaining that training?

You should just train a net from randomly initialized weights for that.

I have no interest in doing augmented approaches *here*. For sure they are interesting and may produce stronger bots. The code is open, you can fork it and try whatever you want.

alreadydone:
> quite a bit of the existing net is going to be trained to count liberties, which you're now making redundant.

Interesting point. I think net2net still serves as a smart initialization though, and the network will forget unnecessary incomplete knowledge about liberties and retain those that remains crucial after introducing the plane (e.g. using liberties to make decisions); at least net2net would be the first thing I try. It seems some people prefer redundancy and even introduce dropout layers which we don't have ...

I haven't looked into the training code yet, but for inference I think `gather_features` is the correct place to insert the plane. `count_rliberties` has been removed from FastBoard, and it looks like `m_libs` is counting the pseudo-liberties? I hope `count_rliberties` found in https://github.com/gcp/leela-zero/blob/5fb3b4d1d7b03b69e71e0729016078e4b3d32cd1/src/FastBoard.cpp still works ...

lightvector:
@alreadydone 
> Footnote: I don't see any point using separate planes for 1,2,3,... liberties as in the first AG paper; liberties are additive when you connect two groups (for this to work better, put the number of empty adjacent points for an empty point in the input plane), and AlphaZero also used only one plane for move count.

Yeah, I also expect a neural net would be able to make use of a single real-valued liberties plane decently well. But it's also not crazy to encode them as separate planes. Encoding it additively in a single plane is sort of like a weak prior that the relevant things the neural net wants to compute should vary smoothly with that value, encoding it in separate planes is sort of like a weak prior that each value behaves distinctly, either way the neural net can learn from data to override that prior. I'm not actually sure which one would perform better. You should try it and compare!

By the way, my experience is that for the policy part of the net at least, only atari, 2, and 3 liberties are important for the neural net to understand, after that you get heavy diminishing returns. For example, when I trained a neural net with about 5 blocks with a typical set of non-"zero" features (liberties, ladder status, etc), experimentally I tried removing single input planes *without* retraining the neural net to see how much unexpectedly losing each input plane alone harmed the prediction accuracy.

The baseline test set accuracy was 52.18%. Losing "on the board", "opponent stone", and "own stone" massively dropped the accuracy to 6.19%, 5.16%, and 4.89%, respectively - obviously these are all critical features. Losing "own group has 1 liberty" and "enemy group has 1 liberty" dropped it to 46.93% and 48.19%. Losing "opp group has 2 liberties" and "enemy group has 2 liberties" dropped it to 50.16% and 50.10%. For 3 liberties, it was 51.45% and 51.40%. And for 4 liberties, it was 52.01%, and 52.23% - so by the time you got to the 4 liberties feature, it was hardly using it at all anyways and barely noticed whether you included the feature or not.

Ttl:
I'll see if I can add increasing the number of input planes functionality. I'm not sure if it makes sense to use it or not, but I don't think it's that hard to implement.

PR #711, opened 2018-01-21
Always copy GameState passed into think or ponder

### LINKED ISSUE #690: time_settings doesn't work
It can set time correctly for the fist time, but the engine doesn't respond to `time_settings` command if I want to change time settings again.   
It seems that this problem occurs in Leela Zero 0.11 only.

killerducky:
Fixes #690.

I hope Travis is ok with doing a search, I guess it will have to do a tuning also.


killerducky:
Travis is still failing, I guess I need to do something different with the path?
`Could not open weights file: ../latest_5x64.txt`

Or maybe it should download a weights file from somewhere? I'm not sure how to do that so I tried the easy way of adding it to the git repo.

src/UCTSearch.h
```diff
@@ -28,6 +28,7 @@
 #include "GameState.h"
 #include "KoState.h"
 #include "UCTNode.h"
+#include "../gtest/googletest/include/gtest/gtest_prod.h"
```
gcp: Ugh no, definitely don't add a GTest dependency into the normal codebase itself. I'd rather add a getter or a function that returns the maximum amount of time a search will run.

src/UCTNode.cpp
```diff
@@ -424,26 +424,32 @@ UCTNode::node_ptr_t UCTNode::find_new_root(const int move) {
...
-        for (auto& child : m_children) {
-            auto move = child->get_move();
-            if (g_new.get_last_move() == move) {
-                g_curr.play_move(move);
-                if (g_curr.board.get_hash() == g_new.board.get_hash()) {
-                    return std::move(child);
+// Also copies the GameState.
+void UCTNode::find_new_root(const GameState& g_new,
+                            GameState& g_curr,
+                            node_ptr_t& root) {
```
gcp: Are there any cases where root wouldn't be this?

gcp:
>Or maybe it should download a weights file from somewhere?

Prefer to include a minimal one in the repo for testing.

killerducky:
@gcp I will update this and use a smaller network. Do you know the portable way to have the gtest find the weights file? 

https://travis-ci.org/gcp/leela-zero/builds/331510253#L2855

Right now I'm doing this: `cfg_weightsfile = "../latest_5x64.txt";` It works locally but not on Travis. Looks like I could do "/src/latest_5x64.txt" but that won't work in a local build. 

There must be something in cmake to handle this I guess?

src/UCTNode.cpp
killerducky: Not sure what you're asking. This function used to return the new root, but I refactored it to include the case were the root is actually the same. I couldn't figure how to get it to return a pointer to itself, it was segfaulting.

So now UCTSearch passes it's current root pointer, and this function puts the new one in if needed.

bood:
@killerducky For now, any tests with GPU support will fail on CI since there is no GPU support there.  I'll need to modify it to test with CPU only build.

src/tests/gtests.cpp
```diff
@@ -149,3 +146,25 @@ TEST_F(LeelaTest, KoSqNotSame) {
     // But ko (square) is not
     EXPECT_NE(hash, maingame.board.get_hash());
 }
+
+TEST_F(LeelaTest, TimeControl) {
+    // Initialize network with a small 32x1 network.
+    cfg_weightsfile = "../test_32x1.txt";
```
killerducky: @bood ok thanks for looking into the GPU issue. But another issue is this file cannot be found by Travis. It works in my local area because I do `mkdir build; cd build; make` and the file is in the dir above. On Travis the build directory is somewhere else so this doesn't work.
https://travis-ci.org/gcp/leela-zero/builds/331628882#L2888

bood: You can change the Dockerfile to do similar thing as well. Use following lines to replace `RUN CXX=g++ CC=gcc cmake CMakeLists.txt`:
```
RUN mkdir -p build
WORKDIR /src/build/
RUN CXX=g++ CC=gcc cmake ..
```

But I suggest you put the weight file somewhere else rather than just drop it in root dir, e.g. src/tests/weights/

killerducky: What does this code do?

```
RUN mkdir -p /src/
WORKDIR /src/
COPY . /src/
```

It looks like the copy will be from/to the same place?

bood:
What `COPY` does is simply copy the git repo dir into the src dir, didn't
try but maybe it can work without this too.

roy7:
Watching the Travis CI build live was interesting (from the Details link). Appears it fails because there's no GPU, is only error left?

Some various compiler warnings I don't get on my own version of g++ when I compile.

killerducky:
Ok well it finally found the weight file at least!

@Bood how will you fix it to do CPU only on Travis? Do you need me to make more changes to this pull?

I will move the weights to a better directory once it actually passes. I don't want to generate more noise until then.

bood:
You can always signup your own travis account and play with this there.

For GPU issue, I'll submit a PR later, but I'll have to sync with @gcp
before I go.
My plan is to deprecate Makefile and use CMake only - all configs are now
hard coded in `config.h` which makes it hard to do multiple builds with
different configs. CMake has this excellent tool called [configure_file](
https://cmake.org/cmake/help/v3.10/command/configure_file.html) which
handles exactly this kind of thing. Or I can try live with the old config
by just not using it first?

src/UCTNode.cpp
gcp: I mean the call is m_root->call(m_root) which suggests that m_root->call() is possible and inside you can do *this = blah.

killerducky: Aha, I will try that way tonight.

gcp:
@bood if you're working on this then note that it seems we build the gtests on Windows, but don't run them. It would be cool if they ran!

gcp:
I'm going to close this because the fix will be merged in another PR, and I assume we'll try a followup with tests.

PR #718, opened 2018-01-22
Always copy GameState passed into think or ponder (fix only no gtest)

killerducky:
Fix only version of #711.

Might take awhile to get Travis working so here is the bug fix only.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -424,26 +424,32 @@ UCTNode::node_ptr_t UCTNode::find_new_root(const int move) {
 }
 
 // Use this version if the child could be anywhere.
-// Also updates the GameState.
 // Gives up after searching the direct children.
-UCTNode::node_ptr_t UCTNode::find_new_root(const GameState& g_new,
-                                           GameState& g_curr) {
-    if (m_has_children) {
-        for (auto& child : m_children) {
-            auto move = child->get_move();
-            if (g_new.get_last_move() == move) {
-                g_curr.play_move(move);
-                if (g_curr.board.get_hash() == g_new.board.get_hash()) {
-                    return std::move(child);
+// Also copies the GameState.
+void UCTNode::find_new_root(const GameState& g_new,
+                            GameState& g_curr,
+                            node_ptr_t& root) {
+    if (g_new.get_komi() != g_curr.get_komi()
+        || g_new.board.get_hash() != g_curr.board.get_hash()) {
+        if (m_has_children) {
+            for (auto& child : m_children) {
+                auto move = child->get_move();
+                if (g_new.get_last_move() == move) {
+                    g_curr.play_move(move);
+                    if (g_curr.board.get_hash() == g_new.board.get_hash()) {
+                        root = std::move(child);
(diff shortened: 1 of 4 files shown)


src/UCTSearch.h
```diff
@@ -76,6 +76,7 @@ class UCTSearch {
 
     UCTSearch();
     void set_gamestate(const GameState& g);
+    GameState get_gamestate() const;
```
gcp: You aren't using this as far as I can see? (And if you were, shouldn't it return a ref?)

killerducky: It's used by the gtest version. I will remove it from this pull and have it only in the gtest pull version.

roy7:
Love you all if we can finish this one up. :)

killerducky:
I couldn't get it to work. Relevant code:
```
void UCTNode::find_new_root(const GameState& g_new,
                                  GameState& g_curr) {
    node_ptr_t root = nullptr;
    root = std::move(child);
    *this = root;


/home/aolsen/projects/lz-tc/leela-zero/src/UCTNode.cpp: In member function ‘void UCTNode::find_new_root(const GameState&, GameState&)’:
/home/aolsen/projects/lz-tc/leela-zero/src/UCTNode.cpp:454:17: error: no match for ‘operator=’ (operand types are ‘UCTNode’ and ‘UCTNode::node_ptr_t {aka std::unique_ptr<UCTNode>}’)
         *this = root;
```

killerducky:
This is ready, unless someone can help me get an alternate way of doing find_new_root working. My Google-fu for "std::move this" didn't find anything helpful.

bood:
#725 is almost ready now, you can try your unit test code again after it's merged.

gcp:
I tried to solve the f->x(f, ...) issue by making find_new_root part of UCTSearch, which actually simplifies things a lot, but then you run into the problem that it fundamentally needs to alter the children of an UCTNode, so it can't do that with the const ref get_children() returns. And it's a sign that UCTSearch shouldn't doing that kind of mucking around too.

src/UCTNode.cpp
```diff
@@ -424,26 +424,32 @@ UCTNode::node_ptr_t UCTNode::find_new_root(const int move) {
...
+                        root = std::move(child);
+                        break;
+                    }
+                    g_curr.undo_move();
                 }
-                g_curr.undo_move();
             }
         }
+        // No match. Create a new root.
+        root = std::make_unique<UCTNode>(FastBoard::PASS, 0.0f, 0.5f);
```
gcp: Isn't this in the wrong place now? If komi or hash mismatches, the root won't be reset. If a move matches, we'll break out of the for and overwrite here?

gcp:
I reworked this a bit. I can't say why, but I prefer x = x->f(...) over x->f(x, ...).

It also makes the signature and usage pattern of both calls match.

src/UCTNode.cpp
killerducky: Oops good catch. I would have found this if I retested self-play. I'll work on a gtest for these cases later.

Issue #701, opened 2018-01-20
Segfault occurs when illegal moves are played


korreman:
The segfault occurred in a human vs. leelaz match, after asking leelaz through GTP to generate a move. [backtrace.txt](https://github.com/gcp/leela-zero/files/1649184/backtrace.txt)

The segfault can be reproduced on my own machine by reusing the set of moves. Moveset and gdb script: [failure.txt](https://github.com/gcp/leela-zero/files/1649178/failure.txt), [failure_gdb.txt](https://github.com/gcp/leela-zero/files/1649180/failure_gdb.txt)

leelaz output: [segfault_log.txt](https://github.com/gcp/leela-zero/files/1649177/segfault_log.txt)

Compiled from source on Arch Linux, commit 47ddf38327b80ec5f2303344d1d9c65e25f5a3fb.

gcp:
```
Passes: 0            Black (X) Prisoners: 0
Black (X) to move    White (O) Prisoners: 0

   a b c d e f g h j k l m n o p q r s t
19 . . . . . . . . . . . . . . . . . . . 19
18 . . . . . . . . . . . . . . . . . . . 18
17 . . . . . O . . . . . . . . O . . . . 17
16 . . . O . . . . . + . . . . . + O . . 16
15 . . . . . . . . . . . . . . . . . . . 15
14 . . O . . . . . . . . . . . . . . . . 14
13 . . . . . . . . . . . . . . . . . . . 13
12 . . . . . . . . . . . . . . . . . . . 12
11 . . . . . . . . . . . . . . . . . . . 11
10 . . . + . . . . . + . . . . . + . . . 10
 9 . . . . . . . . . . . . . . . . . . .  9
 8 . . . . . . . . . . . . . . . . . . .  8
 7 . . . O . . . . . . . . . . X . . . .  7
 6 . . . . . . . . . . . . . . . . . . .  6
 5 . . . . . X X . . . . . . . . . . . .  5
 4 . . . X . O X O . +(O)X . . . X . . .  4
 3 . . X . . X O . . . O . X . . . . . .  3
 2 . . . O . X O . . . . . . . . . . . .  2
 1 . . . . . . . . . . . . . . . . . . .  1
   a b c d e f g h j k l m n o p q r s t

Hash: D07F6BE22364BDE8 Ko-Hash: 7E7889453F751494
```

play b g5

[...]

PR #720, opened 2018-01-22
#701: Do not allow the stone to be placed in an already occupied square.

akdtg:
Fixes issue #701.


gcp:
I did a bit bigger cleanup including a test in https://github.com/gcp/leela-zero/commit/4a95913af0785083a0ace4fe7ec79b8f6b119323

PR #721, opened 2018-01-22
Build autogtp with cmake

thynson:
We just need one build system.


gcp:
@marcocalignano IIRC you added CMake support in the first place so I assume this works for you?

marcocalignano:
Checking out this repo I get the following error. Also be sure to create the _moc files for the QT application.

CMake Error at CMakeLists.txt:20 (add_subdirectory):
  The source directory

    /home/marcuzzo/workspace/thyson/leela-zero/gtest

  does not contain a CMakeLists.txt file.

gcp:
You need to check out the gtest submodule, see README.md.

thynson:
@marcocalignano CMake is able to create _mac files. see https://github.com/gcp/leela-zero/pull/721/commits/891298ac6708fd43cc637169af6c4b5dcab3222a#diff-b4d50ce2ddbfc158ba14cecb49d9991fR9

marcocalignano:
@gcp I always forget that part ;)
@thynson since you are at it add also a cmake for the validation program.

thynson:
@marcocalignano Didn't notice that target, will add it soon.

marcocalignano:
@thynson I would like to have a check for the QT 5 version it has to be at least 5.3 see autogtp project file otherways QAtomicInteger is not supported.

marcocalignano:
I still see a couple of problems:
First the required QT version should go only in the autogtp cmake. Leelaz does not need Qt. I should be able to compile only leelaz even if I do not have Qt. 
Second now I'd like to have a make install option so I do not have to go down the build tree to look for the binaries. 
I am still a fan of qmake that omit the option (introduced from leelaz) "-ffast-math -flto -march=native" and uses -std=gnu++1y instead of  -std=c++14 and  -std=gnu++11. But I guess I am ok with this it the autogtp.pro file stays in the repository.

thynson:
Yeah, it's better to make it optional to build autogtp and validation, and disable them if Qt is not found

PR #722, opened 2018-01-22
OpenCl Network some math optimizations

marcocalignano:


gcp:
I don't measure any speed difference from this at all.

marcocalignano:
I get 4% on my fast GPU and almost 8% on the slower one (tested with 6x128).

PR #725, opened 2018-01-22
Configurable CMake (Enables CPU build and tests on CI)

bood:
This enables:
1. CPU build verifications on CI
2. Unit test with CPU build on CI (Travis for now, Windows later), as discussed in #711 
3. Build with different feature without modify code

Since `config.h` is auto-generated, this will break the old build system (Makefile) so I removed the `Makefile`.

Windows build works on Appveyor too, but I'm not sure how to make it work on local machines (I don't have one), @gcp Could you help find out and update the Windows part of README later? It's mainly about how to setup Boost and pass BOOST_ROOT to CMake, unfortunately nuget version of Boost does not work well with CMake (some hack to CMake file needed)

(I'll go to sleep now so I'll check any feedbacks tomorrow.)


gcp:
This will need rebasing because #721 was merged.

marcocalignano:
I also do not like the config.in method. The one I have suggested [here] (https://github.com/gcp/leela-zero/commit/85062a2961ea7be7d228901b790b41023a924e10#commitcomment-27022326) is more cmake style

bood:
are you sure？configure file seems to be an elegant way doing this, it is a
cmake provided facility too, I cannot see what makes it less cmake style.

@gcp besides rebase, do you want me to change to use your method too? do
you want me to keep the old makefile too? im not a fan of keeping two build
system for one thing though

2018年1月23日 03:02，"marcocalignano" <notifications@github.com>写道：

bood:
I can also update the openblas version in Appveyor, just tell me which version do you want

gcp:
>im not a fan of keeping two build system for one thing though

The issue is this: I don't use CMake on Linux (ok, actually I do use it sometimes, especially for tests, but I mostly just run make), I don't use CMake on Windows, and I don't use the shipped projects on Windows because of the Nuget / versions issue.

On Windows I build via Qt which means qmake, not CMake or the MSVC project (I suspect for marco it's the same on Linux given some past comments).

From my perspective we bundle CMake and the MSVC projects for the benefit of people that don't know how to set up MSVC with include directories, compile the latest versions of the dependencies, etc, and because it's probably more likely to work on configurations we didn't anticipate, and CI.

So I'm not asking you to maintain the old systems - it will happen automatically, because I won't merge things that break my ability to make releases with updated libs. But I will ask you not to do things that make maintaining and switching to another build system more complicated, and the config.in thing is just that.

bood:
ok, I just reverted some of the changes that breaks Makefile.
Also updated gtest to use the latest master commit so unit test on Travis should is working now.

Issue #687, opened 2018-01-20
Is 18th feature plane required? 


Neverever:
## 17th feature plane
From AlphaGo Zero paper:
>  the colour feature C is necessary, because the **_komi_** is not observable

The 17th feature plane is indicating whether komi is given to the opponent. (Black=1, White=0)

## 18th feature plane
The 18th feature plane is added due to

> slight imbalance in that it is easier for the **_white_** player to see the board edge (due to how padding works in neural networks).

AFAIK
the current player's stones are marked as `1` in feature plane 1 to 8
the opponent players stones are marked as `1` in feature plane 9 to 16
There isn't really a concept of black/white players, only myself/opponent

If AlaphaGo Zero's initial design help white player easier to see the board edge, why black is not seeing it? (both players' stones are marked as `1`)

FFLaguna:
I'm also interested in a bit longer of an explanation than I've seen to date on this. I may have missed it somewhere.

lightvector:
When convolutions go past the edge of the board, due to zero-padding it treats it as if the values there were zeros. The presence of a layer that is all ones on the whole board helps the neural net tell what empty spots are on the board versus off the board, which is important for understanding territory and tactics near the edges. Without the 18th feature, only white would have access to such a layer, black would have to implicitly infer the board edge from the fact that nobody has stones there and from the fact that deeper down, convolution channels keep getting "zeroed out" past that edge.

Neverever:
If the Network is rely on the “ones” layer to determine edge of the board
then back in “17 features” days, it’s Black player has the advantage to detmine the edge. 

From AlphaGo Zero paper
> The final feature plane, C, represents the colour to play, and has a constant value of either 1 if black is to play or 0 if white is to play

What/When was the initial observation that White has advantage?
If anyone has issue reference, I’m more than happy to get my hands dirty and dig into this.

lightvector:
Ah, you're right, I got the colors flipped. Yes, it would be black that can see the edge more easily, not white. But the point still stands, either way there will be a weird asymmetry between the colors if you only have 17 and only one of the players gets a plane of 1s.

Neverever:
The 17th plane was designed for Komi indication. 
Whether it’s used by Network internally to determine the board edge? It’s beyond my knowledge. 

The project goal was to replicate AlphaGo Zero
With additional 18th plane, I feel it’s not aligned with the project goal. 

[...]

PR #727, opened 2018-01-22
Wrong advantaged color

TFiFiE:
As per the observation made [here](https://github.com/gcp/leela-zero/issues/687#issuecomment-359142281) (in #687).


gcp:
We fixed this before and it was regressed because another commit completely randomly changed it: https://github.com/gcp/leela-zero/commit/f94392eac63c73806cd7d293a1ec3e4d87d16f79

PR #733, opened 2018-01-23
This allow more parameter to validation.

marcocalignano:
It seems that this feature are needed right now.

With this PR validation can:

- accept different `leelaz` binaries 

- accept different options for the `leelaz` binary

- accept different SPRT hypothesis

### DIFF
--- a/validation/SPRT.cpp
+++ b/validation/SPRT.cpp
@@ -27,194 +27,194 @@ class SprtProbability;
 
 class BayesElo
 {
-	public:
-		BayesElo(double bayesElo, double drawElo);
-		BayesElo(const SprtProbability& p);
+    public:
+        BayesElo(double bayesElo, double drawElo);
+        BayesElo(const SprtProbability& p);
 
-		double bayesElo() const;
-		double drawElo() const;
-		double scale() const;
+        double bayesElo() const;
+        double drawElo() const;
+        double scale() const;
 
-	private:
-		double m_bayesElo;
-		double m_drawElo;
+    private:
+        double m_bayesElo;
+        double m_drawElo;
 };
 
 class SprtProbability
 {
-	public:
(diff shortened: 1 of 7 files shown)


jkiliani:
Does this change anything about the zero win/loss issue @roy7 brought up?

marcocalignano:
@jkiliani no! I am looking at MAXsprt vs SPRT to see if we can use that. The proble is the win/loss is still zero the LLR go to infinity so you get that same result if it is 1-0 or 1000-0

marcocalignano:
I didn't work out all the math but I think the log function are our friends in this one.

marcocalignano:
Ok looking at the math I came up with a termination rule if one network wins or looses only.
[]<a href="https://www.codecogs.com/eqnedit.php?latex=\inline&space;Wins&space;>=&space;e^{UBound}&space;~||~&space;Losses&space;>=&space;e^{-LBound}" target="_blank">

marcocalignano:
Ok looking at the equation we can hav an stop condition when (speacking respect to the first network)

```Wins >= e^(UBound) || Losses >= e^(-LBound)```

**Comment by marcocalignano** [comment]
So @jkiliani and @roy7 here is the stop condition. I hope you can give it a look.

File: validation/SPRT.cpp [hunk 1]
```diff
@@ -166,10 +166,18 @@ Sprt::Status Sprt::status() const
 		0.0,
 		0.0
 	};
-
-	if (m_wins <= 0 || m_losses <= 0 || m_draws <= 0)
-		return status;
-
+    status.lBound = std::log(m_beta / (1.0 - m_alpha));
```
gcp: Indentation seems to be totally off here.

marcocalignano:
Since I took the code from another project there was still TAB inside and that messed up the indentation.
Now should all be clean up.

PR #737, opened 2018-01-24
Tests for TimeControl.

killerducky:
Will add some more test next. Just want to make sure Travis is working for this now.

Bug fixes will go in a separate pull since the test flow is still being worked out.


killerducky:
@bood AppVeyor is dying because the path to 0k.txt is wrong. Can you help fix that?

I added a second commit which catches the error. 

I tried to fix the bug itself but I'm going in circles. The problem is GTP and UCTSearch each have their own GameState objects, with their own TimeControl objects. GTP modifies the time settings (m_maintime etc), and UCTSearch modifies the timers (m_remaining_time etc). It seems like GameState.m_timecontrol should be changed to be a pointer or reference. But I ran into problems figuring out how to get GameState to do the correct copy.

bood:
I believe this fixes the path issue:

https://github.com/bood/leela-zero/commit/1764f2656abeb4078512e31355e18d091da2370e

killerducky:
The tests I added have a problem because they create their own Search object. But GTP::execute has it's own separate copy of Search. I'll have to work on this again tomorrow.

bood:
The path issue will be fixed in #741

gcp:
The problem is GTP and UCTSearch each have their own GameState objects, with their own TimeControl objects.

It's this change that is biting us in the butt right now:
https://github.com/gcp/leela-zero/commit/081bd715bdc2a8ab035f992253e8efd3597338ec#diff-a9f075c2e1a9147471c5ca80794f993fL94

bood:
BTW does it make sense to split the unit tests into different files? One file per module would be ideal.

gcp:
I think I have a fix, doing some local testing and I'll make a separate pull.

killerducky:
Ok good. I woke up early so I will focus on fixing the test in this pull and leave the bug fix to you.

killerducky:
I think this is ready. Locally I merged your fix and the test passes.

src/tests/gtests.cpp
```diff
@@ -174,3 +174,74 @@ TEST_F(LeelaTest, MoveOnOccupiedSq) {
     // Find this error in the output
     EXPECT_NE(output.find("illegal move"), std::string::npos);
 }
+
+
+TEST_F(LeelaTest, TimeControl) {
+    // Initialize network
+    cfg_weightsfile = "../src/tests/0k.txt";
```
gcp: I think you could make a Test Fixture for this, other tests will surely find it useful too.

src/tests/gtests.cpp
```diff
@@ -174,3 +174,74 @@ TEST_F(LeelaTest, MoveOnOccupiedSq) {
...
+    Network::initialize();
+    cfg_max_playouts = 1;
+    cfg_num_threads = 1;
+    std::string output;
+
+    // TODO I tried to make these separate tests,
+    // but it died locally when leelaz ran the second time.
```
gcp: Might want to figure out what's up there or we risk not being able to add any other tests...

src/tests/gtests.cpp
```diff
@@ -174,3 +174,74 @@ TEST_F(LeelaTest, MoveOnOccupiedSq) {
...
+        printf("output\n%s\noutput", output.c_str());
+        EXPECT_NE(output.find("Black time: 00:01:40"), std::string::npos);
+        EXPECT_NE(output.find("White time: 00:01:40"), std::string::npos);
+
+        GTP::execute(maingame, "go");
+        testing::internal::CaptureStderr();
+        GTP::execute(maingame, "showboard");
+        output = testing::internal::GetCapturedStderr();
+        printf("output\n%s\noutput", output.c_str());
+        EXPECT_NE(output.find("Black time: 00:01:38"), std::string::npos);
```
gcp: Pff, what if the VM this inevitably runs on in CI is slow to respond?

I'd just check if the clock has moved from the starting value, not the exact new one.

src/tests/gtests.cpp
```diff
@@ -174,3 +174,74 @@ TEST_F(LeelaTest, MoveOnOccupiedSq) {
...
+        output = testing::internal::GetCapturedStderr();
+        printf("output\n%s\noutput", output.c_str());
+        EXPECT_NE(output.find("Black time: 00:01:40"), std::string::npos);
+        EXPECT_NE(output.find("White time: 00:01:40"), std::string::npos);
+
+        GTP::execute(maingame, "go");
+        testing::internal::CaptureStderr();
+        GTP::execute(maingame, "showboard");
+        output = testing::internal::GetCapturedStderr();
+        printf("output\n%s\noutput", output.c_str());
```
gcp: Debugging printf, can go.

gcp:
Instead of letting the CI machines run for 1.5 seconds, what about setting up a Canadian byo yomi timecontrol, doing a 1 playout search, and checking afterwards that the number of stones remaining has decreased?

gcp:
Might want to rebase with my patch so at the end the CI passes on this pull.

killerducky:
Rebased, it finally passes! I will work on the other cleanups tonight.

killerducky:
Debugging it as two tests instead of one, I narrowed it to hanging on OpenClScheduler.cpp::forward f.get(). I also noticed the second test is getting NNCache hits. So it seems some global objects are not being deleted/rebuilt between tests. Maybe we need some destructor in LeelaTest to delete some of them?

@bood any ideas?

bood:
`GameState` is shared between your two tests, `NNCache` does not appear to be cleared with `clear_board` . Even if you split into two tests, it's still the same object, does not change any thing.

I assume `NNCache.resize(0)` would clear all the caches.

src/tests/gtests.cpp
killerducky: Moved to init_global_objects.

killerducky: Actually moved to LeelaEnv.

src/tests/gtests.cpp
killerducky: Fixed, but see main comments.

src/tests/gtests.cpp
killerducky: done

src/tests/gtests.cpp
killerducky: Using Canadian method instead now.

killerducky:
Ok this is ready for review again. I fixed the globals issue by creating LeelaEnv and only init them once. The advice I read is to dependency injections instead of globals and/or singletons. But I don't want to mess with that for now.

I also created expect_regex to simplify text scraping. Google has gmock which could do the same thing if we want to pull that in as well: `EXPECT_THAT(output, HasSubstr("Black time: 00:01:40"));` and `EXPECT_THAT(output, ContainsRegex("[a-z]blah"));`

src/tests/gtests.cpp
```diff
@@ -174,3 +198,61 @@ TEST_F(LeelaTest, MoveOnOccupiedSq) {
...
+    // clear_board to force GTP to make a new UCTSearch.
+    // This will pickup our new cfg_* settings.
+    GTP::execute(maingame, "clear_board");
+
+    // Absolute time 100s = 1.37s per move in opening.
+    // Enough to be visible on display_times
+    GTP::execute(maingame, "kgs-time_settings canadian 0 120 25");
+    testing::internal::CaptureStderr();
+    GTP::execute(maingame, "showboard");
+    output = testing::internal::GetCapturedStderr();
```
bood: I'm strongly against using internal APIs though, they are private for a reason.

killerducky: I was following the style @gcp used, so let's see what his opinion is. Looks like you can do is by redirecting stdout/stderr to a stringstream:

https://stackoverflow.com/questions/3803465/how-to-capture-stdout-stderr-with-googletest

I haven't tried it yet.

gcp: I basically searched on how to do it and most examples were using that internal API.

src/tests/gtests.cpp
```diff
@@ -174,3 +198,61 @@ TEST_F(LeelaTest, MoveOnOccupiedSq) {
...
+// Basic TimeControl test
+TEST_F(LeelaTest, TimeControl) {
+    auto maingame = get_gamestate();
+    std::string output;
+    // clear_board to force GTP to make a new UCTSearch.
+    // This will pickup our new cfg_* settings.
+    GTP::execute(maingame, "clear_board");
+
+    // Absolute time 100s = 1.37s per move in opening.
+    // Enough to be visible on display_times
```
gcp: Obsolete comment.

src/tests/gtests.cpp
```diff
@@ -174,3 +198,61 @@ TEST_F(LeelaTest, MoveOnOccupiedSq) {
...
+    GTP::execute(maingame, "kgs-time_settings canadian 0 120 25");
+    testing::internal::CaptureStderr();
+    GTP::execute(maingame, "showboard");
+    output = testing::internal::GetCapturedStderr();
+    expect_regex(output, "Black time: 00:02:00, 25 stones left");
+    expect_regex(output, "White time: 00:02:00, 25 stones left");
+
+    GTP::execute(maingame, "go");
+    testing::internal::CaptureStderr();
```
gcp: This capture comes too late, which causes the search to get dumped to screen in-between the tests. (You can see this by running the test locally)

I would rather avoid this, i.e. capture all engine output during the tests, so you get the clean output of gtest only.

killerducky:
@bood I couldn't get this way to work. The printf's showed blank strings, and all output was going to my terminal instead of being captured.
```
    std::pair<std::string, std::string> gtp_execute(std::string cmd) {
        std::stringstream buf_cout;
        std::stringstream buf_cerr;
        std::streambuf * save_cout = std::cout.rdbuf(buf_cout.rdbuf());
        std::streambuf * save_cerr = std::cerr.rdbuf(buf_cerr.rdbuf());
        GTP::execute(get_gamestate(), cmd);
        printf("gtpexe cout: %s\n", buf_cout.str().c_str());
        printf("gtpexe cerr: %s\n", buf_cerr.str().c_str());
        std::cout.rdbuf(save_cout);
        std::cerr.rdbuf(save_cerr);
        return std::make_pair(buf_cout.str(), buf_cerr.str());
    }
```

src/tests/gtests.cpp
killerducky: not working yet. See main thread.

src/tests/gtests.cpp
killerducky: deleted.

src/tests/gtests.cpp
killerducky: fixed with new utility function.

src/tests/gtests.cpp
```diff
@@ -36,33 +37,55 @@
...
+
+        NNCache::get_NNCache().set_size_from_playouts(cfg_max_playouts);
+
+        cfg_weightsfile = "../src/tests/0k.txt";
+        Network::initialize();
+    }
+    void TearDown() {}
+};
+
+::testing::Environment* const leela_env = ::testing::AddGlobalTestEnvironment(new LeelaEnv);
```
gcp: Is there some way to use this without making dummy vars and relying on the initialization for side effects?

killerducky: Looks like we would have to write our own main. I'll keep that in mind but I don't see myself getting to it soon.

https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#global-set-up-and-tear-down

> However, we strongly recommend you to write your own main() and call AddGlobalTestEnvironment() there, as relying on initialization of global variables makes the code harder to read and may cause problems when you register multiple environments from different translation units and the environments have dependencies among them (remember that the compiler doesn't guarantee the order in which global variables from different translation units are initialized).

PR #742, opened 2018-01-24
Make UCTSearch take a reference to the game state.

gcp:
UCTSearch originally took a reference to the current game state. Tree
reuse made it keep its own copy, which is now causing problems to keep
in sync. Revert that and make it take a reference again. We copy the
last GameState at the end of the search and use the copy to find back
the right UCTNode root at the beginning of the next search.

Fixes issue #690.


gcp:
@killerducky Want to have a look?

killerducky:
Quick glance it looks good. m_last_rootstate, nice solution.

PR #745, opened 2018-01-25
Removing kill_superko and let play_simulation invalidate those nodes

sethtroisi:
Pros:
* less code to maintain, consolidates all superko logic to one code location
* eases one hard part of [small_node pull request](https://github.com/gcp/leela-zero/compare/next...sethtroisi:v__small_node#diff-28216771f77a09f4e0782ad5404fdd28R143)

Cons:
* less semantically correct when Dirichlet noise is applied (extra UCTNodes can exist which increases  sample_sum and has an effects on normalization)

Code performs largely the same after removal.
Any if node that would have been removed is explored it get's [invalidated](https://github.com/gcp/leela-zero/blob/next/src/UCTSearch.cpp#L95) and that [playout is ignored](https://github.com/gcp/leela-zero/blob/next/src/UCTSearch.cpp#L363)

Results from competition are even (as expected):

Fast computer with 1600 playouts
```
leelaz-next v leelaz-no-superko (70/70 games)
board size: 19   komi: 7.5
leelaz-next           33 47.14%       19 54.29%     14 40.00%    179.45
leelaz-no-superko     37 52.86%       21 60.00%     16 45.71%    179.72
                                      40 57.14%     30 42.86%
```

Slow Computer with 1000 playouts
```
leelaz-next v leelaz-no-superko (50/50 games)
board size: 19   komi: 7.5
leelaz-next           22 44.00%       11 44.00%     11 44.00%    191.84
leelaz-no-superko     28 56.00%       14 56.00%     14 56.00%    203.58
                                      25 50.00%     25 50.00%
```


gcp:
This seems like a very dangerous change that makes keeping the code correct a lot harder. You now have to make sure that you never pick (or display!) moves that have had no exploration, and hence a chance to be invalidated.

It's very unclear that invariant holds, .e.g: https://github.com/gcp/leela-zero/blob/47ddf38327b80ec5f2303344d1d9c65e25f5a3fb/src/UCTSearch.cpp#L238

It's going to be even more annoying to ensure it keeps holding. dump_stats will display illegal moves as candidates (it also displays them in the PV, but that seems less bad). If we add further analysis at some point, you'll again have the problem that you'll have to remember there might be illegal moves at the root.

sethtroisi:
Those are valid reason to not like this. I'll close and see if I can clean up the smaller node code more easily.

PR #746, opened 2018-01-25
Very simple test set for training

Ttl:
I saw some discussion about the test set for training. I've been missing it too so here's a simple implementation.

parse.py now takes optional "--test" argument for path to chunk to be used as a test set. Symmetry is still applied to it and it loops around during the testing so the tested positions are not identical. I'm not expert in tensorflow and didn't find a quick method to solve those, but this should still be better than nothing.

### DIFF
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -286,6 +286,7 @@ def main():
 
         testset = tf.data.Dataset.from_generator(
             parser_test.parse_chunk, output_types=(tf.string))
+        testset = testset.shuffle(4096)
         testset = testset.map(_parse_function)
         testset = testset.batch(BATCH_SIZE)
         testset = testset.prefetch(4)
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -16,11 +16,10 @@
 #    You should have received a copy of the GNU General Public License
 #    along with Leela Zero.  If not, see <http://www.gnu.org/licenses/>.
 
-import sys
+import argparse
 import glob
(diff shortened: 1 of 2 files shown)


gcp:
See #747. It's similar but I split the chunk list and use a TF conditional to select the right feeder.

Ttl:
Looks better than this.

Does the splitting work correctly when training from supervised data? When chunking from supervised data leelaz loops over the dataset several times so positions from games get split over several chunks. When the split is done positions from same games might end in both sets.

gcp:
FWIW I don't think shuffling the test set is needed, you don't care about the correlation during testing.

gcp:
>Does the splitting work correctly when training from supervised data?

No.

Given the presence of shuffle buffers in TensorFlow, the supervised learning dumper should probably be changed to chunk entire games.

Ttl:
I really don't want to chunk everything again since it takes so long. I would want to have an optional command line argument for chunks to use as a test set instead of splitting. It doesn't seem to be very difficult change.

You can merge your pull request and I'll add the option later. I'll close this one.

gcp:
>It doesn't seem to be very difficult change.

No, and it would work nicely with the code in the other pull. It just needs the chunks split somehow.

gcp:
>FWIW I don't think shuffling the test set is needed, you don't care about the correlation during testing.

Uhm, on the other hand this probably does mess up the MSE measurement. So it probably also needs shuffling.

PR #747, opened 2018-01-25
Implement real validation split for TF.

gcp:
Split the dataset into training and validation parts to get better
information about overfitting. Small fixes to timing.

### DIFF
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -294,8 +294,7 @@ def main(args):
     dataset = dataset.map(_parse_function)
     dataset = dataset.batch(BATCH_SIZE)
     dataset = dataset.prefetch(4)
-    iterator = dataset.make_one_shot_iterator()
-    next_train_batch = iterator.get_next()
+    train_iterator = dataset.make_one_shot_iterator()
 
     test_parser = ChunkParser(test, 1)
     dataset = tf.data.Dataset.from_generator(
@@ -303,10 +302,9 @@ def main(args):
     dataset = dataset.map(_parse_function)
     dataset = dataset.batch(BATCH_SIZE)
     dataset = dataset.prefetch(4)
-    iterator = dataset.make_one_shot_iterator()
-    next_test_batch = iterator.get_next()
+    test_iterator = dataset.make_one_shot_iterator()
 
-    tfprocess = TFProcess(next_train_batch, next_test_batch)
+    tfprocess = TFProcess(dataset, train_iterator, test_iterator)
     if args:
         restore_file = args.pop(0)
         tfprocess.restore(restore_file)
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -274,6 +274,9 @@ def main(args):
     if not chunks:
         return
 
+    # The following assumes positions from one game are not
+    # spread through chunks.
+    random.shuffle(chunks)
     training, test = split_chunks(chunks, 0.1)
     print("Training with {0} chunks, validating on {1} chunks".format(
         len(training), len(test)))
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -296,7 +296,7 @@ def main(args):
     dataset = dataset.prefetch(4)
     train_iterator = dataset.make_one_shot_iterator()
 
-    test_parser = ChunkParser(test, 1)
+    test_parser = ChunkParser(test)
     dataset = tf.data.Dataset.from_generator(
         test_parser.parse_chunk, output_types=(tf.string))
     dataset = dataset.map(_parse_function)
@@ -304,7 +304,8 @@ def main(args):
     dataset = dataset.prefetch(4)
     test_iterator = dataset.make_one_shot_iterator()
 
-    tfprocess = TFProcess(dataset, train_iterator, test_iterator)
+    tfprocess = TFProcess()
+    tfprocess.init(dataset, train_iterator, test_iterator)
     if args:
         restore_file = args.pop(0)
         tfprocess.restore(restore_file)
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -52,7 +52,7 @@ def remap_vertex(vertex, symmetry):
     return y * 19 + x
 
 class ChunkParser:
-    def __init__(self, chunks):
+    def __init__(self, chunks, workers=None):
         # Build probility reflection tables. The last element is 'pass' and is identity mapped.
         self.prob_reflection_table = [[remap_vertex(vertex, sym) for vertex in range(361)]+[361] for sym in range(8)]
         # Build full 16-plane reflection tables.
@@ -65,8 +65,9 @@ class ChunkParser:
         # Build the all-zeros and all-ones flat planes, used for color-to-move.
         self.flat_planes = [ b'\0' * 361, b'\1' * 361 ]
 
-        # Start worker processes, leave 1 for TensorFlow
-        workers = max(1, mp.cpu_count() - 1)
+        # Start worker processes, leave 2 for TensorFlow
+        if workers is None:
+            workers = max(1, mp.cpu_count() - 2)
         print("Using {} worker processes.".format(workers))
         self.readers = []
         for _ in range(workers):
@@ -260,8 +261,11 @@ def benchmark(parser):
         end = time.time()
         print("{} pos/sec {} secs".format( 10000. / (end - start), (end - start)))
 
-def main(args):
+def split_chunks(chunks, test_ratio):
+    splitpoint = 1 + int(len(chunks) * (1.0 - test_ratio))
+    return (chunks[:splitpoint], chunks[splitpoint:])
 
+def main(args):
     train_data_prefix = args.pop(0)
 
     chunks = get_chunks(train_data_prefix)
@@ -270,21 +274,33 @@ def main(args):
     if not chunks:
         return
 
-    parser = ChunkParser(chunks)
+    training, test = split_chunks(chunks, 0.1)
+    print("Training with {0} chunks, validating on {1} chunks".format(
+        len(training), len(test)))
 
-    run_test(parser)
+    #run_test(parser)
     #benchmark(parser)
 
+    train_parser = ChunkParser(training)
     dataset = tf.data.Dataset.from_generator(
-        parser.parse_chunk, output_types=(tf.string))
+        train_parser.parse_chunk, output_types=(tf.string))
     dataset = dataset.shuffle(1 << 18)
     dataset = dataset.map(_parse_function)
     dataset = dataset.batch(BATCH_SIZE)
     dataset = dataset.prefetch(4)
     iterator = dataset.make_one_shot_iterator()
-    next_batch = iterator.get_next()
+    next_train_batch = iterator.get_next()
 
-    tfprocess = TFProcess(next_batch)
+    test_parser = ChunkParser(test, 1)
+    dataset = tf.data.Dataset.from_generator(
+        test_parser.parse_chunk, output_types=(tf.string))
+    dataset = dataset.map(_parse_function)
+    dataset = dataset.batch(BATCH_SIZE)
+    dataset = dataset.prefetch(4)
+    iterator = dataset.make_one_shot_iterator()
+    next_test_batch = iterator.get_next()
+
+    tfprocess = TFProcess(next_train_batch, next_test_batch)
     if args:
         restore_file = args.pop(0)
         tfprocess.restore(restore_file)
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 #
 #    This file is part of Leela Zero.
-#    Copyright (C) 2017 Gian-Carlo Pascutto
+#    Copyright (C) 2017-2018 Gian-Carlo Pascutto
 #
 #    Leela Zero is free software: you can redistribute it and/or modify
 #    it under the terms of the GNU General Public License as published by
@@ -152,7 +152,10 @@ class ChunkParser:
                 with gzip.open(chunk, 'r') as chunk_file:
                     file_content = chunk_file.readlines()
                     item_count = len(file_content) // DATA_ITEM_LINES
-                    for item_idx in range(item_count):
+                    # Pick only 1 in every 8 positions
+                    picked_items = random.sample(range(item_count),
+                                                 (item_count + 7) // 8)
+                    for item_idx in picked_items:
                         pick_offset = item_idx * DATA_ITEM_LINES
                         item = file_content[pick_offset:pick_offset + DATA_ITEM_LINES]
                         str_items = [str(line, 'ascii') for line in item]
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -510,11 +510,11 @@ bool GTP::execute(GameState & game, std::string xinput) {
 
         if (!cmdstream.fail()) {
             auto vec = Network::get_scored_moves(
-                &game, Network::Ensemble::DIRECT, rotation);
+                &game, Network::Ensemble::DIRECT, rotation, true);
             Network::show_heatmap(&game, vec, false);
         } else {
             auto vec = Network::get_scored_moves(
-                &game, Network::Ensemble::DIRECT, 0);
+                &game, Network::Ensemble::DIRECT, 0, true);
             Network::show_heatmap(&game, vec, false);
         }
         gtp_printf(id, "");
--- a/src/Training.cpp
+++ b/src/Training.cpp
@@ -61,8 +61,8 @@ OutputChunker::~OutputChunker() {
 
 void OutputChunker::append(const std::string& str) {
     m_buffer.append(str);
-    m_step_count++;
-    if (m_step_count >= CHUNK_SIZE) {
+    m_game_count++;
+    if (m_game_count >= CHUNK_SIZE) {
         flush_chunks();
     }
 }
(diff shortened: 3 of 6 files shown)


training/tf/parse.py
```diff
@@ -260,8 +261,11 @@ def benchmark(parser):
         end = time.time()
         print("{} pos/sec {} secs".format( 10000. / (end - start), (end - start)))
 
-def main(args):
+def split_chunks(chunks, test_ratio):
+    splitpoint = 1 + int(len(chunks) * (1.0 - test_ratio))
+    return (chunks[:splitpoint], chunks[splitpoint:])
```
killerducky: Maybe do random.shuffle(chunks) first, to help the case where early chunks have weak games and later ones have strong games?

gcp: Yes, that's a good point.

gcp:
Haven't tested the SL changes yet but you get the idea.

training/tf/parse.py
```diff
@@ -270,21 +274,36 @@ def main(args):
...
     dataset = dataset.map(_parse_function)
     dataset = dataset.batch(BATCH_SIZE)
     dataset = dataset.prefetch(4)
     iterator = dataset.make_one_shot_iterator()
-    next_batch = iterator.get_next()
+    next_train_batch = iterator.get_next()
+
+    test_parser = ChunkParser(test, 1)
+    dataset = tf.data.Dataset.from_generator(
+        test_parser.parse_chunk, output_types=(tf.string))
```
gcp: This does need a shuffle buffer too. Without shuffling, testing 800 batches of 512 positions means we test about 1024 different games, which means the errors on the MSE measurement are very correlated.

killerducky:
How about also in the code parsing the chunks, randomly sample a percentage (10%? 1%?) of the positions. Then we don't have to rely so heavily on having a huge dataset.shuffle. I'm afraid Ttl's experiment in the mega-thread shows we have too many positions from the same game even in the 1<<18 size shuffle.

training/tf/parse.py
gcp: Actually, a conditional skip would be much handier here than a shuffle buffer to fix that, and more memory friendly.

gcp:
>How about also in the code parsing the chunks, randomly sample a percentage (10%? 1%?) of the positions.

Yeah I'm thinking along the same lines. If you sample too little you risk getting CPU limited again though.

gcp:
>we have too many positions from the same game even in the 1<<18 size shuffle.

Did that account for rotation?

killerducky:
I don't think it accounted for rotation but presumably the 3x3 filters have already learned all rotations so the values could still be highly correlated.

killerducky:
Link to @Ttl's results:
https://github.com/gcp/leela-zero/issues/591#issuecomment-359908639

roy7:
Forgive me for an overly simple question, but if the training process itself is pretty slow, can't we just fetch games as needed from the chunk files and not even need to shuffle buffer them in ram? Then we can randomly pull from the whole available data without any concerns about over concentration.

I just don't know how many games need loaded or how often, but given the time it takes for 1000 steps to finish, I wasn't sure it'd matter if we did a bunch of reads off my SSD between steps.

gcp:
>can't we just fetch games as needed from the chunk files and not even need to shuffle buffer them in ram? Then we can randomly pull from the whole available data without any concerns about over concentration.

A 250k window is about ~232G uncompressed, so the data needs to say compressed per chunk. This makes picking a single position out of entire chunk very inefficient, though.

gcp:
>I don't think it accounted for rotation but presumably the 3x3 filters have already learned all rotations so the values could still be highly correlated.

I don't understand what you're saying here. The 3x3 filters can't rotate by themselves. They detect features in a specific orientation. Even on higher level features, they can only see a spatial distance of 1 from the previous layer, so they do need to "learn" the rotation too.

killerducky:
That network as a whole could have constructed similar 3x3 filters for all rotations. So for feature A, it could have 8 similar 3x3 filters to detect all rotations of that feature.

Ignoring the details of how the network does it, I think it's safe to say for any position, the value head output will be highly correlated for all 8 rotations of that position. And positions from the same game are also highly correlated even if their positions are fed in with random rotations.

gcp:
>I think it's safe to say for any position, the value head output will be highly correlated for all 8 rotations of that position.

I mean that's the goal of the training so it would be rather sad if it didn't. It doesn't actually support the argument, though.

>And positions from the same game are also highly correlated even if their positions are fed in with random rotations.

Highly correlated *where*? The residual stack has strong spatial locality, so it can't. It's probably only really true for the heads.

(This argument made me find a bug!)

killerducky:
Yes I mean the final value output will be highly correlated for the same positions for all 8 rotations. 

Anyways with the validation set we will be able to find out how much these things matter. Enough talk for now, I leave it to you who is doing the work on this. :)

roy7:
> (This argument made me find a bug!)

This is the great and terrible thing about the NN"s ability to generalize. It can function in spite of bugs... so we don't realize easily that a bug even exists. :)

Ttl:
The earlier problems with value head overfitting might be related to insufficient shuffling. The correlation inside the mini-batch is probably not a problem, since it is quite small. However I think the bigger problem is correlation between the mini-batches. Previously the workers dumped all the positions from chunk to shuffle buffer in order. Big shuffle buffer helps with correlation inside the mini-batch, but even with a big shuffle buffer there is significant correlation between the positions in consecutive mini-batches since they share many of the same games. Since games were dumped in order all the positions from the same games are being trained very closely to each other and the value head outputs are highly correlated overfitting it to those games. Skipping randomly up to 8 positions helps with the problem.

gcp:
>However I think the bigger problem is correlation between the mini-batches.

Yes, I came to the same realization. The shuffle buffer holds about 655 different games. Every mini-batch expels 512 positions from it, causing only...7 (# feeder threads!) new games to enter.

>Skipping randomly up to 8 positions helps with the problem.

It's even a bit more effective than this. It doesn't just skip positions, it ends up taking a randomly ordered subsample out of the *entire* chunk. Given that the chunks contains 32 games, this means each new batch pulls from about 224 different games.

I had some weird performance effects from this, so I can't quantify yet what drop factor we can use without losing speed.

jkiliani:
Would it make sense to make a background process for the training machine that pre-shuffles chunks from the same dataset (i.e. produced by self-play from the same network)? At the moment, the training chunks are completely ordered, does it serve any purpose to keep it that way? If the chunks were not ordered anymore at the time of parsing, maybe you could use a smaller shuffle buffer...

gcp:
>Would it make sense to make a background process for the training machine that pre-shuffles chunks from the same dataset (i.e. produced by self-play from the same network)?

I'm not sure what a background process would gain you?

gcp:
One problem I've got is:

```
        self.next_batch = \
            tf.cond(self.training,
                    lambda: next_train_batch, lambda: next_test_batch)
```

Because of how TensorFlow works, both sides of that branch are evaluated. So I'm burning twice as much CPU as intended.

jkiliani:
> I'm not sure what a background process would gain you? Like, what's the difference with what we're doing now?

Not sure if a background process is the best way to do this, but I'm asking why not read in a number of chunks, shuffle the training data in them both internally and between the chunks, and write them back out again into new, shuffled chunks. That way, any time parsing uses these chunks in the future (quite a number of times, while it remains inside the training window), it would be starting from chunks that are already shuffled, instead of completely ordered ones.

gcp:
>Not sure if a background process is the best way to do this, but I'm asking why not read in a number of chunks, shuffle the training data in them both internally and between the chunks, and write them back out again into new, shuffled chunks.

That's what it's doing now on the fly, except that it doesn't write out the result to disk, which would be slow, and require a pre-processing pass (else the initial learning steps would still have the same problem) which delays things?

>That way, any time parsing uses these chunks in the future (quite a number of times, while it remains inside the training window)

There's 250k * ~400 = 100M positions in the window. 128k steps * 512 positions = 65M positions used.

gcp:
>I realised you'd be giving up the ability to easily filter out particular games from the training data with this, so I guess it depends on whether you have concerns about corrupted data at the time...

I need to take in the data from the original DB (mongo) and write out the chunk files anyhow, which is the point the filtering checks the IP of the source data. But shuffling there takes the same resources as shuffling later on.

gcp:
@ywrt If you're still around I could use your help right about now :)

killerducky:
I guess you already maxed your RAM usage? If not you could double dataset.shuffle a few more times.

gcp:
>I guess you already maxed your RAM usage?

Yeah basically. Hard to fix this with just a shuffle buffer. If I could work out a way to avoid the tf.cond silliness the solution here would be OK.

marcocalignano:
I have a question why just take one random simmentry and not all 8 of them? You could generate more data and distribute them overr othar batches. [L160](https://github.com/gcp/leela-zero/blob/23474b6bc97411752d44058ed605ecbddc17b949/training/tf/parse.py#L160)

gcp:
>and distribute them overr othar batches.

Yes but the question is how. That's exactly the problem here.

ywrt:
@gcp I'm back. I skimmed this thread. Could you maybe summarize what you think the problem is?

My guess after skimming the thread is that the mini-batches are much too internally correlated, and you're looking to have something closer to a random sample over all the moves from all the games in the window?

As an aside: Do you have a pointer to the code you're using to pull from mongo db? (maybe add it to the training directory somewhere?)

One of the challenges I had before is that the format that moves are stored in makes it much harder to run fast.

gcp:
> Could you maybe summarize what you think the problem is?

https://github.com/gcp/leela-zero/pull/747#issuecomment-360871125

I'm open to other ideas to make the shuffling better and fast, but this is currently bothering me. I'm trying to find a clean way to pull from the right dataset, without having TF run both parsers at once.

>As an aside: Do you have a pointer to the code you're using to pull from mongo db? (maybe add it to the training directory somewhere?)

https://pastebin.mozilla.org/9076742

marcocalignano:
> Yes but the question is how.

something like this?

```
    def task(self, chunks, workers):
        while True:
            random.shuffle(chunks)
            for chunk in chunks:
                with gzip.open(chunk, 'r') as chunk_file:
                    file_content = chunk_file.readlines()
                    item_count = len(file_content) // DATA_ITEM_LINES
                    for item_idx in range(item_count):
                        pick_offset = item_idx * DATA_ITEM_LINES
                        item = file_content[pick_offset:pick_offset + DATA_ITEM_LINES]
                        str_items = [str(line, 'ascii') for line in item]
                        symmetry = 0
                        for write in writes:
                            success, data = self.convert_train_data(str_items, symmetry)
                            if success:
                                # Send it down the pipe.
                                writer.send_bytes(data)
                            symmetry = (symmetry + 1) % 8

```

gcp:
>something like this?

That doesn't distribute them over batches. The inability of the shuffle buffer to do so amounts to most of the discussion in this thread.

ywrt:
(After a closer read)

I think you'd find this easier if the partitioning was done on the mongodb export side?

My first though would be to hash-partition the moves into separate chunk sets, and then use a feedable iterator to switch between the training and validation datasets.

So the export side, something like:
```
  for m in moves:
    if hash(m) < MAX_HASH_VALUE * 0.01:  # 1% validation sample.
      write_to_chunk_with_prefix("validate", m)
    else:
      write_to_chunk_with_prefix("train", m)
```

and then on the TF side, build two separate Datasets and use a placeholder to feed them separately into training and validation.

If you like, I can sketch out a patch for this on the TF side?

Re shuffling: I agree that the limited shuffling is a real concern. It might be worth doing a random shuffle on export? 

This is really a RAM consideration: Ideally, we'd have `number_of_partitions = available_ram / (2* 1024 * 1024)`, which spreads as wide as possible while still being able to write to disk in 2MB chunks (so we stay throughput bound rather than spindle bound).

Which translates to: maintain N buffers in RAM, add each move to a random buffer. When a buffer reaches ~2MB, flush it to a random chunk file that's not yet full. 

The next step is to do the reverse on the TF side: read from K chunks in parallel, reading into 2MB buffers (to stay disk efficient), and merge the chunks to produce the final dataflow in TF.
This would reduce in a not-completely-terrible approximation to uniform sampling, without overly much CPU overhead. (Obv, we're trading off increased RAM to reduce the CPU cost of the shuffle).

Depends on how crazy you want to get :)

isty2e:
My programming skill is somewhere between abysmal and mediocre, so this might be a silly question. If I were to do this, I probably would separated the dataset first and let the train and test dataset to have different iterators, which also probably could bypass the double evaluation of tf.cond()... I suppose. Is there a reason the split is not implemented this way?

marcocalignano:
@gcp But maybe you can create pipes between threads and each threads write a different symmetry on a pipe from a different thread. Could that work?

gcp:
> If I were to do this, I probably would separated the dataset first and let the train and test dataset to have different iterators, which also probably could bypass the double evaluation of tf.cond()

This is exactly how it's implemented.

killerducky:
Is it true that the main reason the chunks are written as contiguous games is for the bad game filtering process? How about the bad game filter process is run completely separately one time only, and each game is marked good/bad in the database. Then when reading out from the database and creating the chunks they can be written out pre-randomized.

 BTW are there still bad games now that the OpenCL bug is detected automatically?

gcp:
>build two separate Datasets and use a placeholder to feed them separately into training and validation.

That's what the code does. Look in this pull request.

gcp:
>Then when reading out from the database and creating the chunks they can be written out pre-randomized.

As explained previously here, this makes no difference to the computing time required, and it ruins the compression.

ywrt:
Sorry, I was unclear. The placeholder should be for the iterator. Something like:
In main:
```
  train_handle = sess.run(next_train_batch.string_handle())
  valid_handle = sess.run(next_test_batch.string_handle())

  handle = tf.placeholder(tf.string, shape=[])
  iterator = tf.data.Iterator.from_string_handle(
      handle, dataset.output_types, dataset.output_shapes)
  next_batch = iterator.get_next()

  TFProcess(next_batch, train_handle, valid_handle)
```

and in TFProcess for training:
```
  # Run training for this batch
        policy_loss, mse_loss, reg_term, _, _ = self.session.run(
            [self.policy_loss, self.mse_loss, self.reg_term, self.train_op,
                self.next_batch],
                feed_dict={self.training: True, handle: train_handle})
```
and then validation
```
         for _ in range(0, 10):
                train_accuracy, train_mse, _ = self.session.run(
                    [self.accuracy, self.mse_loss, self.next_batch],
                    feed_dict={self.training: False, handle: valid_handle})
```

Is that clearer? As I mentioned, happy to sketch out a patch for this if you want.

gcp:
Thanks, that gives me some idea how to solve it.

ywrt:
Re the shuffle: If you want something dumb and easy, changing the workers as follows would help reduce intra-batch correlation.
```
    CHUNKS_TO_MERGE=8
    def chunk_convert(self, chunk):
        with gzip.open(chunk, 'r') as chunk_file:
            file_content = chunk_file.readlines()
            item_count = len(file_content) // DATA_ITEM_LINES
            for item_idx in range(item_count):
                pick_offset = item_idx * DATA_ITEM_LINES
                item = file_content[pick_offset:pick_offset + DATA_ITEM_LINES]
                str_items = [str(line, 'ascii') for line in item]
                # Pick a random symmetry to apply
                symmetry = random.randrange(8)
                success, data = self.convert_train_data(str_items, symmetry)
                if success:
                    yield data

    def task(self, chunks, writer):
        while True:
            random.shuffle(chunks)
            chunk_list = chunks
            while chunk_list:
                c = chunk_list[:CHUNKS_TO_MERGE]
                chunk_list = chunk_list[CHUNKS_TO_MERGE:]
                iters = [ self.chunk_convert(i) for i in c]
                for m in zip(iters):
                    writer.send_bytes(m)

```

This is pretty dumb as it needlessly pulls the entire chunk file into RAM.

Note also, that the shuffle buffer in the Dataset pipeline was deliberately Dataset.shuffle().map(), not Dataset.map().shuffle(), so that the shuffle operates on packed strings.

If you want a larger shuffle buffer, right now the ChunkParser assumes that the 'probs' can be arbitrary floats, when in reality it's a one-hot vector (all zero except a single one), so the string representing the move can be shrunk to ~ 25% of it's current size, similarly making the shuffle buffer a quarter of it's current size (it would also make both the parsing faster and the TF graph faster but that's by-the-by).

gcp:
>right now the ChunkParser assumes that the 'probs' can be arbitrary floats, when in reality it's a one-hot vector (all zero except a single one)

That's only true for supervised learning data. RL data has a true probability map.

gcp:
Ok together with your hint and https://www.tensorflow.org/programmers_guide/datasets#creating_an_iterator I think I got something.

training/tf/parse.py
```diff
@@ -270,21 +277,34 @@ def main(args):
...
     dataset = dataset.batch(BATCH_SIZE)
     dataset = dataset.prefetch(4)
-    iterator = dataset.make_one_shot_iterator()
-    next_batch = iterator.get_next()
+    train_iterator = dataset.make_one_shot_iterator()
+
+    test_parser = ChunkParser(test, 1)
+    dataset = tf.data.Dataset.from_generator(
+        test_parser.parse_chunk, output_types=(tf.string))
+    dataset = dataset.map(_parse_function)
```
ywrt: Might be a good idea to shuffle the validation dataset too?

gcp: Not sure, didn't we just discover the shuffle buffer may not help all that much unless its made much larger?

marcocalignano:
@gcp and if you do all the task work in one task and move only `self.convert_train_data(str_items, symmetry)` in a different thread you could distribute the data already shuffle and with a different symmetry to each task.

gcp:
>@gcp and if you do all the task work in one task

Then it'll bottleneck even harder in the ungzip+skip, which is that actual bottleneck...

PR #754, opened 2018-01-26
#697 Add a high level comment on the GTP class. Also give a web refer…

barrybecker4:
I followed @wctgit  suggestion and just added a comment and web reference. No code changes.

### DIFF
--- a/src/GTP.h
+++ b/src/GTP.h
@@ -51,6 +51,14 @@ extern FILE* cfg_logfile_handle;
 extern bool cfg_quiet;
 extern std::string cfg_options_str;
 
+/*
+    A list of all valid GTP2 commands is defined here:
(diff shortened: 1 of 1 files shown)


killerducky:
Your commit log message is all on one line. I think this can happen if you do not leave the 2nd line blank.

barrybecker4:
I don't think I can amend it because it's already pushed to the branch. Is it bad to have it all on once line?
 Are there certain conventions that need to be followed?

killerducky:
You can do a force push, or just let gcp fix it. 

Check this for some guidelines:
https://github.com/erlang/otp/wiki/writing-good-commit-messages

gcp:
>I don't think I can amend it because it's already pushed to the branch

git push -f

PR #756, opened 2018-01-26
convolve3 Output transform / input transform fusion

ihavnoid:
Looking at the code, the output transformation and the input transformation of the next convolve3() call can be merged to a single kernel.  The benefit is that we don't need to store the output transformation results to global memory on roughly half of the cases, hence gaining some performance.  Another bonus is that we reduce the number of kernel calls by a lot.

Downside is that it adds quite a bit of mess to the code.

I may not have perfect understanding of the current code so I will need a lot of nitpicking and comments.  One thing is that there is a 'magic number' on the kernel group size, which simply seemed to yield the best results... would like comments on how to address that magic number.

Roughly +10% on netbench observed on a GTX1080.

### DIFF
--- a/src/OpenCL.cpp
+++ b/src/OpenCL.cpp
@@ -54,7 +54,6 @@ static std::string sourceCode_config = R"(
 static std::string sourceCode_convolve3 = R"(
 void __in_transform_eq(float x[4][4], __global float *V, int offset, int CPpad) {
     float T1[4][4];
-    float T2[4][4];
 
     T1[0][0] = x[0][0] - x[2][0];
     T1[0][1] = x[0][1] - x[2][1];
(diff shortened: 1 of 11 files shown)


src/OpenCL.cpp
```diff
@@ -260,6 +258,150 @@ __kernel void out_transform_fused_bn(__global float *M,
...
+
+    const int yin = 2 * block_y - 1;
+    const int xin = 2 * block_x - 1;
+
+
+    const int x = 2*block_x;
+    const int y = 2*block_y;
+    int a_ind = (y)*W + (x);
+
+    const int kHW = k * W * H;
```
Hersmunch: Is this the same as chT? or do the brackets around T do something I don't know about?

src/OpenCL.cpp
```diff
@@ -260,6 +258,150 @@ __kernel void out_transform_fused_bn(__global float *M,
...
+    if (block < P && k < K) {
+        const int CPpad = Ppad * Cpad;
+        // Cache input tile and handle zero padding
+        float xx[4][4];
+        for (int i = 0; i < 4; i++) {
+            for (int j = 0; j < 4; j++) {
+                int a = xin + j;
+                int b = yin + i;
+                if (b >= 0 && a >= 0 && b < H && a < W) {
+                    xx[i][j] = ybuf[kg * T + b*W + a];
```
Hersmunch: So from what I can tell, up to 16 locations are read from ybuf[] but up on line 333 only up to 4 locations are written to. Are the other locations initialised elsewhere? Or should the code here be calling vload_net_t() for those that weren't loaded into ybuf[]?

Hersmunch:
I don't have any experience with the definition of the kernels etc. but is it possible to write functions for them all to share code rather than repeating it?

src/OpenCL.cpp
Hersmunch: OK, I just read what barrier(CLK_LOCAL_MEM_FENCE) does :)

ihavnoid:
Added a commit which creates local functions for the transforms.  Also addressed some of the comments too.

Hersmunch:
Nice! Thank you for making it much more manageable. I hope that using functions doesn't incur a significant performance hit.

One more thought that occurred to me would be whether ybuf[] could be initialised with zeros such that no edge testing/zero padding handling has to be done when reading from it into xx[][]. But maybe that could be looked at in a later pull.

Ttl:
Nice feature. I think I mentioned this optimization in the Winograd pull request, but I didn't start working on it myself.

It looks like the code can be simplified a bit by handling the input convolution with fused_bn kernel. Then the normal out_transform and batchnorm kernels can be removed completely. I did the changes while looking at the changes locally and uploaded them to my fork. I'll see if I can make a pull request to pull request.

One limitation I found was that out_transform fused with in_transform assumes that number of inputs equals number of outputs. It doesn't work correctly for the input convolution at the moment. Is this limitation intentional?

ihavnoid:
Yes it was intentional.  I dont think we have a usage case where they dont match, and if it doesn't I find it quite difficult to fuse the two ops becase it would need to change the number of concurrent threads.  I was considering that to be something the caller should assume, though.

ihavnoid:
If you cant create a pull request on a pull request, pleas let me know where your changes are.  I will manually add it to this pull request.

killerducky:
My googling doesn't show this, but I think you should pull the ihavnoid:out_in_fusion branch into your repo, and do a pull request to ihavnoid's repo based on that branch. If he accepts that pull his branch will be updated, which should update this pull request automatically.

Interested to see if it works.

ihavnoid:
Okay I merged the pull request... seems to work. :)

Ttl:
I was talking about the input convolution which has different number of input and outputs. However the fused output transformation always has inputs equal to just calculated outputs. Turns out that it didn't work because of a bug. I submitted another pull request to fix it.

PR #757, opened 2018-01-26
Implement an N-way merge for training games.

ywrt:
This improve shuffling and thereby reduces intra-batch and batch-to-batch correlation

(just because I get nerd-sniped into thinking about this)

### DIFF
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -303,7 +303,7 @@ def main(args):
     parser = ChunkParser(chunks, n_way_merge=64)
 
     run_test(parser)
-    benchmark(parser)
+    #benchmark(parser)
 
     dataset = tf.data.Dataset.from_generator(
         parser.parse_chunk, output_types=(tf.string))
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -19,6 +19,8 @@
 import sys
 import glob
 import gzip
+import itertools
+import io
 import random
 import math
 import multiprocessing as mp
@@ -52,7 +54,7 @@ def remap_vertex(vertex, symmetry):
     return y * 19 + x
 
 class ChunkParser:
-    def __init__(self, chunks):
+    def __init__(self, chunks, n_way_merge=1):
         # Build probility reflection tables. The last element is 'pass' and is identity mapped.
         self.prob_reflection_table = [[remap_vertex(vertex, sym) for vertex in range(361)]+[361] for sym in range(8)]
         # Build full 16-plane reflection tables.
@@ -67,12 +69,12 @@ class ChunkParser:
 
         # Start worker processes, leave 1 for TensorFlow
         workers = max(1, mp.cpu_count() - 1)
-        print("Using {} worker processes.".format(workers))
+        print("Using {} worker processes with a {}-way merge.".format(workers, n_way_merge))
         self.readers = []
         for _ in range(workers):
             read, write = mp.Pipe(False)
             mp.Process(target=self.task,
-                       args=(chunks, write)).start()
+                       args=(chunks, write, n_way_merge)).start()
             self.readers.append(read)
 
     def convert_train_data(self, text_item, symmetry):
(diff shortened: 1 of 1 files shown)


killerducky:
So before each worker's data was put into the buffer in some normal order (round robin I guess?) and now it is shuffled? But I think the dataset.shuffle accomplishes the same thing except much better since it is very large. Am I missing something here?

roy7:
For what it's worth, default num_workers on my Ryzen is 15, I assume same for GCP. (We have similar setups.) So for even more randomness, it's time for a Threadripper and 32 workers? ;)

ywrt:
Hmm. I just realized this patch may be in error. Running the numbers:

At ~ 200 moves per game, a 2^18 element shuffle buffer holds ~1300 games. With a batch size of 512, it's highly likely that any given batch has multiple moves from the same game. As in, the probability of this _not_ happening is around 10^-51.

Even with 10,000 games in the buffer, the probability of moves from the same game going into a batch are > 99.9998%

Which I think says that this patch is insufficient? Going to this about this some more.

killerducky:
Is there an easy way to add some metadata to the data that gets yielded to the shuffle buffer that records which chunk, which line number the data came from? And then we could analyze how random it is coming out of the shuffle buffer. I tried this a long time ago, my first step was to try to just print the output of shuffle.buffer but I couldn't get that working.

ywrt:
TL;DR: This patch isn't really useful for increasing batch diversity.

Not an easy way per-se. It would need to be plumbed through all the way through chunk parsing et al making it something of a pain.

But we can get an easier indication by simulation.

Let's assume 250 moves per game (because I'm too lazy to actually measure it).
A 2^18 move shuffle buffer filled sequentially then holds 2^(18 - lg(250)) = ~1048 games.

With a batch size of 512, we expect to get about 4 moves from the same game in the batch.

We can flip this around and ask, how many distinct games per batch do we see?
For 512 samples over 1048 games, **we expect to see only 405 distinct games per batch**(!) (+/- ~ 3%).

If we can grow the number of games in the shuffle buffer to 5000, these numbers become an expected 2.6 moves from the same game, and 486 distinct games in the batch.

At 15,000 games, we expect to see ~503 distinct games per 512 element batch which is much better.

At the moment, the games are contiguous in the chunks, and the workers are feeding 15 chunks in parallel into a shuffle buffer hold ~1050 games, so we expect that the buffer is a sample over something close to 1050.

Sadness: This patch doesn't change that! It does a merge over more chunks, but not nearly enough to seriously change the expected number of games in the shuffle buffer. In the best case it maybe doubles the number of games.

Instead, we can flip around again and say that the shuffle buffer should be ~15 times larger than it is.

At the moment, the shuffle buffer holds strings that are roughly 8K in size (2^13 bytes). A 2^18 element shuffle buffer will eat 18+13 = 31 == ~2GB of RAM.

So a 5x larger shuffle buffer is 10GB of RAM, and a 15x buffer is 30GB.

So, roy7, how much RAM do you have? ;)

The alternative is to change https://pastebin.mozilla.org/9076742 so that it doesn't write out games contiguously, so it spreads each game over multiple chunks which would have the same improvement without eating a ton of RAM.

Trivial simulation code FWIW:
```
#include <stdio.h>
#include <random>
#include <vector>

int main(void) {
  const int k = 1050;  // number of games
  const int n = 512;    // batch size

  std::random_device r;
  std::default_random_engine e1(r());
  std::uniform_int_distribution<int> uniform_dist(0, k); 
  std::vector<int> maxes(20);
  std::vector<int> games(n);
  for (int loop = 0; loop < 1000; ++loop) {
    std::vector<int> count(k);
    // sample with replacement n times. 
    // Note! Assumes games-per-buffer much less than moves-per-buffer
    for (int i = 0; i < n; ++i) count[uniform_dist(e1)]++;
    // Find max number of duplicates
    int max = 0;
    for (const auto v : count) max = std::max(max, v); 
    maxes[max]++;

    int c = 0;
    for (const auto v : count) if (v) c++;
    games[c]++;
  }

  // generate stats.
  int sum = 0;
  int count = 0;
  for (int i = 0; i < maxes.size(); ++i) {
    if (!maxes[i]) continue;
    //printf("%3d: %d\n", i, maxes[i]);
    sum += i * maxes[i];
    count += maxes[i];
  }
  printf("mean same-game moves per batch: %f\n", 1. * sum / count);

  sum = 0;
  count = 0;
  for (int i = 0; i < games.size(); ++i) {
    //if (games[i]) printf("%3d: %d\n", i, games[i]);
    sum += i * games[i];
    count += games[i];
  }
  printf("mean distinct games per batch: %f\n", 1. * sum / count);
  return 0;
}
```

ywrt:
This is all mildly annoying because if the move record size out of ChunkParser was better compressed, this issue would pretty much go away. But tensorflow lacks a fast way to split a bit vector, so the planes are passed as bytes and the whole thing is ~4x larger than it should be. Sigh.

killerducky:
@ywrt - @ttl did some simulations also: https://github.com/gcp/leela-zero/issues/591#issuecomment-359908639

roy7:
Is there not some way to simply include 1 position from each game in the batch? Or chunk out positions instead of whole games (I realize this will result in way more disk space) such that each chunk is filled only with single positions from games with never more than 1 position from any single game?

roy7:
Actually is the training data in the chunks already broken up by position? I guess I was thinking of an SGF file where we create board positions on the fly from the game record, but that's not the case, we have a full set of network targets/etc for individual positions, right?

roy7:
Basically what I'm thinking is this. Games are capped at some number of moves, I don't recall exactly what it is, say 750. Assume average moves per game is 250. So let's not worry about a training window of exactly 250K games. Let's make it a training window of 250K * 250 positions. (62.5 million.)

Make 750 round robin database files (or more if these are too large), to hold game positions. The round robin is so they are a fixed size as more data is added later on. As we go through game data to initialize this, or are adding new games as they arrive, iterate over 0..749 in a loop writing out one position at a time to each database in order. Since games can't exceed 750 moves (whatever # it is), no chunk ever has a game repeated. (If there's a concern about writing the positions out to the chunks sequentially, the ordering could be randomized in some way.)

End result is 750 chunks of 62.5M positions / 750 = 83K positions guaranteed different games. 

From your "2^(18 - lg(250)) = ~1048 games" comment it sounds like shuffle buffers hold about 260K game positions currently. I'm not sure if it's a problem to use smaller shuffle buffers, if we know every one of the 83K positions in the buffer are different games or not.

So, now, when you need to load a shuffle buffer, just read in one of the chunks. As new game data flows to the training machine, write it to the chunk round robin databases. This will automatically replace the oldest data with new data, giving you a sliding scale without needing to add/delete chunk files or re-dump the whole database 250K games each time.

ywrt:
@roy7 I think it might be a little messy to do in the db. I'm thinking that shuffling the moves when going from db to chunk might be enough. e.g. efficiently writing 250 chunks in parallel would required "only" 500MB of RAM, but would increase the number of games that the shuffle buffer samples from by ~250x.

The biggest drawback is it would require parsing the game data on the way out of the db, where it's currently just copied.

gcp:
>The biggest drawback is it would require parsing the game data on the way out of the db, where it's currently just copied.

Splitting by lines really, not full parsing. Given the discussion, I think doing a parallel write-out of the chunks during the database dump is indeed the best option, but it does not solve the supervised learning case, so meh.

The performance impact of skipping positions (i.e. what it does right now) is noticeably high despite most of the parsing being unnecessary even for skipped positions.

I slightly regret not making the training format fixed-size binary records now. (But text is easy for debugging, and it did help me find some bugs...)

Closing this as, as you said, it doesn't help enough. I'll put the mongodb dumper in the repo.

gcp:
https://stackoverflow.com/a/6077078/909836

Maybe one simple optimization is to split the readlines() into a read, and then skip forward the newlines manually.

Given that we only need every 1/x lines, this probably avoids a whole bunch of memory allocation etc.

gcp:
It's a massive speedup, LOL: https://github.com/gcp/leela-zero/commit/9e9fd4909f26c787ea065b0f720ba8faa452d4ad

roy7:
LOL, reminds me how --logfile used to be nightmarishly slow. ;)

What about converting the text info in db to fixed-size binary records during export? Then on the training side of things, you can more easily access records using the fixed-size boundaries.

gcp:
Yes, that might be an overall gain. The chess people did this (they also had about 10 times more input planes and thus data to deal with...).

PR #769, opened 2018-01-28
New Stop and resume Feature

marcocalignano:
Two new option introduced:

`-t` _n_ run `autogtp` for _n_ minutes and then save the current games on the disk
`-s` run `autogtp` for just one game and exit

Also, in normal with `q+Enter` saves all games on the disk and exit. 
Every time autogtp starts look for previous stored game and restore them.

### DIFF
--- a/autogtp/main.cpp
+++ b/autogtp/main.cpp
@@ -36,6 +36,7 @@
 #include <iostream>
 #include "Game.h"
 #include "Management.h"
+#include "Console.h"
 
 int main(int argc, char *argv[]) {
     QCoreApplication app(argc, argv);
@@ -99,15 +100,17 @@ int main(int argc, char *argv[]) {
         }
     }
     QMutex mutex;
+    Console cons;
     Management *boss = new Management(gpusNum, gamesNum, gpusList, AUTOGTP_VERSION, parser.value(keepSgfOption), parser.value(keepDebugOption), &mutex);
     QObject::connect(&app, &QCoreApplication::aboutToQuit, boss, &Management::storeGames);
     QTimer *timer = new QTimer();
     boss->checkStoredGames();
     boss->giveAssignments();
-    if(parser.isSet(timeoutOption))
-    {
+    if(parser.isSet(timeoutOption)) {
         QObject::connect(timer, &QTimer::timeout, &app, &QCoreApplication::quit);
         timer->start(parser.value(timeoutOption).toInt() * 60000);
+    } else {
+        QObject::connect(&cons, &Console::sendQuit, &app, &QCoreApplication::quit);
     }
     return app.exec();
 }
--- a/autogtp/main.cpp
+++ b/autogtp/main.cpp
@@ -26,6 +26,7 @@
 #include <QFile>
 #include <QFileInfo>
 #include <QDir>
+#include <QtWidgets/QShortcut>
 #include <QDebug>
 #include <chrono>
 #ifdef WIN32
@@ -41,7 +42,7 @@ int main(int argc, char *argv[]) {
     app.setApplicationName("autogtp");
     app.setApplicationVersion(QString("v%1").arg(AUTOGTP_VERSION));
 
-    QTimer::singleShot(0, &app, SLOT(quit()));
+//    QTimer::singleShot(0, &app, SLOT(quit()));
 
     QCommandLineParser parser;
     parser.addHelpOption();
@@ -78,7 +79,6 @@ int main(int argc, char *argv[]) {
     }
 
     // Map streams
-    QTextStream cin(stdin, QIODevice::ReadOnly);
     QTextStream cout(stdout, QIODevice::WriteOnly);
 #if defined(LOG_ERRORS_TO_FILE)
     // Log stderr to file
@@ -126,13 +126,14 @@ int main(int argc, char *argv[]) {
         }
     }
     QMutex mutex;
-    Management boss(gpusNum, gamesNum, gpusList, AUTOGTP_VERSION,
-                    parser.value(keepSgfOption), parser.value(keepDebugOption),
-                    &mutex);
-    boss.giveAssignments();
-    mutex.lock();
+    Management *boss = new Management(gpusNum, gamesNum, gpusList, AUTOGTP_VERSION, parser.value(keepSgfOption), parser.value(keepDebugOption), &mutex);
+    boss->checkStoredGames();
+    boss->giveAssignments();
+    int r = app.exec();
+    // Synchornize on the work_thread.  Letting it run to completion.
+    boss->wait();
     cerr.flush();
     cout.flush();
-    mutex.unlock();
-    return app.exec();
+    app.exit();
+    return r;
 }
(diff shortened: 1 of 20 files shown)


autogtp/main.cpp
```diff
@@ -62,11 +62,20 @@ int main(int argc, char *argv[]) {
     QCommandLineOption keepDebugOption(
         { "d", "debug" }, "Save training and extra debug files after each self-play game.",
                           "output directory");
+    QCommandLineOption timeoutOption(
+        { "t", "timeout" }, "Save running games after the timeout (in minutes) is passed.",
+                          "time in minutes");
```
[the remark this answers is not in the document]
marcocalignano: You mean to add it in the comment?

marcocalignano:
The Travis CI build failed. Can someone add the two new files that are missing in Docker?

autogtp/main.cpp
```diff
@@ -76,39 +85,13 @@ int main(int argc, char *argv[]) {
...
+        gpusNum = 1;
     }
-    QTextStream cerr(&caFile);
-#else
+    // Map streams
     QTextStream cerr(stderr, QIODevice::WriteOnly);
-#endif
-#ifdef WIN32
-    // We need to make sure these files we need are there before calling them.
-    // Otherwise it will result in a crash.
```
gcp: This check was nice to have I think?

gcp:
Does this work correctly with relation to restoring the tree search data for the training? I don't see how it can. The data can't be dumped until the winner of the game is known.

autogtp/main.cpp
marcocalignano: I guess they disappeared because of a wrong rebase. I will check which and fix it when I get home

marcocalignano:
Yes you are correct. I just store and reload the SGF file.

gcp:
It still works for match games, but I guess you shouldn't try to restore training games. But match games are minority.

So I'm not really sure how useful this is then :-/

Dumping the partial training data and then filling in the correct match result when the game is done would be a solution, but then autogtp is parsing the training data format, yikes.

marcocalignano:
Maybe Leelaz should be able to dump and restore training data the same way it does with SGF.

gcp:
That would be another way to deal with it. dump_training/restore_training/dump_final?

marcocalignano:
We could do something like: `save_training()` write in a file all the `TimeStep` in `m_data`, `read_training()` reads back all the `TimeStep` from the file and `dump_training()` stays unmodified.

marcocalignano:
Ok like from the previous comment I did the save_training and load_trainig function. I tested saving loading and resaving the same data an comparing the two files: they are identical. If anyone has a better test please let me know. Now only the integration in autogtp is missing.

marcocalignano:
Ok I coded the integration but I have no strength to test it. Tomorrow!

marcocalignano:
@gcp If I upload a game with training data "interrupted" ;) can you see if the data are still useful?

marcocalignano:
@gcp ok I tested it stoped it after move 198 and resumed with move 199 at the end it uploaded the game:

`Game data 9baa29b6f3f5b3c1069c9adb132942e505752fc17cf090aeb2f4d0c71a6e086b stored in database
`
If you can confirm it is working, then you can merge this.

gcp:
Dump out the training data before sending it to the server, and the number of lines (wc -l) divided by 19 should match the number of moves in the game. (That's really all I can do too)

Maybe do a sanity check that the number every 19th line (the winner) alternates.

JDJoe:
can you also add run number of games and then quit. 

`-x`  _n_ run `autogtp` for _n_ times and then quit etc

marcocalignano:
@JDJoe what you want is more complicated (especially with parallel threads) so I want to get this merge and then implement your request.

zediir:
Would it be easy to port stop and resume to validation? Or add a parameter to pass already run win/loss results?

marcocalignano:
> Or add a parameter to pass already run win/loss results

what do you mean by that?

zediir:
I mean if I've had validation running and it is now 64 wins 45 losses but I have to interrupt it, I would like to be able to pass those results ro the process when I start it again

marcocalignano:
@zediir This should be easy enough. After this get merge I try to add `q+Enter` even for Validation. If this doesn't work I'll create a new option to initialize the SPRT.

JDJoe:
@marcocalignano just to clarify, I mean _n_  completed games
autogtp reports how many games it has completed 

for example: 

`20 game(s) (20 self played and 0 matches) played in 138 minutes = 415 seconds/game, 1249 ms/move, last game took 279 seconds.
`

I assumed this information can be cached and used to stop the game. Hmm... i wonder can I do this with bash scrip / sed something.

marcocalignano:
@gcp I tested like you said and both tast passed (wc -l /19  == moves)  and every 19 lines alternates -1 and 1. So I think I am done.

marcocalignano:
@JDJoe If you do like that when you start `autogtp` with `-g` _n_ when you reach the number you want you already started _n-1_ threads that then will be interrupted you have to do it differently. I have a solution but not for this PR.

alreadydone:
> I tested saving loading and resaving the same data an comparing the two files: they are identical. 

@marcocalignano I'd like to make sure whether you meant that an interrupted game and a non-interrupted one will generate identical *.sgf and *.train files in the end, if they use the same random seed. Because of tree reuse (#632) and since we are not saving the tree, how could we ensure that the subsequent searches are identical? Will the few moves after resuming be weaker than normal? Also there's the issue of reproducibility, though it's not a big concern. (Sorry if I misunderstand how tree reuse works.)

marcocalignano:
@alreadydone My test was 
- start leelaz 
- generate some moves 
- save the training file and the sgf file
- exit leelaz 
- restart leelaz
- load the training and the sgf file
- save them immediately on different files 
- make sure that they are the same 

In this way I tested both saving adn loading function for the training file.

alreadydone:
I did the following test:
- start a game using autogtp and save the game after move 1;
- resume the game and save it after move 50;
- resume the game again, this time letting it finish, get sgf#1;
- resume from move 51 and let the game finish, get sgf#2.

I found that sgf#1 and sgf#2 differ starting from move 72.
@gcp So this indeed affect reproducibility. Maybe the random seed is used the same way in both games; is tree reuse the sole reason for the difference? Not recording the tree probably make the play a little weaker than normal, but if we just want to ensure reproducibility, we could record the moves at which the game is interrupted instead of saving the tree with the *.sgf and *.train files. Of course, reproducibility may not be a big concern, or there may be other factors, which make the fixes hard or not worth implementing.

alreadydone:
@marcocalignano I find that after resuming, the correct time setting is not set; it seems that leelaz thinks at most 50 seconds and doesn't necessarily do all of the requested playouts/visits.

marcocalignano:
@alreadydone The time settings are the same, so no idea why. Can you check if before the resume is really different? Can you post some debugs and the options?

alreadydone:
I discovered the problem when looking at the search probabilities in the
training data (.train or .txt.0); for a game resumed on a slow CPU or GPU,
they may not be multiples of 1/3199. You should be able to reproduce the
problem easily by resuming sufficiently many games on your CPU so that each
move takes about 1 minute.

marcocalignano <notifications@github.com> 于 2018年3月16日周五 05:20写道：

PR #792, opened 2018-02-01
Change play_move to handle FastBoard::PASS

sethtroisi:
IMO this is more generic code and
We're able to removing the nearly identical play_pass from FastState.cpp

### DIFF
--- a/src/FastState.cpp
+++ b/src/FastState.cpp
@@ -69,14 +69,22 @@ bool FastState::is_move_legal(int color, int vertex) {
                 !board.is_suicide(vertex, color));
 }
 
-void FastState::play_pass(int color) {
+void FastState::play_move(int vertex) {
+    play_move(board.m_tomove, vertex);
+}
+
+void FastState::play_move(int color, int vertex) {
+    board.m_hash ^= Zobrist::zobrist_ko[m_komove];
+    if (vertex == FastBoard::PASS) {
+        // No Ko move
+        m_komove = 0;
+    } else {
+        m_komove = board.update_board(color, vertex);
+    }
+    board.m_hash ^= Zobrist::zobrist_ko[m_komove];
+
+    m_lastmove = vertex;
     m_movenum++;
-
-    m_lastmove = FastBoard::PASS;
-
-    board.m_hash ^= Zobrist::zobrist_ko[m_komove];
-    m_komove = 0;
-    board.m_hash ^= Zobrist::zobrist_ko[m_komove];
 
     if (board.m_tomove == color) {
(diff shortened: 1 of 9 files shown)


src/FastState.cpp
```diff
@@ -69,51 +69,30 @@ bool FastState::is_move_legal(int color, int vertex) {
...
 
-void FastState::play_pass(int color) {
-    m_movenum++;
-
-    m_lastmove = FastBoard::PASS;
+void FastState::play_move(int vertex) {
+    play_move(board.m_tomove, vertex);
+}
 
+void FastState::play_move(int color, int vertex) {
```
sethtroisi: This diff isn't the best, I'd suggest comparing this by hand with the code that used to start at line 97

src/FastState.cpp
```diff
@@ -69,51 +69,30 @@ bool FastState::is_move_legal(int color, int vertex) {
...
-            board.m_hash ^= Zobrist::zobrist_pass[0];
-        }
-    } else if (vertex == FastBoard::PASS) {
-        play_pass(color);
+    if (vertex == FastBoard::PASS) {
+        increment_passes();
+    } else {
+        set_passes(0);
     }
+    board.m_hash ^= Zobrist::zobrist_pass[get_passes()];
```
oni-link: When passing,  `0` and the number of passes are xored in.
When not passing, the hash contains no `0` passes modification. 
Would it not be more consistent if `calc_hash()` would also xor in "`0` passes" so that the hash would only contain the right number of passes?

sethtroisi: I think it would be nice to fix the intercoupled of zobrist logic between FastState and FullBoard, so that calc_hash() == m_hash at all times.

That being said I don't think this PR is the right place for changing it as I'd like to keep it a no-op change.

src/FullBoard.cpp
```diff
@@ -110,6 +110,11 @@ void FullBoard::set_to_move(int tomove) {
 }
 
 int FullBoard::update_board(const int color, const int i) {
+    if (i == FastBoard::PASS) {
+        // No ko
+        return 0;
```
gcp: Might be a matter of preference but I think it's cleaner to have this logic in the caller.

sethtroisi: Changed.

sethtroisi:
Slightly verified by comparing result of

(echo "auto" | ./leelaz -w ../weights.txt --noponder -p 100 -t 1 -r 0 -s 129; echo)
(echo "auto" | ./leelaz -w ../weights.txt --noponder -p 100 -t 1 -r 1 -s 123; echo)

before and after in debug mode

Issue #788, opened 2018-01-31
Time management


TFiFiE:
From the looks of it, the current time management code doesn't appear to be particularly sophisticated and as I can't find any existing thread dedicated to this subject, let me so bold as to create one and ask what plans and/or ideas there are concerning this.

Some existing literature:
https://www.remi-coulom.fr/Publications/TimeManagement.pdf
https://dke.maastrichtuniversity.nl/m.winands/documents/time_management_for_monte_carlo_tree_search.pdf

jkiliani:
The implementation of --visits is a form of time management, since it allocates more resources to positions with relatively free move choices, and less when responses are forced. So far, it hasn't been used for self-play however.

odeint:
From the paper:
> Comparison with fixed time per move. To illustrate the
effect of successful time management, two additional experiments
were conducted with OREGO using fixed time
per move in 13×13 Go. In the first experiment, the time
per move (650 msec) was set so that approximately the
same win rate against GNU GO was achieved as with EXPSTONES
and STOP at 30 seconds per game. The result of
2500 games demonstrated that the average time needed per
game was 49.0 seconds—63% more than needed by our timemanagement
strategy. In the second experiment, the time per
move (425 msec) was set so that the average time per game
was approximately equal to 30 seconds. In 2500 games under
these conditions, OREGO could only achieve a 27.6% win
rate, 11.5% less than with EXP-STONES and STOP.

To be honest, 27.6% vs 39.1% versus a benchmark doesn't seem that big a deal. That was comparing their best strategy against the pretty stupid equal-time-per move. That's just about 1-2 generations of new networks for LZ. Even AlphaGo against Lee Sedol didn't seem to use fancy time management, so DeepMind also didn't think it worth working too much on. Of course, we're all volunteers and do it for fun and learning, so if someone wants to implement time management strategies, that would still be cool!

gcp:
>The implementation of --visits is a form of time management, since it allocates more resources to positions with relatively free move choices, and less when responses are forced. So far, it hasn't been used for self-play however.

I didn't want to change this mid run, but now that all clients should have --visits capability it would be possible. But IIRC previous results were inconclusive and of course tree reuse + NNCache=TTable have a pretty big impact on those two?

gcp:
Fancy time management is nice for non-self play games, but indeed you just can't expect to get too much out of this.

One very obvious thing you can do is to terminate the search early if no other move can possibly overcome the current best.

[...]

PR #797, opened 2018-02-02
Add new option --timemanage.

killerducky:
Stop thinking when it's unlikely for the best move to be overtaken. A simple first step for #788.

I only did a very quick sanity check on this, will test more later. Looking for feedback on switch name etc.

For --visits and --playouts I think it should be impossible for this code to mistakenly stop thinking too early, assuming TM_MARGIN of 0.9f is much larger than virtual loss and other thread weirdness. For time controls there could be issues if the computer load changes while thinking, but that doesn't seem like a big problem to me.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -50,6 +50,7 @@ bool cfg_allow_pondering;
 int cfg_num_threads;
 int cfg_max_playouts;
 int cfg_max_visits;
+bool cfg_timemanage;
 int cfg_lagbuffer_cs;
 int cfg_resignpct;
 int cfg_noise;
@@ -76,6 +77,7 @@ void GTP::setup_default_parameters() {
     cfg_num_threads = std::max(1, std::min(SMP::get_num_cpus(), MAX_CPUS));
     cfg_max_playouts = std::numeric_limits<decltype(cfg_max_playouts)>::max();
     cfg_max_visits = std::numeric_limits<decltype(cfg_max_visits)>::max();
+    cfg_timemanage = false;
     cfg_lagbuffer_cs = 100;
 #ifdef USE_OPENCL
     cfg_gpus = { };
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -50,7 +50,7 @@ bool cfg_allow_pondering;
 int cfg_num_threads;
 int cfg_max_playouts;
 int cfg_max_visits;
-timemanage_t cfg_timemanage;
+UCTSearch::timemanage_t cfg_timemanage;
 int cfg_lagbuffer_cs;
 int cfg_resignpct;
 int cfg_noise;
@@ -77,7 +77,7 @@ void GTP::setup_default_parameters() {
     cfg_num_threads = std::max(1, std::min(SMP::get_num_cpus(), MAX_CPUS));
     cfg_max_playouts = std::numeric_limits<decltype(cfg_max_playouts)>::max();
     cfg_max_visits = std::numeric_limits<decltype(cfg_max_visits)>::max();
-    cfg_timemanage = AUTO;
+    cfg_timemanage = UCTSearch::AUTO;
     cfg_lagbuffer_cs = 100;
 #ifdef USE_OPENCL
     cfg_gpus = { };
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -50,7 +50,7 @@ bool cfg_allow_pondering;
 int cfg_num_threads;
 int cfg_max_playouts;
 int cfg_max_visits;
-bool cfg_timemanage;
+int cfg_timemanage;
 int cfg_lagbuffer_cs;
 int cfg_resignpct;
 int cfg_noise;
@@ -77,7 +77,7 @@ void GTP::setup_default_parameters() {
     cfg_num_threads = std::max(1, std::min(SMP::get_num_cpus(), MAX_CPUS));
     cfg_max_playouts = std::numeric_limits<decltype(cfg_max_playouts)>::max();
     cfg_max_visits = std::numeric_limits<decltype(cfg_max_visits)>::max();
-    cfg_timemanage = false;
+    cfg_timemanage = -1;
     cfg_lagbuffer_cs = 100;
 #ifdef USE_OPENCL
     cfg_gpus = { };
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -50,7 +50,7 @@ bool cfg_allow_pondering;
 int cfg_num_threads;
 int cfg_max_playouts;
 int cfg_max_visits;
-int cfg_timemanage;
+timemanage_t cfg_timemanage;
 int cfg_lagbuffer_cs;
 int cfg_resignpct;
 int cfg_noise;
@@ -77,7 +77,7 @@ void GTP::setup_default_parameters() {
     cfg_num_threads = std::max(1, std::min(SMP::get_num_cpus(), MAX_CPUS));
     cfg_max_playouts = std::numeric_limits<decltype(cfg_max_playouts)>::max();
     cfg_max_visits = std::numeric_limits<decltype(cfg_max_visits)>::max();
-    cfg_timemanage = -1;
+    cfg_timemanage = AUTO;
     cfg_lagbuffer_cs = 100;
 #ifdef USE_OPENCL
     cfg_gpus = { };
--- a/src/GTP.h
+++ b/src/GTP.h
@@ -32,7 +32,7 @@ extern bool cfg_allow_pondering;
 extern int cfg_num_threads;
 extern int cfg_max_playouts;
 extern int cfg_max_visits;
-extern bool cfg_timemanage;
+extern int cfg_timemanage;
 extern int cfg_lagbuffer_cs;
 extern int cfg_resignpct;
 extern int cfg_noise;
--- a/src/GTP.h
+++ b/src/GTP.h
@@ -26,16 +26,14 @@
 #include <vector>
 
 #include "GameState.h"
+#include "UCTSearch.h"
 
 extern bool cfg_gtp_mode;
 extern bool cfg_allow_pondering;
 extern int cfg_num_threads;
 extern int cfg_max_playouts;
 extern int cfg_max_visits;
-enum timemanage_t : int {
-    AUTO = -1, OFF = 0, ON = 1
-};
-extern timemanage_t cfg_timemanage;
+extern UCTSearch::timemanage_t cfg_timemanage;
 extern int cfg_lagbuffer_cs;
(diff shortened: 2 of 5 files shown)


gcp:
My thoughts:

1) I wouldn't bother about doing this if playouts or visits is given. We most likely are going to be moving quickly anyway. 
2) This shouldn't be a flag. If we're not running in a fixed playouts or visits mode, we always want this to be active.
3) The rationale for having a magic TM_MARGIN isn't clear.

roy7:
On 1- it's possible people could be using a large playouts or visits counter. Like a bot playing a correspondence game with +1 day fisher time control.

killerducky:
In addition to roy7's example, if we end up switching self-play and matches to --visits, we can use this option for matches. For self-play we cannot use it because we need all the visits to make the policy output. But for matches we only need to find the first move. I think for the --visits case this is guaranteed to pick the same moves (except differences due to random numbers).

The time saved is around 20%:
```

lzbase-v-2600-t-1 v lztm-v-2600-t-1 (2/1000 games)
board size: 19   komi: 7.5
                    wins              black       white       avg cpu
lzbase-v-2600-t-1      1 50.00%       0 0.00%     1 100.00%    256.61
lztm-v-2600-t-1        1 50.00%       0 0.00%     1 100.00%    208.26
                                      0 0.00%     2 100.00%
```

About TM_MARGIN, I'm aiming for a change that is close to a mathematical guarantee we pick the same move with and without this. So for the time case it allows for cases like the user's computer was running slow early, and then speeds up later (maybe 2nd leela instance was killed). For the visits and playouts case it covers virtual losses distorting Nfirst and Nsecond. Also it doesn't lock the root so the algorithm searching Nfirst and Nsecond is not atomic. I don't know how much performance we would lose from locking to do this test, but I didn't want to pay it for this heuristic.

gcp:
>I'm aiming for a change that is close to a mathematical guarantee 

No need to aim for this, especially if it makes the code simpler (or avoids magic variables that don't guarantee correctness either).

>we can use this option for matches

You misunderstand my point. This shouldn't be an option, it should be the default. If it affects self-play (by making affect visits/playouts), then unfortunately that means autogtp has to be aware of it. The match thing is a good point, that's at least a second use case in addition to @roy7's (which I was aware of when I made the original post :-)

Maybe I can cheat and change "dumbpass" to just "dumb" and have that cover all human-added heuristics. Or disable the heuristic if "-m" options are given. After all, that one will NOT work correctly when this is enabled.

killerducky:
Magic number: Ok I will remove it. One other option would be `Nfirst - Nsecond + threads*VIRTUAL_LOSS_COUNT` to cover that case. More correct and doesn't introduce any *new* magic variables. But I'll assume you want the simple version.

For defaults: how about default off when `-m` is given, default on when is is not. But we keep a switch to allow the user to override.

If this change goes in new clients will start playing matches with the cutoff, and if the matches still use `-p` it will change the results (push less moves into the tree for reuse). It will change for both sides so maybe it's not a problem.

roy7:
I can see this working like expected on forced moves with -v, but with -p say we have 4 forced moves coming up, -p 1600, and 1000 search is in the PV. So we move since the PV can't change. This will apply for all 4 of the next moves in the PV forced move chain. But that's 600 + 6400 playouts we've skipped where normally we'd have done more search each step of the way. Is this a problem?

roy7:
Please note my thought here:

https://github.com/gcp/leela-zero/issues/696#issuecomment-362729180

That stopping search early will change dump_training's results and bias learning against moves discovered to be good but too late to actually make the move. Ie: Early stop at 80%/20% move split vs normal stop at 55%/45% move split. First move still chosen, but learning stage will see the 2nd move as a high importance prior.

killerducky:
@roy7 yes and it also effects what `-m` does. That's why we cannot use timemanage mode this in self-play games.

Similar for -p mode and a forced sequence. This mode *will* change the moves selected for -p mode because it changes how many moves are held over in the tree reuse. For -v mode it will not change the moves selected, but it will change the training outputs. That's why the current plan is to make the timemanage mode dependent on `-m`, which is always used in self-play.

killerducky:
@gcp ok updated mostly as you asked, except I left --timemanage to allow the user to override the auto behavior.

I also added a description of what `--resignpct -1` means.

src/GTP.cpp
```diff
@@ -50,6 +50,7 @@ bool cfg_allow_pondering;
 int cfg_num_threads;
 int cfg_max_playouts;
 int cfg_max_visits;
+int cfg_timemanage;
```
gcp: This should probably be a enum rather than magic numbers.

killerducky: Ok I'll make it an enum, something like AUTO, ON, OFF. Should the command line take string arguments `--timemanage [on|off|auto]`?

gcp: I think that's prettier than the magic -1 = auto yes.

killerducky:
I did some more testing, including a match at `-v 800`. It should be exactly the same strength, but run faster. Looks like that's true within statistical noise:

```
lzbase-v-800-t-1 v lztm-v-800-t-1 (550/1000 games)
board size: 19   komi: 7.5
lzbase-v-800-t-1    258 46.91%       128 46.55%     130 47.27%     82.27
lztm-v-800-t-1      292 53.09%       145 52.73%     147 53.45%     63.82
                                     273 49.64%     277 50.36%
```

src/UCTSearch.cpp
```diff
@@ -352,8 +352,48 @@ bool UCTSearch::is_running() const {
...
+        auto time_left = time_for_move - elapsed_centis;
+        auto est_playouts_left = playout_rate * time_left;
+        if (est_playouts_left < Nfirst - Nsecond) {
+            stop = true;
+            if (!cfg_quiet) {
+                myprintf("%.1fs left\n", time_left/100.0f);
+            }
+        }
+    }
+    if (stop && !cfg_quiet) {
```
gcp: same

src/UCTSearch.cpp
```diff
@@ -352,8 +352,48 @@ bool UCTSearch::is_running() const {
...
+    if ((m_maxplayouts - playouts < Nfirst - Nsecond)
+        || (m_maxvisits - visits < Nfirst - Nsecond)) {
+        stop = true;
+    } else if (elapsed_centis > 0) {
+        auto playout_rate = 1.0f * playouts / elapsed_centis;
+        auto time_left = time_for_move - elapsed_centis;
+        auto est_playouts_left = playout_rate * time_left;
+        if (est_playouts_left < Nfirst - Nsecond) {
+            stop = true;
+            if (!cfg_quiet) {
```
gcp: myprintf already takes cfg_quiet into account.

src/GTP.h
```diff
@@ -32,6 +32,10 @@ extern bool cfg_allow_pondering;
 extern int cfg_num_threads;
 extern int cfg_max_playouts;
 extern int cfg_max_visits;
+enum timemanage_t : int {
```
gcp: I'd move the definition of timemanage_t elsewhere. UCTSearch.h perhaps?

killerducky: done

src/UCTSearch.cpp
killerducky: done

src/UCTSearch.cpp
killerducky: done

killerducky:
@bood do you know why the appveyor build failed here?

```
    5>OpenBLAS_AfterBuild:
         Creating hard link to copy "C:\projects\leela-zero\pkgs\OpenBLAS.0.2.14.1\build\native\..\..\lib\native\bin\x64\libgcc_s_seh-1.dll" to "C:\projects\leela-zero\build\Release\libgcc_s_seh-1.dll".
     4>OpenBLAS_AfterBuild:
         Creating hard link to copy "C:\projects\leela-zero\pkgs\OpenBLAS.0.2.14.1\build\native\..\..\lib\native\bin\x64\libgcc_s_seh-1.dll" to "C:\projects\leela-zero\build\Release\libgcc_s_seh-1.dll".
     5>OpenBLAS_AfterBuild:
         Creating hard link to copy "C:\projects\leela-zero\pkgs\OpenBLAS.0.2.14.1\build\native\..\..\lib\native\bin\x64\libgfortran-3.dll" to "C:\projects\leela-zero\build\Release\libgfortran-3.dll".
         Creating hard link to copy "C:\projects\leela-zero\pkgs\OpenBLAS.0.2.14.1\build\native\..\..\lib\native\bin\x64\libopenblas.dll" to "C:\projects\leela-zero\build\Release\libopenblas.dll".
         Creating hard link to copy "C:\projects\leela-zero\pkgs\OpenBLAS.0.2.14.1\build\native\..\..\lib\native\bin\x64\libquadmath-0.dll" to "C:\projects\leela-zero\build\Release\libquadmath-0.dll".
       FinalizeBuildStatus:
         Deleting file "tests.dir\Release\tests.tlog\unsuccessfulbuild".
         Touching "tests.dir\Release\tests.tlog\tests.lastbuildstate".
     5>Done Building Project "C:\projects\leela-zero\build\tests.vcxproj" (default targets).
     4>OpenBLAS_AfterBuild:
         Could not use a link to copy "C:\projects\leela-zero\pkgs\OpenBLAS.0.2.14.1\build\native\..\..\lib\native\bin\x64\libgcc_s_seh-1.dll" to "C:\projects\leela-zero\build\Release\libgcc_s_seh-1.dll". Copying the file instead. Cannot create a file when that file already exists. (Exception from HRESULT: 0x800700B7)
         Copying file from "C:\projects\leela-zero\pkgs\OpenBLAS.0.2.14.1\build\native\..\..\lib\native\bin\x64\libgcc_s_seh-1.dll" to "C:\projects\leela-zero\build\Release\libgcc_s_seh-1.dll".
     4>
```

gcp:
Looks like a race in CI. I retriggered it.

gcp:
I pulled the enum out of the UCTSearch class and into its own namespace. I think that looks prettier especially to avoid the class-wide ON/OFF/AUTO enums.

zediir:
This is also dropping my n/s from 800 to 380 when running  `-w leelaz-model-3008000.txt -r1 -t4 -g --gpu 0 --gpu 1`. Both in linux and windows. Both when I have timesettings on and off. Tested by reverting the commit.

killerducky:
@zediir when you said timesettings off, do you mean `--timemanage off`?

zediir:
Yes. Was a typo that I edited.

killerducky:
Can you paste the logfile here? Did it do only 1 playout? That is fixed in #858.

killerducky:
Ok nevermind I've reproduced it here. Looking into it now.

PR #799, opened 2018-02-02
Refactor find_new_root

sethtroisi:
**Improvements**
* Can now find new_root at >1 depth,
* Uses m_rootstate history to find moves
* Prints stats if #ifndef NDEBUG
* Supports 100% tree reuse after a single undo

tested with 

    make -j8 debug
    echo -e "time_settings 0 8 1\ngo\ngo\ngo\ngo\nkomi 8\ngo\nclear_board\ngo\nplay white D4\nplay black R4\nplay white D16\ngo\nundo\ngo" | ./leelaz -w ../weights.txt -s 123 -t 1; echo

|command|explanation|output|
|---|---|---|
|go|no tree reuse on first move||
|go||Leela: update_root Q16, 2857321 -> 423165 nodes (14.8% reused)|
|go||Leela: update_root D4, 2724778 -> 936392 nodes (34.4% reused)|
|go||Leela: update_root R4, 2865117 -> 959579 nodes (33.5% reused)|
|komi 8|||
|go|no reuse because of komi changed||
|clear_board|||
|go|no tree reuse because of clearing board||
|play white D4||
|play black R4||
|play white D16||
|go|tree reuse from depth 3(4?)|Leela: update_root Q16 D4 R4 D16, 5833086 -> 26126 nodes (0.4% reused)|
|go||Leela: update_root F3, 4919913 -> 2289485 nodes (46.5% reused)|
|undo|tree is still usable because of not automatically playing move in in think||
|go||Leela: update_root , 4312540 -> 4312540 nodes (100.0% reused)|

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -410,8 +410,8 @@ size_t UCTNode::count_nodes() const {
     return nodecount;
 }
 
-// Use this version if you know the child is directly under the parent.
-UCTNode::node_ptr_t UCTNode::find_new_root(const int move) {
+// Used to find new root in UCTSearch
+UCTNode::node_ptr_t UCTNode::find_child(const int move) {
     if (m_has_children) {
         for (auto& child : m_children) {
             if (child->get_move() == move) {
@@ -419,44 +419,9 @@ UCTNode::node_ptr_t UCTNode::find_new_root(const int move) {
             }
         }
     }
-    // Can happen for example if we resigned. Return a clean
-    // root for the next game or position.
-    return std::make_unique<UCTNode>(FastBoard::PASS, 0.0f, 0.5f);
-}
 
-// Use this version if the child could be anywhere.
-void UCTNode::find_new_root(node_ptr_t& root,
-                            const GameState& g_curr,
-                            std::unique_ptr<GameState>&& g_old) {
-    auto found = false;
-
-    if (g_old) {
-        if (g_curr.get_komi() == g_old->get_komi()) {
-            if (g_curr.board.get_hash() == g_old->board.get_hash()) {
-                // root is already set correctly
-                found = true;
-            } else {
-                // search the direct children
-                for (auto& child : root->m_children) {
-                    auto move = child->get_move();
-                    if (g_curr.get_last_move() == move) {
-                        g_old->play_move(move);
-                        if (g_curr.board.get_hash()
-                            == g_old->board.get_hash()) {
-                            root = std::move(child);
-                            found = true;
-                            break;
-                        }
-                        g_old->undo_move();
-                    }
-                }
-            }
-        }
-    }
-
-    if (!found) {
-        root = std::make_unique<UCTNode>(FastBoard::PASS, 0.0f, 0.5f);
-    }
+    // Can happen for example if we resigned.
+    return nullptr;
 }
 
 UCTNode* UCTNode::get_nopass_child(FastState& state) const {
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -420,7 +420,7 @@ UCTNode::node_ptr_t UCTNode::find_child(const int move) {
         }
     }
 
-    // Can happen for example if we resigned.
+    // Can happen if we resigned or children are not expanded
     return nullptr;
 }
 
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -46,77 +46,79 @@ UCTSearch::UCTSearch(GameState& g)
     m_root = std::make_unique<UCTNode>(FastBoard::PASS, 0.0f, 0.5f);
 }
 
+bool UCTSearch::advance_to_new_rootstate() {
+    if (!m_root || !m_last_rootstate) {
+        // No current state
+        return false;
+    }
+
+    if (m_rootstate.get_komi() != m_last_rootstate->get_komi()) {
+        return false;
+    }
+
+    auto depth =
+        (int) (m_rootstate.get_movenum() - m_last_rootstate->get_movenum());
+
+    if (depth < 0) {
+        return false;
+    }
+
+    auto test = std::make_unique<GameState>(m_rootstate);
+    for (auto i = 0; i < depth; i++) {
+        test->undo_move();
+    }
+
+    if (m_last_rootstate->board.get_ko_hash() != test->board.get_ko_hash()) {
+        // m_rootstate and m_last_rootstate don't match
+        return false;
+    }
+
+    // Try to replay moves advancing m_root
+    for (auto i = 0; i < depth; i++) {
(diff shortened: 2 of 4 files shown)


src/UCTNode.cpp
```diff
@@ -410,53 +410,18 @@ size_t UCTNode::count_nodes() const {
     return nodecount;
 }
 
-// Use this version if you know the child is directly under the parent.
-UCTNode::node_ptr_t UCTNode::find_new_root(const int move) {
+// Used to find new root in UCTSearch
+UCTNode::node_ptr_t UCTNode::find_child(const int move) {
```
sethtroisi: This might be better name 'return_child' or 'find_child_destructive' to indicate that it destroys part of the subtree

src/UCTNode.cpp
```diff
@@ -410,53 +410,18 @@ size_t UCTNode::count_nodes() const {
...
-                    }
-                }
-            }
-        }
-    }
 
-    if (!found) {
-        root = std::make_unique<UCTNode>(FastBoard::PASS, 0.0f, 0.5f);
-    }
+    // Can happen for example if we resigned.
```
gcp: I guess there's quite some more cases now where this is true.

src/UCTSearch.cpp
```diff
@@ -49,12 +50,71 @@ void UCTSearch::update_root() {
     // Definition of m_playouts is playouts per search call.
     // So reset this count now.
     m_playouts = 0;
-    // See if the current state of the game is a direct succesor of
-    // our stored tree's root node. If so, make that the new root.
-    UCTNode::find_new_root(m_root, m_rootstate, std::move(m_last_rootstate));
-    // If the above succeeded, we'll have a search tree already. See
-    // how big it is.
+
+    bool found = false;
```
gcp: nit: auto all the things

src/UCTSearch.cpp
```diff
@@ -49,12 +50,71 @@ void UCTSearch::update_root() {
...
+    #ifndef NDEBUG
```
gcp: nit: indent this to the left

src/UCTSearch.cpp
```diff
@@ -49,12 +50,71 @@ void UCTSearch::update_root() {
...
+            // Verify m_rootstate and m_last_rootstate have same history.
+            if (m_last_rootstate->board.get_ko_hash()
+                    != test->board.get_ko_hash()) {
+                break;
+            }
+
+            test->forward_move();
+            auto move = test->get_last_move();
+            m_root = m_root->find_child(move);
+            if (!m_root->has_children()) {
```
gcp: You will crash here if it's a nullptr.

src/UCTSearch.cpp
```diff
@@ -49,12 +50,71 @@ void UCTSearch::update_root() {
...
+                break;
+            }
+            m_last_rootstate->play_move(move);
+            # ifndef NDEBUG
+            replay += m_last_rootstate->board.move_to_text(move) + " ";
+            # endif
+        }
+
+        // Check if advancing last_rootstate was possible.
+        if (m_root->has_children()
```
gcp: I think this separate check can go away. The code that finds the match and sets the new root can set found. If it did not, nothing was found.

src/UCTSearch.cpp
sethtroisi: I think you always hit an expanded node first but it's easy to add.

src/UCTSearch.cpp
sethtroisi: I think I would lose the last ko_hash check if I do it that way + I have to set found = true at the start of the if which isn't as pretty to me.

src/UCTSearch.cpp
sethtroisi: Done.

src/UCTSearch.cpp
sethtroisi: done

src/UCTSearch.cpp
sethtroisi: I thought about this some more.

I believe I only need the first ko hash check, if a position has the same ko_hash then you play a series of identical moves they will have the same end ko_hash.

But this made me realize that this potentially returns trees which are transposes of the real position.

if you
play Q16, D4, R4, D16
undo x4
play R4, D4, Q16, D16
it will reuse the nodes (but they will have been generate with the moves in a different order).

This isn't an issue for the autogtp, or any of the bots but for humans doing self play it can technically happen.

gcp: Yes, that would be undesirable.

sethtroisi: I found a case where this can fail.

If the user has entered two black moves in a row the node will find the child but the hashes won't match :)

src/UCTSearch.cpp
```diff
@@ -45,16 +46,79 @@ UCTSearch::UCTSearch(GameState& g)
...
+        // No current state
+        return false;
+    }
+
+    if (m_rootstate.get_komi() != m_last_rootstate->get_komi()) {
+        return false;
+    }
+
+    auto depth =
+        (int) (m_rootstate.get_movenum() - m_last_rootstate->get_movenum());
```
sethtroisi: I believe this works but It's not as pretty as I'd like, a quick Google didn't suggest anything better.

src/UCTSearch.cpp
```diff
@@ -45,16 +46,79 @@ UCTSearch::UCTSearch(GameState& g)
...
+        if (!m_root) {
+            // Tree hasn't been expanded this far
+            return false;
+        }
+        m_last_rootstate->play_move(move);
+    }
+
+    assert(m_rootstate.get_movenum() == m_last_rootstate->get_movenum());
+
+    if (m_last_rootstate->board.get_ko_hash() != test->board.get_ko_hash()) {
```
gcp: You pointed out earlier this will hit even if there's a transposition, and that was a valid concern.

At the very least, it should use hash, not ko-hash, because the latter does not hash ko squares. (This is perhaps less confusing if one remembers that ko-hash is the hash used for detecting ko, not encoding it)

src/UCTSearch.cpp
gcp: Agree, this kinda sucks in C++.

src/UCTSearch.cpp
sethtroisi: I think it's not a problem in practice as it can only happens if you have to have multiple 4+ undos (which bots and self play don't have) and the consequence is tiny on the infrequent occasions it happens (eval for that position A will have been generated for position A with a different history).

Also it's not a regression from the current code (which also uses get_hash to verify positions are the same but doesn't check that all of m_lastroot moves match m_root moves).

If you still think it's a problem I suggest that we clear UCTSearch on "undo" or verify all move history between m_root and m_lastroot.

PR #801, opened 2018-02-02
add newline on normal exit

sethtroisi:
When file input is used with Leela, and leelaz exits normally the prompt ("Leela: ") without a newline will often be the last output. this shifts the bash prompt and if you press up for the last command and that command is longer than one line it will put [bash in a weird state](https://askubuntu.com/a/24422/767394) because it thinks the command should wrap at a different place.

This prints a newline before exiting in some cases

|command|behavior|
|---|---|
| exit | unchanged (no newline) |
| runtime error | unchanged |
| quit | print an extra, unneeded newline |
| pipe | print a newline for "Leela:" line|
| file redirect | print a newline for "Leela:" line|

This does print the extra newline even in GTP mode but I'm assuming a training newline at EXIT isn't going to break anything (I did test sabaki)

### DIFF
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -306,6 +306,7 @@ int main (int argc, char *argv[]) {
             GTP::execute(*maingame, input);
         } else {
             // eof or other error
+            std::cout << std::endl;
             break;
         }
 


src/Leela.cpp
```diff
@@ -316,5 +316,6 @@ int main (int argc, char *argv[]) {
         }
     }
 
+    std::cout << std::endl;
```
gcp: Maybe it should go in the "eof or other error" branch? That's the case you're hitting, right?

I checked GTP quit vs exit, and:
https://github.com/gcp/leela-zero/blob/47ddf38327b80ec5f2303344d1d9c65e25f5a3fb/src/GTP.cpp#L209

https://github.com/gcp/leela-zero/blob/47ddf38327b80ec5f2303344d1d9c65e25f5a3fb/src/GTP.cpp#L233

So they won't hit this case. quit is handled correctly and it needs to be according to: 
"The full response of this command must be sent before the                 engine closes the connection. The controller must receive                 the response before the connection is closed on its side."

Maybe "exit" should just be removed, OTOH it's handy if someone starts the binary and gets the typical 'how do I get out?!' reaction.

sethtroisi: I moved it under the "eof or other error" branch, which probably covers 90% of my problems (control-c being the others and that's harder to catch/fix).

this is a good compromise (fixes most of this problem and doesn't affect GTP at all)

PR #804, opened 2018-02-03
Rename Utils::lcm and add test

sethtroisi:
sedReplace "lcm" "ceilMultiple"

### DIFF
--- a/src/tests/utils_unittest.cpp
+++ b/src/tests/utils_unittest.cpp
@@ -1,0 +1,45 @@
+/*
+    This file is part of Leela Zero.
+    Copyright (C) 2018 Gian-Carlo Pascutto
+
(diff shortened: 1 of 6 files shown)


Ttl:
lcm is not actually least common multiple function. It's needed in OpenCL functions to round arrays to multiple of work group. It should return smallest number that is bigger or equal to the first argument and divisible by the second one. The original functionality was correct for that purpose. It should probably be renamed instead.

src/tests/utils_unittest.cpp
```diff
@@ -0,0 +1,45 @@
+/*
+    This file is part of Leela Zero.
+    Copyright (C) 2018 Gian-Carlo Pascutto
```
gcp: Probably want to put your own name here next time.

PR #805, opened 2018-02-03
Faster network parsing and error checking

Ttl:
Replaces stringstream with faster Boost Spirit X3 parser.

Previously giving invalid weight file caused segmentation fault. The new parser checks for errors.

Benchmark:
```
time ./leelaz -w ../autogtp/257aeeb863dc51bfc598838361225459257377a4b2c9abd3e1ac6cdba1fcc88f -t 1 <<< "exit"

Previous:
real	0m1.004s
user	0m0.876s
sys	0m0.276s

X3:
real	0m0.487s
user	0m0.304s
sys	0m0.248s
```

### DIFF
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -30,6 +30,7 @@
 #include <string>
 #include <boost/utility.hpp>
 #include <boost/format.hpp>
+#include <boost/spirit/home/x3.hpp>
 
 #ifdef __APPLE__
 #include <Accelerate/Accelerate.h>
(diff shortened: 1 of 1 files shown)


l1t1:
may be we could use fixed length number format

gcp:
FWIW I would like to limit dependencies to what is reasonably easy to install on Ubuntu LTS (with some leeway if it's just been released).

But it seems this was already in Boost 1.58.0. (It's a bit confusing because "Spirit v2.5" seems to include X3.

PR #806, opened 2018-02-03
Zobrist cleanup

sethtroisi:
* Add a new constant for empty zobrist state,
* Add a new method in Random to getting a uint64_t to cleanup Zobrist init.
* Clean up a warning, some new lines, and a less obvious code pattern

### DIFF
--- a/src/FullBoard.cpp
+++ b/src/FullBoard.cpp
@@ -56,7 +56,7 @@ int FullBoard::remove_string(int i) {
 }
 
 std::uint64_t FullBoard::calc_ko_hash(void) {
-    auto res = std::uint64_t{0x1234567887654321ULL};
+    auto res = Zobrist::zobrist_empty;
 
     for (int i = 0; i < m_maxsq; i++) {
         if (m_square[i] != INVAL) {
@@ -64,14 +64,13 @@ std::uint64_t FullBoard::calc_ko_hash(void) {
         }
     }
 
-    m_ko_hash = res;
-
     /* Tromp-Taylor has positional superko */
+    m_ko_hash = res;
     return res;
 }
 
 std::uint64_t FullBoard::calc_hash(int komove) {
-    auto res = std::uint64_t{0x1234567887654321ULL};
+    auto res = Zobrist::zobrist_empty;
 
     for (int i = 0; i < m_maxsq; i++) {
         if (m_square[i] != INVAL) {
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -282,7 +282,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
     } else if (command.find("clear_board") == 0) {
         Training::clear_training();
         game.reset_game();
-        std::make_unique<UCTSearch>(game).swap(search);
+        search = std::make_unique<UCTSearch>(game);
         gtp_printf(id, "");
         return true;
     } else if (command.find("komi") == 0) {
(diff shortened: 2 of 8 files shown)


src/FullBoard.cpp
```diff
@@ -132,7 +131,7 @@ int FullBoard::update_board(const int color, const int i) {
     auto eyeplay = (m_neighbours[i] & s_eyemask[!color]);
 
     auto captured_stones = 0;
-    int captured_sq;
+    auto captured_sq = -1;
```
gcp: This is not required. (Despite what the stupid compiler says).

src/GTP.cpp
```diff
@@ -282,7 +282,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
     } else if (command.find("clear_board") == 0) {
         Training::clear_training();
         game.reset_game();
-        std::make_unique<UCTSearch>(game).swap(search);
+        search.reset(new UCTSearch(game));
```
gcp: make_unique, never "new".

src/Zobrist.cpp
```diff
@@ -25,28 +25,30 @@ std::array<std::uint64_t, FastBoard::MAXSQ>                    Zobrist::zobrist_
 std::array<std::array<std::uint64_t, FastBoard::MAXSQ * 2>, 2> Zobrist::zobrist_pris;
 std::array<std::uint64_t, 5>                                   Zobrist::zobrist_pass;
 
+std::uint64_t getRandHash(Random& rng) {
```
gcp: Let's add this to Random() instead. Additionally adding a randuint64 that takes an argument (for consistency with the others) is trickier to get right but we can just rely on the C++11 random lib and pass our own gen as UniformRandomBitGenerator.

src/Zobrist.h
```diff
@@ -28,11 +28,13 @@
 
 class Zobrist {
 public:
+    static constexpr auto zobrist_empty = 0x1234567887654321ULL;
```
gcp: Not 100% sure but I think as long as there's an 0x prefix then the ULL part is not actually required.

src/GTP.cpp
sethtroisi: Fixed, re-reading it appears reset(new ...) is a c++11 suggestion from before make_unique.

src/FullBoard.cpp
sethtroisi: I know it's not required but I'm a big fan of reducing warnings so that when they do point to real bugs people pay attention to them.

src/Zobrist.cpp
sethtroisi: Moved the current zero arg implementation to Random, this passes the current tests.

I'd prefer to do a 2nd check in that changes the random algorithm and tests (so that it's clean what changed the test and doesn't cast doubt on bad casts in this commit)

src/Zobrist.h
sethtroisi: fixed

src/FullBoard.cpp
gcp: We can disable that compiler warning. It's even generating warnings in boost code for me, to clearly it's a bit too spammy.

src/Random.cpp
```diff
@@ -68,6 +68,13 @@ std::uint32_t Random::randuint32() {
     return gen() >> 32;
 }
 
+std::uint64_t Random::randuint64() {
+    // Current implementation for consistency with old zobrist method.
```
gcp: No real need to have consistency with the old one?

sethtroisi:
Thank you for writing the follow up change (30c1eacd573c2b4d06800aaea6665e84cefd73ef)

PR #811, opened 2018-02-04
Use pinned memory for device -> host transfers

Ttl:
Currently getting data out of the GPU is the biggest bottleneck. This transfer can be made faster with pinned (mapped) memory. Just replacing the transfer gives about 6 % performance increase in netbench.

I also moved the 1x1 convolutions of the output heads to GPU to reduce the amount of memory that needs to be moved. That increased performance gain to 15 % compared to the original.

```
Current:
10000 evaluations in 14.40 seconds -> 694 n/s

Pinned memory:
10000 evaluations in 13.48 seconds -> 741 n/s

Pinned memory + OpenCL head convolutions:
10000 evaluations in 12.41 seconds -> 805 n/s
```

It still seems that the memory transfer is limiting the performance. If I comment it out I get `10000 evaluations in 11.47 seconds -> 871 n/s`. Host -> device memory transfer doesn't seem to be as important.

### DIFF
--- a/src/OpenCL.cpp
+++ b/src/OpenCL.cpp
@@ -51,6 +51,100 @@ static std::string sourceCode_config = R"(
     #define vstore_net_t(data,offset,p) (((p)[(offset)])=(data))
 )";
 
+static std::string sourceCode_convolve1 = R"(
+    __kernel
+    __attribute__((work_group_size_hint(8, 16, 1)))
+    void convolve1(
+                   __global const net_t * in,
+                   __global net_t * merge,
+                   __global const net_t * weights,
+                   __local float * channel_buff,
+                   __local float * row_buff) {
+        // cl::NDRange global(channels, outputs, row);
+        const int c   = get_global_id(0);  // channel
+        const int o   = get_global_id(1);  // output
+        const int row = get_global_id(2);  // row
+        const int channels = get_global_size(0);
+        const int outputs  = get_global_size(1);
+        // cl::NDRange local(2, (1->32), 1);
+        const int lx = get_local_id(0);
+        const int ly = get_local_id(1);
+        const int chan_buff_size = 8;
+        const int out_buff_size  = get_local_size(1);
+        const int row_buff_size  = 7;
+        const int chan_shift     = 3;
+        // input = channels * height * width
+        // output = outputs * height * width
+        // weights = output * channels * filter
+        // merge = channels * outputs * height * width
+        const int width = 19;
+        const int height = 19;
+        const int strip_size = width;
+        // Copy the input channels (strips) locally
+        if (out_buff_size < 19 && ly == 0) {
+            // strip-row
+            for (int w = 0; w < width; w++) {
+                channel_buff[lx * width + w] =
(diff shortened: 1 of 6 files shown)


zediir:
Windows 10 GTX 1080

```
1600 playouts current net -s 1:
Current:
1291 n/s
Your pinned_memory branch:
3202 n/s
```

Whoah.

zediir:
I first got something like 1200 n/s -> 3000 n/s improvement on windows when comparing my next build and your branch but when I took the latest changes from next and build leelaz again it was at around 3000 n/s also so something else has had a huge improvement for windows builds.

Ttl:
The latest commit is technically unrelated to others, but I didn't want to open yet another pull request. It adds output buffer for convolve3 instead of overwriting input allowing the input to be saved for residual connection. Slight performance increase:

`10000 evaluations in 12.22 seconds -> 818 n/s`

src/OpenCL.cpp
```diff
@@ -448,16 +453,22 @@ void OpenCL_Network::forward(const std::vector<net_t>& input,
...
 
-    const auto finalSize = m_layers.back().outputs * one_plane;
-    queue.enqueueReadBuffer(inBuffer, CL_FALSE, 0, finalSize, output.data());
+    auto pinnedOutBufferHost = queue.enqueueMapBuffer(
+        opencl_thread_data.m_pinnedOutBuffer, CL_FALSE,
+        CL_MAP_READ, 0, finalSize);
 
     std::lock_guard<std::mutex> lock(m_queue_finish_mutex);
     queue.finish();
+
```
ihavnoid: The original intention was that we unlock m_queue_finish_mutex right after queue.finish() -  the mutex was there for preventing too many threads from waiting on the finish() call (which is a busy wait on many OpenCL implementations)

gcp:
>when I took the latest changes from next and build leelaz again it was at around 3000 n/s also so something else has had a huge improvement for windows builds.

Probably fused input/output transforms.

src/OpenCL.cpp
```diff
@@ -367,12 +468,21 @@ void OpenCL_Network::forward(const std::vector<net_t>& input,
             m_ceil * m_ceil *  max_channels * sizeof(net_t);
         const auto alloc_vm_size =
             WINOGRAD_TILE * m_ceil * n_ceil * sizeof(net_t);
+
+#ifndef NDEBUG
+        //Check that VBuffer can be used as a merge buffer for convolve1
```
gcp: So uh, in which cases would this not be true (and hence we crash/give wrong output) in release?

Ttl: It's always enough. mergebuffer needs to be max_channels/4 planes, while Vbuffer is always at least max_channel planes. I'll remove this check.

alreadydone:
> > when I took the latest changes from next and build leelaz again it was at around 3000 n/s also so something else has had a huge improvement for windows builds.
> 
> Probably fused input/output transforms, which lower the amount of calls into the driver.

On Big Red II it goes from 4.5 s/move to 3.6 s/move.

PR #812, opened 2018-02-04
Fix ignored biases

Ttl:
Newer networks have zero biases since they are cancelled by the batchnorm.
However some older networks, such as the 6x128 human best network have non-zero
biases. Biases can be moved to batchnorm means to fix the output.

This fixes the mysterious different output of human best 6x128 network with Winograd convolution. Now the heatmap matches the previous implementation.


gcp:
Nice find!

PR #816, opened 2018-02-05
Small cleanup to visit counting

sethtroisi:
if `UCTSearch::play_simulation` expands a leaf node it counts it as a visit. 
if the root node is a leaf node it gets expanded in `UCTSearch::think` but isn't counted as a visit.

I unified this by counting the initial "visit" in `UCTSearch::think` which is a much smaller change because it only happens when playouts is TINY or when opponent play weird moves that weren't expected.

After this behavior is unified I'll write a 2nd change ([preliminary draft](https://github.com/sethtroisi/leela-zero/compare/c__tt_cleanup...sethtroisi:c__tt_cleanup_v2)) to not count initial expansion as visits per "Mastering the game of Go" which says "Edge edge traversed in the simulation is updated to increment its visit count" ie leaf nodes have zero visits but this requires updating `get_pv` and `get_best_move`. In the meantime this is accomplished by subtracting 1 from visit counts in `uct_select_node`

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -208,16 +208,17 @@ void UCTNode::dirichlet_noise(float epsilon, float alpha) {
 }
 
 void UCTNode::randomize_first_proportionally() {
-    auto accum = std::uint32_t{0};
+    auto accum = 0;
     auto accum_vector = std::vector<decltype(accum)>{};
     for (const auto& child : m_children) {
         accum += child->get_visits();
         accum_vector.emplace_back(accum);
     }
+    assert(accum + 1 == get_visits());
 
-    auto pick = Random::get_Rng().randuint32(accum);
+    auto pick = int(Random::get_Rng().randuint32(accum));
     auto index = size_t{0};
-    for (size_t i = 0; i < accum_vector.size(); i++) {
+    for (auto i = size_t{0}; i < accum_vector.size(); i++) {
         if (pick < accum_vector[i]) {
             index = i;
             break;
@@ -254,10 +255,6 @@ void UCTNode::update(float eval) {
 
 bool UCTNode::has_children() const {
     return m_has_children;
-}
-
-void UCTNode::set_visits(int visits) {
-    m_visits = visits;
 }
 
 float UCTNode::get_score() const {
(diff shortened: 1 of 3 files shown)


sethtroisi:
Additional verification by

comparing 
`(yes "go" | head -n50) | ./leelaz -w ../weights.txt -t1 -r0 -p 1000 --noponder -s 20 |& tee visit_change_1000`
(yes "go" | head -n50) | ./leelaz-next-d53ede -w ../weights.txt -t1 -r0 -p 1000 --noponder -s 20 |& tee next_branch_1000

and loadsgf super_ko_test.sgf and making sure the assertion didn't fire

src/UCTNode.cpp
```diff
@@ -208,16 +208,17 @@ void UCTNode::dirichlet_noise(float epsilon, float alpha) {
 }
 
 void UCTNode::randomize_first_proportionally() {
-    auto accum = std::uint32_t{0};
+    auto accum = 0;
```
gcp: This really wants to be a 64-bit but it requires a randuint64(arg), see the remarks to the Zobrist pull.

I guess our current tree limit means we can't hit this being > 32-bit signed int anyway.

gcp:
Note that this will require re-tuning PUCT. Not sure if this change will end up being worth it.

src/UCTNode.cpp
sethtroisi: we use int which would be like 2 billion nodes or ~10 million playouts which I think is more than anyone is doing right now :)

I'll work on randuint64(arg) tomorrow

sethtroisi:
I believe in practice this is a no-op, in self-play mode I believe it adds a single visit to the initial root (tree reuse means that no other nodes should have 0 visits at the start of think) and I verified the heatmap is unchanged for the first move

gcp:
The reason I'm saying that is that the UCT formula has child()->get_visits() + 1 in it, which will have a different slope now.

sethtroisi:
No, if you look at line 325 I assert that the count used by UCT formula is the same (I do have to add subtract one to get it to be the same)

Technically `first_visit` on the root_node might be affected by this, but that's only in the case that there was no tree reuse (bots, undo, komi change, highly unexpected move, low playouts?) and first_visit is only used by get_no_pass_child (i.e. not on the root node)

gcp:
I ended up reverting this because of the issue with the assertion.

This drew my attention to another problem, which exists with the current code: if multiple threads entered subnodes of this node, and are applying virtual loss, I think we could end up doing sqrt(negative number).

gcp:
Actually no, that probably can't happen because the virtual loss score is kept separately, so get_visits() is always sane.

gcp:
Probably safe to reland with just the asserts gone and rebase to your Random() changes.

PR #817, opened 2018-02-05
simplify get_best_move

sethtroisi:
use more generic get_eval and removes the only non-debug uses of first_visit.

This code dates back to GitHub inception prior to FPU changes.

This increases the likelyhood of the initial eval of the no_pass child being less than the resign threshold and Leela resigning*, this can already happen if the no_pass child node has visits.

This only changes behavior in the rare case that search spent **all** of its playouts evaluating PASS and the 2nd move had such a low initial eval that it's resign-able. This frequently happens when playouts = 1, less frequently with p = 2, 3, ...

*You might consider preventing Leela from resigning if NO_PASS is set because it only happens in gen_handicap_moves and kgs_cleanup.

**Pros**
*  Doesn't affect self-play (as dumppass is used)
*  Seth's OCD is improved

**Cons**
*  Changes resign likelihood in `kgs_cleanup` command
*  Changes behavior when `playouts = 1` and top move was pass

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -227,22 +227,17 @@ int UCTSearch::get_best_move(passflag_t passflag) {
     if (passflag & UCTSearch::NOPASS) {
         // were we going to pass?
         if (bestmove == FastBoard::PASS) {
-            UCTNode * nopass = m_root->get_nopass_child(m_rootstate);
-
+            auto nopass = m_root->get_nopass_child(m_rootstate);
             if (nopass != nullptr) {
                 myprintf("Preferring not to pass.\n");
                 bestmove = nopass->get_move();
-                if (nopass->first_visit()) {
-                    bestscore = 1.0f;
-                } else {
-                    bestscore = nopass->get_eval(color);
-                }
+                bestscore = nopass->get_eval(color);
             } else {
                 myprintf("Pass is the only acceptable move.\n");
             }
         }
-    } else {
-        if (!cfg_dumbpass && bestmove == FastBoard::PASS) {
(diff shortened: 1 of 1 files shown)


gcp:
Note that these "rare" cases are not rare at all if you play with a single playout, i.e. with a "network only" setup.

gcp:
Also note that the "low initial eval" in that case is dependent on the FPU reduction. With the default value a winrate of say 25% at the root might cause us to resign if we try to pass and it loses on the count. That's probably borderline OK.

But I think in general this idea is flawed.

sethtroisi:
The flawed behavior you describe will still happen at any other playout count.

`echo "auto" | ./leelaz -w ../autogtp/223737476718d58a4a5b0f317a1eeeb4b38f0c06af5ab65cb9d76d68d9abadb6 -p 2 --noponder -t1 -s 58`

```
Thinking at most 50.0 seconds...
NN eval=0.830256

pass ->       5 (V: 14.84%) (N:  1.62%) PV: pass L9 pass C11 P1
  P1 ->       1 (V:  8.59%) (N:  1.66%) PV: P1 
7 visits, 1196 nodes, 2 playouts, 200 n/s

Passing loses :-(
Avoiding pass because it loses.
Eval (8.59%) looks bad. Resigning.
```

I'll reflect on the edge case more (and try to generate some data about frequency)

gcp:
I think what I'd consider a flaw is that if we at any point use a "score" that has actually been fudged by FPU reduction. If at some point FPU reduction = 0.4 turns out to be better, you could have some pretty bad cases.

The example you gave is fine I think because the 8.59% is a true value, not 0.3-0.22=8%?

PR #819, opened 2018-02-05
autogtp: adding option -m

marcocalignano:
with the he option `-m` _n_ (--maxgames) autogtp stops after _n_ games are completed. `-m` and `-s` are the same thing.


alreadydone:
You mean `-m 1` and `-s` are the same thing? Will `-m` without value produce an error?

Issue #818, opened 2018-02-05
OpenCL self-check is too strict when value is very close to zero


Ttl:
Got this error earlier today: `Error read 119 'Error in OpenCL calculation: expected -0.000000 got 0.000001 (error=34028234663852885981170418348451692544000.000000%)`. Surely this is accurate enough?

Self-check currently gives error if the sign is different even if the absolute difference is minimal: https://github.com/gcp/leela-zero/blob/63517b19bbcba5d07ff68bb908994dfd835d229d/src/Network.cpp#L783-L785

gcp:
The close to zero was supposed to be caught by https://github.com/gcp/leela-zero/blob/63517b19bbcba5d07ff68bb908994dfd835d229d/src/Network.cpp#L790 but the sign test is ahead of it.

PR #820, opened 2018-02-05
Don't check sign differences near zero.

gcp:
We could report a self check failure if two very small values differed
in sign, which makes no sense.

Fixes issue #818.


gcp:
@Ttl Do you want to review?

Ttl:
I don't see any issues.

PR #821, opened 2018-02-05
UCTNode: some optimisations

marcocalignano:


src/UCTNode.cpp
```diff
@@ -181,28 +181,23 @@ void UCTNode::dirichlet_noise(float epsilon, float alpha) {
...
-    for (auto& v: dirichlet_vector) {
-        v /= sample_sum;
-    }
-
     child_cnt = 0;
     for (auto& child : m_children) {
         auto score = child->get_score();
-        auto eta_a = dirichlet_vector[child_cnt++];
-        score = score * (1 - epsilon) + epsilon * eta_a;
+        auto eta_a = dirichlet_vector[child_cnt++]/sample_sum - score;
```
gcp: I'm not convinced any "speedup" here is worth the readability loss. This function isn't even in a critical loop.

src/UCTNode.cpp
```diff
@@ -376,8 +371,7 @@ class NodeComp : public std::binary_function<UCTNode::node_ptr_t&,
 
 void UCTNode::sort_children(int color) {
     LOCK(get_mutex(), lock);
-    std::stable_sort(begin(m_children), end(m_children), NodeComp(color));
-    std::reverse(begin(m_children), end(m_children));
+    std::stable_sort(rbegin(m_children), rend(m_children), NodeComp(color));
```
gcp: This one is more interesting. I vaguely remember that it wasn't done this way because they don't give the same result for equally ordered nodes.  But that was probably referring to a cleanup of the existing code, and making sure that had no functional change, not that this has functional benefit now.

https://github.com/gcp/leela-zero/pull/301#discussion_r155961620
@sethtroisi I think we can fix this now, right?

sethtroisi: Good memory :)
The original comments are in https://github.com/gcp/leela-zero/pull/301/files#diff-28216771f77a09f4e0782ad5404fdd28, we didn't change it then because having refactor change be a no-op is nice.

This will slightly change the order nodes that have same visit count and same score (which is fine). I'm not sure how often scores are the same (I assume some large number of children may have score = 0) I know it occurs at least frequently enough to mess up reproducibility between the two.

A test that verifies top move is expande, sorted, evaled correctly might have high value for preventing regressions in this code. I'll try to find the time to write one.

gcp:
I pulled the latter part of this. I doubt the former has a real performance benefit and IMHO it reduces the readability of that code.

marcocalignano:
No problem, so now my next branch is free for the next commit in Validation (coming soon I hope)

PR #824, opened 2018-02-05
Saving parent_visit in the parent node to optimize uct_select_child

marcocalignano:
This commit saves a lot. On my GTX 1050 Ti I go from `2015 ms/move` to `1630 ms/move`. 
The only thing I am not sure, and I ask for your review is what happens on an GTP Undo command.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -44,8 +44,8 @@
 
 using namespace Utils;
 
-UCTNode::UCTNode(int vertex, float score, float init_eval)
-    : m_move(vertex), m_score(score), m_init_eval(init_eval) {
+UCTNode::UCTNode(int vertex, float score, float init_eval, UCTNode* parent)
+    : m_parent(parent), m_move(vertex), m_score(score), m_init_eval(init_eval) {
 }
 
 bool UCTNode::first_visit() const {
@@ -131,7 +131,7 @@ void UCTNode::link_nodelist(std::atomic<int> & nodecount,
     m_children.reserve(nodelist.size());
     for (const auto& node : nodelist) {
         m_children.emplace_back(
-            std::make_unique<UCTNode>(node.second, node.first, init_eval)
+            std::make_unique<UCTNode>(node.second, node.first, init_eval, this)
         );
     }
 
@@ -249,6 +249,9 @@ void UCTNode::virtual_loss_undo() {
 
 void UCTNode::update(float eval) {
     m_visits++;
+    if(m_parent != nullptr) {
+        m_parent->add_parent_visit(1);
+    }
     accumulate_eval(eval);
 }
 
@@ -258,6 +261,10 @@ bool UCTNode::has_children() const {
 
 void UCTNode::set_visits(int visits) {
     m_visits = visits;
+    if(m_parent != nullptr) {
+        m_parent->add_parent_visit(visits);
+    }
+    
 }
 
 float UCTNode::get_score() const {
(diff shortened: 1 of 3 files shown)


src/UCTNode.cpp
```diff
@@ -258,6 +261,10 @@ bool UCTNode::has_children() const {
 
 void UCTNode::set_visits(int visits) {
     m_visits = visits;
+    if(m_parent != nullptr) {
+        m_parent->add_parent_visit(visits);
```
oni-link: Don't we need to also subtract the old value of `m_visits`?

marcocalignano: I guess you are right!

killerducky:
@marcocalignano can you try testing by doing a self-play game with `-t 1 -s ###` with and without this change to verify it plays the same game? It should right?

marcocalignano:
also without -n but do you need only the SGF file to be the same?

killerducky:
You can include `-n -m 30`, as long as you use `-t 1 -s ###` all games should be 100% the same. Yes just diffing the sgf should be sufficient.

marcocalignano:
@killerducky No the are different. I Only the first 30 moves are the same, did you see something wrong in the code? I think there might be a problem when the root changes.

sethtroisi:
I have a similar change open in https://github.com/gcp/leela-zero/pull/816, I be suprised if this change had a positive performance impact and it increases node size by ~8%

Also as a general rule of thumb most of time (70-90%) is spent on GPU so a 20% improvement from CPU only changes would mean something like a 60% reduction in CPU usage.

marcocalignano:
@sethtroisi So you mean it is not worth to save 20%?

sethtroisi:
@marcocalignano I don't think it has any performance impact.
This code isn't that expensive (~361 adds) x play_simulation depth and there is other hideously expensive code (coping gamestate, constructing network features, coping Network to GPU)

I ran a small test with ringmaster

```
leelaz-next-feb-05 v leelaz-pull824 (15/200 games)
board size: 19   komi: 7.5
                     wins              black        white       avg cpu
leelaz-next-feb-05      7 46.67%       2 25.00%     5  71.43%     11.47
leelaz-pull824          8 53.33%       2 28.57%     6  75.00%     11.54
                                       4 26.67%     11 73.33%
```
You can see they use ~identical~ CPU

marcocalignano:
@sethtroisi You know how to kill my enthusiasmus :smile: 
So why do you think I got that improvement?

sethtroisi:
My best guess is you were comparing an older branch vs newer branch + your changes.
Or that the released network changed while you were testing and that affected speed.

marcocalignano:
No on both count but if it is really CPU relate, maybe my PC was doing something more during the first game. So should I just close this PR?

sethtroisi:
Yes, I think https://github.com/gcp/leela-zero/pull/816 does the same thing but avoids the need to add a new field (this info is already tracked in m_visits) and setters / getters

marcocalignano:
@sethtroisi Can you name the function that should be optimize to really make self-play faster? I'd like to have alook if I can do something.

Ttl:
@marcocalignano I suggest using profiler to measure the performance. I have used the following command: `valgrind --tool=callgrind ./leelaz --noponder -p 1600 -w ../autogtp/257aeeb863dc51bfc598838361225459257377a4b2c9abd3e1ac6cdba1fcc88f < moves`

moves file has some genmoves and quit at the end:

```
genmove b
genmove w
genmove b
genmove w
genmove b
genmove w
quit
```

kcachegrind is a good program for visualizing the output.

PR #825, opened 2018-02-06
Make use of <random> in Random.cpp + add tests

sethtroisi:
None of has real impact (assuming this doesn't break things) because PRNG is good enough for all our uses (Zobrist and sampling)

* Added comments about inclusive vs exclusive
* `uniform_int_distribution` better handles the one arg case.
`randuint16_t(60000)` was twice as likely to choose some numbers (0, 10, 21, 32, ...) as others.
* Replaced the only use of randflt with randfix

News tests can probabilistic fail (they choose a random seed).

**Pros**

* The failure rate is very low (4 in 10,000 theoretically and ~1 in 1,300 in practice).
* Helps detect bad PRNG
* Avoids the problem where everything is a mock and nothing is actually tested

**Cons**

* Possible that if PRNG is bad it will fail rarely but often enough to cause confusion

---

**Testing**

No impact on performance
```
leelaz-next-feb-05 v leelaz-random (34/200 games)
board size: 19   komi: 7.5
leelaz-next-feb-05     16 47.06%       7  41.18%     9  52.94%     10.72
leelaz-random          18 52.94%       8  47.06%     10 58.82%     10.67
                                       15 44.12%     19 55.88%
```

Failure rate is < 1 in 1000
`i=0; while [ $? -eq 0 ]; do ./tests --gtest_filter=UtilsTest.* && echo "$((i++))"; done`

### DIFF
--- a/src/Random.cpp
+++ b/src/Random.cpp
@@ -22,6 +22,7 @@
 #include <climits>
 #include <cstdint>
 #include <thread>
+#include <random>
 
 #include "GTP.h"
 #include "Utils.h"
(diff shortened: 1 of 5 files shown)


gcp:
I like the cleanup part. But obvious question: is there much use to us testing the RNG in CI, given that we are supplying a known good one?

sethtroisi:
I'm a fan of keeping them (they can always be removed later if they cause problems)

I think the advantage is that if the code changes it validates these changes:
* RNG is not broken (in obvious or non-obvious ways)
* exclusivity of max parameter e.g. [0, max) 
* source is still fairly random

The disadvantage is
* have to be maintained during api refactors / class refactors
* may give false confidence that a chance is safe
* more code in the codebase

PR #826, opened 2018-02-06
Tensorboard tag: Must capitalize the L in "Policy Loss"

FFLaguna:
With the wrong capitalization, Tensorboard outputs the "Policy loss" (lowercase L) onto a different scalar chart than "Policy Loss". Instead, the policy loss logs for both test and train should be output onto the same chart, as is already the case for the Tensorboard MSE Loss chart.


FFLaguna:
Example of current Tensorboard output: [https://i.imgur.com/bn77MJW.png](https://i.imgur.com/bn77MJW.png)

Example of fixed Tensorboard output: [https://i.imgur.com/qKeqKur.png](https://i.imgur.com/qKeqKur.png)

PR #830, opened 2018-02-07
validation: adding Stop Resume feature

marcocalignano:
This will save the results and Sprt data. This won't save the running games.

### DIFF
--- a/validation/CMakeLists.txt
+++ b/validation/CMakeLists.txt
@@ -3,7 +3,7 @@ CMAKE_MINIMUM_REQUIRED(VERSION 3.1)
 
 ADD_EXECUTABLE(validation
         main.cpp ../autogtp/Game.cpp SPRT.cpp Validation.cpp Results.cpp
-        ../autogtp/Game.h SPRT.h Validation.h Results.h)
+        ../autogtp/Game.h SPRT.h Validation.h Results.h ../autogtp/Console.h)
 SET_TARGET_PROPERTIES(validation PROPERTIES AUTOMOC 1)
 TARGET_LINK_LIBRARIES(validation Qt5::Core)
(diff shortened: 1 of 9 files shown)


marcocalignano:
@zediir Yes! This is all for you :smiley:

WhiteHalmos:
Thank you! This will be very useful.

WhiteHalmos:
By the way, a suggestion, can intermediate results be printed in a nice format after each game rather than just at the end of validation? Would `m_results.printResults(m_firstNet, m_secondNet);` work when status is `Sprt::Continue`?

marcocalignano:
It works but wouldn't be verbose especially if you have more than one thread?

WhiteHalmos:
True... perhaps when the games are stopped and resumed it would be a good time to output intermediate results? That way validation doesn't have to run till the end to have nice results available on demand?

Question: How does stop work? Does it respond to Ctrl+C?

marcocalignano:
The stop funcion is the same like autogtp ```q+Enter```. And it prints the result when it is resumed. I need to add it when it stops. Give me a few mins.

marcocalignano:
@WhiteHalmos Ok now you see the print out when you stop it and when you resume it.

WhiteHalmos:
Thank you very much!

marcocalignano:
@gcp When you merge this I have another PR ready. Using QLockFiles while loading saved files or uploading games.

marcocalignano:
Or should I just add the commit here?

gcp:
I prefer multiple small pulls. If I have a remark on one part it won't block the other parts.

(Maybe less applicable to autogtp/validation as you're de facto the maintainer for those)

PR #843, opened 2018-02-09
autogtp: adding QLockFile for multiple processing safety on saved files

marcocalignano:


zediir:
On windows (don't know about linux) both this and the version in next crashes after the first played match game.

zediir:
On windows at least (didn't test linux) both this and the version in next crash after the first match game has finished.

marcocalignano:
It crashes without messages?
Can you take the `-q` option from the leelaz commands and see if you get a more verbose error message?

zediir:
It's not a leelaz error it's validation error.

It doesn't give any messages and the crash happens after leelaz processes have quit and finished the game.

Ran a debug build and got some more information

```
Score: W+Resign
Winner: white
Stopping engine.
1 wins, 0 losses
1 games played.
Status: 0 LLR 0 Lower Bound -2.94444 Upper Bound 2.94444
ASSERT: "copy" in file thread\qmutex.cpp, line 608

This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.
```

zediir:
It crashes after this

breakpoint

marcocalignano:
Try to start validation with -g 2

marcocalignano:
The lock() call is missing (damn rebase!) which version of QT do you have? @zediir

zediir:
5.9.3

marcocalignano:
@zadiir try #848

PR #836, opened 2018-02-08
Replace fpu reduction with best child eval

WhiteHalmos:
* Use minimum of best visited child's eval and parent eval when a node
has not been visited.
* Update AUTHORS (to be compliant with my company policy which allows me to contribute).
* Based on @Eddh's totalvp work in #827.

PR #827, opened 2018-02-06
modify fpu reduction depending on visited policy

Eddh:
This is a fpu reduction method that seems like it is as strong as parent's net_eval for weak network while being as strong as flat fpu reduction of 0.22 for strong networks.

PR #850, opened 2018-02-10
Use average visited child eval as fpu

WhiteHalmos:
* Use average of visited child eval for child nodes with 0 visits.
* Also does not play out ladder in #838 on latest net (6665).
* Based on work in #836 and @Eddh's totalvp work in #827.
* Update AUTHORS (to be compliant with my company policy which allows me to contribute).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -26,3 +26,5 @@ cheshirecats
 gaieepo
 zediir
 Пахотин Иван
+
+Google LLC  # Please keep.
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -64,7 +64,6 @@ bool cfg_tune_only;
 #endif
 float cfg_puct;
 float cfg_softmax_temp;
-float cfg_fpu_reduction;
 std::string cfg_weightsfile;
 std::string cfg_logfile;
 FILE* cfg_logfile_handle;
@@ -86,7 +85,6 @@ void GTP::setup_default_parameters() {
 #endif
     cfg_puct = 0.8f;
     cfg_softmax_temp = 1.0f;
-    cfg_fpu_reduction = 0.22f;
     // see UCTSearch::should_resign
     cfg_resignpct = -1;
     cfg_noise = false;
--- a/src/GTP.h
+++ b/src/GTP.h
@@ -47,7 +47,6 @@ extern bool cfg_tune_only;
 #endif
 extern float cfg_puct;
 extern float cfg_softmax_temp;
-extern float cfg_fpu_reduction;
 extern std::string cfg_logfile;
 extern std::string cfg_weightsfile;
 extern FILE* cfg_logfile_handle;
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -91,7 +91,6 @@ static void parse_commandline(int argc, char *argv[]) {
 #ifdef USE_TUNER
         ("puct", po::value<float>())
         ("softmax_temp", po::value<float>())
-        ("fpu_reduction", po::value<float>())
 #endif
         ;
     // These won't be shown, we use them to catch incorrect usage of the
@@ -142,9 +141,6 @@ static void parse_commandline(int argc, char *argv[]) {
     }
     if (vm.count("softmax_temp")) {
         cfg_softmax_temp = vm["softmax_temp"].as<float>();
-    }
-    if (vm.count("fpu_reduction")) {
-        cfg_fpu_reduction = vm["fpu_reduction"].as<float>();
     }
 #endif
 
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -289,14 +289,12 @@ float UCTNode::get_eval(int tomove) const {
         }
         return score;
     } else {
-        // If a node has not been visited yet,
-        // the eval is that of the parent, potentially
-        // minus a constant.
+        // If a node has not been visited yet, the eval is that of the parent.
         auto eval = m_init_eval;
         if (tomove == FastBoard::WHITE) {
             eval = 1.0f - eval;
         }
-        return eval - cfg_fpu_reduction;
+        return eval;
     }
 }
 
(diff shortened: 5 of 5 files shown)


WhiteHalmos:
Will update this comment as more test games come in. Nets used: 6665 and a115.

```
leelaz-avg-eval_66 v leelaz-next_66 (9/400 games)       
board size: 19   komi: 7.5  
leelaz-avg-eval_66      6 66.67%       3 60.00%     3 75.00%    169.66                                           
leelaz-next_66          3 33.33%       1 25.00%     2 40.00%    146.36                                           
                                       4 44.44%     5 55.56%                                                     

leelaz-avg-eval_a1 v leelaz-next-fpu0_a1 (9/400 games)  
board size: 19   komi: 7.5  
leelaz-avg-eval_a1       8 88.89%       5 100.00%     3 75.00%    326.01                                         
leelaz-next-fpu0_a1      1 11.11%       1  25.00%     0  0.00%    333.60                                         
                                        6  66.67%     3 33.33%
```

<details>
<summary>gomill config</summary>

```
competition_type = 'playoff'
LZ_ARGS = '/projects/lztest/{} -p1600 --noponder -r1 -g -t1 -q -d -w {} {}'
def LZ(binary, weights, extra_args=''):
  return Player(LZ_ARGS.format(binary, weights, extra_args))
players = dict()
p1 = 'leelaz-avg-eval_66'
p2 = 'leelaz-next_66'
p3 = 'leelaz-avg-eval_a1'
p4 = 'leelaz-next-fpu0_a1'
players[p1] = LZ(p1.split('_')[0], p1.split('_')[1])
players[p2] = LZ(p2.split('_')[0], p2.split('_')[1])
players[p3] = LZ(p3.split('_')[0], p3.split('_')[1])
players[p4] = LZ(p4.split('_')[0], p4.split('_')[1])
board_size = 19
komi = 7.5
matchups = [
  Matchup(p3, p4, alternating=True, number_of_games=400),
]
```
</details>

MaxMaki:
Should test games be run with timemanage on, off, or auto? I was getting extremely lopsided results, but with very different cpu times (the next i used didn't have timemanage and whitehalmos did). So seems like -p doesn't disable timemanage? (Don't know if it's supposed to or not)

roy7:
--playouts doesn't alter timemanage. So if you have -p 1600 and 801 are into move X and none into other moves, it'll make move X and stop search since it can't change off of move X no matter what it finds.

MaxMaki:
Is that true even with these different fpu tests we are doing? That it won't change if 801 playouts go into one move it won't change? And also, what should test games be run with, auto on or off?

roy7:
Training and test games use randomness, and the randomness option automatically turns timemanage off.

MaxMaki:
I'm refering to test/validation games run on these different commits to check if the change is stronger than next or not, not to general training or network test games

roy7:
If their version of /next is new enough to have the timemanage code included, then it's on by default for what they are testing. What effect this has on strength/results I can't comment on.

killerducky:
@MaxMaki --help describes the default behavior of --timemanage. 

```
  --timemanage arg (=auto)      [auto|on|off] Enable extra time management 
                                features.
                                auto = off when using -m, otherwise on

  -m [ --randomcnt ] arg (=0)   Play more randomly the first x moves.
```

First of course make sure whatever test code is rebased to include the timemanage so they both run the same code base minus the code being tested.

About how to test going forward, I don't know what the standard should be. If we switch to `-v` instead of `-p` for self-play and matches then the cutoff will have no impact on playing strength, it will just play faster. But with `-p` there is some interaction with tree reuse and timemanage. 

I think we need @gcp to give some guidance on this.

WhiteHalmos:
@roy7 I didn't realize that match tests also used randomness. Mind linking the code where the server sends out match args? Sorry I'm not familiar with server code.

roy7:
Apologies, randomcnt does default to 0 for matches in the server. The matches have to be queued with a randomcnt setting to override it, which we don't do.

Maybe I was confusing myself with the resignation command. Sorry!

WhiteHalmos:
I have a more theoretically sound PR being tested right now, i.e. rather than a simple average, use a weighted average weighted by visits instead. This is basically the average of all W(s, a) for all legal moves a from the paper: https://github.com/WhiteHalmos/leela-zero/tree/weighted_eval

Eddh:
For the weighted average, couldn't it be more practical to use get_eval of the parent ? It should be basically the same.

WhiteHalmos:
Good point. I'll close this for now and wait until I have test results before making a PR.

killerducky:
Why close this one already? Your last update showed it was better than /next right? Until proven otherwise it's the best thing we have?

I think this one could be better. Weighted average will skew the FPU prior to be closer to the best ones, because that's where the search focuses. I don't think you want to put extra weight on only the best siblings.

WhiteHalmos:
Agreed, this is better as a pure average. Will keep this open in case someone else also wants to test it.

WhiteHalmos:
As the first comment shows, this branch is a flop for strong networks.

MaxMaki:
@WhiteHalmos , i was waiting to get more games before posting, but sadly my test showed similar results to yours.

````text

WhiteHalmos_avg_eval_1600p v Leela_Next_1600p (88/400 games)
board size: 19   komi: 7.5
WhiteHalmos_avg_eval_1600p     33 37.50%       14 31.82%     19 43.18%     94.77
Leela_Next_1600p               55 62.50%       25 56.82%     30 68.18%     79.49
                                               39 44.32%     49 55.68%

````

**Comment by WhiteHalmos** [comment]
Thanks @MaxMaki ! Our two tests together make this statistically significant. :)

52 for this branch vs 86 for next

**Comment by Eddh** [comment]
Honestly I still have no clue if this is a better starting point than parent's net eval to reduce fpu afterward.

**Comment by WhiteHalmos** [comment]
Probably this is more generalizable than parent eval minus fpu reduction, which can be very bad for early nets. Still to be tested though, since without sufficient test data, it's hard to justify accepting something that makes strong nets weaker.

**Comment by WhiteHalmos** [comment]
And by test data in this case I mean tuned with fpu reduction. :)

**Comment by Splee99** [comment]
I'm testing this interesting code now with leela 0.11. I had feeling that this build has solid defending skill which only can be demonstrated with stronger opponents.

**Comment by Splee99** [comment]
Unfortunately, I can't set the number of playouts or time per move. It very often stopped search earlier (showing stopping early). This might give unfair advantage to the opponent during the previous tests.

**Comment by WhiteHalmos** [comment]
@Splee99 You can add "--timemanage off", but I think it's okay to just leave it off and take the speed increase. It shouldn't decrease strength by a significant amount and we're testing almost everywhere without that option anyways.

**Comment by gcp** [comment]
```
leelaz-next v leelaz-wh (473/1000 games)
board size: 19   komi: 7.5
leelaz-next    291 61.52%       115 48.52%     176 74.58%    113.95
leelaz-wh      182 38.48%       60  25.42%     122 51.48%    127.83
                                175 37.00%     298 63.00%

leelaz-next-weak v leelaz-wh-weak (472/1000 games)
board size: 19   komi: 7.5
leelaz-next-weak    207 43.86%       101 42.80%     106 44.92%    413.48
leelaz-wh-weak      265 56.14%       130 55.08%     135 57.20%    415.91
                                     231 48.94%     241 51.06%

```

PR #851, opened 2018-02-10
Turn off TimeManagement only for initial random moves

Mardak:
Instead of tuning off time management completely for self play, those can gain the benefits of stopping early when there's a dominant best move.


killerducky:
It's done the way it is because for self-play games the training output require all playouts to be done, no short cutting.

PR #858, opened 2018-02-11
Small Timemanage fix.

killerducky:
Wait for at least 1 second and 100 playouts so we get a reliable playout_rate.

roy7 saw this on his machine. Most moves he gets over 2000n/s. But on the very first playout the measured average is much lower. This fix adds a minimum time and playout count so we have a reliable average playout_rate to work with.

Example of the problem: http://termbin.com/9zrg
After only 1st playout it measured 33n/s instead of 2000n/s, and triggered the stop.
```
Thinking at most 25.0 seconds...
NN eval=0.492506
25.0s left
Stopping early.

 G14 ->    1483 (V: 51.69%) (N: 44.95%) PV: G14 H11 J13 K13 J18 H18 K17 J16 H17 J19 K18 G17 H16 H15 G16 G15 F16 F14 J15 K15 F15
 H11 ->     211 (V: 50.01%) (N:  8.07%) PV: H11 G14 H14 G15 F14 H10 G10 F11 H9 H12 H13

1996 visits, 359952 nodes, 1 playouts, 33 n/s

= G14
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,8 +115,8 @@ source and remove the line that says "#define USE_OPENCL".
     Double-click the leela-zero2015.sln or leela-zero2017.sln corresponding
     to the Visual Studio version you have.
     # Build from Visual Studio 2015 or 2017
-    # Download and extract <https://sjeng.org/zero/best_v1.txt.zip> to msvc/x64/Release
-    # msvc/x64/Release/leela-zero --weights weights.txt
+    # Download and extract <https://sjeng.org/zero/best_v1.txt.zip> to msvc\x64\Release
+    msvc\x64\Release\leelaz.exe --weights weights.txt
 
 ## Example of compiling and running - CMake (macOS/Ubuntu)
 
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -460,7 +460,7 @@ void UCTWorker::operator()() {
         if (result.valid()) {
             m_search->increment_playouts();
         }
-    } while(m_search->is_running() && !m_search->stop_thinking());
+    } while(m_search->is_running());
 }
 
 void UCTSearch::increment_playouts() {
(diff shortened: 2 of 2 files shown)


killerducky:
I added a fix for the issue reported in #797. @zediir can you try this?

Test method:
```
./leelaz -w latest.txt -r1 -t4 
time_settings 0 10 1
genmove b
```
Before fix:
`21481 visits, 7577371 nodes, 21481 playouts, 2397 n/s`
After fix:
`39634 visits, 13962160 nodes, 39634 playouts, 4418 n/s`

zediir:
@killerducky That worked! 👍

killerducky:
Ok I'm not sure why it impacted performance to that extent. I guess all threads were contending for access to atomic variables, but I didn't think it would be so bad, this check is only done once per playout in the root node only. 

Also UCTNode::uct_select_child does a mutex lock, should stop_thinking use one also? I figured it was not critical to get the counts perfect, but maybe there are other reasons to use it?

Do we need to review other code such as FPU changes for too much lock contention?

gcp:
> maybe there are other reasons to use it?

uct_select child traverses the m_children array. If you try to do that you should take the lock.

gcp:
Though you can get away without one because the start of the search guarantees m_root has been expanded and cleared of superko already, so m_children won't change underneath you.

Ttl:
Is this expected behaviour:

```
Thinking at most 49.9 seconds...
NN eval=0.490593
Stopping early.

 S17 ->    1600 (V: 50.94%) (N: 98.84%) PV: S17 R3 Q3 R4 R5 S5 S6 R6 Q5 S7 S4 T6 S3 R2 S2 Q6 C17 D17 C16 C15 B15 B14 C14 D15 B13 B16 A14 B17
 C17 ->       0 (V: 50.80%) (N:  0.29%) PV: C17 
1601 visits, 535716 nodes, 1 playouts, 100 n/s
```

There are 1600 visits from tree reuse, but only one new playout is made.

leelaz arguments: `./leelaz -t 1 -p 1600 --noponder -w ../autogtp/5ac1d898e0a4f169eb30669c389f74331371a4eb68dc84cfeed6d98563ee937d`. With `auto` it will play these fast moves sometimes early in the opening.

killerducky:
It's working as intended. S17 has 1600 playouts, so it's impossible for C17 to catch up in the next 1600 playouts, and therefore a waste of time to do them.

PR #866, opened 2018-02-12
WIP: Replace FPU reduction with dynamic parent eval

WhiteHalmos:
**Status: Work In Progress**
* Use parent node's eval when a child node has 0 visits.
* Based on @Eddh's experiments and discussion in issue #696.
* Update AUTHORS (company policy).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -26,3 +26,5 @@ cheshirecats
 gaieepo
 zediir
 Пахотин Иван
+
+Google LLC  # Please keep.
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -64,6 +64,7 @@ bool cfg_tune_only;
 #endif
 float cfg_puct;
 float cfg_softmax_temp;
+float cfg_fpu_reduction;
 std::string cfg_weightsfile;
 std::string cfg_logfile;
 FILE* cfg_logfile_handle;
@@ -85,6 +86,7 @@ void GTP::setup_default_parameters() {
 #endif
     cfg_puct = 0.8f;
     cfg_softmax_temp = 1.0f;
+    cfg_fpu_reduction = 0.25f;
     // see UCTSearch::should_resign
     cfg_resignpct = -1;
     cfg_noise = false;
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -64,7 +64,6 @@ bool cfg_tune_only;
 #endif
 float cfg_puct;
 float cfg_softmax_temp;
-float cfg_fpu_reduction;
 std::string cfg_weightsfile;
 std::string cfg_logfile;
 FILE* cfg_logfile_handle;
@@ -86,7 +85,6 @@ void GTP::setup_default_parameters() {
 #endif
     cfg_puct = 0.8f;
     cfg_softmax_temp = 1.0f;
-    cfg_fpu_reduction = 0.25f;
     // see UCTSearch::should_resign
     cfg_resignpct = -1;
     cfg_noise = false;
--- a/src/GTP.h
+++ b/src/GTP.h
@@ -47,7 +47,6 @@ extern bool cfg_tune_only;
 #endif
 extern float cfg_puct;
 extern float cfg_softmax_temp;
-extern float cfg_fpu_reduction;
 extern std::string cfg_logfile;
 extern std::string cfg_weightsfile;
 extern FILE* cfg_logfile_handle;
--- a/src/GTP.h
+++ b/src/GTP.h
@@ -47,6 +47,7 @@ extern bool cfg_tune_only;
 #endif
 extern float cfg_puct;
 extern float cfg_softmax_temp;
+extern float cfg_fpu_reduction;
 extern std::string cfg_logfile;
 extern std::string cfg_weightsfile;
 extern FILE* cfg_logfile_handle;
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -143,9 +143,6 @@ static void parse_commandline(int argc, char *argv[]) {
     if (vm.count("softmax_temp")) {
         cfg_softmax_temp = vm["softmax_temp"].as<float>();
     }
-    if (vm.count("fpu_reduction")) {
-        cfg_fpu_reduction = vm["fpu_reduction"].as<float>();
-    }
 #endif
 
     if (vm.count("logfile")) {
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -143,6 +143,9 @@ static void parse_commandline(int argc, char *argv[]) {
     if (vm.count("softmax_temp")) {
         cfg_softmax_temp = vm["softmax_temp"].as<float>();
     }
+    if (vm.count("fpu_reduction")) {
+        cfg_fpu_reduction = vm["fpu_reduction"].as<float>();
+    }
 #endif
 
     if (vm.count("logfile")) {
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -289,8 +289,7 @@ float UCTNode::get_eval(int tomove) const {
         }
         return score;
     } else {
-        // If a node has not been visited yet,
-        // the eval is that of the parent.
+        // If a node has not been visited yet, the eval is that of the parent.
         auto eval = m_init_eval;
         if (tomove == FastBoard::WHITE) {
             eval = 1.0f - eval;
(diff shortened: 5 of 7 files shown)


WhiteHalmos:
Next is after @Eddh's PR was merged. Will stop running 1a since it's a blowout.
5a is 5ac1d898, elo 7998. 1a is 1ac2638d, elo 543
```
leelaz-parent-eval_5a v leelaz-next-e06f_5a (96/400 games)
board size: 19   komi: 7.5
leelaz-parent-eval_5a     46 47.92%       20 41.67%     26 54.17%    158.98
leelaz-next-e06f_5a       50 52.08%       22 45.83%     28 58.33%    143.61
                                          42 43.75%     54 56.25%

leelaz-parent-eval_1a v leelaz-next-e06f_1a (95/400 games)
board size: 19   komi: 7.5
leelaz-parent-eval_1a     88 92.63%       46 95.83%     42 89.36%    196.85
leelaz-next-e06f_1a        7  7.37%       5  10.64%     2   4.17%    184.78
                                          51 53.68%     44 46.32%
```

Before recent FPU change with @Eddh's PR:
```
cd is cde9c8d4, elo 7957
leelaz-parent-eval_cd v leelaz-next_cd (222/400 games)
board size: 19   komi: 7.5
leelaz-parent-eval_cd    107 48.20%       47 42.34%     60  54.05%    236.13
leelaz-next_cd           115 51.80%       51 45.95%     64  57.66%    205.61
                                          98 44.14%     124 55.86%
```

WhiteHalmos:
Re-iterating notes from: https://github.com/gcp/leela-zero/issues/696#issuecomment-364964865

Advantages:
* Simpler to understand, reason about without more tuning parameters.
* Improves in FPU prediction as the search progresses rather than being a static value throughout.
* Vastly better for flatter (weak) policy networks, indicating generalization is better across nets.
* Allows memory reduction per node since we can remove `m_init_eval`

Disadvantages:
* Not as fully tested out as current next
* Nothing to tune
* Only about even for current (strong) policy networks

Eddh:
Is leelaz-next-e06f_1a  using fpu reduction = 0 ?

WhiteHalmos:
Opps! Re-running with fpu0 now. Thanks for catching that @Eddh! 😃

Eddh:
It's good to know that it's better with fpu reduction on too.

WhiteHalmos:
Preliminary results:
```
leelaz-parent-eval_1a v leelaz-next-e06f-fpu0_1a (26/400 games)
board size: 19   komi: 7.5
leelaz-parent-eval_1a        20 76.92%       9  69.23%     11 84.62%    177.96
leelaz-next-e06f-fpu0_1a      6 23.08%       2  15.38%     4  30.77%    182.42
                                             11 42.31%     15 57.69%
```

wctgit:
```
leelaz-parent-eval_5a     48 48.48%       22 44.00%     26 53.06%    160.63
leelaz-next-e06f_5a       51 51.52%       23 46.94%     28 56.00%    145.01
```

It's possible that the _apparent_ weakening of 5a with parent eval is more due to the NN having been trained under FPU reduction, and that after training for some time with parent eval, adding back in the FPU reduction code would actually cause a weakening effect also. I.e. It may be the case that the NN is **already** implicitly tuning **itself** to whatever FPU calculation method we impose on it.

jkiliani:
@wctgit I would agree with this hypothesis if it were the case. However, afaik most clients run the master branch, and only a minority part of clients use /next with FPU reduction. Once any FPU reduction code is merged to master and enforced for both all training games and distributed testing, I would agree that instead of choosing neural networks that perform well with the UCT code before FPU, nets that perform well with FPU will be picked.

killerducky:
Wow, in another thread I told @WhiteHalmos a similar idea would be bad, but the results look promising. I hope this works out because it's easier to understand what it is doing. I am ready with a rationalization for why this is "obviously" good. ;-)

alreadydone:
@jkliani FYI Big Red II has been running /next with FPU reduction for a few days (maybe a week, I don't remember). It contributes more than half of the self-play games in the good days, but never do any matches.

Eddh:
I can see one reason why it is not bad : it can self adapt in case of overexploration of useless move. In that case get_eval will go down, and fpu with it, causing less exploration. In case a bad move with high policy is overexploited instead, fpu will go down, but there will be at least some amount of exploration because get_eval include parent's eval. So it adapts well to these situations.

WhiteHalmos:
Updated results and analysis of advantages/disadvantages in top comments.

jkiliani:
> - PUCT may have to be re-tuned

This can be seen as an opportunity, not a burden. PUCT tuning may easily compensate the current very small regression with strong networks. From the mix of tests you posted so far, this branch looks amazingly good.

Eddh:
I'm thinking it could be good to tune down cfg_puct with this change. If there is not much much exploration because of it, parent's action value will get close to best child's action value, making it explore more over time.
That's just a guess though. CLOP will tell the truth.

WhiteHalmos:
Makes sense. Anyone who used gomill to tune parameters mind sharing some configs to help others (me especially, haha) to try some tuning runs? Does gomill use CLOP?

WhiteHalmos:
@jkiliani Agreed, those tests basically pit this one change against four different other fpu algorithms and either came close or beat them by a large margin, so it has some potential to do well in general.

gcp:
```
leelaz-next v leelaz-wh (378/1000 games)
board size: 19   komi: 7.5
leelaz-next    228 60.32%       104 55.03%     124 65.61%    224.91
leelaz-wh      150 39.68%       65  34.39%     85  44.97%    252.13
                                169 44.71%     209 55.29%
```

Doesn't look good. The idea that we don't need to store m_initial_eval is good though, I will try both variations of that (use parent static eval, use current parent eval).

Eddh:
I think the optimum fpu reduction might be very different if you use current parent eval though. I don't think you can directly compare without tuning

gcp:
Will have to leave that for after the release then. Eliminating m_initiall_eval should still be possible, though.

WhiteHalmos:
Thanks for testing this some more! Here are some initial results from some PUCT tuning. I'll leave this branch open for comments.
```
leelaz-parent-eval-puct-0.9_5a v leelaz-next-e06f_5a (89/100 games)
board size: 19   komi: 7.5                                                                                       
leelaz-parent-eval-puct-0.9_5a     31 34.83%       14 31.11%     17 38.64%    211.18                             
leelaz-next-e06f_5a                58 65.17%       27 61.36%     31 68.89%    184.41                             
                                                   41 46.07%     48 53.93%
                                                                                                                 
leelaz-parent-eval-puct-0.7_5a v leelaz-next-e06f_5a (89/100 games)
board size: 19   komi: 7.5                                                                                       
leelaz-parent-eval-puct-0.7_5a     38 42.70%       17 37.78%     21 47.73%    220.12                             
leelaz-next-e06f_5a                51 57.30%       23 52.27%     28 62.22%    198.74                             
                                                   40 44.94%     49 55.06%
                                                   
leelaz-parent-eval-puct-0.75_5a v leelaz-next-e06f_5a (89/100 games)                                             
board size: 19   komi: 7.5                                                                                       
leelaz-parent-eval-puct-0.75_5a     39 43.82%       15 33.33%     24 54.55%    221.25                            
leelaz-next-e06f_5a                 50 56.18%       20 45.45%     30 66.67%    197.20                            
                                                    35 39.33%     54 60.67%                                      

leelaz-parent-eval-puct-0.85_5a v leelaz-next-e06f_5a (89/400 games)
board size: 19   komi: 7.5
leelaz-parent-eval-puct-0.85_5a     42 47.19%       19 42.22%     23 52.27%    219.37
leelaz-next-e06f_5a                 47 52.81%       21 47.73%     26 57.78%    193.04
                                                    40 44.94%     49 55.06%
```

WhiteHalmos:
I'll try to see if I can just eliminate m_init_eval without changing anything.

WhiteHalmos:
Just realized that the parent node's get_eval() has virtual loss incorporated in it. That may be why it works fairly well (already incorporates some sense of fpu reduction since denominator is increased by virtual loss) and also why good fpu reduction is different on it. Testing calculating parent_eval for uct_select_child without virtual loss now...

Eddh:
Ohh I didn't think about that !

WhiteHalmos:
After I get a reasonable test for real (no virtual loss) parent eval vs current next, next idea to test is real parent eval - fpu_reduction against init_eval - fpu_reduction. :)

Eddh:
From my previous experience I expect ok results from parent's action value - fpu_reduction. But I've seen it have problems with ladder situation, because it can know that the ladder is bad, and not explore any other option, because fpu went down with the evaluation, and the ladder has 99% policy, leaving every other moves in the dust because of low policy + fpu reduction. So it can cause only 1 bad move to be searched. In the meantime I think I'll go back to test more with average child eval with fpu reduction, which I like because it is more resilient to these situations.

WhiteHalmos:
I think it's pretty clear that it was the virtual loss doing the bulk of the work making parent eval match fpu reduction.
```
leelaz-parent-eval-no-vloss_5a v leelaz-next-e06f_5a (40/400 games)                                              
board size: 19   komi: 7.5  
leelaz-parent-eval-no-vloss_5a     13 32.50%       6  30.00%     7  35.00%    221.69                             
leelaz-next-e06f_5a                27 67.50%       13 65.00%     14 70.00%    187.53                             
                                                   19 47.50%     21 52.50%
```

Eddh:
Don't the good results on weak networks hint at another kind of fpu reduction which may work for weak networks and dynamic fpu ? From what I understand the more the parent node is visited the less virtual loss has an effect right ? Maybe decaying fpu may work when using a dynamic fpu such as real parent eval or the average value of the children.

WhiteHalmos:
Agreed, @Eddh that there's something here that we can improve upon since the results vs the weaker nets are so good. I'm just a little sad that a solution that seemed so elegant turned out to not be such.

WhiteHalmos:
Results are extremely positive, but since @gcp has all the relevant changes in cleaner code and more in #872 I'm closing this one.
```
leelaz-parent-eval-fpu-reduct_5a v leelaz-next-e06f_5a (107/400 games)
board size: 19   komi: 7.5
leelaz-parent-eval-fpu-reduct_5a     61 57.01%       25 46.30%     36 67.92%    194.28
leelaz-next-e06f_5a                  46 42.99%       17 32.08%     29 53.70%    200.36
                                                     42 39.25%     65 60.75%
```

ThomasCabaret:
@WhiteHalmos What do you use(tool, script) to get automatically those stats between 2 different leelaz binary?

WhiteHalmos:
@ThomasCabaret Please see https://github.com/gcp/leela-zero/issues/870#issuecomment-365340090

PR #872, opened 2018-02-14
Use parent evaluation instead of m_init_eval.

gcp:
Instead of storing the original parent eval in all the child nodes, get
the current parent evaluation during node selection.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -44,8 +44,8 @@
 
 using namespace Utils;
 
-UCTNode::UCTNode(int vertex, float score, float init_eval)
-    : m_move(vertex), m_score(score), m_init_eval(init_eval) {
+UCTNode::UCTNode(int vertex, float score)
+    : m_move(vertex), m_score(score) {
 }
 
 bool UCTNode::first_visit() const {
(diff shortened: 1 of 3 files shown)


killerducky:
Am I reading the code right? It looks like missing m_root->update was a bug for awhile now. If the tree is reused and so already has_children it won't matter. But if the opponent makes an unexpected move not in the tree the old code would start with the old default 0.5 value I guess? Same thing for the very first move of the game I suppose. Hard to predict what kind of impact this had.

gcp:
@sethtroisi pointed that out a while ago in a pull request that got reverted. I needed this fix for the visits > 0 assertions not to fire.

It's only relevant for the very first playout so I don't think it makes a difference, aside from code cleanliness.

Unfortunately the idea seems to be simply not good.

```
leelaz-next v leelaz-par (458/1000 games)
board size: 19   komi: 7.5
leelaz-next    258 56.33%       123 53.71%     135 58.95%    225.61
leelaz-par     200 43.67%       94  41.05%     106 46.29%    229.25
                                217 47.38%     241 52.62%
```

gcp:
And note we don't use the root winrate for move selection, only the children of the root.

WhiteHalmos:
Strange but not impossible that your results are so different from mine in #866. Both are SPRT significant in opposite directions, I guess this is the 5%?

WhiteHalmos:
Ah, it's probably because your PR is using total_visited_policy in conjunction with this while mine is flat subtraction. I'll test some more and open a new PR if I get actually good results.

WhiteHalmos:
@gcp What command-line args do you use for your tests? Namely do you use `-r1`? I use the same as matches `-p1600 --noponder -r1 -g -t1 -q -d -w {}`.

Issue #860, opened 2018-02-11
MCTS alternatives


roy7:
Since a lot of people are working on tuning FPU at the moment and some people are exploring tweaks to the search algorithm I wanted to share a few areas of research I was looking over this evening, in case the more mathematical programmers among us were interested in trying these with LZ.

After reviewing [Multi Armed Bandits and Exploration Strategies](https://sudeepraja.github.io/Bandits/) I was interested in the mention of Thompson Sampling. It's an alternative to UCB that also has a logarithmic bound on regret and has been known since the 1930s, but the theoretical groundwork on it has only recently been finished in 2012 in the paper [Analysis of Thompson Sampling for the Multi-armed Bandit Problem](http://proceedings.mlr.press/v23/agrawal12/agrawal12.pdf). That paper also covers the stochastic case, which makes it more applicable for us by people like us.

So I searched around some more if anyone has used this with MCTS yet, and it turns out this is a pretty hot current research area. There's a 2014 paper [Thompson Sampling Based Monte-Carlo Planning in POMDPs](https://aijunbai.github.io/publications/ICAPS14-Bai.pdf), and two 2017 master's thesis I found, [Thompson Sampling for Monte Carlo Tree Search and Maxi-min Action Identification](https://www.universiteitleiden.nl/binaries/content/assets/science/mi/scripties/master/2017-2018/master-thesis_mingxi-li_12okt2017.pdf) and [Monte Carlo Tree Search with Thompson Sampling in the Settlers of Catan](http://eprints.fri.uni-lj.si/3504/1/63100285-Katja_Tuma-Drevesno_preiskovanje_Monte_Carlo_s_Thompsonovim_vzor%C4%8Denjem_pri_igri_Prebivalci_otoka_Catan_(3).pdf) (two languages). 

A recent paper following up on these things is [Enhancements in Monte Carlo Tree Search
[...]

Issue #855, opened 2018-02-11
Is it reasonable to just average evaluated value?


wpstmxhs:
As you know, there are some different approaches to utilize value network's evaluation.

For example, An open sourced computer go program RN, just averages value network evaluations and uses it, like Leela Zero. but another open sourced computer go program AQ, doesn't simply average, but averages in depth-weighted manner. It makes sense, because deeper evaluation of value network is more accurate than shallower one.

I want to discuss about this and listen everyone's opinions.

jkiliani:
How does the UCT search done by AQ work? Also, what neural network architecture do they currently use? (2.1.1)

wpstmxhs:
And I have an new idea. See the below picture below first.

This game my supervised 20-block Leela Zero bot lost by tygem 9D human, because bot played a severe blunder.

![board](https://user-images.githubusercontent.com/28377280/36072840-43932df2-0f6a-11e8-9699-32cc282a089b.png) (Result: W+Resign)

And the following is problematic situation which my bot caused a bug and the next one is its evaluation of candidate moves.

![board2](https://user-images.githubusercontent.com/28377280/36072848-78082cfe-0f6a-11e8-83c5-b6f55dd9e132.png) (Black to play)

```
 D19 ->    3139 (V: 89.48% / MV: 88.05%) (N:  2.99%) PV: D19 H1 J2 G1 J7 B19 G15 L14 F15 E16 N15 P12 Q12 P5 O6 C9 B9 A15 A14 M16
  S2 ->    2712 (V: 89.79% / MV: 89.79%) (N:  0.13%) PV: S2 S3 G1 J2 J3 H1 J4 J7 G2 H2 G15
  H1 ->    2665 (V: 89.54% / MV: 89.13%) (N:  9.07%) PV: H1 P12 Q12 L14 G15 N15 N16 P5 O6 F15 G13 C9 B9 A15 A14 B19
 G15 ->    1369 (V: 88.33% / MV: 86.84%) (N: 18.28%) PV: G15 P12 Q12 H1 J2 G1 J7 C9 B9 P5 O6 N15 N16 L14 S19 R19 T17
  A6 ->    1226 (V: 88.11% / MV: 88.09%) (N: 19.87%) PV: A6 A4 D19 B19 G1 J2 J7 P5
[...]

PR #883, opened 2018-02-17
Choose move at root using highest lower confidence bound instead of visit count

roy7:
From discussion in threads https://github.com/gcp/leela-zero/issues/860 and https://github.com/gcp/leela-zero/issues/855 I came up with some ideas I wanted to try. This is the first of 3 experiments. In summary:

Step 1- Choose a move at the root based on best lower confidence bound (LCB) instead of visit count once search has completed.
Step 2- Choose a move by searching the tree using best LCBs once search has completed, instead of only using the totals at the root level.
Step 3 - Adjust search so that the value of a node is (sum of all child nodes wins) / (sum of all child node visits) but only for children where their UCB is higher than the LCB of the best child. Don't let moves that can't possibly be best (as far as we know) drag node value up or down.

This PR is just for Step 1, for others to review and experiment with. This change is a SPRT pass vs /next for strong and weak networks. Options used were "--timemanage off -r 1 --playouts 1600 -w 8f00 --noponder --gtp" for the strong network, and -w 1ac2 for the weak network.

Strong network:
```
next v max_lcb (223/400 games)
board size: 19   komi: 7.5
next        88 39.46%       43  36.75%     45  42.45%    179.39
max_lcb    135 60.54%       61  57.55%     74  63.25%    188.32
                            104 46.64%     119 53.36%
```

Weak network:
```
next v max_lcb (226/400 games)
board size: 19   komi: 7.5
next        91 40.27%       59  52.21%     32 28.32%    229.52
max_lcb    135 59.73%       81  71.68%     54 47.79%    246.61
                            140 61.95%     86 38.05%
```

Should CI_ALPHA be stronger than .001? I just tried that first but it worked so well I didn't change it. We could change it to something like .00001 if someone wanted to retest.


roy7:
For the moment I'd recommend any use of this PR to add --timemanage off to your options. That is designed to exploit how visits used to work to save clock time, but this PR might choose a different moves than the most visited. It'd be best to use normal full playout/visit counts before moving.

roy7:
Someone suggested fitting a Student's t-distribution instead of comparing confidence bounds, so that's a potential experiment.

Eddh:
Even though it would be a big departure from the AlphagoZero paper, would it be possible to train the policy to predict lcb instead of visit count ?

roy7:
Giving this a fresh validation run on the new strong network 0fb6. Initial start isn't good (13-19) but I know that's too small to trust (3 different results away from 50/50). Will update this post once I have real results. Just making sure this technique still works as network gets stronger.

roy7:
If this did happen to get merged one day, we may want to make timemanage default to off so we'll use more time/playouts/visits when available to potentially find better moves than the most visited mode. If so, should I add that change as part of this PR?

I suspect one reason the win %s in head to head play isn't a big increase is because this code should rarely actually override the current default move. It's possible most games it is never used at all. I should have thought to put a note in the log files to grep for after a big validation was finished to get a count. Maybe I'll do that next time I run a round of tests. I might also move confidence test to 99.999% instead of 99.9% to be sure we do it as rarely as possible, and only when we're highly confident.

Ttl:
I don't think the exact value of CI_ALPHA matters in this pull request. Confidence bound is only used in sorting and if one move has higher confidence bound with one alpha it also has higher bound with other alpha values. Sorting order and the chosen move doesn't change if alpha is changed. If tree search is changed to use this method then it might behave differently with different alpha values.

roy7:
@Ttl Thank you, great call. Glad I didn't waste my time, hah. :)

On latest 0fb6:
```
next v max_lcb (400/400 games)
board size: 19   komi: 7.5
next       180 45.00%       84  42.00%     96  48.00%    213.10
max_lcb    220 55.00%       104 52.00%     116 58.00%    212.37
                            188 47.00%     212 53.00%
```

killerducky:
I think CI_ALPHA does matter. Test code and results:

```
    auto ci = 0.001;
    float a_lb = binomial_distribution<>::find_lower_bound_on_p( 1000, 0.2*1000, ci);
    float b_lb = binomial_distribution<>::find_lower_bound_on_p(   10, 0.8*  10, ci);
    myprintf("lb test CI=%f %f %f\n", ci, a_lb, b_lb);
    ci = 0.0001;
    a_lb = binomial_distribution<>::find_lower_bound_on_p( 1000, 0.2f*1000, ci);
    b_lb = binomial_distribution<>::find_lower_bound_on_p(   10, 0.8f*  10, ci);
    myprintf("lb test CI=%f %f %f\n", ci, a_lb, b_lb);
    ci = 0.00001;
    a_lb = binomial_distribution<>::find_lower_bound_on_p( 1000, 0.2f*1000, ci);
    b_lb = binomial_distribution<>::find_lower_bound_on_p(   10, 0.8f*  10, ci);
    myprintf("lb test CI=%f %f %f\n", ci, a_lb, b_lb);
    ci = 0.000001;
    a_lb = binomial_distribution<>::find_lower_bound_on_p( 1000, 0.2f*1000, ci);
    b_lb = binomial_distribution<>::find_lower_bound_on_p(   10, 0.8f*  10, ci);
    myprintf("lb test CI=%f %f %f\n", ci, a_lb, b_lb);

lb test CI=0.001000 0.162528 0.281543
lb test CI=0.000100 0.155485 0.206708
lb test CI=0.000010 0.149513 0.152821
lb test CI=0.000001 0.144280 0.113471
```

roy7:
Yes thank you! That didn't occur to me that a varying # of trials means the bounds will change at different speeds based on the confidence we want.

jkiliani:
A less strict confidence interval, like 0.02 or even up to 0.1 may actually work better. I think the reason your test at 0fb6 was less conclusive may have been the change in CI_ALPHA from 0.01 to 0.001. Very strict intervals probably do more to suppress late found stronger moves than increase confidence. CI_ALPHA approaching zero will asymptotically lead to always choosing the move with higher visit count.

roy7:
All tests have been done with .001, I don't think I used .01 in any of these tests. The only change with 0fb6 network was the network itself. I used the same leelaz binaries as the prior test.

My initial decision to use a high number was to err on the side of "using the established behavior we already rely on". Only using this technique to try and detect/avoid blunder moves found late in search when visit counts are unable to avoid the danger but LCB tests can.

It might be fun to try a head to head experiment though with a far different confidence, like .02 vs .00001.

roy7:
I've started a gomill of CI_ALPHA .02 vs .00001 for a sense of how much that would matter at all, at the simple root move selection level.

For changes to actual search algorithm, it might matter a whole lot more.

roy7:
```
max_lcb_02 v max_lcb_00001 (400/400 games)
board size: 19   komi: 7.5
max_lcb_02       204 51.00%       96  48.00%     108 54.00%    215.57
max_lcb_00001    196 49.00%       92  46.00%     104 52.00%    215.64
                                  188 47.00%     212 53.00%
```

jkiliani:
Interesting result, I expected lcb_02 to win with a more convincing margin. But it appears the optimal CI_ALPHA must be in the middle, maybe 0.001 was a good choice after all...

roy7:
I don't consider 196-204 statistically significant. :)

roy7:
I wish when I ran these tests I kept a log file of each game, not just the SGFs. I'd like to know what % of games this behavior is even used, and how the win/loss rate looks in games when it was used vs games when it never came up at all.

gcp:
Are you planning any further work on this, or should I make CI_ALPHA  a tunable and do a CLOP run?

roy7:
I do plan to do more experiments and try using LCB at deeper levels of the tree instead of just the root, but I haven't gotten there yet.

If you think this root level change by itself is worthwhile we could implement it (timemanage should default to off though?) and I can try further experiments off the new /next. Or if you want to let this float longer I'll keep experimenting in case I have a better/larger change to do all at once later.

roy7:
But note .00001 vs .02 didn't have statistically significant difference in win rate. I'm not sure it needs tuned. I suspect because it should only rarely come into play, those rare moments are probably obvious enough any reasonable confidence amount would work. I personally learn towards .001 or .00001 just from a "change as little as possible" mindset.

WhiteHalmos:
Personally I think --timemanage default on is very valuable since it saves so much time outside of self-plays.

jkiliani:
This pull shouldn't deactivate time management, but maybe you could try to actually implement a different way of time management as part of the confidence interval calculation? Say, if the PVs LCB is higher than the root UCB of all other moves, stop evaluating?

roy7:
Once I get into deeper parts of the tree it might not be so easy, but perhaps for the root only selection we could calculate some edges like "what if all remaining playouts were wins or loss for the top 2 nodes, would their order change?". If not, then I guess it's safe to stop early. That could require sorting the children each time we've going to make this check however, which will take a bit of cpu? But of course less than the moves we shave off.

Although the savings this way might be fairly small. The reason --timemanage saves so much time is because LZ often sends most of its visits down the PV, so the typical move will be made only using half the playout/visit cap. The problem when we double down in this way on "use visit # to choose moves" is I feel like this is a weak move selection basis to begin with and timemanage just exploits it.

Eddh:
I think this paper Rémi Coulom could interest you https://www.remi-coulom.fr/CG2006/CG2006.pdf

roy7:
An idea for timemanage + root_lcb is that, when timemanage kicks on, direct all remaining search to non-PV moves. Perhaps going by lcb instead of winrate to decide where those visits go.

Stated another way, the "move by # of visits" has made up its mind, so spend remaining time trying to find an alternate better move (based on higher lcb).

Just putting this down for the record, my personal interest is still more about changing search itself.

roy7:
Rebasing broke my PR. I'm unsure why atm. I now get this:

```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<std::domain_error> >'
  what():  Error in function boost::math::binomial_distribution<double>::find_lower_bound_on_p: Number of Successes argument is -nan, but must be >= 0 !
```

When it tries to sort. It's at line
`float b_lb = binomial_distribution<>::find_lower_bound_on_p( b->get_visits(), floor(b->get_eval(m_color) * b->get_visits()), CI_ALPHA);`
so get_eval is coming back 'nan'.

roy7:
Must be related to https://github.com/gcp/leela-zero/commit/642cbec221cda64e09a62a8b0dfaf169e884aab1 since recent changes mean get_eval() can sometimes return 'nan'. Must be trying to sort child where there's only 1 child that has been expanded?

roy7:
For anyone interested I'm starting down the rabbit hole here:

https://github.com/roy7/leela-zero/tree/max_lcb

Current focus is getting recalculate_sig to work, then change search to use get_visits_sig and get_eval_sig instead.

roy7:
Seems like this has gained a bit of strength with the 10 layer network, although I don't have 400 results yet. Will update once it finishes. This isn't the latest /next as of today, I 'froze' the versions when I started by copying the binaries to my gomill directory. So this is the /next from ~13 days ago.

```
next v max_lcb (94/400 games)
board size: 19   komi: 7.5
next        29 30.85%       12 25.53%     17 36.17%    207.74
max_lcb     65 69.15%       30 63.83%     35 74.47%    207.45
                            42 44.68%     52 55.32%
```

roy7:
https://github.com/roy7/leela-zero/tree/max_lcb

Now has a functional implementation of my full idea. I think. Was tested a bit on tiny # of playouts so I could examine everything, and had it play a game vs DarkGo that it won, so no major "forgot how to play" bugs on the surface. Will need to run a gomill to see what happens.

What this does is chooses final move based on best LCB like this PR does, but it also does the MCTS search only counting "significant" visits and evals. Children who we believe have no statistical significance to the moves that will actually be played are removed from data MCTS is using. This way if we find a terrible refutation to our PV, we don't need large # of visits to get off of that PV. The whole PV will drop in winrate quickly to the "real" winrate...

roy7:
Because there's many loops called often, nodes/sec is about half of before. Max GPU I can generate is like 25%. I'm sticking with -t 1 though for fear I need to do some locks someplace. Of course, faster/smarter ways to handle updates are possible. I'm just aiming for a proof of concept atm. :)

roy7:
Full max-lcb based search vs same /next as all above tests. This is using my max_lcb branch and not simply max_lcb_root PR.

```
next v max_lcb (260/400 games)
board size: 19   komi: 7.5
next       122 46.92%       63  48.46%     59  45.38%    311.39
max_lcb    138 53.08%       71  54.62%     67  51.54%    378.84
                            134 51.54%     126 48.46%
```

So not quite as strong as before in self play. However -p 1600 I suspect really restricts what my search can do because there's a minimum amount of data needed to determine significance between branches in the search tree. CPU usage is higher because I use lots of loops with no optimization.

I'm going to consider this result a "it doesn't make strength worse" confirmation.  I'm going to start a second gomill with a much higher cap, perhaps 30000 visits. My hope is that'll show a higher win rate impact.

TFiFiE:
As this is in competition with #968, the latter should be part of the baseline to test against, unless you're also planning to first add the equivalent of it for this.

roy7:
There are a few things being worked on that will interfere but I will cope somehow once those are merged.

roy7:
Slow going since I can't mutli-thread my branch yet, and the machine runs other bots too. But inching closer to an SPRT pass. This is with "--timemanage off -t 1 -q -r 10 --visits 10000 -w 5b90.w --noponder --gtp" for /next from start of this thread, and the full lcb_search idea. I started at first with 30000 visits but it took way too long (2.5 times /next cpu since I do lots of loops) and it wsan't playing as well, I suspect I was running into the ~55 second move cap.

```
next v max_lcb_search (90/400 games)
board size: 19   komi: 7.5
next               32 35.56%       13 28.89%     19 42.22%    565.67
max_lcb_search     58 64.44%       26 57.78%     32 71.11%    971.13
                                   39 43.33%     51 56.67%
```

OmnipotentEntity:
Just a passing thought, if we decide to go with this, it is also possible to change how visits are scheduled.  Currently it seems that the visits uses selection scheme based on number of visits and evaluations and some magic constants.

Perhaps this could be changed to simply selecting the branch with the highest upper confidence bound?

roy7:
@OmnipotentEntity Check out https://github.com/gcp/leela-zero/issues/860 if you haven't. There are strong alternatives to the MCTS we use. My goal with max_lcb_root and max_lcb_search was to just build upon/improve MCTS itself, not totally replace it. I may try to implement Thompson Sampling one day as an alternative.

roy7:
Fresh text of /next and max_lcb_root with a next that is only 1-2 days old. I speculated that timemanage would reduce the ability of this code to find significantly better moves to override. This test is with timemanaage on for both players:

```
leelaz-next-bce0e3d v leelaz-max_lcb_root-0db8c82 (400/400 games)
board size: 19   komi: 7.5
leelaz-next-bce0e3d            191 47.75%       89  44.50%     102 51.00%    144.11
leelaz-max_lcb_root-0db8c82    209 52.25%       98  49.00%     111 55.50%    142.99
                                                187 46.75%     213 53.25%
```

I'm running another test presently of max_lcb_root without timemanage vs normal /next with timemanage on.

gcp:
>I speculated that timemanage would reduce the ability of this code to find significantly better moves to override. 

You can stop timemanage from acting if your code would choose another move?

roy7:
> You can stop timemanage from acting if your code would choose another move?

My thinking is that the extra time would be required to find the better move in the first place. That's just speculation on my part. I could perhaps do a test where timemanage is ignored if the 2nd best move has a better bound than the 1st move. By that point the 2nd best move would be chosen if timemanage stops searching though. So it'd only help refine the bounds or change back to first move if 2nd move was a false positive. Good idea though.

We'll see how the other two ringmasters end up. Early results of tmoff vs tm for max_lcb_root on both:

```
leelaz-max_lcb_root-0db8c82 v leelaz-max_lcb_root-0db8c82-tmoff (149/400 games)
board size: 19   komi: 7.5
leelaz-max_lcb_root-0db8c82           63 42.28%       28 37.33%     35 47.30%    169.31
leelaz-max_lcb_root-0db8c82-tmoff     86 57.72%       39 52.70%     47 62.67%    224.84
                                                      67 44.97%     82 55.03%
```

WhiteHalmos:
I think you'll want to test tmoff vs tmoff and tm vs tm, since with tree-reuse the extra visits actually make a difference in strength.

roy7:
@WhiteHalmos I'll run all the tests. :) Just takes a couple days for 400 matches, I only use -j 1 since all my bots run on this machine.

afalturki:
@roy7 Thanks for doing this. Is it possible to make a test with a higher amount of visits to check if this method becomes more effective the more visits you have?

roy7:
@afalturki I think that was with the full max_lcb_search which goes beyond want this PR does. I should redo this PR at 30K and see what happens. I have 2 tests running now and 1 more to do after, then I'll run a 30K visit test. Should it be with or without timemanage I wonder.

roy7:
2nd of my 4 tests is complete. We could simply say more search = stronger, so this doesn't show much. But I wanted to test all scenarios.

```
leelaz-max_lcb_root-0db8c82 v leelaz-max_lcb_root-0db8c82-tmoff (400/400 games)
board size: 19   komi: 7.5
leelaz-max_lcb_root-0db8c82          175 43.75%       77  38.50%     98  49.00%    202.15
leelaz-max_lcb_root-0db8c82-tmoff    225 56.25%       102 51.00%     123 61.50%    267.58
                                                      179 44.75%     221 55.25%
```

roy7:
```
leelaz-next-bce0e3d v leelaz-max_lcb_root-0db8c82-tmoff (400/400 games)
board size: 19   komi: 7.5
leelaz-next-bce0e3d                  174 43.50%       80  40.00%     94  47.00%    209.07
leelaz-max_lcb_root-0db8c82-tmoff    226 56.50%       106 53.00%     120 60.00%    276.53
                                                      186 46.50%     214 53.50%
```

WhiteHalmos:
I assume the large CPU increase might be reduced when the code is optimized?

roy7:
There's not much to optimize. Note -tmoff means timemanage is off, so the # of visits might be twice as high in some cases. In the first of the 3 benchmarks, both versions used timemanage and the cpu usage was almost identical. My final test running is next vs PR with timemanage off for both.

roy7:
```
leelaz-next-bce0e3d-tmoff v leelaz-max_lcb_root-0db8c82-tmoff (400/400 games)
board size: 19   komi: 7.5
leelaz-next-bce0e3d-tmoff            158 39.50%       68  34.00%     90  45.00%    194.13
leelaz-max_lcb_root-0db8c82-tmoff    242 60.50%       110 55.00%     132 66.00%    193.54
                                                      178 44.50%     222 55.50%
```

Here we go, If there's more thinking time done, max_lcb_root has more opportunity to find alternate moves that are superior vs most visited move. This is with time manage off for both clients.

roy7:
Given the full lcb based search code I wrote is a bit of a mess, needs optimized, and isn't thread-safe atm perhaps my efforts would be better directed to leave that alone and just experiment with whole new tree search algorithms instead (KL-UCB or Thomson Sampling). I still thing the theory is sound and it's based right off traditional PUCT search, but it isn't like it's ever been tested by actual researchers or published on like these other systems have.

gcp:
>Here we go, If there's more thinking time done, max_lcb_root has more opportunity to find alternate moves that are superior vs most visited move. This is with time manage off for both clients.

And a fixed time per game?

roy7:
I've done no tests with time controls, only visit caps. How would I set up gomill to do a test with a specific time control? Would we want something like 15 seconds byoyomi? Time based tests will be difficult for me to schedule since I run these on my same server running the bots, so actual time to move can vary based on load.

WhiteHalmos:
@roy see config here for gomill: https://github.com/gcp/leela-zero/pull/1049#issuecomment-373991479

roy7:
Would anyone else with an idle system be interested in running a fixed time test? So I wouldn't need to take the bots offline for a day to run it myself. :)

gcp:
If you have a clean and debugged pull I have free system time now.

roy7:
Using this PR should be clean and stable. Or clone https://github.com/roy7/leela-zero/tree/max_lcb_root into a new directory if it's easier.

roy7:
The messy complex experiment is in my "max_lcb" branch and hasn't infected this PR. ;) This PR keeps things clean and simple to only adjust final move choice at the end.

roy7:
@gcp If you happened to start this, it might be interesting to redo it after it finishes using the self-play settings with the new randomness. Just to see if it adds more strength in general or not to the self play vs matches.

gcp:
It makes no sense to test in self-play as the move choice there depends on the probability distribution over the spent search effort - i.e. this code has no effect there.

roy7:
I'll run a fresh test this weekend. I want to use the strongest 20 layer network we've got out there. Is it 00335835?

roy7:
Oh, the auto testing is right. My code segfaults now. Somehow the change to UCTNodePointer broke it. Hmm.

roy7:
Somehow this line
`if (a->get_visits() && b->get_visits()) {`
causes a seg fault. Gotta take off for the day though. I'll try later tonight to figure this out if noone beats me to it.

roy7:
Oh, sneaky. The -> changed to . in that section of code. Doh.

```
float UCTNodePointer::get_lcb(int color) const {
    if (is_inflated()) return read_ptr()->get_lcb(color);
    return 0.0f;
}
```

For this sort of thing, should I assert is_infalted()? This is only called on NodePointer during sorting and it is after a test that both get_visits() for a and b are >0. I assume you can't have a visits greater than 0 if the NodePointer isn't inflated?

roy7:
@gcp I assume you didn't end up trying a fixed time test on this branch? If that goes well, is it possible we might actually merge this? :)

roy7:
Since the change itself adds virtually no cpu time as shown in the tests, I guess the time parity question is going to be patch without time manage vs next with time manage? To see if next client's time savings is stronger than the patch without the time savings?

OmnipotentEntity:
As discussed on the Discord, I'm running a gomill on this.  Here's the config file.

```
competition_type = 'playoff'

record_games = True
stderr_to_log = False

players = {
    'leelaz-next' : Player("./leelaz-next -d -q -t 3 -r 10 --gpu 0 --timemanage on -w e8601c38 --gtp "),
    'leelaz-lcb' : Player("./leelaz-lcb -d -q -t 3 -r 10 --gpu 1 --timemanage off -w e8601c38 --gtp "),
}

board_size = 19
komi = 7.5

matchups = [
  Matchup('leelaz-next', 'leelaz-lcb', alternating=True, board_size=19, scorer='players', number_of_games=400),
]
```

roy7:
It's cool that @OmnipotentEntity has two identical GPUs, so he can use pondering in these matches. :)

roy7:
Early time based results not looking good though. Maybe extra clock from timemanage is worth more than occasional better move strength. If it ends up a SPRT fail in the end, I'll have to see how I might be able to adapt time manage to be more max_lcb_root compatible.

OmnipotentEntity:
```
leelaz-next v leelaz-lcb (83/400 games)
board size: 19   komi: 7.5
leelaz-next     53 63.86%       21 50.00%     32 78.05%    712.91
leelaz-lcb      30 36.14%       9  21.95%     21 50.00%    716.05
                                30 36.14%     53 63.86%
```

This is an SPRT fail, so I'm stopping it early.

Here are the generated sgfs: http://xed.space/rumble.7z

roy7:
Darn. Thanks for doing this though! My gpu isn't good for time based games since I run the bots and GPU availability changes from moment to moment.

I'll need to see what more can be done regarding early moves with lcb involved.

OmnipotentEntity:
Because I suspect there might be a disparity between my GPU speeds due to some weird hardware voodoo with my system, maybe, perhaps, (probably not, honestly I'm doing this just to rule it out), I'm running two 200 game matches instead of one 400 game match, with one engine on each GPU and then swapping GPUs for the second match.  This should control for hardware nonsense.

Hersmunch:
The difference may be due to `--timemanage on` vs `off` - see #1127.

OmnipotentEntity:
leelaz-next-0 v leelaz-lcb-1 (68/200 games)                        
```
board size: 19   komi: 7.5                                         
leelaz-next-0     40 58.82%       17 50.00%     23 67.65%    752.51                                                                    
leelaz-lcb-1      28 41.18%       11 32.35%     17 50.00%    754.90                                                                    
                                  28 41.18%     40 58.82%          

leelaz-next-1 v leelaz-lcb-0 (67/200 games)                        
board size: 19   komi: 7.5                                         
leelaz-next-1     40 59.70%       17 50.00%     23 69.70%    730.67                                                                    
leelaz-lcb-0      27 40.30%       10 30.30%     17 50.00%    729.51                                                                    
                                  27 40.30%     40 59.70%
```

Another SPRT fail, at least this time we can be sure that there's probably no appreciable difference between which graphics card it's running on.  It's almost uncannily similar.

> The difference may be due to `--timemanage on` vs `off` see #1127 

Essentially, the purpose of this test was to determine if the lcb PR without timemanage (because it's more difficult to implement timemanage with lcb) could outperform just timemanage.  This is per @gcp's statement upthread:

> The critical test for this idea seems to be fixed time per game, timemanage off for lcb vs timemanage on, but I think that's a scenario that wasn't tested.

roy7:
Running a Time based test with time manage on for both clients shouldn't be much different than the tests I did previously. But do you want to go ahead and do one for the heck of it? Just so we have explicitly tested it.

OmnipotentEntity:
Ok, I've got that running now.

Hersmunch:
Comparing the results to a test run of just next with time management on and off (with the same net) will allow you to see if there is a significant difference

roy7:
I suspect https://github.com/gcp/leela-zero/pull/883#issuecomment-383089113 is simply a matter of timemanage vs no timemanage. Where timemanage is a huge help in absolute small clocks. The search time per move was probably small enough max_lcb_root could never find better moves anyway?

It might be interesting to leave max_lcb_root out, just test next+tm vs next-tm in absolute 7.5 minute time and see if you have the same results as before.

The "fail safe" of max_lcb_root is that if there isn't enough time or search done for it to make a difference, behavior is identical to /next. I believe with "all things being equal", max_lcb_root won't result in weaker general play.

roy7:
@ttreva did some tests over at https://github.com/gcp/leela-zero/issues/860#issuecomment-383369094.

OmnipotentEntity:
Results for 450 second clock with timemanage on, -lcb vs -next 

```
leelaz-next v leelaz-lcb (394/400 games)                
board size: 19   komi: 7.5                              
leelaz-next    202 51.27%       97  49.24%     105 53.30%    577.86                                             
leelaz-lcb     192 48.73%       92  46.70%     100 50.76%    579.45                                             
                                189 47.97%     205 52.03%     
```     

This is an SPRT false result, and the difference likely to simply be random chance.  Because the entire point of timemanage is it will stop searching after one move has a huge lead on others, likely in this sort of situation there would be no opportunity for there to be a difference between -lcb's move selection and -next's.

gcp:
So is the conclusion from both tests that with fixed time and timemanage on, it's not a gain, but for fixed visits or playouts, it is?

roy7:
@gcp Correct, but with the caveat that the only fixed time test was with absolute clock. Maybe that's what you meant. @OmnipotentEntity is doing more tests, including one with byoyomi time. I think in byoyomi setting, lcb with timemanage off would be stronger. @OmnipotentEntity is also testing timemanage on vs timemanage off (without lcb used at all) to confirm if that test result was mostly just a result of the tight clock.

gcp:
Yes, fixed clock or at least a large base time. On a given set of hardware, Japanese byo yomi without a large base time is equivalent to fixed visits.

I want to make a patch that disables timemanage in Japanese byo yomi because it can't help, only hurt.

OmnipotentEntity:
Here are the results from the timemanage-on vs timemanage-off bout without lcb.  Extremely similar to the first bout.

```
leelaz-next-on v leelaz-next-off (147/400 games)                                         
board size: 19   komi: 7.5                  
leelaz-next-on      93 63.27%       44 59.46%     49 67.12%    724.02                    
leelaz-next-off     54 36.73%       24 32.88%     30 40.54%    727.06                    
                                    68 46.26%     79 53.74%
```

(This demonstrates that the difference in the first test was likely not due to lcb.)

Hersmunch:
@OmnipotentEntity, just for the record, what weights did you use for your --timemanage on/off test?

OmnipotentEntity:
I'm using the LZ125 net.  I haven't updated since I started the test because I wanted stability in the results, and I didn't want to have another unknown.

roy7:
@OmnipotentEntity When running a 450 second test do you know roughly what sort of visits per move you were generating? With the -v 200 showing such strong benefits to max_lcb_root, it seems odd the 450 second match was break even. Although that's tight on time per move, surely it's still getting ~200 visits a move?

TFiFiE:
The pruning/time-management code could be generalized by calculating the best score over all children with the remaining number of visits all added as losses for each of them (which, depending on the scoring function, would not necessarily be that of the current best child) and who would be able to reach that if all the visits were added as wins instead. It could be, though, that this would be too conservative to be effective here.

gcp:
>But when we know it's stronger in other situations and the event you're entering has longer clocks than 450 seconds, it doesn't seem wrong to use it?

It's not like there's been a test at > 450 seconds, as far as I know?

Stronger in other situations is not convincing, i.e. we know the issue with byo yomi times, which was just fixed in "next" now. From looking at the relevant thread, the situation with timemanage and fixed visits/playouts is still totally confused, i.e. #1127. I'd like to try to clear that more up before thinking about LCB in fixed visit/playout situations...

tterava:
@gcp What kind of test would you like to see to confirm lcb_root is stronger than regular next? Maybe I can help out and run it or at least contribute if it requires many playouts .

Afaik these are the results so far:
-t 1 -v 200 tm off: LCB is clear winner
-t 1 -v 1600 tm off: LCB is winner
-t 11 -v 1600 tm off: LCB is clear winner
fixed time game with tm on: No clear winner
fixed time game with tm off: LCB is winner

On equal playouts I've never seen any test where Next would beat LCB, and you can always force the same amount of playouts by using the same settings. Current time manage scheme doesn't make much sense for LCB. It's like an arbitrary cut off point. It still makes LCB stronger by allowing more time to think in spots with multiple options.

roy7:
@gcp I believe @OmnipotentEntity it looks for more specific tests to try out if there's anything you'd like him to run. Related to this PR or otherwise.

gcp:
>Afaik these are the results so far:
>-t 1 -v 200 tm off: LCB is clear winner
>-t 1 -v 1600 tm off: LCB is winner
>-t 11 -v 1600 tm off: LCB is clear winner
>fixed time game with tm on: No clear winner
>fixed time game with tm off: LCB is winner

Against what were these tested? Next with time management on or off? Which next version? We've done some time control related fixes.

tterava:
All tests were ran with equal settings on both LCB client and regular "next" client. I used one of our recent 192x15 networks. If different settings would be used then you're not comparing just move selection anymore and other factors come into play. The point was to show that under exactly the same settings, LCB should never be worse than regular "next".

Time management changes affect both clients equally and it works the same way in both move selection schemes. The idea is that both clients still get the same amount of visits regardless of what sort of time management is used, but LCB is better at choosing the move which will be played.

If you need more data let me know and I can also update clients to most recent version we have.

OmnipotentEntity:
@gcp, At the time when the tests were run, I was using the same hash that I specified in the time management thread.  Issue #1127.  ( 5b2460c )  All previous matches that I've run have been against this commit.

However, the current match I am running (900 sec absolute time without time management turned on) is against 51641a9 as that was the current commit a few days ago when the match began.

tterava:
I ran the test with this command line:
leelaz.exe -g -t 11 -p 1600 -q -r 5 -d --noponder -w ecab83bb

Final result was 126 - 84 (60.00 %) in favor of LCB_root, which is an SPRT pass.
I was using LZ 0.14.

gcp:
```
leelaz-next v leelaz-lcb (294/1000 games)
board size: 19   komi: 7.5
leelaz-next    157 53.40%       65  44.22%     92  62.59%    202.29
leelaz-lcb     137 46.60%       55  37.41%     82  55.78%    203.32
                                120 40.82%     174 59.18%

leelaz-next-v3200 v leelaz-lcb-v3200 (292/1000 games)
board size: 19   komi: 7.5
leelaz-next-v3200    154 52.74%       64  43.84%     90  61.64%    526.40
leelaz-lcb-v3200     138 47.26%       56  38.36%     82  56.16%    527.16
                                      120 41.10%     172 58.90%
```

gcp:
I tested:
https://github.com/gcp/leela-zero/tree/max_lcb_root
which is jroy7's branch rebased onto next, versus next.

First test was game in 5 minutes, second test -v 3200, ecab network.

The result at fixed time is (barring statistical noise) similar to https://github.com/gcp/leela-zero/pull/883#issuecomment-383380871, but the result at fixed visits looks unlike previous reports that claim an advantage for fixed visits/playouts.

tterava:
This is very bizarre. Roy7 even replicated the result I had with -v 200. Could threat count have an effect?
I always run my tests with -t 11 because it's been tested to be most effective with 1080 ti cards.

What's your exact command line for the -v 3200 test?

gcp:
I am testing -v 400 now to see if it's visits amount related.

gcp:
```
leelaz-next-v400 v leelaz-lcb-v400 (190/1000 games)
board size: 19   komi: 7.5
leelaz-next-v400     86 45.26%       32 33.68%     54  56.84%     38.11
leelaz-lcb-v400     104 54.74%       41 43.16%     63  66.32%     37.95
                                     73 38.42%     117 61.58%
```

Seems like it's visits dependent, and drops off at longer times.

roy7:
Hmm in the past high visits weren't a hindrance at all. I'll have to rerun some tests myself. Maybe @OmnipotentEntity will too.

roy7:
I always did my tests with a command line like "-t 1 -q -r 10 --visits 3600 -w 1934.w --noponder --gtp". Although I guess -r 10 is default these days anyway.

roy7:
Looking back here was a 3600 visit with timemanage on test I'd done.

> Fresh test of /next and max_lcb_root with a next that is only 1-2 days old. I speculated that timemanage would reduce the ability of this code to find significantly better moves to override. This test is with timemanaage on for both players, 3600 visits, 10% resign.

```
leelaz-next-bce0e3d v leelaz-max_lcb_root-0db8c82 (400/400 games)
board size: 19   komi: 7.5
leelaz-next-bce0e3d            191 47.75%       89  44.50%     102 51.00%    144.11
leelaz-max_lcb_root-0db8c82    209 52.25%       98  49.00%     111 55.50%    142.99
                                                187 46.75%     213 53.25%
```

I'll re-run this exact same test with same weights and options, just latest code, and see what happens. It was the least successful test of all the scenarios I think I ran.

tterava:
I'm currently running tests with
`-g -t 1 -v 3200 -q -r 5 -d --noponder -w ecab`

I'm 50 games in with score 30-20 and while still not significant, it seems consistent with previous tests I've done. I'll update when there's more progress.

@gcp Silly question but have you double checked that your executables are actually different? It seems weird that you don't get SPRT passes in almost any test.

roy7:
I"m only 3 games into my re-test but I'm finding it odd max_lcb_root is using less cpu time.

```
leelaz-next-6fb405f v leelaz-max_lcb_root-40a3d93 (3/400 games)
board size: 19   komi: 7.5
leelaz-next-6fb405f              1 33.33%       1  50.00%     0  0.00%    255.44
leelaz-max_lcb_root-40a3d93      2 66.67%       1 100.00%     1 50.00%    237.20
                                                2  66.67%     1 33.33%
```

roy7:
My current re-test is going poorly. I have no theories at the moment why this might be.

```
leelaz-next-6fb405f v leelaz-max_lcb_root-40a3d93 (64/400 games)
board size: 19   komi: 7.5
leelaz-next-6fb405f             36 56.25%       18 56.25%     18 56.25%    146.13
leelaz-max_lcb_root-40a3d93     28 43.75%       14 43.75%     14 43.75%    143.02
                                                32 50.00%     32 50.00%
```

tterava:
200 games in with -v 3200 and score is now 104 - 96.

It does look like higher visit counts make lcb_root have less of a difference. However from my previous tests it also seems that increasing -t makes lcb more significant.

roy7:
Final result mostly ended up a wash.
```

leelaz-next-6fb405f v leelaz-max_lcb_root-40a3d93 (400/400 games)
board size: 19   komi: 7.5
leelaz-next-6fb405f            207 51.75%       94  47.00%     113 56.50%    125.12
leelaz-max_lcb_root-40a3d93    193 48.25%       87  43.50%     106 53.00%    124.16
                                                181 45.25%     219 54.75%
```

roy7:
@tterava or @OmnipotentEntity Before we punt on this, would anyone like to run a final test with -t 4 or something?

But yeah we can close it and if things change in the future we could re-open or I submit a new PR.

gcp:
If it doesn't work at -t1 and works with -t4 then we need to look at the search & threading behavior, not LCB, I think :-)

roy7:
Looking back over this whole thread and every posted result, it seems it's much better with timemanage off than with timemanage on. And we know timemange on is stronger in many situations because the saved time really makes a difference.

My homework eventually will be to try and find an improved way to do this and still benefit from timemanage. Maybe a fast optimized max_lcb_full is the future.

roy7:
Has anyone tried running a byoyomi period test after all the timemanage changes were put into place? Since timemanage doesn't fire during byoyomi (right?) in theory max_lcb_root should still win well there.

AncalagonX:
I'd like to add some anecdotal evidence first, as I've been using the max_lcb_root branch (not full) extensively over the past week and a half. Then I'll go into a little more detail and offer to track some stats if it would be helpful. None of this should have strong bearing on -v 3200, selfplay, or match games. My games are typically played with 20-30 seconds per move plus pondering, so LZ can frequently accumulates well over 30k visits per move performed with the 15x192 weights by the time I place a stone. I use a modified Lizzie which displays the LCB value and total visits for every move with >=1 visit, so I am able to watch these counts update smoothly at a rate of 10 times per second.

Right up front, I regularly notice many reasonable LCB overrides in my games, at all levels of visits counts—sometimes at 5k visits, or at 15k visits, or even beyond 75k visits. An "LCB override" here is what I define as a move which received enough confident visits to score it highest in LCB terms, despite _not_ having received the most visits overall. Below I'll add two sets of observations on LCB, as well as a suggestion for the overall discussion here:

1) I've noticed that, anecdotally, approximately 3 in every 4 early LCB overrides turn out to be preferred by the vanilla "total visits" selection method if given enough time. The vast majority of these LCB overrides occur thousands or even tens of thousands of visits prior to acquiring enough visits for the "total visits" selection method to have chosen them. As stated, I estimated that around 75% of these early LCB overrides end up eventually the _overwhelming_ majority of visits when search is given enough time to do so. Somehow, sorting by max LCB has given LZ the ability to save large amounts of time in finding better moves for the majority of cases I've anecdotally experienced in very-high-visit-count games.

2) I estimate that around 1 in 4 LCB overrides end up with the originally preferred move eventually ending with a _higher_ LCB, and as an indirect consequence many more visits than, the previous override move. These "reversed" or "failed" overrides fall into a couple of categories that I've. The first category reaches an endstate after tens of thousands or a hundred thousand visits with each move receiving roughly an equal number of total visits, and a roughly equal LCB score—essentially interchangeable as far as LZ winning the game is concerned. The second category is more predictable: a solid LCB override lasts for a few thousand visits more, but eventually a deeper read identifies that the original high-visit move turned out to be vastly more worthwhile, in which case its LCB _and_ visit count reverse the original LCB override.

3) So. I can't donate the raw processing time to perform the 200-game gauntlets people have graciously donated the CPU and GPU time to conduct above. Is it then worthwhile for me to record various game positions as I come across them which demonstrate solid examples of "LCB overrides", collect some basic details such as visit counts at override, and whether the LCB override was eventually chosen by LZ's vanilla search, etc.? If there is value in me collating this data, I will do so as I play against 9dans on WBaduk server.

afalturki:
@OmnipotentEntity Very encouraging results. May I ask what options were chosen for the match? How many threads for example? Also, was this test started before or after the time management fixes mentioned by @gcp?

OmnipotentEntity:
```
competition_type = 'playoff'

record_games = True
stderr_to_log = False

players = {
    'leelaz-next' : Player("./leelaz-next -d -q -t 3 -r 10 --gpu 0 --timemanage off -w 18e6a6c5 --gtp", startup_gtp_commands=['time_settings 900 0 0']),
    'leelaz-lcb' : Player("./leelaz-lcb -d -q -t 3 -r 10 --gpu 1 --timemanage off -w 18e6a6c5 --gtp", startup_gtp_commands=['time_settings 900 0 0']),
}

board_size = 19
komi = 7.5

matchups = [
  Matchup('leelaz-next', 'leelaz-lcb', alternating=True, board_size=19, scorer='players', number_of_games=400),
]
```

The git commit for -next is 51641a9 which was HEAD when I began this test.

gcp:
It's missing 7e8ab1 which has a fairly large impact, but may not change the result? Still, not very clear to do with these results so far, which are contradictory to some extent. It seems obvious that if we make time management smart about extending time, then considering LCB seems potentially very useful.

The results in this thread seem mostly a wash.

OmnipotentEntity:
Is there anything left to test in your mind @gcp?  Would you like me to redo the 900 second test with current head of -next and with -lcb rebased onto head of -next?

gcp:
@OmnipotentEntity I would be surprised if the outcome was very different. The thing that would get me to merge this is if there would be a clear, important scenario (so not so much very low visit games ONLY) where LCB *clearly* helps. But all results right now are mostly a wash, with even the best or worse ones not really being all that far off from random noise.

gcp:
>Since timemanage doesn't fire during byoyomi (right?) in theory max_lcb_root should still win well there

FWIW it's identical to a fixed playouts test.

afalturki:
Am I misunderstanding something, but 56.25% seems statistically significant for a 400 games match, right?

If LCB is the same for some scenarios but better in others, shouldn't that still be considered better overall?

gcp:
@afalturki it's not the only result at fixed time, and the others were the opposite way.

>If LCB is the same for some scenarios but better in others, shouldn't that still be considered better overall?

There's no proof or clarify even on this. Most results show it worse. There are some conflicted ones where if you average everything together it's probably a wash.

TFiFiE:
What I fail to understand is why @OmnipotentEntity tested with time management turned off in `next`.

tterava:
I finally finished my test with this command line:
`-g -t 1 -v 3200 -q -r 5 -d --noponder -w ecab`

Result:
```220 wins, 178 losses
398 games played.
Status: 0 LLR 2.21549 Lower Bound -2.94444 Upper Bound 2.94444```

First engine in the test was 0.14 Master combined with LCB_root while second one was 0.14 Master.

I believe these were same parameters @gcp used while his test run ended 47% - 53%. I'm not sure why I seem to get SPRT pass almost any time i use LCB_root. Do you think this is just variance with the tests or is there possibly some other reason affecting the results?

gcp:
>What I fail to understand is why @OmnipotentEntity tested with time management turned off in next.

Oh, I missed this. It explains the discrepancy.

gcp:
>as mentioned upthread, the time management code assumptions do not necessarily hold for lcb play.

But then you need to make a pull with those changes and consider the combined impact. It makes no sense to first weaken next and then say there's an improvement.

gcp:
>Do you think this is just variance with the tests or is there possibly some other reason affecting the results?

Could be some variance in the tests. Note all results are pretty close to 50%, which means a little nudge either way changes the apparent conclusion.

gcp:
With that discrepancy at least somewhat explained, I'm going to close this. It seems at best a wash.

My recommendation would be to rework TimeControl a bit so it can return both a base time and a maximum reasonable time from max_time_for_move. Then, adapt stop_thinking so if elapsed_centis >= time_for_move, we do a check if there's something interesting (LCB not on top? highest winrate move not on top?) and if so use the extended time (if it's bigger than base time).

prune_noncontenders and friends will need some updates to take the potentially extended time into account.

OmnipotentEntity:
@gcp Would you accept a pull request with lcb + adapted time management code, provided it can defeat -next with time management?

gcp:
Yes, that's what I'm proposing to do. I personally would focus on using time management to bring LCB and most visits in agreement, but whatever is clearly stronger...

roy7:
So I just realized all of my LCB and UCB calculations have used get_eval which has virtual losses mingling in there, after looking at https://github.com/gcp/leela-zero/pull/1412. Could this be why the # of threads was maybe giving us different results?

roy7:
Can't simply use the Normal distribution to choose moves since it can higher LCBs turn to less variance even with much lower visit counts.

```Playouts: 1929, Win: 42.90%, PV: G15 G13 E15 F16 G16 F17 G17 F15 G14
Playouts: 2814, Win: 42.89%, PV: G15 G13 E15 F16 G16 F17 G17 F15 G14
84.0s left, stopping early.

 G15 ->     637 (V: 43.68%) (N:  8.02%) (LCB-N: 23.68%) (UCB-N: 63.67%) (LCB-Bi: 37.62%) (UCB-Bi: 49.87%) PV: G15 G13 E15 F16 G16 F17 G17 F15 G14 F14
 K18 ->     600 (V: 43.27%) (N: 11.98%) (LCB-N: 21.64%) (UCB-N: 64.90%) (LCB-Bi: 37.04%) (UCB-Bi: 49.65%) PV: K18 L18 G17 G15 H14 H12 G14 F15 F14 E15 F12 H16 F10
 E15 ->     563 (V: 43.43%) (N:  9.14%) (LCB-N: 25.77%) (UCB-N: 61.09%) (LCB-Bi: 37.00%) (UCB-Bi: 50.02%) PV: E15 F16 G14 G12 H16 J12 E13 D15 E11 D10
 G14 ->     344 (V: 43.86%) (N:  2.71%) (LCB-N: 26.94%) (UCB-N: 60.78%) (LCB-Bi: 35.63%) (UCB-Bi: 52.33%) PV: G14 H16 G11 J12 H12 J13 G16 G17 F16
 H13 ->     290 (V: 42.81%) (N:  9.42%) (LCB-N: 22.43%) (UCB-N: 63.19%) (LCB-Bi: 33.91%) (UCB-Bi: 52.05%) PV: H13 D18 C18 H15 H11 J12 H12 J13 H14
 G17 ->     191 (V: 42.06%) (N:  9.68%) (LCB-N: 24.41%) (UCB-N: 59.71%) (LCB-Bi: 31.20%) (UCB-Bi: 53.48%) PV: G17 G15 K18 L18 H14 H12 G14 F15 F14 E15 F12 H16 F10
 G16 ->     181 (V: 41.45%) (N: 11.85%) (LCB-N: 24.97%) (UCB-N: 57.94%) (LCB-Bi: 30.35%) (UCB-Bi: 53.20%) PV: G16 H14 K18 L18 E15 D15 F17 D18 C18 G13
 F17 ->     166 (V: 41.60%) (N: 10.43%) (LCB-N: 22.25%) (UCB-N: 60.95%) (LCB-Bi: 30.01%) (UCB-Bi: 53.87%) PV: F17 D18 C18 G15 E15 E16 H14 L12
 J12 ->      47 (V: 39.63%) (N:  5.34%) (LCB-N: 24.70%) (UCB-N: 54.56%) (LCB-Bi: 19.28%) (UCB-Bi: 62.91%) PV: J12 L12 G15 G13 G17 F16 H14
 H12 ->      33 (V: 41.67%) (N:  2.01%) (LCB-N: 21.78%) (UCB-N: 61.55%) (LCB-Bi: 17.51%) (UCB-Bi: 69.11%) PV: H12 H14 L12 K13 N11 O12
 L12 ->      41 (V: 38.65%) (N:  5.67%) (LCB-N: -3.56%) (UCB-N: 80.86%) (LCB-Bi: 17.34%) (UCB-Bi: 63.62%) PV: L12 K12 L11 L10 K11 J11 K10 K9 J10 H10 J9 J8
  Q2 ->      24 (V: 41.08%) (N:  1.85%) (LCB-N: 29.52%) (UCB-N: 52.63%) (LCB-Bi: 13.89%) (UCB-Bi: 72.98%) PV: Q2 S7 R8 R2 K18 L18 G17
 H14 ->      23 (V: 41.70%) (N:  1.48%) (LCB-N: 28.50%) (UCB-N: 54.89%) (LCB-Bi: 13.85%) (UCB-Bi: 74.11%) PV: H14 H12 G17 G15 K18 L18 G14
 F16 ->      11 (V: 36.67%) (N:  2.14%) (LCB-N: 21.53%) (UCB-N: 51.82%) (LCB-Bi:  4.55%) (UCB-Bi: 82.26%) PV: F16 D18 C18 G17 G16 H17
 G13 ->       6 (V: 39.59%) (N:  0.65%) (LCB-N: 26.67%) (UCB-N: 52.51%) (LCB-Bi:  1.65%) (UCB-Bi: 93.04%) PV: G13 G15 G10 L12 H14
 J13 ->       6 (V: 38.70%) (N:  0.81%) (LCB-N: 31.64%) (UCB-N: 45.77%) (LCB-Bi:  1.51%) (UCB-Bi: 92.72%) PV: J13 J11 G16 H14
7.1 average depth, 16 max depth
2189 non leaf nodes, 1.44 average children
3164 visits, 906195 nodes, 3163 playouts, 532 n/s```

roy7:
@tterava If you're still interested, try using get_pure_eval instead of get_eval in the lcb function call. That will ignore virtual losses. Oops. Or update to my changes in my personal branch.

roy7:
In retrospect, it turns out the reason Normal looked like a bad choice in https://github.com/leela-zero/leela-zero/pull/883#issuecomment-389378213 is because I had implemented the function wrong to provide the Normal bounds. Thanks to @Ttl for reviewing my old code to explain why my old conclusion to stop testing Normal was flawed.

PR #878, opened 2018-02-16
Add edges

WhiteHalmos:
Split change into 3+ commits to make it easier to review.
* First commit is #857, will be rebased out once that is merged.
* Second commit is #888, has no logic change, paves the way for edges.
* Third commit and onwards adds edges (16 bytes) to replace the vast majority of nodes (56 bytes).

Playouts | Nodes (KB) | Nodes (MB) | n/s | Edges (KB) | Edges (MB) | n/s | Mem | Speed
-- | -- | -- | -- | -- | -- | -- | -- | --
1000 | 158928 | 155 | 1063 | 150324 | 147 | 1086 | -5.41% | 2.16%
5000 | 263584 | 257 | 1213 | 209280 | 204 | 1262 | -20.60% | 4.04%
10000 | 391964 | 383 | 1312 | 284100 | 277 | 1360 | -27.52% | 3.66%
50000 | 1419948 | 1387 | 1350 | 886796 | 866 | 1392 | -37.55% | 3.11%
100000 | 2692292 | 2629 | 1326 | 1625284 | 1587 | 1394 | -39.63% | 5.13%

<details>
<summary>Test details (default 1600)</summary>

next (with benchmark PR) sizeof(UCTNode) == 56
```
/usr/bin/time -v ./leelaz-next -w w --benchmark            
Thinking at most 2678400.0 seconds...                                                                            
NN eval=0.476396                                        

  D4 ->     511 (V: 50.36%) (N: 28.82%) PV: D4 D16 Q4 C3 C4 D3 E3 E2 F2 F3 E4 G2 D2 F1 C2 R3 R4 Q3 P3 P2 O2 O3 P4
 N2 Q2 O1                                               
  Q4 ->     418 (V: 50.25%) (N: 25.25%) PV: Q4 D4 D16 R3 Q3 R4 R5 S5 S6 R6 Q5 S7 S4 T6 S3 C17 D17 C16 C15 B15 B14
 C14 D15                                                
 D16 ->     389 (V: 50.23%) (N: 23.48%) PV: D16 Q4 D4 C17 D17 C16 C15 B15 B14 C14 D15 B13 B16 A14 B17 C3 D3 C4 C5
 B5 B6 C6 D5                                            
  R4 ->      57 (V: 50.09%) (N:  3.76%) PV: R4 D4 D16 C17 D17 C16 C15
  Q3 ->      51 (V: 50.32%) (N:  2.86%) PV: Q3 D4 D16 C17 C16 D17 E17 E18
[...]

PR #887, opened 2018-02-19
Reduce the memory consumption of nodes v2

sethtroisi:
Status: **Fixing problems with get_children()**

This is an updated version of @ywrt  https://github.com/gcp/leela-zero/pull/574

The core idea is to only create `UCTNode` children nodes (which cost ~60 bytes of memory) for the root node and just before they are used in `play_simulation` otherwise children (a move and a score) are stored in a `vector<pair<float, short>>` of their parent (~6 bytes).

**Results**
* **66%** reduction in memory (From ~60 Bytes to ~20 Bytes) 
* increase in max_tree_size from **100 to 300 million positions** (allowing for closer to 1M playouts per position, maybe ~500K with tree reuse)
* **+6-12% speed**, measured with --benchmark -p 25000 on three machines

This change adds more complexity to UCTNode (having to lock the mutex in a many more methods, having `expand` calls mixed in with uct_select_child). As UCTNode becomes more complex the risk of low frequency rarely seen threading bugs increases. After this is checked in I promise to do a pass through UCTSearch and UCTNode and see if I can reduce the complexity and number of methods

----
**Testing** all tests performed in both debug and normal build

**--benchmark** matches for several thousand but then diverges because of network eval.
`./leelaz -w weights --noponder --benchmark -p 2323`

Hand inspected the first `uct_select_child` instance that don't match with both gpu and cpu paths, difference is in 5th sig fig of winrate calculation

0.519917 = 0.503896 + 0.234812, 564 D16
0.519903 = 0.503882 + 0.234812, 564 D16

**--benchmark** with -p 10000 and -p 100,000 to verify memory usage

	Command being timed: "./leelaz -w weights --noponder --benchmark -p 10000"
	Maximum resident set size (kbytes): 178920
        vs
	Maximum resident set size (kbytes): 388100
        and
	Maximum resident set size (kbytes): 612792
        vs
	Maximum resident set size (kbytes): 2686648

**low playouts**
`(yes "go" | head -n 60) | ./leelaz -w weights -s1 -t1 --noponder -p {1,2,5,10,20}`
Hash and Ko-Hash match for all playouts
(this discovered https://github.com/gcp/leela-zero/pull/948)

**Ringmaster** with 0fb68ccf (400 playouts)
```
leelaz-next v leelaz-small-node (200/200 games)
board size: 19   komi: 7.5
leelaz-next          101 50.50%       56  56.00%     45 45.00%     11.36
leelaz-small-node     99 49.50%       55  55.00%     44 44.00%     10.81
                                      111 55.50%     89 44.50%
```

**Ringmaster** with weights.txt (400 playouts)
```
leelaz-next v leelaz-small-node (200/200 games)
board size: 19   komi: 7.5
leelaz-next          101 50.50%       55  55.00%     46 46.00%     13.50
leelaz-small-node     99 49.50%       54  54.00%     45 45.00%     12.81
                                      109 54.50%     91 45.50%
```

**Ringmaster** with weights.txt  (800 playouts on debug with -t4)
```
leelaz-next v leelaz-small-node (400/400 games)
board size: 19   komi: 7.5
leelaz-next          202 50.50%       115 57.50%     87  43.50%     36.06
leelaz-small-node    198 49.50%       113 56.50%     85  42.50%     36.12
                                      228 57.00%     172 43.00%
```

Ringmaster had two failures earlier, I found a possible race condition with an assert and fixed that. I tested 600 more games with ringmaster and separately tested with gdb for 10 hours with no failures.

----
**Backup**

Edges (https://github.com/gcp/leela-zero/pull/878) theoretically requires 30% of original memory and in practice requires 60% of the old memory.

Unexpanded theoretically requires 15% of original memory and in practice requires 30% BUT comes at the cost of more complexity in `uct_select_child` and `UCTSearch` (having to call expand_child in a couple of extra places)

I believe the "practical limits" for playouts (on my UNIX 16 gigs machine) are
Next ~200K playouts
Edges ~300K playouts
Unexpanded ~1500K playouts

IMHO none of this really matters for self-play only for people who want to run large number of playouts.

|playouts|leelaz-next|leelaz-edges|leelaz-unexpanded|
|:---|---:|---:|---:|
|5,000|253144|199096|148932|
|50,000|1409680|874620|369920|
|250,000|6291676|3630720|1070976|

None of this has a notable impact on speed on either my fast GPU or slow GPU

----

**Peculiarities**
Currently `uct_select_child` depends on m_children sort order more than IMHO it should which makes verifying this is a no-op slightly harder. If you want to test

add to uct_select_child on line ~343 
```    if (parentvisits == 0) {
        // sort by score.
        parentvisits = 1;
    }
```

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -56,8 +56,8 @@ SMP::Mutex& UCTNode::get_mutex() {
 }
 
 bool UCTNode::create_children(std::atomic<int>& score_count,
-                              GameState & state,
-                              float & eval) {
+                              GameState& state,
+                              float& eval) {
     // check whether somebody beat us to it (atomic)
     if (has_children()) {
         return false;
@@ -302,8 +302,8 @@ void UCTNode::accumulate_eval(float eval) {
 
 // Expand nth child and remove from m_child_scores.
 UCTNode* UCTNode::expand(size_t child) {
-    // Relies on caller to hold a lockode.
     assert(child < m_child_scores.size());
+    // Relies on caller to hold a lock.
 
     // Swap score to the back.
     std::iter_swap(begin(m_child_scores) + child,
@@ -331,7 +331,8 @@ void UCTNode::expand_all() {
 
 // Index of child with highest score.
 size_t UCTNode::best_child() {
-    assert(!m_child_scores.size().empty());
+    assert(!m_child_scores.empty());
+    // Relies on caller to hold a lock.
 
     auto best = size_t{0};
     auto bestScore = -1.0f;
@@ -347,7 +348,7 @@ size_t UCTNode::best_child() {
 
 UCTNode* UCTNode::uct_select_child(int color) {
     LOCK(get_mutex(), lock);
-    assert(!m_expanded.size() && !m_child_scores.empty());
+    assert(!(m_expanded.empty() && m_child_scores.empty()));
 
     // Count parentvisits manually to avoid issues with transpositions.
     auto total_visited_policy = 0.0f;
@@ -371,7 +372,6 @@ UCTNode* UCTNode::uct_select_child(int color) {
     // Estimated eval for unknown nodes = original parent NN eval - reduction
     auto fpu_eval = get_net_eval(color) - fpu_reduction;
 
-    // positive index mean expanded nodes, negative index for unexpanded.
     auto best_index = 0;
     auto best_value = -1000.0f;
 
@@ -399,9 +399,9 @@ UCTNode* UCTNode::uct_select_child(int color) {
     }
 
     if (!m_child_scores.empty()) {
+        // Unexpanded child
         auto best_unexpanded = best_child();
 
-        // Unexpanded child
         auto winrate = fpu_eval;
         auto psa = m_child_scores[best_unexpanded].second;
         auto puct = cfg_puct * psa * numerator;
@@ -505,6 +505,8 @@ UCTNode::node_ptr_t UCTNode::find_child(const int move) {
     if (m_has_children) {
         for (auto& child : m_expanded) {
             if (child->get_move() == move) {
+                // This causes child to because null and subsequent calls to
+                // this node to segfault.
                 return std::move(child);
             }
         }
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -118,6 +118,9 @@ bool UCTNode::create_children(std::atomic<int>& score_count,
 
     assert(!nodelist.empty());
 
+    // Relock before modifying
+    lock.lock();
+
     m_child_scores.reserve(nodelist.size());
     for (const auto& node : nodelist) {
         m_child_scores.emplace_back(node.second, node.first);
@@ -341,7 +344,7 @@ size_t UCTNode::best_child_score() {
             bestScore = score;
             best = i;
         }
-    };
+    }
     return best;
 }
 
@@ -467,6 +470,8 @@ UCTNode& UCTNode::get_best_root_child(int color) {
 }
 
 UCTNode* UCTNode::get_first_child() {
+    LOCK(get_mutex(), lock);
+
     if (!m_expanded.empty()) {
         return m_expanded.front().get();
     }
@@ -503,8 +508,7 @@ UCTNode::node_ptr_t UCTNode::find_child(const int move) {
     if (m_has_children) {
         for (auto& child : m_expanded) {
             if (child->get_move() == move) {
-                // This causes child to because null and subsequent calls to
-                // this node to segfault.
+                // Child is now a nullptr, subsequent calls will cause segfault
                 return std::move(child);
             }
         }
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -197,7 +197,6 @@ void UCTNode::dirichlet_noise(float epsilon, float alpha) {
     for (auto& child : m_expanded) {
         auto score = child->get_score();
         auto eta_a = dirichlet_vector[child_cnt++];
-
         score = score * (1 - epsilon) + epsilon * eta_a;
         child->set_score(score);
     }
@@ -330,7 +329,7 @@ void UCTNode::expand_all() {
 }
 
 // Index of child with highest score.
-size_t UCTNode::best_child() {
+size_t UCTNode::best_child_score() {
     assert(!m_child_scores.empty());
     // Relies on caller to hold a lock.
 
@@ -372,14 +371,10 @@ UCTNode* UCTNode::uct_select_child(int color) {
     // Estimated eval for unknown nodes = original parent NN eval - reduction
     auto fpu_eval = get_net_eval(color) - fpu_reduction;
 
-    auto best_index = 0;
+    UCTNode* best = nullptr;
     auto best_value = -1000.0f;
 
-    FastBoard board;
-    board.reset_board(19);
-
-    for (auto i = size_t{0}; i < m_expanded.size(); i++) {
-        auto child = m_expanded[i].get();
+    for (const auto& child : m_expanded) {
         if (!child->valid()) continue;
 
         float winrate = fpu_eval;
@@ -393,14 +388,14 @@ UCTNode* UCTNode::uct_select_child(int color) {
         assert(value > -1000.0);
 
         if (value > best_value) {
-            best_index = i;
+            best = child.get();
             best_value = value;
         }
     }
 
     if (!m_child_scores.empty()) {
         // Unexpanded child
-        auto best_unexpanded = best_child();
+        auto best_unexpanded = best_child_score();
 
         auto winrate = fpu_eval;
         auto psa = m_child_scores[best_unexpanded].second;
@@ -413,7 +408,7 @@ UCTNode* UCTNode::uct_select_child(int color) {
         }
     }
 
-    return m_expanded[best_index].get();
+    return best;
 }
 
 class NodeComp : public std::binary_function<UCTNode::node_ptr_t&,
@@ -461,7 +456,7 @@ UCTNode& UCTNode::get_best_root_child(int color) {
 
     if (any_visits == false && !m_child_scores.empty()) {
         // Expand best child for inclusion in search.
-        expand(best_child());
+        expand(best_child_score());
     }
 
     assert(!m_expanded.empty());
@@ -471,11 +466,14 @@ UCTNode& UCTNode::get_best_root_child(int color) {
                               NodeComp(color))->get());
 }
 
-UCTNode* UCTNode::get_first_child() const {
-    if (m_expanded.empty()) {
-        return nullptr;
+UCTNode* UCTNode::get_first_child() {
+    if (!m_expanded.empty()) {
+        return m_expanded.front().get();
     }
-    return m_expanded.front().get();
+    if (!m_child_scores.empty()) {
+        return expand(best_child_score());
+    }
+    return nullptr;
 }
 
 const std::vector<UCTNode::node_ptr_t>& UCTNode::get_children() const {
@@ -522,6 +520,7 @@ UCTNode* UCTNode::get_nopass_child(FastState& state) {
     expand_all();
 
     LOCK(get_mutex(), lock);
+    assert(m_child_scores.empty());
 
     for (const auto& child : m_expanded) {
         auto child_move = child->get_move();
@@ -530,7 +529,6 @@ UCTNode* UCTNode::get_nopass_child(FastState& state) {
             return child.get();
         }
     }
-
     return nullptr;
 }
 
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -55,9 +55,9 @@ SMP::Mutex& UCTNode::get_mutex() {
     return m_nodemutex;
 }
 
-bool UCTNode::create_children(std::atomic<int>& nodecount,
-                              GameState& state,
-                              float& eval) {
+bool UCTNode::create_children(std::atomic<int>& score_count,
+                              GameState & state,
+                              float & eval) {
     // check whether somebody beat us to it (atomic)
     if (has_children()) {
         return false;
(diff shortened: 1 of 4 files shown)


src/UCTNode.cpp
```diff
@@ -112,53 +111,55 @@ bool UCTNode::create_children(std::atomic<int> & nodecount,
...
+    m_child_scores.reserve(nodelist.size());
     for (const auto& node : nodelist) {
-        m_children.emplace_back(
-            std::make_unique<UCTNode>(node.second, node.first, init_eval)
-        );
+        m_child_scores.emplace_back(node.second, node.first);
     }
 
-    nodecount += m_children.size();
+    m_child_init_eval = init_eval;
```
WhiteHalmos: Rather than add m_child_init_eval, rebase on #888 to save this complexity.

src/UCTNode.cpp
```diff
@@ -303,17 +308,43 @@ void UCTNode::accumulate_eval(float eval) {
     atomic_add(m_blackevals, (double)eval);
 }
 
-UCTNode* UCTNode::uct_select_child(int color) {
-    UCTNode* best = nullptr;
-    auto best_value = -1000.0f;
 
+// Expand nth child and remove from m_child_scores.
+UCTNode* UCTNode::expand(size_t child) {
```
WhiteHalmos: I think `expand` needs to be locked.

src/UCTNode.cpp
```diff
@@ -322,13 +353,21 @@ UCTNode* UCTNode::uct_select_child(int color) {
...
 
-    for (const auto& child : m_children) {
-        if (!child->valid()) {
-            continue;
-        }
+    // positive index mean expanded nodes, negative index for unexpanded.
+    auto best_index = 0;
+    auto best_value = -1000.0f;
+
+    for (auto i = size_t{0}; i < m_expanded.size(); i++) {
```
WhiteHalmos: Rather than the complexity of positive and negative indices, just initialize best_index to -1 and use that to mean it's a child of m_expanded, and use `for (auto child : m_expanded)` instead. It's simpler than changing the meaning of index.

src/UCTNode.cpp
```diff
@@ -376,43 +437,58 @@ class NodeComp : public std::binary_function<UCTNode::node_ptr_t&,
...
 UCTNode& UCTNode::get_best_root_child(int color) {
     LOCK(get_mutex(), lock);
-    assert(!m_children.empty());
 
-    return *(std::max_element(begin(m_children), end(m_children),
+    if (m_expanded.size() < 2) {
+        // NodeComp first checks visits so only need to consider m_expanded,
+        // unless few things has been expanded then for simplicity expand_all.
+        // This can be improved by expanding and returning node with max score.
+        expand_all();
```
WhiteHalmos: Not sure about the need to `expand_all` here. If m_expanded is empty, then this should just return the first child since all get_evals would be equal in that case.

sethtroisi:
Looking at memory consumption of only one move isn't a good indication of how many playouts the game can be played with. Tree reuse means you carry on average ~40% of those playouts forward and even more when time management is off and responding to a sente move.

**leela-zero/next with --timemanagement off**

|playouts|benchmark|10 moves|full game (-r0)|
|:-----|----|----|----|
|1600|161|322|616
|5000|247|640|1693|
|10000|374|1124|3270|
|20000|624|2484||

------

**Games with high playouts**

timemanegement on, 250000 playouts
Maximum resident set size (kbytes): 2148272
http://eidogo.com/#1GoYKgms7

src/UCTNode.cpp
sethtroisi: Done.

src/UCTNode.cpp
sethtroisi: in fixing this I introduced a subtle bug which worries my, I'm backing out some of the clever optimizations and changing a number of places to call expand_all()

If we want to optimize performance later I think that's a fine thing but for now I'm more worried about making mistakes.

gcp:
I'll pull #888 first, after that has been rebased.

gcp:
Can you rebase this please?

sethtroisi:
Rebased but please hold off on merging for now. I want to extensively test this PR as it has a number of edge cases.

src/UCTNode.cpp
WhiteHalmos: @sethtroisi Friendly ping for removing m_child_init_eval now that m_net_eval is available.

src/UCTSearch.h
```diff
@@ -73,12 +73,12 @@ class UCTSearch {
...
-        Maximum size of the tree in memory. Nodes are about
-        48 bytes, so limit to ~1.2G on 32-bits and about 5.5G
-        on 64-bits.
+        Maximum size of the tree in memory.
+        Nodes + Scores average less than 20 bytes.
+        limit to ~1.1G on 32-bits and about 5.6G on 64-bits.
     */
     static constexpr auto MAX_TREE_SIZE =
-        (sizeof(void*) == 4 ? 25'000'000 : 100'000'000);
+        (sizeof(void*) == 4 ? 60'000'000 : 300'000'000);
```
WhiteHalmos: 60 million nodes for 32-bit, i.e. 4GB max RAM?

sethtroisi: The original number was set to use ~ 1.2G of ram (see the now updated comment above).

sethtroisi:
@WhiteHalmos I'm seeing modest performance gains (8-15%). Would you mind patching and comparing --benchmark (which I'm loving) with next branch?

src/UCTNode.cpp
```diff
@@ -322,15 +362,25 @@ UCTNode* UCTNode::uct_select_child(int color) {
         }
     }
 
+    if (parentvisits == 0) {
+        // sort by score.
+        parentvisits = 1;
```
sethtroisi: I believe this is a no-op.

Before children had been sorted by score then would have equal eval and first child (with best score) would have been chosen.

src/UCTNode.cpp
```diff
@@ -377,64 +441,96 @@ class NodeComp : public std::binary_function<UCTNode::node_ptr_t&,
...
+    assert(!m_expanded.empty());
+    std::stable_sort(rbegin(m_expanded), rend(m_expanded), NodeComp(color));
 }
 
 UCTNode& UCTNode::get_best_root_child(int color) {
     LOCK(get_mutex(), lock);
-    assert(!m_children.empty());
 
-    return *(std::max_element(begin(m_children), end(m_children),
+    // Check if any of m_expanded have visits.
```
Hersmunch: Can this loop be avoided by doing something like:
if expanded is empty return expanded best child
best_expanded = max(expanded)
if best_expanded visits == 0 return expanded best child
return best_expanded
?

src/UCTNode.cpp
```diff
@@ -377,64 +441,96 @@ class NodeComp : public std::binary_function<UCTNode::node_ptr_t&,
...
+
 // Used to find new root in UCTSearch
 UCTNode::node_ptr_t UCTNode::find_child(const int move) {
+    LOCK(get_mutex(), lock);
+
     if (m_has_children) {
-        for (auto& child : m_children) {
+        for (auto& child : m_expanded) {
             if (child->get_move() == move) {
+                // This causes child to because null and subsequent calls to
```
Hersmunch: typo "because"?

src/UCTNode.cpp
sethtroisi: Almost works,

when best_expanded.visits == 0 && best_expanded.score > best_unexpanded.score current code would return best_expanded, your pseudo code would return best_unexpanded

src/UCTSearch.cpp
```diff
@@ -175,12 +175,19 @@ void UCTSearch::dump_stats(KoState & state, UCTNode & parent) {
...
-
-
-    if (parent.get_first_child()->first_visit()) {
+    auto first_child = parent.get_first_child();
+    if (first_child == nullptr || first_child->first_visit()) {
         return;
     }
 
+    if (parent.get_children().size() < 2) {
+        // There's a small change only one child has been expanded and debug
```
Hersmunch: typo "change"

src/UCTNode.cpp
```diff
@@ -304,16 +300,60 @@ void UCTNode::accumulate_eval(float eval) {
...
+
+    auto best = size_t{0};
+    auto bestScore = -1.0f;
+    for (auto i = size_t{0}; i < m_child_scores.size(); i++) {
+        auto score = m_child_scores[i].second;
+        if (score > bestScore) {
+            bestScore = score;
+            best = i;
+        }
+    };
```
Hersmunch: spurious ;

src/UCTNode.cpp
Hersmunch: Whoops, so it does. In that case the best_expanded and best_unexpanded scores can be compared to decide which to return.

src/UCTNode.cpp
```diff
@@ -322,15 +362,25 @@ UCTNode* UCTNode::uct_select_child(int color) {
...
 
-    for (const auto& child : m_children) {
-        if (!child->valid()) {
-            continue;
-        }
+    auto best_index = 0;
+    auto best_value = -1000.0f;
+
+    FastBoard board;
+    board.reset_board(19);
```
Hersmunch: What is this doing?

WhiteHalmos:
Here's performance on my machine. About 6% increase in performance. I am slightly concerned that it doesn't generate the same results.

With node v2:
```
Thinking at most 2678400.0 seconds...
NN eval=0.507272

  D4 ->     563 (V: 51.55%) (N: 32.09%) PV: D4 Q4 D16 C3 D3 C4 C5 B5 B6 C6 D5 B7 B4 A6 B3 C17 D17 C16 C15 B15 B14 C14 D15 B13
 D16 ->     403 (V: 51.35%) (N: 25.21%) PV: D16 Q4 D4 C3 D3 C4 C5 B5 B6 C6 D5 B7 B4 A6 B3 C17 D17 C16 C15 B15 B14
  Q4 ->     265 (V: 51.36%) (N: 16.52%) PV: Q4 D4 D16 R3 R4 Q3 P3 P2 O2 O3 P4 N2 Q2 O1 R2 C17 D17 C16
 D17 ->      83 (V: 52.06%) (N:  3.30%) PV: D17 D4 Q4 R3 R4 Q3 P3 P2 O2
 R17 ->      68 (V: 51.05%) (N:  4.72%) PV: R17 R16 Q17 P17 P18 O18 O17 P16 N18 Q18 O19 R18 D16 Q4 D4
 C16 ->      63 (V: 51.85%) (N:  2.99%) PV: C16 D4 Q4 R3 Q3 R4 R5 S5 S6
  D3 ->      51 (V: 51.59%) (N:  2.66%) PV: D3 D16 Q4 R3 R4 Q3 P3 P2
  Q3 ->      41 (V: 51.68%) (N:  2.09%) PV: Q3 D4 D16 C17 C16 D17 E17 E18
  C4 ->      32 (V: 51.19%) (N:  2.15%) PV: C4 D16 Q4 R3 R4 Q3 P3 P2
  R4 ->      31 (V: 50.94%) (N:  2.23%) PV: R4 D4 D16 C17 C16 D17
1601 visits, 1961 nodes (563610 scores), 1600 playouts, 1103 n/s
```

With next
```
Thinking at most 2678400.0 seconds...
NN eval=0.507272

  D4 ->     563 (V: 51.55%) (N: 32.09%) PV: D4 Q4 D16 C3 D3 C4 C5 B5 B6 C6 D5 B7 B4 A6 B3 C17 D17 C16 C15 B15 B14 C14 D15 B13
 D16 ->     403 (V: 51.35%) (N: 25.21%) PV: D16 Q4 D4 C3 D3 C4 C5 B5 B6 C6 D5 B7 B4 A6 B3 C17 D17 C16 C15 B15 B14
  Q4 ->     265 (V: 51.36%) (N: 16.52%) PV: Q4 D4 D16 R3 R4 Q3 P3 P2 O2 O3 P4 N2 Q2 O1 R2 C17 D17 C16
 D17 ->      83 (V: 52.06%) (N:  3.30%) PV: D17 D4 Q4 R3 R4 Q3 P3 P2 O2
 R17 ->      68 (V: 51.05%) (N:  4.72%) PV: R17 R16 Q17 P17 P18 O18 O17 P16 N18 Q18 O19 R18 D16 Q4 D4
 C16 ->      63 (V: 51.85%) (N:  2.99%) PV: C16 D4 Q4 R3 Q3 R4 R5 S5 S6
  D3 ->      51 (V: 51.59%) (N:  2.66%) PV: D3 D16 Q4 R3 R4 Q3 P3 P2
  Q3 ->      41 (V: 51.68%) (N:  2.09%) PV: Q3 D4 D16 C17 C16 D17 E17 E18
  C4 ->      32 (V: 51.19%) (N:  2.15%) PV: C4 D16 Q4 R3 R4 Q3 P3 P2
  R4 ->      31 (V: 50.94%) (N:  2.23%) PV: R4 D4 D16 C17 C16 D17
1601 visits, 565571 nodes, 1600 playouts, 1038 n/s
```

sethtroisi:
@WhiteHalmos I lost all of yesterday to tracking down mismatches in benchmark

I added this comment in the testing section:

Hand inspected the first `uct_select_child` instance that don't match with both gpu and cpu paths, difference is in 5th sig fig of winrate (AKA neural net eval) calculation

on such example
0.519917 = 0.503896 + 0.234812, 564 D16
0.519903 = 0.503882 + 0.234812, 564 D16

src/UCTNode.cpp
sethtroisi: Fixed

src/UCTNode.cpp
sethtroisi: I was using this for debug and it didn't get pulled out. Thanks for noticing

src/UCTSearch.cpp
sethtroisi: changed to chance

src/UCTSearch.h
sethtroisi: Those are **nodes**, max_tree_size counts both expanded and unexpanded nodes.
With 250,000 playouts, tree reuse, and timemanage off you'll probably see 400,000 playouts after some sente move.
~400,000 nodes * ~300 scores/node = 120,000,000 (nodes+scores) * 20 bytes = 2.4 GB


 there (360-move)

src/UCTNode.cpp
sethtroisi: I think that would make the score sufficiently more complex (I'd either have to have a duplicate code of NodeComp logic or expand(best_child_score) either temporarily with logic to handle that or permanently in which case it's the same as this code.

src/UCTNode.cpp
```diff
@@ -377,61 +439,97 @@ class NodeComp : public std::binary_function<UCTNode::node_ptr_t&,
...
-    return m_children.front().get();
+    if (!m_child_scores.empty()) {
+        return expand(best_child_score());
+    }
+    return nullptr;
 }
 
 const std::vector<UCTNode::node_ptr_t>& UCTNode::get_children() const {
-    return m_children;
+    return m_expanded;
```
sethtroisi: This method is no longer safe and fixing will take sometime

src/UCTNode.cpp
Hersmunch: Agreed. Oh well.
Maybe I’ll have a play to see if there is any worthwhile performance gain by not expanding the best unexpanded to warrant e.g. making a function to compare an unexpanded to an expanded node that calls a function with the common code from NodeComp pulled out.
Shall I remove this review comment?

gcp:
Going to close this a seth seems to be busy with other work, and some smart way of doing tree trimming may be more efficient / easier to implement.

PR #888, opened 2018-02-19
Simplify m_init_eval to m_net_eval

WhiteHalmos:
Rather than have each child store its parent's net eval, have every node store the net eval of its own game state. No change in behaviour (verified with --benchmark).


gcp:
Could you rebase this?

WhiteHalmos:
Done.

PR #917, opened 2018-02-25
V2pipeline

ywrt:
Refactor the training input pipeline.

1. Introduce a binary format (v2).
2. Use the v2 format for the shuffle buffer (shrinks RAM by ~2x)
3. Support for loading v2 format off disk (much faster to parse)
4. Amend test to be full end-to-end test, including worker interfaces.
5. Add benchmark for the tensorflow frontend.
6. Change worker->main process communication to be fixed size binary format instead of protobuf.

Net result is that the pipeline now saturates at about 45,000 moves/sec into tensorflow (aka ~88 batches/sec with n=512).

### DIFF
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -43,7 +43,7 @@ BATCH_SIZE = 512
 
 # Use a random sample of 1/16th of the input data read. This helps
 # improve the spread of games in the shuffle buffer.
-DOWN_SAMPLE = 8
+DOWN_SAMPLE = 16
 
 def remap_vertex(vertex, symmetry):
     """
@@ -107,7 +107,7 @@ class ChunkParser:
         # set the down-sampling rate
         self.sample = sample
 
-        # V2 Format (see 'PackedMove' in Training.h)
+        # V2 Format
         # int32 version (4 bytes)
         # (19*19+1) float32 probabilities (1448 bytes)
         # 19*19*16 packed bit planes (722 bytes)
@@ -184,7 +184,6 @@ class ChunkParser:
         probs = probabilities.tobytes()
         assert(len(probs) == 362 * 4)
 
-
         # Load the game winner color.
         winner = float(text_item[18])
         assert winner == 1.0 or winner == -1.0
@@ -256,7 +255,7 @@ class ChunkParser:
             #print("V2 chunkdata")
             items = [ chunkdata[i:i+self.v2_struct.size]
                         for i in range(0, len(chunkdata), self.v2_struct.size) ]
-            if self.sample > 0:
+            if self.sample > 1:
                 # Downsample to 1/Nth of the items.
                 items = random.sample(items, len(items) // self.sample)
             return items
@@ -266,7 +265,7 @@ class ChunkParser:
 
             result = []
             for i in range(0, len(file_chunkdata), DATA_ITEM_LINES):
-                if self.sample > 0:
+                if self.sample > 1:
                     # Downsample, using only 1/Nth of the items.
                     if random.randint(0, self.sample-1) != 0:
                         continue  # Skip this record.
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -247,13 +247,6 @@ class ChunkParser:
 
         return self.raw_struct.pack(winner, probs.tobytes(), planes)
 
-    def load_chunk_file(self, filename):
-        """
-            Read chunk file off disk
-        """
-        with gzip.open(filename, 'r') as chunk_file:
-            return chunk_file.read()
-
     def convert_chunkdata_to_v2(self, chunkdata):
         """
             Take chunk of unknown format, and return it as a list of
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -16,16 +16,22 @@
 #    You should have received a copy of the GNU General Public License
 #    along with Leela Zero.  If not, see <http://www.gnu.org/licenses/>.
 
-import sys
+
+from tfprocess import TFProcess
+import binascii
 import glob
 import gzip
-import random
+import itertools
 import math
 import multiprocessing as mp
 import numpy as np
-import time
+import queue
+import random
+import struct
+import sys
 import tensorflow as tf
-from tfprocess import TFProcess
+import time
+import threading
 
 # 16 planes, 1 side to move, 1 x 362 probs, 1 winner = 19 lines
 DATA_ITEM_LINES = 16 + 1 + 1 + 1
@@ -34,6 +40,10 @@ DATA_ITEM_LINES = 16 + 1 + 1 + 1
 # of RAM in your GPU and the network size. You need to adjust the learning rate
 # if you change this.
 BATCH_SIZE = 512
+
+# Use a random sample of 1/16th of the input data read. This helps
+# improve the spread of games in the shuffle buffer.
+DOWN_SAMPLE = 8
 
 def remap_vertex(vertex, symmetry):
     """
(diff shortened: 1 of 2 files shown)


ywrt:
I got frustrated by the parsing speed when I was trying to run some experiments, so I changed a few things.

I think the only feature that it useful with v1 format training files is the larger shuffle buffer, albeit with the tradeoff that the buffer is split over the workers.
But if training files are available the (binary) v2 format, then the shuffle buffer loads at ~ 80,000 moves/second (aka ~6 seconds to fill).

Alderi-Tokori:
@ywrt 45'000 positions per second? So a x10 speedup in training? Or have I misunderstood something? If that speeds training up that much it's incredible.

ywrt:
No, the training itself is still GPU limited. This just reduces up the amount of CPU that's spend on data ingestion.

My motivation was primarily because I wanted to play around with training drastically simpler networks, where the training time was dominated by the data ingestion rate. It just happens that this code also reduces the memory footprint of the shuffle buffer et al

training/tf/parse.py
```diff
@@ -52,7 +62,36 @@ def remap_vertex(vertex, symmetry):
...
+            The data is represented in a number of formats through this dataflow
+            pipeline. In order, they are:
+
+            chunk: The name of a file containing chunkdata
+
+            chunkdata: type Bytes. Either mutiple records of v1 format, or multiple records
+            of v2 format.
+
+            v1: The original text format describing a move. 19 lines long. VERY slow
+            to decode. Typically around 2500 bytes long. Used only for backward
```
sethtroisi: should this be 27568 (4 + 4 * 393 + 4 * 6498)?

zediir:
Looking at the code this should still be backwards compatible with the v1 chuck files? I'm getting the following error:

```text
D:\My Folders\Documents\leela-zero-ywrt\models>..\training\tf\parse.py ..\..\leela-zero\traindatatygem\train.out
Found 18350 chunks
Training with 16516 chunks, validating on 1834 chunks
Using 6 worker processes.
  File "D:\My Folders\Documents\leela-zero-ywrt\training\tf\parse.py", line 535, in <module>
    main(sys.argv[1:])
  File "D:\My Folders\Documents\leela-zero-ywrt\training\tf\parse.py", line 507, in main
    shuffle_size=1<<19, sample=DOWN_SAMPLE)
  File "D:\My Folders\Documents\leela-zero-ywrt\training\tf\parse.py", line 138, in __init__
    args=(chunkdatagen, write)).start()
  File "C:\Users\zedii\AppData\Local\Programs\Python\Python36\lib\multiprocessing\process.py", line 105, in start
    self._popen = self._Popen(self)
  File "C:\Users\zedii\AppData\Local\Programs\Python\Python36\lib\multiprocessing\context.py", line 223, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "C:\Users\zedii\AppData\Local\Programs\Python\Python36\lib\multiprocessing\context.py", line 322, in _Popen
    return Popen(process_obj)
  File "C:\Users\zedii\AppData\Local\Programs\Python\Python36\lib\multiprocessing\popen_spawn_win32.py", line 65, in __init__
    reduction.dump(process_obj, to_child)
  File "C:\Users\zedii\AppData\Local\Programs\Python\Python36\lib\multiprocessing\reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
TypeError: can't pickle Struct objects

D:\My Folders\Documents\leela-zero-ywrt\models>Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Users\zedii\AppData\Local\Programs\Python\Python36\lib\multiprocessing\spawn.py", line 105, in spawn_main
    exitcode = _main(fd)
  File "C:\Users\zedii\AppData\Local\Programs\Python\Python36\lib\multiprocessing\spawn.py", line 115, in _main
    self = reduction.pickle.load(from_parent)
EOFError: Ran out of input
```

training/tf/parse.py
```diff
@@ -104,76 +165,195 @@ def convert_train_data(self, text_item, symmetry):
...
+        probabilities = np.array(text_item[17].split()).astype(np.float32)
+        if np.any(np.isnan(probabilities)):
+            # Work around a bug in leela-zero v0.3, skipping any
+            # positions that have a NaN in the probabilities list.
+            return False, None
+        assert len(probabilities) == 362
+
+        probs = probabilities.tobytes()
+        assert(len(probs) == 362 * 4)
+
```
sethtroisi: nit -nl

training/tf/parse.py
```diff
@@ -104,76 +165,195 @@ def convert_train_data(self, text_item, symmetry):
 
         # We flatten to a single array of len 16*19*19, type=np.uint8
         planes = np.concatenate(planes)
+        # and then to a byte string
+        planes = np.packbits(planes).tobytes()
 
+        # Get the 'side to move'
+        stm = text_item[16][0]
+        assert stm == "0" or stm == "1"
```
sethtroisi: consider: `assert stm in ("0", "1"), stm`

training/tf/parse.py
```diff
@@ -104,76 +165,195 @@ def convert_train_data(self, text_item, symmetry):
...
+    def convert_chunkdata_to_v2(self, chunkdata):
+        """
+            Take chunk of unknown format, and return it as a list of
+            v2 format records.
+        """
+        if chunkdata[0:4] == b'\1\0\0\0':
+            #print("V2 chunkdata")
+            items = [ chunkdata[i:i+self.v2_struct.size]
+                        for i in range(0, len(chunkdata), self.v2_struct.size) ]
+            if self.sample > 0:
```
sethtroisi: `self.sample > 1`

training/tf/parse.py
```diff
@@ -104,76 +165,195 @@ def convert_train_data(self, text_item, symmetry):
...
+                # Downsample to 1/Nth of the items.
+                items = random.sample(items, len(items) // self.sample)
+            return items
+        else:
+            #print("V1 chunkdata")
+            file_chunkdata = chunkdata.splitlines()
+
+            result = []
+            for i in range(0, len(file_chunkdata), DATA_ITEM_LINES):
+                if self.sample > 0:
```
sethtroisi: `if self.sample > 1`

training/tf/parse.py
```diff
@@ -35,6 +41,10 @@
 # if you change this.
 BATCH_SIZE = 512
 
+# Use a random sample of 1/16th of the input data read. This helps
+# improve the spread of games in the shuffle buffer.
+DOWN_SAMPLE = 8
```
gcp: Comment does not match actual value.

training/tf/parse.py
```diff
@@ -65,28 +104,50 @@ def __init__(self, chunks, workers=None):
         # Build the all-zeros and all-ones flat planes, used for color-to-move.
         self.flat_planes = [ b'\0' * 361, b'\1' * 361 ]
 
+        # set the down-sampling rate
+        self.sample = sample
+
+        # V2 Format (see 'PackedMove' in Training.h)
```
gcp: I think this part of the pull request is missing.

training/tf/parse.py
ywrt: Not quite sure where that number came from? The minimum size is 2200 (everything zero). The max size depends on the number of non-zero probabilities. The average I measured was about 2500 bytes per move?

training/tf/parse.py
ywrt: tnx. fixed.

training/tf/parse.py
ywrt: done.

training/tf/parse.py
ywrt: done.

training/tf/parse.py
ywrt: fixed (me messing with it for testing)

training/tf/parse.py
ywrt: I removed the mention. If you're interested in those changes, I'll put them into a separate pull request (that code is still a little messier, and not backward compatible yet).

gcp:
>Looking at the code this should still be backwards compatible with the v1 chuck files? I'm getting the following error:

I have the same problem now. The code isn't compatible with Windows.

Unless a quick fix is possible, this will be backed out.

ywrt:
that's annoying. What version of python is that?

(I'll send through a fix later today)

Alderi-Tokori:
I have the same problem, python 3.5.2

PR #919, opened 2018-02-25
Optimize debug experience and release build

Error323:
`-march=native` is now part of Release builds only. This ensure all debuggingsymbols remain.


gcp:
I don't see why -march=* would change this. I guess it's more likely related to -flto, but that should then be fixed in LINKER_EXE_FLAGS too.

PR #923, opened 2018-02-25
Normalize policy to uniform when the sum is denorm.

killerducky:
See also https://github.com/glinscott/leela-chess/pull/72

I guess we avoided this issue because LZ's initial random net was very small. @Ttl showed a new random 5x32 net was not so uniform, but the heatmap of the first LZ 1x8 random net looked fine (totally uniform policy, near 50% value).

But this code seems like a good safeguard.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -104,11 +104,16 @@ bool UCTNode::create_children(std::atomic<int> & nodecount,
         }
     }
 
-    // If the sum is 0 or a denormal, then don't try to normalize.
     if (legal_sum > std::numeric_limits<float>::min()) {
         // re-normalize after removing illegal moves.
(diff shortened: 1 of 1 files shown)


Error323:
Hmmm now that you mention that, perhaps we should also simpify the network massively for the first gen dataset. It safes a lot of useless compute time.

ywrt:
Additionally, it may be useful to disable de-norms. something like
```
#include <xmmintrin.h>
    _mm_setcsr(_mm_getcsr() | (_MM_FLUSH_ZERO_ON | _MM_DENORMALS_ZERO_ON));
```
presumably the _mm_setcsr() to be somewhere near main().

(unfortunately, both FTZ and DAZ need to be set to be reasonably confident of actually preventing processing denorms on Intel CPUs)

PR #928, opened 2018-02-26
Adding BOARD_SIZE define in config.h

Alderi-Tokori:
Allows compiling for any reasonable odd board size (not tested with size = 1)

### DIFF
--- a/src/FastBoard.cpp
+++ b/src/FastBoard.cpp
@@ -26,11 +26,11 @@
 #include <string>
 
 #include "Utils.h"
+#include "config.h"
 
 using namespace Utils;
 
 const int FastBoard::NBR_SHIFT;
-const int FastBoard::MAXBOARDSIZE;
 const int FastBoard::MAXSQ;
 const int FastBoard::BIG;
 const int FastBoard::PASS;
@@ -50,8 +50,8 @@ int FastBoard::get_boardsize(void) const {
 }
 
 int FastBoard::get_vertex(int x, int y) const {
-    assert(x >= 0 && x < MAXBOARDSIZE);
-    assert(y >= 0 && y < MAXBOARDSIZE);
+    assert(x >= 0 && x < BOARD_SIZE);
+    assert(y >= 0 && y < BOARD_SIZE);
     assert(x >= 0 && x < m_boardsize);
     assert(y >= 0 && y < m_boardsize);
 
--- a/src/FastBoard.cpp
+++ b/src/FastBoard.cpp
@@ -45,15 +45,9 @@ const std::array<FastBoard::square_t, 4> FastBoard::s_cinvert = {
     WHITE, BLACK, EMPTY, INVAL
 };
 
-int FastBoard::get_boardsize(void) const {
-    return m_boardsize;
-}
-
 int FastBoard::get_vertex(int x, int y) const {
     assert(x >= 0 && x < BOARD_SIZE);
     assert(y >= 0 && y < BOARD_SIZE);
-    assert(x >= 0 && x < m_boardsize);
-    assert(y >= 0 && y < m_boardsize);
 
     int vertex = ((y + 1) * m_squaresize) + (x + 1);
 
@@ -63,12 +57,12 @@ int FastBoard::get_vertex(int x, int y) const {
 }
 
 std::pair<int, int> FastBoard::get_xy(int vertex) const {
-    //int vertex = ((y + 1) * (get_boardsize() + 2)) + (x + 1);
+    //int vertex = ((y + 1) * (BOARD_SIZE + 2)) + (x + 1);
     int x = (vertex % m_squaresize) - 1;
     int y = (vertex / m_squaresize) - 1;
 
-    assert(x >= 0 && x < m_boardsize);
-    assert(y >= 0 && y < m_boardsize);
+    assert(x >= 0 && x < BOARD_SIZE);
+    assert(y >= 0 && y < BOARD_SIZE);
     assert(get_vertex(x, y) == vertex);
 
     return std::make_pair(x, y);
@@ -98,7 +92,6 @@ void FastBoard::set_square(int x, int y, FastBoard::square_t content) {
 }
 
 void FastBoard::reset_board(int size) {
-    m_boardsize = size;
     m_squaresize = size + 2;
     m_maxsq = m_squaresize * m_squaresize;
     m_tomove = BLACK;
@@ -243,8 +236,8 @@ int FastBoard::calc_reach_color(int color) const {
     auto reachable = 0;
     auto bd = std::vector<bool>(m_maxsq, false);
     auto open = std::queue<int>();
-    for (auto i = 0; i < m_boardsize; i++) {
-        for (auto j = 0; j < m_boardsize; j++) {
+    for (auto i = 0; i < BOARD_SIZE; i++) {
+        for (auto j = 0; j < BOARD_SIZE; j++) {
             auto vertex = get_vertex(i, j);
             if (m_square[vertex] == color) {
                 reachable++;
@@ -278,28 +271,26 @@ float FastBoard::area_score(float komi) const {
 }
 
 void FastBoard::display_board(int lastmove) {
-    int boardsize = get_boardsize();
-
     myprintf("\n   ");
     print_columns();
-    for (int j = boardsize-1; j >= 0; j--) {
+    for (int j = BOARD_SIZE-1; j >= 0; j--) {
         myprintf("%2d", j+1);
         if (lastmove == get_vertex(0, j))
             myprintf("(");
         else
             myprintf(" ");
-        for (int i = 0; i < boardsize; i++) {
+        for (int i = 0; i < BOARD_SIZE; i++) {
             if (get_square(i,j) == WHITE) {
                 myprintf("O");
             } else if (get_square(i,j) == BLACK)  {
                 myprintf("X");
-            } else if (starpoint(boardsize, i, j)) {
+            } else if (starpoint(BOARD_SIZE, i, j)) {
                 myprintf("+");
             } else {
                 myprintf(".");
             }
             if (lastmove == get_vertex(i, j)) myprintf(")");
-            else if (i != boardsize-1 && lastmove == get_vertex(i, j)+1) myprintf("(");
+            else if (i != BOARD_SIZE-1 && lastmove == get_vertex(i, j)+1) myprintf("(");
             else myprintf(" ");
         }
         myprintf("%2d\n", j+1);
@@ -310,7 +301,7 @@ void FastBoard::display_board(int lastmove) {
 }
 
 void FastBoard::print_columns() {
-    for (int i = 0; i < get_boardsize(); i++) {
+    for (int i = 0; i < BOARD_SIZE; i++) {
         if (i < 25) {
             myprintf("%c ", (('a' + i < 'i') ? 'a' + i : 'a' + i + 1));
         }
@@ -408,8 +399,8 @@ std::string FastBoard::move_to_text(int move) const {
     column--;
     row--;
 
-    assert(move == FastBoard::PASS || move == FastBoard::RESIGN || (row >= 0 && row < m_boardsize));
-    assert(move == FastBoard::PASS || move == FastBoard::RESIGN || (column >= 0 && column < m_boardsize));
+    assert(move == FastBoard::PASS || move == FastBoard::RESIGN || (row >= 0 && row < BOARD_SIZE));
+    assert(move == FastBoard::PASS || move == FastBoard::RESIGN || (column >= 0 && column < BOARD_SIZE));
 
     if (move >= 0 && move <= m_maxsq) {
         result << static_cast<char>(column < 8 ? 'A' + column : 'A' + column + 1);
@@ -434,11 +425,11 @@ std::string FastBoard::move_to_text_sgf(int move) const {
     column--;
     row--;
 
-    assert(move == FastBoard::PASS || move == FastBoard::RESIGN || (row >= 0 && row < m_boardsize));
-    assert(move == FastBoard::PASS || move == FastBoard::RESIGN || (column >= 0 && column < m_boardsize));
+    assert(move == FastBoard::PASS || move == FastBoard::RESIGN || (row >= 0 && row < BOARD_SIZE));
+    assert(move == FastBoard::PASS || move == FastBoard::RESIGN || (column >= 0 && column < BOARD_SIZE));
 
     // SGF inverts rows
-    row = m_boardsize - row - 1;
+    row = BOARD_SIZE - row - 1;
 
     if (move >= 0 && move <= m_maxsq) {
         if (column <= 25) {
@@ -536,8 +527,8 @@ std::string FastBoard::get_string(int vertex) const {
 std::string FastBoard::get_stone_list() const {
     std::string result;
 
-    for (int i = 0; i < m_boardsize; i++) {
-        for (int j = 0; j < m_boardsize; j++) {
+    for (int i = 0; i < BOARD_SIZE; i++) {
+        for (int j = 0; j < BOARD_SIZE; j++) {
             int vertex = get_vertex(i, j);
 
             if (get_square(vertex) != EMPTY) {
--- a/src/FastBoard.cpp
+++ b/src/FastBoard.cpp
@@ -45,9 +45,15 @@ const std::array<FastBoard::square_t, 4> FastBoard::s_cinvert = {
     WHITE, BLACK, EMPTY, INVAL
 };
 
+int FastBoard::get_boardsize(void) const {
+    return m_boardsize;
+}
+
 int FastBoard::get_vertex(int x, int y) const {
     assert(x >= 0 && x < BOARD_SIZE);
     assert(y >= 0 && y < BOARD_SIZE);
+    assert(x >= 0 && x < m_boardsize);
+    assert(y >= 0 && y < m_boardsize);
 
     int vertex = ((y + 1) * m_squaresize) + (x + 1);
 
@@ -57,12 +63,12 @@ int FastBoard::get_vertex(int x, int y) const {
 }
 
 std::pair<int, int> FastBoard::get_xy(int vertex) const {
-    //int vertex = ((y + 1) * (BOARD_SIZE + 2)) + (x + 1);
+    //int vertex = ((y + 1) * (get_boardsize() + 2)) + (x + 1);
     int x = (vertex % m_squaresize) - 1;
     int y = (vertex / m_squaresize) - 1;
 
-    assert(x >= 0 && x < BOARD_SIZE);
-    assert(y >= 0 && y < BOARD_SIZE);
+    assert(x >= 0 && x < m_boardsize);
+    assert(y >= 0 && y < m_boardsize);
     assert(get_vertex(x, y) == vertex);
 
     return std::make_pair(x, y);
@@ -92,6 +98,7 @@ void FastBoard::set_square(int x, int y, FastBoard::square_t content) {
 }
 
 void FastBoard::reset_board(int size) {
+    m_boardsize = size;
     m_squaresize = size + 2;
     m_maxsq = m_squaresize * m_squaresize;
     m_tomove = BLACK;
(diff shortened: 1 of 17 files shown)


Alderi-Tokori:
This is a basic change adding a define into config.h allowing users to compile Leelaz with various board sizes. Due to the way winograd tiles are handled, I can only guarantee for it to be working on odd sizes (3x3, 5x5, 7x7, 9x9, etc.).

src/OpenCL.cpp
```diff
@@ -187,8 +187,8 @@ void __in_transform_eq(float x[4][4], __global float * restrict V, int offset, i
 __kernel void in_transform(__global net_t * restrict in, __global float * restrict V,
                            const int C, const int Cpad,
                            const int Ppad) {
-    const int W = 19;
-    const int H = 19;
+    const int W = BOARD_SIZE;
```
sethtroisi: consider adding BOARD_SQUARES = BOARD_SIZE*BOARD_SIZE;
which can be substituted in dozens of places here.

src/Training.cpp
```diff
@@ -234,15 +234,15 @@ void Training::dump_training(int winner_color, OutputChunker& outchunk) {
                               | plane[bit + 3] << 0;
                 out << std::hex << hexbyte;
             }
-            // 361 % 4 = 1 so the last bit goes by itself
+            // BOARD_SIZE * BOARD_SIZE % 4 = 1 so the last bit goes by itself
```
sethtroisi: This is only true for odd sized boards.

src/config.h
```diff
@@ -29,6 +29,11 @@
 #define HAVE_SELECT
 #endif
 
+/*
+ * BOARD_SIZE: Define size of the board to compile Leela with, must be an odd number due to winograd tiles
+ */
+#define BOARD_SIZE 19
+
```
sethtroisi: #if (BOARD_SIZE % 2 == 0)
#error code assumes odd board size remove at your own risk
#endif

jkiliani:
Is it actually necessary to use a define I config.h here? What about coding it so that board size is automatically read from the weight file?

gcp:
There is no ambiguity (the filter count in the policy head should identify it), but parts of the code size some arrays statically at compile time based on the max board size, so unless you want to compile in a huge value by default (wasting memory and slowing things down), much more changes are needed to support dynamic board sizes.

src/FastBoard.h
```diff
@@ -40,7 +40,7 @@ class FastBoard {
     /*
         largest board supported
     */
-    static constexpr int MAXBOARDSIZE = 19;
+    static constexpr int MAXBOARDSIZE = BOARD_SIZE;
```
gcp: This no longer has any reason of existence.

src/Im2Col.h
```diff
@@ -70,7 +70,7 @@ template <>
 void im2col<1>(const int channels,
                const std::vector<net_t>& input,
                std::vector<float>& output) {
-    constexpr unsigned int boardsize = 19;
+    constexpr unsigned int boardsize = BOARD_SIZE;
```
gcp: Same. Pointless now.

src/FastBoard.h
gcp: Maybe defining something here or elsewhere that is BOARD_SIZE * BOARD_SIZE (SQUARE_COUNT?) or so would save a lot of redundancy.

Square count sounds a bit odd because go is played on the intersections, but I can't think of anything better right now.

src/SGFTree.cpp
```diff
@@ -123,7 +123,7 @@ void SGFTree::populate_states(void) {
         } else {
             if (!m_properties.count("SZ")) {
                 // No size, but SGF spec defines default size for Go
-                m_properties.insert(std::make_pair("SZ", "19"));
+                m_properties.insert(std::make_pair("SZ", "BOARD_SIZE"));
```
gcp: This is wrong, see comment right above the change.

src/SGFTree.cpp
```diff
@@ -153,8 +153,8 @@ void SGFTree::populate_states(void) {
         float komi;
         strm >> komi;
         int handicap = m_state.get_handicap();
-        // last ditch effort: if no GM or SZ, assume 19x19 Go here
-        int bsize = 19;
+        // last ditch effort: if no GM or SZ, assume BOARD_SIZExBOARD_SIZE Go here
+        int bsize = BOARD_SIZE;
```
gcp: Wrong for same reason as above.

src/Network.cpp
```diff
@@ -643,7 +643,7 @@ void convolve(size_t outputs,
 
     // Weight shape (output, input, filter_size, filter_size)
     // 96 22 3 3
-    // outputs[96,19x19] = weights[96,22x3x3] x col[22x3x3,19x19]
+    // outputs[96,BOARD_SIZExBOARD_SIZE] = weights[96,22x3x3] x col[22x3x3,BOARD_SIZExBOARD_SIZE]
```
gcp: This was a an example with specific values, so this doesn't make a lot of sense now. (Also Zero has 18, not 22 inputs...)

src/SGFTree.cpp
```diff
@@ -290,7 +290,7 @@ int SGFTree::string_to_vertex(const std::string& movestring) const {
         return FastBoard::PASS;
     }
 
-    if (m_state.board.get_boardsize() <= 19) {
+    if (m_state.board.get_boardsize() <= BOARD_SIZE) {
```
gcp: Wrong.

Alderi-Tokori:
Alright, I'm going to work on the requested changes this evening.

src/FastBoard.h
Alderi-Tokori: Do you mean I can simply remove the line, or should I keep it with 19?

src/Network.cpp
Alderi-Tokori: Same question, should I just remove the comment then or simply put back 19 in it?

src/config.h
Alderi-Tokori: Nice, thank you for reviewing the code

src/FastBoard.h
Alderi-Tokori: As for the constant for BOARD_SIZE * BOARD_SIZE, maybe BOARD_SQUARES as @sethtroisi proposed, or something like INTERSECTION_COUNT?

src/Im2Col.h
Alderi-Tokori: I don't understand this, the variable is used and the function itself is called somewhere in the code, so it doesn't seem useless to me, but if I put back 19, it still works (despite the following assert that should stop the program). 

Should I remove the 3 lines ?
```CPP
    constexpr unsigned int boardsize = 19;
    auto outSize = size_t{channels * boardsize * boardsize};
    assert(output.size() == outSize);
```

gcp:
Dirichlet noise should also be rescaled according to board size (see AZ paper).

src/FastBoard.h
gcp: Let's take BOARD_SQUARES. The line above can go as MAXBOARDSIZE==BOARD_SIZE.

src/Im2Col.h
gcp: I mean that it's setting boardsize==BOARD_SIZE.

Just use BOARD_SIZE.

src/Network.cpp
gcp: I'd put the 19x19 back in.

Alderi-Tokori:
Good catch, it never occurred to me.

Alderi-Tokori:
I think a crude linear proportionality to the number of intersections might do it (after all, I've trained mine with the Dirichlet noise intended for 19x19 and I haven't had any major problem).

gcp:
I think the AZ paper set out their reasoning?

" Dirichlet noise Dir(α) was added to the prior probabilities in the
root node; this was scaled in inverse proportion to the approximate number of legal moves in a
typical position, to a value of α = {0.3, 0.15, 0.03} for chess, shogi and Go respectively."

Chess ~35 and 19x19 Go ~361

Alderi-Tokori:
Yes, but I'm not sure the average number of legal moves is actually linearly proportional to the number of intersections in go (might be, might not be), and I can't really have a look-up table in the program to have a perfect Dirichlet noise for each board size.

Alderi-Tokori:
There we go, everything should be in order.

l1t1:
nice job

src/FastBoard.cpp
```diff
@@ -50,8 +50,8 @@ int FastBoard::get_boardsize(void) const {
 }
 
 int FastBoard::get_vertex(int x, int y) const {
-    assert(x >= 0 && x < MAXBOARDSIZE);
-    assert(y >= 0 && y < MAXBOARDSIZE);
+    assert(x >= 0 && x < BOARD_SIZE);
+    assert(y >= 0 && y < BOARD_SIZE);
     assert(x >= 0 && x < m_boardsize);
```
sethtroisi: long term we can remove m_boardsize.

src/Im2Col.h
```diff
@@ -27,8 +27,8 @@ template <unsigned long filter_size>
 void im2col(const int channels,
             const std::vector<net_t>& input,
             std::vector<float>& output) {
-    constexpr unsigned int height = 19;
-    constexpr unsigned int width = 19;
+    constexpr unsigned int height = BOARD_SIZE;
+    constexpr unsigned int width = BOARD_SIZE;
     constexpr unsigned int channel_size = height * width;
```
sethtroisi: BOARD_SQUARES

src/OpenCL.cpp
```diff
@@ -49,7 +49,7 @@ static std::string sourceCode_config = R"(
     typedef float net_t;
     #define vload_net_t(offset,p) ((p)[(offset)])
     #define vstore_net_t(data,offset,p) (((p)[(offset)])=(data))
-)";
+    #define BOARD_SIZE )" + std::to_string(BOARD_SIZE);
```
sethtroisi: This doesn't look right

src/OpenCL.cpp
```diff
@@ -127,14 +127,14 @@ __kernel void merge(
                         __global const net_t * restrict in,
                         __global net_t * restrict out,
                         __private const int channels) {
-        // cl::NDRange global(outputs, 19*19);
+        // cl::NDRange global(outputs, BOARD_SIZE*BOARD_SIZE);
```
sethtroisi: BOARD_SQUARES

src/OpenCL.cpp
```diff
@@ -127,14 +127,14 @@ __kernel void merge(
...
         const int gx = get_global_id(0);
         const int gy = get_global_id(1);
         const int output = gx;
         const int b = gy;
         const int outputs = get_global_size(0);
-        const int width = 19;
-        const int height = 19;
+        const int width = BOARD_SIZE;
+        const int height = BOARD_SIZE;
         const int boardsize = width * height;
```
sethtroisi: BOARD_SQUARES

src/SGFTree.cpp
```diff
@@ -136,7 +136,7 @@ void SGFTree::populate_states(void) {
         std::istringstream strm(size);
         int bsize;
         strm >> bsize;
-        if (bsize <= FastBoard::MAXBOARDSIZE) {
+        if (bsize <= BOARD_SIZE) {
             // Assume 7.5 komi if not specified
             m_state.init_game(bsize, 7.5f);
             valid_size = true;
```
sethtroisi: line 157 has a 19 on it that should be replaced

src/OpenCL.cpp
Alderi-Tokori: In OpenCL, I haven't defined BOARD_SQUARES : it seems not that useful to me given the only place it can be used is in this one commented line (Only place in OpenCL where there's a BOARD_SIZE*BOARD_SIZE). 

Do you really want me to replace this by BOARD_SQUARE?

src/OpenCL.cpp
Alderi-Tokori: It's the neatest way I've found in order to pass the BOARD_SIZE define into the OpenCl code source (the code doesn't have access to external defines). My aim with this is so the user only has to modify the board size in one place (namely config.h).

I'm ready to change this to some other thing if you have a better idea, but I don't have any other idea myself if I want to keep this "only has to modify the size in config.h" constraint.

src/SGFTree.cpp
Alderi-Tokori: I'd think so, yes.

src/FastBoard.cpp
Alderi-Tokori: That'd mean removing get_boardsize() and replacing every call with BOARD_SIZE. I'll do this tonight.

Alderi-Tokori:
@sethtroisi There you go, removing m_boardsize is quite a big change so I'm waiting for a review from @gcp before doing anything else.

sethtroisi:
@Alderi-Tokori Agreed, m_boardsize is quite a large change and I totally agree with holding off. I'd suggest doing it in a 2nd pull request because it will touch parameters and a dozen files which would certainly clutter this PR

Alderi-Tokori:
@sethtroisi Well, I've already done it in this one ^^' If it is rejected, I could always undo it though.

src/GameState.cpp
```diff
@@ -39,7 +39,7 @@ void GameState::init_game(int size, float komi) {
     game_history.clear();
     game_history.emplace_back(std::make_shared<KoState>(*this));
 
-    m_timecontrol.set_boardsize(board.get_boardsize());
+    m_timecontrol.set_boardsize(BOARD_SIZE);
```
sethtroisi: My bet is that this should be hardcoded in m_timecontrol now

Alderi-Tokori: Hmm true, I'll do that if gcp validates the removing of m_boardsize.

src/Network.cpp
```diff
@@ -749,7 +749,7 @@ void Network::forward_cpu(std::vector<float>& input,
     auto M = std::vector<float>(WINOGRAD_TILE * output_channels * tiles);
 
     winograd_convolve3(output_channels, input, conv_weights[0], V, M, conv_out);
-    batchnorm<361>(output_channels, conv_out,
+    batchnorm<BOARD_SQUARES>(output_channels, conv_out,
```
gcp: Fix the alignment of the line below.

src/Network.cpp
```diff
@@ -762,15 +762,15 @@ void Network::forward_cpu(std::vector<float>& input,
         std::copy(begin(conv_in), end(conv_in), begin(res));
         winograd_convolve3(output_channels, conv_in,
 	                       conv_weights[i], V, M, conv_out);
-        batchnorm<361>(output_channels, conv_out,
+        batchnorm<BOARD_SQUARES>(output_channels, conv_out,
```
gcp: Same.

src/Network.cpp
```diff
@@ -762,15 +762,15 @@ void Network::forward_cpu(std::vector<float>& input,
...
                        batchnorm_means[i].data(),
                        batchnorm_stddivs[i].data());
 
         output_channels = conv_biases[i + 1].size();
         std::swap(conv_out, conv_in);
         winograd_convolve3(output_channels, conv_in,
 			               conv_weights[i + 1], V, M, conv_out);
-        batchnorm<361>(output_channels, conv_out,
+        batchnorm<BOARD_SQUARES>(output_channels, conv_out,
```
gcp: etc

src/OpenCL.cpp
```diff
@@ -750,11 +750,11 @@ void OpenCL_Network::convolve1(int channels, int outputs,
                               cl::Buffer& bufferOutput,
                               cl::Buffer& bufferMerge,
                               weight_slice_t weights) {
-    // fixed for 19x19
-    constexpr int width = 19;
-    constexpr int height = 19;
+    // fixed for BOARD_SIZExBOARD_SIZE
```
gcp: It's no longer fixed now is it? :-)

src/SGFTree.cpp
```diff
@@ -290,13 +291,13 @@ int SGFTree::string_to_vertex(const std::string& movestring) const {
...
 
-    if (m_state.board.get_boardsize() <= 19) {
+    if (BOARD_SIZE <= 19) {
         if (movestring == "tt") {
             return FastBoard::PASS;
         }
     }
 
-    int bsize = m_state.board.get_boardsize();
+    int bsize = BOARD_SIZE;
```
gcp: This makes no sense. This code was checking whether we correctly found a usable board size at any point.

src/OpenCL.cpp
Alderi-Tokori: Ah yeah, I'll modify the comment. By the wawy, should I keep these variables in those functions or should I remove them and directly use BOARD_SIZE/BOARD_SQUARES in the code?

src/SGFTree.cpp
Alderi-Tokori: The removing of m_boardsize (and the corresponding getter, was done after one of seth's comments, suggesting it wasn't necessary anymore.  Should I put it back in?

Alderi-Tokori: (Or should I just remove this test, since board sizes != BOARD_SIZE will throw an exception in populate_state now)

Alderi-Tokori:
@gcp I'm assuming you're validating removing m_boardsize. I'll do another commit hardcoding BOARD_SIZE in TimeControl (as @sethtroisi suggested) next if you think this is a good idea?

gcp:
Yes, prefer separate pull requests.

src/OpenCL.cpp
gcp: Keep the variables for now.

src/SGFTree.cpp
gcp: > I don't think a 19x19 net would handle correctly a 9x9 board.

Indeed it would not due to the policy heads etc.

>if we haven't found either GM or SZ, it makes sense as a last effort to assume it's the right size

No, you should assume it's 19 because that is what the SGF spec says. If that's not BOARD_SIZE, give up.

>So in that case should we reject it if not specified and BOARD_SIZE != 19 to be consistent and follow SGF spec?

Exactly.

Alderi-Tokori:
Alright, that's what I was thinking. I'm going to revert the changes and clean that up.

gcp:
Sorry about changing my mind about half of this code, but I was so happy about seeing the hardcoded 19's go that it clouded my judgement.

Alderi-Tokori:
Haha no problem, I'm not particularly in a hurry to see this PR merged, and when I committed these changes without having consulted you beforehand, I was already prepared to undo them.

Alderi-Tokori:
@gcp Ok, I've reverted the last two commits then nitpicked the other changes (aside from removing m_boardsize) from those reverted commits and added a few modifications.

src/Im2Col.h
```diff
@@ -70,8 +69,7 @@ template <>
 void im2col<1>(const int channels,
                const std::vector<net_t>& input,
                std::vector<float>& output) {
-    constexpr unsigned int boardsize = 19;
-    auto outSize = size_t{channels * boardsize * boardsize};
+    auto outSize = size_t{channels * static_cast<size_t>(BOARD_SQUARES)};
```
gcp: Does this actually need both size_t casts?

src/SGFTree.cpp
```diff
@@ -124,7 +124,9 @@ void SGFTree::populate_states(void) {
             if (!m_properties.count("SZ")) {
                 // No size, but SGF spec defines default size for Go
                 m_properties.insert(std::make_pair("SZ", "19"));
-                valid_size = true;
+                if (19 == BOARD_SIZE) {
+                    valid_size = true;
```
gcp: valid_size only says whether m_boardsize has been set, not whether we can deal with it or not. So I think you don't need changes here. Later on, you use valid_size to see what the board has been set at, and there's a check that it was either set or defaulted to BOARD_SIZE.

gcp:
Looks great, I had one last round of nits.

src/Im2Col.h
Alderi-Tokori: If I don't cast the value, it raises a warning which makes it fail the AppVeyor check.

Alderi-Tokori:
@gcp Aside from the casting (which I don't think I can remove), everything should be done.

Issue #920, opened 2018-02-25
Only cpu version for linux?


emdio:
So I see in the release section an only cpu version exists for windows, but I can't see a linux one.
Am I missing something?

I do know it can be worthless in order to generate training games, but it can allow more people to be able to play against leelaz, so make the project more popular.

jkiliani:
For both Mac and Linux you have to compile yourself, whether you have a GPU or not. If you want to deactivate the GPU, you have to comment out #define USE_OPENCL in config.h before compiling leelaz.

emdio:
"By the way, it's not worthless for training game generation"

In this case, shouldn't exist at least some info in the README file? We could be missing some potential contributors here.

jkiliani:
Here is how my config.h looks after the modification:

```
/*
    This file is part of Leela Zero.
    Leela Zero is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation, either version 3 of the License, or
    (at your option) any later version.
    Leela Zero is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU General Public License for more details.
    You should have received a copy of the GNU General Public License
    along with Leela Zero.  If not, see <http://www.gnu.org/licenses/>.
*/

#ifndef CONFIG_INCLUDED
#define CONFIG_INCLUDED

/*
 * We need to check for input while we are thinking.
 * That code isn't portable, so select something appropriate for the system.
 */
#ifdef _WIN32
#undef HAVE_SELECT
#define NOMINMAX
#else
#define HAVE_SELECT
#endif

/*
 * Features
 *
 * USE_BLAS: Use a basic linear algebra library.
 * We currently require this, as not all operations are performed on
 * the GPU - some operations won't get any speedup from it.
 * Also used for OpenCL self-checks.
 */
#define USE_BLAS

/*
 * We use OpenBLAS by default, except on macOS, which has a fast BLAS
 * built-in. (Accelerate)
 */
#if !defined(__APPLE__) && !defined(__MACOSX)
#define USE_OPENBLAS
#endif

/*
 * USE_MKL: Optionally allows using Intel Math Kernel library as
 * BLAS implementation. Note that MKL's license is not compatible with the GPL,
 * so do not redistribute the resulting binaries. It is fine to use it on your
 * own system.
 */
[...]

PR #942, opened 2018-02-27
Update README for compiling CPU-olny build

akdtg:
Related to #920 issue


gcp:
We support other build systems than CMake and the current lien in the README is still correct.

PR #948, opened 2018-02-28
Fix segfault in dump stats

sethtroisi:
`echo "go" | ./leelaz -w weights -s1 -t1 --noponder -p 1`

used to throws a segfault (from `get_eval`), now it doesn't


gcp:
I rather like the 2nd move display for sanity checking purposes, so I fixed this another way.

PR #949, opened 2018-02-28
Cleanups to dump_stats

sethtroisi:
Avoid making dozens of copies of KoState, avoid copying Gamestate which used to be expensive but isn't anymore.

Some minor fixes that I noticed while making https://github.com/gcp/leela-zero/pull/948

If you don't like these sort of changes feel free to just close this.

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -26,10 +26,10 @@
 #include <type_traits>
 
 #include "FastBoard.h"
+#include "FastState.h"
 #include "FullBoard.h"
 #include "GTP.h"
 #include "GameState.h"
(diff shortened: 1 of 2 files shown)


WhiteHalmos:
Nice, any performance or RAM numbers based on `--benchmark`?

gcp:
Needs a minimal rebase because I didn't take the other pull.

sethtroisi:
@WhiteHalmos It's possible it decreases average usage but I can't measure that [[1]](https://stackoverflow.com/questions/21138012/usr-bin-time-always-give-0-for-average-resident-set-size) maximum usage isn't decreased

@gcp rebased

PR #957, opened 2018-03-01
Turn off the quiet mode in case of critical error.

### LINKED ISSUE #935: Benchmark command's quiet mode
#857 sets the quiet mode:
```
+    if (vm.count("benchmark")) {
+        cfg_quiet = true;  // Set this early to avoid unnecessary output.
+    }
```
This way any error output will be omitted.
For example, if you missed the weight file option, leelaz will stop without any visible clue:

```
~/studio/leela-zero/build$ ./leelaz --benchmark

~/studio/leela-zero/build$ 
```

akdtg:
The critical errors should never be silenced.
Fixes #935

### DIFF
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -112,6 +112,7 @@ static void parse_commandline(int argc, char *argv[]) {
                   .options(all).positional(p_desc).run(), vm);
         po::notify(vm);
     }  catch(const boost::program_options::error& e) {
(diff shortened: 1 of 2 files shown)


gcp:
A simpler fix is to change myprintf to printf.

akdtg:
I think the best way to manage the output is providing a severity level of the message as a parameter of the print function, but this needs more changes.

PR #959, opened 2018-03-02
Change mongo_training.py to v2 format.

ywrt:
Change the MongoDB dumper to dump v2 records, dump separate train/test, and shuffle on the way.

The V2 records make the training files much faster to parse.
The shuffle gives a much wider spread of games into training.
The separate train/test means that test games are never used for training, avoiding
contaminating the test.

NB! I don't have a handy MongoDB around with all the training data, so this isn't tested to the degree that I'd like. I've tested it using a fake source FWIW.

(This is based against #958 so the diff will be noisy until that's picked up. just ignore everything other than mongo_training.py

### DIFF
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -19,32 +19,23 @@
 
 from tfprocess import TFProcess
 from chunkparser import ChunkParser
-import binascii
+import argparse
 import glob
 import gzip
-import itertools
-import math
 import multiprocessing as mp
-import numpy as np
-import queue
 import random
 import shufflebuffer as sb
-import struct
 import sys
 import tensorflow as tf
 import time
-import threading
-
-# 16 planes, 1 side to move, 1 x 362 probs, 1 winner = 19 lines
-DATA_ITEM_LINES = 16 + 1 + 1 + 1
 
 # Sane values are from 4096 to 64 or so. The maximum depends on the amount
 # of RAM in your GPU and the network size. You need to adjust the learning rate
 # if you change this.
 BATCH_SIZE = 512
 
-# Use a random sample of 1/16th of the input data read. This helps
-# improve the spread of games in the shuffle buffer.
+# Use a random sample input data read. This helps improve the spread of
+# games in the shuffle buffer.
 DOWN_SAMPLE = 16
 
 def get_chunks(data_prefix):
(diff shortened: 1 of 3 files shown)


ywrt:
Brought up-to-date with 'next'.

NB! Forgot to mention that 'parse.py' now takes *three* args: 
```
parse.py <train_prefix> <test_prefix> [snapshot_prefix]
```

ywrt:
FWIW: Shuffling before compressing produces a ~3x blowup with a v2 destination.
```
$ du -sh rl-v1 rl
8.5G	rl-v1
26G	rl
```

training/tf/parse.py
```diff
@@ -35,17 +35,14 @@
...
 # of RAM in your GPU and the network size. You need to adjust the learning rate
 # if you change this.
 BATCH_SIZE = 512
 
-# Use a random sample of 1/16th of the input data read. This helps
-# improve the spread of games in the shuffle buffer.
-DOWN_SAMPLE = 16
+# Use a random sample input data read. This helps improve the spread of
+# games in the shuffle buffer.
+DOWN_SAMPLE = 2
```
gcp: Are good values of this format dependent?

training/tf/parse.py
```diff
@@ -108,17 +105,14 @@ def _parse_function(planes, probs, winner):
...
 def main(args):
     train_data_prefix = args.pop(0)
+    test_data_prefix = args.pop(0)
 
-    chunks = get_chunks(train_data_prefix)
-    print("Found {0} chunks".format(len(chunks)))
-
-    if not chunks:
+    training = get_chunks(train_data_prefix)
+    test = get_chunks(test_data_prefix)
```
gcp: This breaks the script from working with v1 input?

gcp:
This should be changed such that the name is mongo_training_v2.py and the existing script stays.

Because of the blow-up, it is important to keep supporting both formats for now.

training/tf/parse.py
ywrt: No, they're dependent on the amount of shuffling being done. The data shuffle in v2_write_training combined with the data shuffle in parse.py is enough that down sampling isn't required to get good move diversity in mini-batches.

training/tf/parse.py
ywrt: Yes. I added argument parsing to allow specifying just train files.

ywrt:
I called it 'v2_writing_training.py' as it takes both MongoDB and V1 chunkfiles as inputs.

training/tf/parse.py
```diff
@@ -19,47 +19,46 @@
...
 # of RAM in your GPU and the network size. You need to adjust the learning rate
 # if you change this.
 BATCH_SIZE = 512
 
-# Use a random sample of 1/16th of the input data read. This helps
-# improve the spread of games in the shuffle buffer.
-DOWN_SAMPLE = 16
+# Use a random sample input data read. This helps improve the spread of
+# games in the shuffle buffer.
+DOWN_SAMPLE = 1
```
gcp: Is this an appropriate default when using v1 data?

ywrt: You're quite right, it isn't! :)

I've reverted it back to 16. If you end up using the v2_write_training.py, it would be reasonable to drop this to 1.

gcp:
I'm closing this because I'm assuming it's superseded by your next pull request.

PR #961, opened 2018-03-02
Split UCTNode into UCTFullNode and UCTNode

sethtroisi:
UCTFullNode contains all the methods that should/are only called on root nodes (kill_superkos, dirichlet_noise, randomize_first_proportionally). I'm slightly prefer the name "UCTRootNode" but I went with the project convention.

Pros
* UCTNode was getting quite large: 450 lines with 7 helper methods used once each from UCTSearch
* UCTFullNode code is UCTSearch specific and disjoint generally disjoint from UCTNode logic.
* Helpful for small_node pull (UCTFullNode can have the invariant that it's fully expanded)

Cons:
* UCTFullNode doesn't have additional data, in practice every UCTNode is actually a UCTFullNode.
* dump_stats  (useful when debugging) can only be called on UCTFullnode and will require a static_cast when debugging.
* std::unique_ptr don't support automatically casting to a base class which can complicate things later.[[1]](https://stackoverflow.com/questions/17473900/unique-ptr-to-a-base-class)

### DIFF
--- a/msvc/VS2015/leela-zero.vcxproj
+++ b/msvc/VS2015/leela-zero.vcxproj
@@ -93,6 +93,7 @@
     <ClCompile Include="..\..\src\Training.cpp" />
     <ClCompile Include="..\..\src\Tuner.cpp" />
     <ClCompile Include="..\..\src\UCTNode.cpp" />
+    <ClCompile Include="..\..\src\UCTNodeRoot.cpp" />
     <ClCompile Include="..\..\src\UCTSearch.cpp" />
     <ClCompile Include="..\..\src\Utils.cpp" />
     <ClCompile Include="..\..\src\Zobrist.cpp" />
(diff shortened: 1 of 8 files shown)


gcp:
> I'm slightly prefer the name "UCTRootNode" but I went with the project convention.

I'd also prefer UCTRootNode, not sure where the convention would be?

I think you'd get much of the pros and none of the cons if you would split up the implementation of UCTNode into 2 .cpp files, and put the root specific methods into UCTNodeRoot.cpp or something like that. (The reversal of the name being for cleaner ordering in file views)

The methods would all still go in the same object, but the header isn't that big yet.

The invariant can be checked with asserts() in each method that relies on it.

roy7:
I currently use dump_stats to descend the tree to a depth I specify to help me with testing/debugging some changes. (I added a depth argument and call it recursively to that depth.) This way I can see the entire tree to whatever depth I need. So if it's still possible to do that after these changes that'll be appreciated. :)

gcp:
> not sure where the convention would be?

FWIW FastBoard vs FullBoard exist because FastBoard originally only had the bare minimum required to support a Monte Carlo playout (no undo, no smart counting, no superko, etc).

sethtroisi:
@GCP, splitting implementation is a great idea.

Issue #956, opened 2018-03-01
evals after MAX_TREE_SIZE decrease move quality 


sethtroisi:
After the search tree is "full" (size >= MAX_TREE_SIZE) UCTSearch repeatedly adds playout "visits" to nodes biasing their score towards the initial neural net eval (and away from any playout data)

I artificially lowered the limit to demonstrate this here 

```NN eval=0.478301
Playouts: 3271, Win: 49.57%, PV: D4 Q16 Q4 D16 R17 R16 Q17 P17
Playouts: 6515, Win: 49.47%, PV: D4 Q16 Q4 D16 R17 R16 Q17 P17
Playouts: 9781, Win: 49.37%, PV: D4 Q16 Q4 D16 R17 R16 Q17 P17
Playouts: 13037, Win: 49.30%, PV: D4 Q16 Q4 D16 R17 R16 Q17 P17
Playouts: 16283, Win: 49.22%, PV: D4 C3
Playouts: 19520, Win: 49.15%, PV: D4 C3
...
Playouts: 62027, Win: 48.77%, PV: D4 C3
```

This is what I saw when doing high playout exploration of D4
initially the NN scores of Q16 the highest but after 1M playouts we see it has low combined score than D16.
After tree is full UCTSearch alternates between adding playouts to D16 and Q16 this slowly biases the score of Q16 up till it's greater than D16.

More theoretically I believe PV will converge to the nodes with the highest initial score (which might be the same as using `-p 1`)

I suggest immediately exiting UCTSearch after the max number of positions that fit in memory have been expanded. This means limiting leela-zero to ~100K playouts in the short term (and ~500K after small_nodes goes in)

RavnaBergsndot:
I believe there was a similar phenomenon discussed in the computer-go mailing list. In that case, a traditional rollout based MCTS engine becomes weaker when fixing the tree size and increasing the playouts indefinitely. I don't remember if the behavior was converging back to playout=1 or not though.

gcp:
>UCTSearch repeatedly adds playout "visits" to nodes biasing their score towards the initial neural net eval (and away from any playout data)

I don't understand. Why does this happen?

sethtroisi:
I think the biggest culprit is that we can re visit a leaf node, add evaluations*, and increase it's playout count multiple times which UTCNode interprets as having more confidence in the eval.

Eddh:
Are you running this on multiple threads ? What depth does the tree reach at max_tree_size ?

[...]

PR #963, opened 2018-03-02
Remove eval of states without expansion

sethtroisi:
See discussion and data in https://github.com/gcp/leela-zero/issues/956


roy7:
With this gone we'll no longer generate an eval on a new node with no children?

roy7:
Thanks. I overlooked create_children was generating the initial eval for the node. In a branch I'm working on I needed the initial eval when the node was made so I was calling eval_state to generate it as needed, but I now realize I can just get_net_eval instead.

PR #967, opened 2018-03-03
Reduced processing of unvisited nodes in UCTNode::uct_select_child()

Hersmunch:
For unvisited nodes the score is proportional to the calculated value. Therefore for the unvisited nodes the best score can be found whilst counting the parent visits and used once to calculate the best value for those nodes. This then allows the second loop to only process nodes that have visits > 0.

$ time ./leelaz -w weights -t 1 --noponder -p 100000 --benchmark
Before:
100001 visits, 34766701 nodes, 100000 playouts, 950 n/s
real    1m48.873s
After:
100001 visits, 34766701 nodes, 100000 playouts, 960 n/s
real    1m47.830s

Edit: with latest next and 75603ad2
Before:
100001 visits, 34320429 nodes, 100000 playouts, 458 n/s
real    3m41.455s
After:
100001 visits, 34320429 nodes, 100000 playouts, 460 n/s
real    3m40.657s

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -211,7 +211,7 @@ void UCTNode::accumulate_eval(float eval) {
 
 UCTNode* UCTNode::uct_select_child(int color) {
     UCTNode* best = nullptr;
-    auto best_value = -1000.0;
+    auto best_unvisited_psa = -1000.0;
 
     LOCK(get_mutex(), lock);
 
@@ -220,27 +220,38 @@ UCTNode* UCTNode::uct_select_child(int color) {
     auto parentvisits = size_t{0};
     for (const auto& child : m_children) {
         if (child->valid()) {
-            parentvisits += child->get_visits();
-            if (child->get_visits() > 0) {
+            if (!child->first_visit()) {
+                parentvisits += child->get_visits();
                 total_visited_policy += child->get_score();
+            } else if (child->get_score() > best_unvisited_psa) {
+                best_unvisited_psa = child->get_score();
+                best = child.get();
             }
         }
     }
 
+    if (parentvisits == 0) {
+        assert(best != nullptr);
+        return best;
+    }
+
     auto numerator = std::sqrt((double)parentvisits);
-    auto fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy);
-    // Estimated eval for unknown nodes = original parent NN eval - reduction
-    auto fpu_eval = get_net_eval(color) - fpu_reduction;
+    auto best_value = -1000.0;
+
+    if (best_unvisited_psa > -1000.0) {
+        // Estimated eval for unknown nodes = original parent NN eval - reduction
+        auto fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy);
+        auto fpu_eval = get_net_eval(color) - fpu_reduction;
+        auto best_unvisited_puct = cfg_puct * best_unvisited_psa * numerator;
+        best_value = fpu_eval + best_unvisited_puct;
+    }
 
     for (const auto& child : m_children) {
-        if (!child->active()) {
+        if (!child->valid() || child->first_visit()) {
             continue;
         }
 
-        float winrate = fpu_eval;
-        if (child->get_visits() > 0) {
-            winrate = child->get_eval(color);
-        }
+        auto winrate = child->get_eval(color);
         auto psa = child->get_score();
         auto denom = 1.0 + child->get_visits();
         auto puct = cfg_puct * psa * (numerator / denom);
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -211,7 +211,7 @@ void UCTNode::accumulate_eval(float eval) {
 
 UCTNode* UCTNode::uct_select_child(int color) {
     UCTNode* best = nullptr;
-    auto best_unvisited_psa = -1000.0;
+    auto best_unvisited_psa = -DBL_MAX;
 
     LOCK(get_mutex(), lock);
 
@@ -236,9 +236,10 @@ UCTNode* UCTNode::uct_select_child(int color) {
     }
 
     auto numerator = std::sqrt((double)parentvisits);
-    auto best_value = -1000.0;
+    auto best_value = -DBL_MAX;
+    auto best_unvisited_value = -DBL_MAX;
 
-    if (best_unvisited_psa > -1000.0) {
+    if (best_unvisited_psa > -DBL_MAX) {
         // Estimated eval for unknown nodes = original parent NN eval - reduction
         auto fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy);
         auto fpu_eval = get_net_eval(color) - fpu_reduction;
@@ -256,7 +257,7 @@ UCTNode* UCTNode::uct_select_child(int color) {
         auto denom = 1.0 + child->get_visits();
         auto puct = cfg_puct * psa * (numerator / denom);
         auto value = winrate + puct;
-        assert(value > -1000.0);
+        assert(value > -DBL_MAX);
 
         if (value > best_value) {
             best_value = value;
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -26,6 +26,7 @@
 #include <functional>
 #include <iterator>
 #include <limits>
+#include <cfloat>
 #include <numeric>
 #include <utility>
 #include <vector>
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -237,7 +237,6 @@ UCTNode* UCTNode::uct_select_child(int color) {
 
     auto numerator = std::sqrt((double)parentvisits);
     auto best_value = -DBL_MAX;
-    auto best_unvisited_value = -DBL_MAX;
 
     if (best_unvisited_psa > -DBL_MAX) {
         // Estimated eval for unknown nodes = original parent NN eval - reduction


src/UCTNode.cpp
```diff
@@ -300,29 +301,43 @@ UCTNode* UCTNode::uct_select_child(int color) {
...
     for (const auto& child : m_children) {
         if (!child->valid()) {
             continue;
         }
-
-        float winrate = fpu_eval;
-        if (child->get_visits() > 0) {
-            winrate = child->get_eval(color);
+        auto visits = child->get_visits();
+        if (visits == 0) {
```
sethtroisi: caching visits isn't a speed up. this would be more cleanly written as

if (child->first_visit()) {
   continue;
}

auto denom = 1.0 + child->get_visits();

Eddh:
Are nodes sorted depending on policy ? If so can't we stop the loop at first unvisited child ? Not on my computer right now

Hersmunch:
@Eddh So I've tried only looking at the first unvisited node and it appears to give the same --benchmark result. Is it guaranteed that unvisited nodes scores will be less than visited nodes? Is that sufficient testing?

The latest commit gives me a fair saving to:
100001 visits, 34766701 nodes, 100000 playouts, 995 n/s
real 1m44.018s

src/UCTNode.cpp
```diff
@@ -300,36 +300,49 @@ UCTNode* UCTNode::uct_select_child(int color) {
     auto parentvisits = size_t{0};
     for (const auto& child : m_children) {
         if (child->valid()) {
-            parentvisits += child->get_visits();
-            if (child->get_visits() > 0) {
+            if (!child->first_visit()) {
+                parentvisits += child->get_visits();
                 total_visited_policy += child->get_score();
+            } else if (child->get_score() > best_unvisited_psa) {
```
Hersmunch: I don't know if this check is really needed anymore

Eddh:
@Hersmunch as long as nodes are sorted in descending order of policy, and that fpu is the same for every unvisited node, it should be the case. Highest policy unvisited node is always going to be the one to be selected first. This might not work if later we use previous knowledge of nodes though.

Hersmunch:
@Eddh Ok, thanks for confirming. I figured that we might as well optimise the approach we have even if it does get replaced later as it will give some immediate benefit. If we're likely to change approach really soon though then perhaps this isn't worth doing.

killerducky:
Did you verify nodes are actually sorted at this point? I thought they only get sorted after the search is done.

Hersmunch:
@killerducky I didn't check directly, instead I played out a seeded game and got the same result. I forgot to turn on noise though, and I do get different results with the standard selfplay noise settings. Sorry about that. I'll remove the assumption that the children are sorted.

Eddh:
Ok nvm then, sorry for the confusion. Since there is nothing to gain from that apparently there is no need to sort them too.

gcp:
Children are added in order sorted by policy value.

src/UCTNode.cpp
```diff
@@ -300,36 +300,47 @@ UCTNode* UCTNode::uct_select_child(int color) {
...
+                best = child.get();
             }
         }
     }
 
-    auto numerator = std::sqrt((double)parentvisits);
-    auto fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy);
-    // Estimated eval for unknown nodes = original parent NN eval - reduction
-    auto fpu_eval = get_net_eval(color) - fpu_reduction;
+    if (parentvisits > 0) {
```
gcp: It's cleaner to inverse the condition and exit immediately. This also removes the need to indent the following big block.

src/UCTNode.cpp
```diff
@@ -291,7 +291,7 @@ void UCTNode::accumulate_eval(float eval) {
 
 UCTNode* UCTNode::uct_select_child(int color) {
     UCTNode* best = nullptr;
-    auto best_value = -1000.0;
+    auto best_unvisited_psa = -1000.0;
```
gcp: Maybe instead of hardcoding the -1000 in several places we should use <limit> to get some minimal double value and use a named constant/constexpr.

Hersmunch:
New to git and github. Failed at rebasing. Anyone got any hints?

src/UCTNode.cpp
Hersmunch: I see why you say that. Just within the scope of this function?

Hersmunch:
Hmm. My validation run (with standard settings and 5ac1d898) just ended:
```
                 wins        black       white
next             59 42.14%   28 41.79%   31 42.47%
opt_uct_select   81 57.86%   39 58.21%   42 57.53%
                             67 47.86%   73 52.14%
```
Any got any ideas? Could anyone try to reproduce this? (with gomill to get cpu stats please)

src/UCTNode.cpp
gcp: Sure.

gcp:
That looks good right? :-)

Hersmunch:
Yes but I don’t understand why it is better. I only tried to optimise existing behaviour, not improve it. So where is the improvement coming from?
The only thing I can think of is that if there is a tie for best child then the unvisited one now always gets picked.

gcp:
Was 81-59 an actual SPRT pass?

Eddh:
> Yes but I don’t understand why it is better. I only tried to optimise existing behaviour, not improve it. So where is the improvement coming from?
> The only thing I can think of is that if there is a tie for best child then the unvisited one now always gets picked.

That's very interesting. Maybe it's just luck on the games, but if it is confirmed, it could indicate that fpu reduction is currently set too high. But would this be enough for > 35 elo improvement ? How often do ties happen ?

Hersmunch:
It looks like I just need a #include <cfloat> but can anyone else confirm this since it compiles anyway for me?

src/UCTNode.cpp
```diff
@@ -220,32 +220,44 @@ UCTNode* UCTNode::uct_select_child(int color) {
...
+        assert(best != nullptr);
+        return best;
+    }
+
     auto numerator = std::sqrt((double)parentvisits);
-    auto fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy);
-    // Estimated eval for unknown nodes = original parent NN eval - reduction
-    auto fpu_eval = get_net_eval(color) - fpu_reduction;
+    auto best_value = -DBL_MAX;
+    auto best_unvisited_value = -DBL_MAX;
```
Hersmunch: I'll get rid of this left over test variable on next commit too

ywrt:
NB: this should be fully backward compatible with current scripts.

Eddh:
>  I played a few games with fairly low playouts and it didn't happen..

Your tests were on 1 thread right ?

Hersmunch:
So including cfloat made the checks pass..

> Your tests were on 1 thread right ?

Yes.
I still cannot see how any change could have altered behaviour. Maybe it was luck but my second validation run is still going and is fairly close is at ("wins" = next):
```
68 wins, 78 losses
146 games played.
Status: 0 LLR -1.76402 Lower Bound -2.94444 Upper Bound 2.94444
```

src/UCTNode.cpp
```diff
@@ -220,32 +221,43 @@ UCTNode* UCTNode::uct_select_child(int color) {
...
+    if (parentvisits == 0) {
+        assert(best != nullptr);
+        return best;
+    }
+
     auto numerator = std::sqrt((double)parentvisits);
-    auto fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy);
-    // Estimated eval for unknown nodes = original parent NN eval - reduction
-    auto fpu_eval = get_net_eval(color) - fpu_reduction;
+    auto best_value = -DBL_MAX;
```
gcp: ```std::numeric_limits<double>::min()```

Eddh:
Could you share the command you used for validation ? It'd be interesting to know how many threads were used etc. We should try to understand why this gives better result given same visits settings.

gcp:
I backed it out for now because I can't measure any speed benefit in CPU or GPU mode, so it's not worth complicating this code. 

I'd be interesting to understand the validation results, though.

src/UCTNode.cpp
Hersmunch: Ok but I think min is the smallest positive number?
http://en.cppreference.com/w/cpp/types/numeric_limits
So lowest should be used instead.

Hersmunch: Ok but I think min is the smallest positive number?
http://en.cppreference.com/w/cpp/types/numeric_limits
So lowest should be used instead.

Hersmunch:
@Eddh it was the default settings. Will confirm them when I get home after work.

src/UCTNode.cpp
```diff
@@ -220,32 +221,43 @@ UCTNode* UCTNode::uct_select_child(int color) {
...
+    if (best_unvisited_psa > -DBL_MAX) {
+        // Estimated eval for unknown nodes = original parent NN eval - reduction
+        auto fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy);
+        auto fpu_eval = get_net_eval(color) - fpu_reduction;
+        auto best_unvisited_puct = cfg_puct * best_unvisited_psa * numerator;
+        best_value = fpu_eval + best_unvisited_puct;
+    }
 
     for (const auto& child : m_children) {
-        if (!child->active()) {
```
WhiteHalmos: Any reason for switching from active() to valid()? This would remove the benefit of root child pruning.

src/UCTNode.cpp
gcp: You're right (but the difference probably doesn't matter for this code?).

src/UCTNode.cpp
Hersmunch: Whoops! That was a merge mistake..
So that means I have to add checking whether unvisited nodes are active in the first loop too. Ok.

Hersmunch:
@gcp I'm totally with you on keeping the code complication as small as possible and that the benefit is at least < 0.5% from my measurements, so definitely questionable. I'll avoid making optimisation PRs for anything less than say 3-5% in future.
I will run another validation run after fixing up the review comments. It's possible the active/valid accident was the cause.

@Eddh The first run I did was just defaults but I can't remember which net so:
`validation.exe -b leelaz_next -n weights -o "-g -p 1600 --noponder -t 1 -q -d -r 0 -w" -b leelaz_opt_uct_select -n weights -o "-g -p 1600 --noponder -t 1 -q -d -r 0 -w"`
and the second one was:
`validation.exe -b leelaz_next -n 75603ad2 -o "-g -v 3200 --noponder -t 1 -q -d -r 10 -w" -b leelaz_opt_uct_select -n 75603ad2 -o "-g -v 3200 --noponder -t 1 -q -d -r 10 -w"`

Hersmunch:
So the second of those two runs has finished and the first is still ongoing:
next vs leelaz_opt_fixed:
```
117 wins, 111 losses
228 games played.
Status: 0 LLR -0.55802 Lower Bound -2.94444 Upper Bound 2.94444
```
leelaz_opt_fixed vs leelaz_opt_broke:
```
1ccb7342 v 1ccb7342 ( 226 games)
              wins        black       white
1ccb7342  104 46.02%   45 45.45%   59 46.46%
1ccb7342  122 53.98%   54 54.55%   68 53.54%
                       99 43.81%  127 56.19%
```
This is suggesting that not doing root child pruning is stronger. Perhaps another validation run should be done for next vs next with child root pruning cleanly disabled? It would be great if someone else could try and reproduce my results but I am happy to do another validation run if needs be.

Issue #946, opened 2018-02-27
hilarious game of LeelaZero against Leela 0.11 


pw31:
If you're interested: https://online-go.com/game/11845631:
LeelaZero 92dd0397 (white, alias RoyalZeroSlow on OGS) against my Leela 0.11 (black, on 8 CPUs at 10sec/move ~ 30000-100000 visits).  Leela 0.11 performs a clean kill of a huge LeelaZero group as  early as move 47.o2!  Soon thereafter, game ends with a ladder. Feels like crying.

wctgit:
:-D 
@roy7
If RoyalZeroSlow's using 80,000 visits, why does it rarely get close to 80,000? Time management or something?

pw31:
I did set the time to 30sec/move, to RoyalZeroSlow accelerate a bit. She is using that time fully each move.

OmnipotentEntity:
@wctgit I've noticed something recently where evaluation will stop early.  What's going on is in UTCSearch.cpp there's a function called `stop_thinking` and one of the things it does is estimate the number of playouts it has left to think.  If the best move has so many playouts that, if another move gets 100% of the playouts from this point forward, it will still have the most, then the evaluation aborts early.

It's designed to speed up moving when there is very little chance of the selected move changing.

roy7:
Yeah that's the "timemanage" feature KillerDucky added. :)

For those interested, here's the full log from RoyalZeroSlow's side of that game. http://termbin.com/vvod

wctgit:
> if another move gets 100% of the playouts from this point forward, it will still have the most, then the evaluation aborts early.
>
> It's designed to speed up moving when there is very little chance of the selected move changing.

Hmm, while it seems like a good idea on the surface, I think it may backfire a bit because spending more time even on the **best** move will still sometimes find out better responses to the opponent's responses, hence different PVs, for example.

Instead of stopping thinking altogether, it would be better just to devote all the remaining visits/playouts to the top move (since no other move will be able to displace it). **Those** playouts would not be wasted.

PR #968, opened 2018-03-03
Stop visiting doomed root children

TFiFiE:
This generalizes the time management code such that it stops visiting any root child the moment it can no longer overtake the current best. Might provide a slight speed and/or performance boost.

### DIFF
--- a/src/UCTNode.h
+++ b/src/UCTNode.h
@@ -47,7 +47,9 @@ public:
                          GameState& state, float& eval);
     void kill_superkos(const KoState& state);
     void invalidate();
+    void set_active(const bool active);
     bool valid() const;
+    bool active() const;
     int get_move() const;
     int get_visits() const;
     float get_score() const;
@@ -73,6 +75,11 @@ public:
     SMP::Mutex& get_mutex();
 
 private:
+    enum Status {
+        INVALID, // superko
+        PRUNED,
+        ACTIVE
+    };
     void link_nodelist(std::atomic<int>& nodecount,
                        std::vector<Network::scored_node>& nodelist);
     // Note : This class is very size-sensitive as we are going to create
@@ -88,8 +95,7 @@ private:
     float m_score;
     float m_net_eval{0};  // Original net eval for this node (not children).
     std::atomic<double> m_blackevals{0};
-    // node alive (not superko)
-    std::atomic<bool> m_valid{true};
+    std::atomic<Status> m_status{ACTIVE};
     // Is someone adding scores to this node?
     // We don't need to unset this.
     bool m_is_expanding{false};
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -20,6 +20,7 @@
 #include "UCTSearch.h"
 
 #include <cassert>
+#include <cmath>
 #include <cstddef>
 #include <limits>
 #include <memory>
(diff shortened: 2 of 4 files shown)


alreadydone:
Looks better and more than what @wctgit and I suggested. Nice work!

src/UCTSearch.cpp
```diff
@@ -401,48 +402,50 @@ bool UCTSearch::is_running() const {
...
+        const auto time_left = time_for_move - elapsed_centis;
+        return std::min(static_cast<float>(playouts_left), std::ceil(playout_rate * time_left));
+    }
+    else {
+        return playouts_left;
     }
+}
 
+int UCTSearch::prune_noncontenders(int elapsed_centis, int time_for_move) {
     auto Nfirst = 0;
```
sethtroisi: Is this the same as

`const auto Nfirst = get_best_root_child().get_visits()`

src/UCTSearch.cpp
```diff
@@ -401,48 +402,50 @@ bool UCTSearch::is_running() const {
...
-        // Wait for at least 1 second and 100 playouts
-        // so we get a reliable playout_rate.
-        auto playout_rate = 1.0f * playouts / elapsed_centis;
-        auto time_left = time_for_move - elapsed_centis;
-        auto est_playouts_left = playout_rate * time_left;
-        if (est_playouts_left < Nfirst - Nsecond) {
-            stop = true;
-            myprintf("%.1fs left\n", time_left/100.0f);
+    const auto min_required_visits = Nfirst - est_playouts_left(elapsed_centis, time_for_move);
+    auto pruned_nodes = 0;
```
sethtroisi: size_t{0} and you can avoid two static_casts

src/UCTSearch.cpp
```diff
@@ -401,48 +402,50 @@ bool UCTSearch::is_running() const {
...
+        const auto playout_rate = 1.0f * playouts / elapsed_centis;
+        const auto time_left = time_for_move - elapsed_centis;
+        return std::min(static_cast<float>(playouts_left), std::ceil(playout_rate * time_left));
+    }
+    else {
+        return playouts_left;
     }
+}
 
+int UCTSearch::prune_noncontenders(int elapsed_centis, int time_for_move) {
```
sethtroisi: Might be easier to have this return if all nodes have been trimmed

src/UCTSearch.cpp
TFiFiE: Maybe it is, but what is there says straight up we just want the most number of visits.

src/UCTSearch.cpp
```diff
@@ -401,48 +402,50 @@ bool UCTSearch::is_running() const {
     return m_run && m_nodes < MAX_TREE_SIZE;
 }
 
-bool UCTSearch::stop_thinking(int elapsed_centis, int time_for_move) const {
-    auto visits = m_root->get_visits();
+int UCTSearch::est_playouts_left(int elapsed_centis, int time_for_move) const
+{
     auto playouts = static_cast<int>(m_playouts);
```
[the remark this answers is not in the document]
TFiFiE: Fair enough.

src/UCTSearch.cpp
```diff
@@ -401,48 +402,50 @@ bool UCTSearch::is_running() const {
...
-    if (stop || cfg_timemanage == TimeManagement::OFF) {
-        return stop;
+    const auto playouts_left = std::min(m_maxplayouts - playouts,
+                                        m_maxvisits - m_root->get_visits());
+    if (elapsed_centis > 100 && playouts > 100) {
+        // Wait for at least 1 second and 100 playouts
+        // so we get a reliable playout_rate.
+        const auto playout_rate = 1.0f * playouts / elapsed_centis;
+        const auto time_left = time_for_move - elapsed_centis;
+        return std::min(static_cast<float>(playouts_left), std::ceil(playout_rate * time_left));
```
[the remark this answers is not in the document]
TFiFiE: OK.

src/UCTSearch.cpp
```diff
@@ -401,48 +402,50 @@ bool UCTSearch::is_running() const {
...
+    }
+    else {
+        return playouts_left;
```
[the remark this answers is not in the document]
TFiFiE: One could argue that sibling outcomes should normally have the same indentation, but in this case it's probably justified because it's a test whether to cut short a computation.

src/UCTSearch.cpp
TFiFiE: 1. From its name, the most straightforward expectation of what this function might return is the correct one. Anything different would be unintuitive and would at least have to be documented.
1. With this return value the function supplies more info, which might prove useful down the line.

sethtroisi:
quick and dirty testing with 800 playouts is inconclusive, waiting for 100 games then I'll starting a 2nd run with 1600 playouts

```
leelaz-next v leelaz-pull968 (63/200 games)
board size: 19   komi: 7.5
leelaz-next        30 47.62%       13 40.62%     17 54.84%     36.62
leelaz-pull968     33 52.38%       14 45.16%     19 59.38%     36.54
                                   27 42.86%     36 57.14%
```

wctgit:
Nice. Does it work solely based on time (via playout_rate) or can it also work with -p 8000 or -v 80000? In other words, if you specify -v 80000, with no time controls, will it always (by default) use all 80000 visits, or do time controls have to be in place first? I read through the changes, but I'm not familiar enough with the codebase to draw this particular conclusion. It looks to me like it only works when time controls are set, and only takes those time controls into consideration, without considering -v 80000 or whatever. [The key game which caused me to suggest this was https://online-go.com/game/11845631 (from https://github.com/gcp/leela-zero/issues/946#issue-300802223), which seemed like LZ had lots of extra time available, but simply wouldn't use up all 80,000 visits which were specified with the -v option. Just want to know if this change would 'fix' that kind of game or if it only comes into play when time is running low.]

src/UCTNode.h
```diff
@@ -88,8 +95,7 @@ class UCTNode {
     float m_score;
     float m_net_eval{0};  // Original net eval for this node (not children).
     std::atomic<double> m_blackevals{0};
-    // node alive (not superko)
-    std::atomic<bool> m_valid{true};
+    std::atomic<Status> m_status{ACTIVE};
```
sethtroisi: pop quiz: What is `sizeof(Status)`?

wctgit: Good question. For me, I've been out of C++ programming for too long, but I imagine it's some form of standard integer by default. Therefore, I'll guess 4 bytes, maybe 8? Could easily be reduced to a single byte, I imagine. Or, if bit fields are possible (I can't remember) it would only require 2 bits. If booleans are only 1 bit, you could code it as two booleans. Do data structures have to conform to integer number of bytes? I forget.

wctgit:
@sethtroisi 
> Quick and dirty testing with 800 playouts is inconclusive, waiting for 100 games then I'll starting a 2nd run with 1600 playouts

I imagine that any effects would be more obvious and pronounced with more playouts/visits. So if it were up to me, I'd test it with something more like 5000 or 10,000 playouts. Again, the game in question which started this for me was set to 80,000 visits. Not sure how to compare that to playouts, but a guess based on using 2600 visits instead of 1600 playouts makes me think something very roughly on the order of 2:1, so 10,000 playouts might be somewhere close to 20,000 visits.

This is just a hunch though. Maybe with tree reuse it doesn't matter? But I just have a strong feeling more playouts will lead to a significantly more drastic effect.

Alderi-Tokori:
I think this should be a toggleable option through a command line argument (same as time manage), so it can be turned of in self-play.

src/UCTNode.h
TFiFiE: That does replace a 1-byte variable with a 4-byte one, but sizeof(UCTNode) stays unchanged.

TFiFiE:
It already only kicks in when time management is on, but it could become a separate option.

src/UCTNode.h
sethtroisi: In UCTSearch.h for passflag_t GCP created his own constants, but I really don't like that code (and might do clean it up right now)

You can define the enum as char
http://en.cppreference.com/w/cpp/language/enum

TFiFiE: Given that it doesn't affect the total size of UCTNode, not having the underlying type be of the native CPU word size would more likely be detrimental than beneficial to performance.

wctgit: @TFiFiE: Since I've been out of the C++ loop (no pun intended) for so long, could you briefly (very briefly, not to bother you) explain why the total sizeof(UCTNode) doesn't change? Is it because data structures are already allocated in multiples of 4 bytes (words), and UCTNode was already 4k + 1 bytes in size, so increasing to 4k + 4 doesn't actually require an extra 4 byte word, since 4k + 1 already must be rounded up to 4k + 4 _[which = 4(k + 1) ]_? Just want to check my understanding; nothing super-important. Thanks. :-)

Hersmunch: see https://www.cprogramming.com/tutorial/size_of_class_object.html

Issue #971, opened 2018-03-04
Resign Analysis Data


WhiteHalmos:
@gcp @roy7 Please consider increasing resign percent to at least 10.

Incorrectly resigned games for 15 is 4.93%, 10 is 2.11%, 1 is 0.7%. According to AGZ paper we should choose 15 or at least 10 since the rate is lower than 5%.

Compared to current 1 percent resign, 15 is a 21.5% speedup and 10 is a 16% speedup. I will continue to run this analysis on my machine to ensure the rate does not creep up with new nets.

Data:
```
gunzip build/debug/*.gz; scripts/resign_analysis/resign_analysis.py build/debug/*.debug.txt.0 --r 0.2,0.15,0.1,0.01                                                                    
Total games to analyze: 142                                                                                      
Resign rate: 0.20 - Black won 62/142 (43.66%)
Black wins - Incorrect: 6/62 (9.68%) Correct: 56/62 (90.32%) No Resign: 0/62 (0.00%)
Black wins - Average game length: 177/386 (54.10% reduction)                                                     
White wins - Incorrect: 7/80 (8.75%) Correct: 73/80 (91.25%) No Resign: 0/80 (0.00%)                             
White wins - Average game length: 196/390 (49.73% reduction)                                                     
Overall    - Incorrect: 13/142 (9.15%) Correct: 129/142 (90.85%) No Resign: 0/142 (0.00%)                        
Overall    - Average game length: 187/388 (51.63% reduction)

Resign rate: 0.15 - Black won 62/142 (43.66%)
Black wins - Incorrect: 1/62 (1.61%) Correct: 61/62 (98.39%) No Resign: 0/62 (0.00%)
Black wins - Average game length: 211/386 (45.31% reduction)
White wins - Incorrect: 6/80 (7.50%) Correct: 74/80 (92.50%) No Resign: 0/80 (0.00%)
White wins - Average game length: 211/390 (45.93% reduction)
Overall    - Incorrect: 7/142 (4.93%) Correct: 135/142 (95.07%) No Resign: 0/142 (0.00%)
Overall    - Average game length: 211/388 (45.66% reduction)

Resign rate: 0.10 - Black won 62/142 (43.66%)
Black wins - Incorrect: 0/62 (0.00%) Correct: 62/62 (100.00%) No Resign: 0/62 (0.00%)
Black wins - Average game length: 225/386 (41.71% reduction)
[...]

PR #972, opened 2018-03-04
More concise resign script output

WhiteHalmos:
See #971 for example.

Friendly ping @killerducky for review.


killerducky:
Looks good.

Issue #505, opened 2017-12-27
cannot run two instances of leelaz on a supercomputer


alreadydone:
Please see the following terminal output. Does anyone has an idea what the cause may be? I am able to run one instance of autogtp on this supercomputer, which generates moves at a rate of 3 seconds/move, but I don't think this completely exploits the full power of the Tesla K20.

`~/leela-zero> autogtp/leelaz -w weights.txt & autogtp1/leelaz -w weights.txt
[1] 27460

Wed Dec 27 06:58:08 2017: [unset]:_pmi_alps_get_apid:alps_app_lli_put_request failed
Wed Dec 27 06:58:08 2017: [unset]:_pmi_alps_get_appLayout:pmi_alps_get_apid returned with error: Bad file descriptor
**(these don't matter when running one instance)**

Using 2 thread(s).
RNG seed: 11884490720694353691
Leela Zero  Copyright (C) 2017  Gian-Carlo Pascutto
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; see the COPYING file for details.

Initializing OpenCL

Wed Dec 27 06:58:08 2017: [unset]:_pmi_alps_get_apid:alps_app_lli_put_request failed
Wed Dec 27 06:58:08 2017: [unset]:_pmi_alps_get_appLayout:pmi_alps_get_apid returned with error: Bad file descriptor

Using 2 thread(s).
RNG seed: 1535599010717395473
Leela Zero  Copyright (C) 2017  Gian-Carlo Pascutto
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; see the COPYING file for details.

Initializing OpenCL

Detected 1 OpenCL platforms
Platform version: OpenCL 1.2 CUDA 7.5.23
Platform profile: FULL_PROFILE
Platform name:    NVIDIA CUDA
Platform vendor:  NVIDIA Corporation
Device ID:     0
Device name:   Tesla K20
Device type:   GPU
Device vendor: NVIDIA Corporation
Device driver: 352.68
Device speed:  705 MHz
Device cores:  13 CU
Device score:  1112
Selected platform: NVIDIA CUDA
Selected device: Tesla K20
with OpenCL 1.2 capability

Detected 1 OpenCL platforms
Platform version: OpenCL 1.2 CUDA 7.5.23
Platform profile: FULL_PROFILE
[...]

PR #973, opened 2018-03-04
Autogtp: Add no resign and self play only options

WhiteHalmos:
- No resign makes it easier to keep resign analysis running without
  maintaining local changes.
- Self play only allows clients who only want to run self play games to
  do so.

### DIFF
--- a/autogtp/Management.cpp
+++ b/autogtp/Management.cpp
@@ -42,7 +42,9 @@ Management::Management(const int gpus,
                        const int ver,
                        const int maxGames,
                        const QString& keep,
-                       const QString& debug)
+                       const QString& debug,
+                       const bool noResign,
+                       const bool onlySelfPlay)
 
     : m_syncMutex(),
     m_gamesThreads(gpus * games),
@@ -54,6 +56,8 @@ Management::Management(const int gpus,
     m_gamesPlayed(0),
     m_keepPath(keep),
     m_debugPath(debug),
+    m_noResign(noResign),
+    m_onlySelfPlay(onlySelfPlay),
     m_version(ver),
     m_fallBack(Order::Error),
     m_gamesLeft(maxGames),
@@ -125,7 +129,7 @@ void Management::giveAssignments() {
                 m_gamesThreads[thread_index]->order(getWork(finfo));
             } else {
                 m_gamesThreads[thread_index]->order(getWork());
-            }            
+            }
             m_gamesThreads[thread_index]->start();
         }
     }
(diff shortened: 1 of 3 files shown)


WhiteHalmos:
@marcocalignano I've added an option to request "only self play" matches. Please take a look at the code when you get a chance. Thanks!

gcp:
I won't take the no resign flag into the official distribution. I don't want anyone to get the idea that this is flag they *can* or *should* fiddle with.

Only self-play is a bit more tricky, but AFAIK the only real use case for it were some very strange hardware/driver setups.

Not too keen on taking this.

marcocalignano:
I am not for only-self play either. I think if you want to support the project doing matches is also part of it. Why someone would want to disable it?.

gcp:
There's apparently some shared supercomputer toggles in nvidia's driver that limits the access to the GPU to a single process. This works for self-play, but breaks in amtches.

WhiteHalmos:
If you have a slow or flaky network connection, even with no driver issues or a decently fast computer, you also don't want to keep downloading big networks, especially now that we are on 10 blocks and possibly more in the future.

Agreed about resign, will back that out.

gcp:
Matches cause much more networks downloads.

I hear you about the second part though. It only takes one "smart" person to advertise the option as "bandwidth optimization" and before you know it we're stuck because there are no match clients. I have some experience in my day job with this kind of "optimization guides"...

marcocalignano:
If this is a problem of supercomputer only, and give that a supercomputer has multiple GPU I could teach autogtp to start the tho process on different GPUs what a match is requested. Would be this a good solution?

alreadydone:
People using supercomputers likely compile by themselves, so they can modify the source themselves. (Are there any other people? I hope my proposal will attract some.) My solution is to add a global variable selfPlayOnly = true as in https://github.com/alreadydone/lz/blob/br2gpu/autogtp/Management.cpp
P.S. Each job on Big Red II requests one node with one GPU. I had no luck making one job utilize 2 GPUs.

WhiteHalmos:
@marcocalignano The number of match networks is an order of magnitude higher than the number of best networks.

@gcp If someone wanted to post an optimization guide they can already do so. The code is open source, nothing prevents them from compiling their own binaries or sharing patch diffs. I don't think this is a valid argument against making it easier for legitimate contributors whose autogtp fails often due to network (we have actual reported network issues, not theoretical adversaries).

marcocalignano:
So the only arguments for not doing matches is that the network is to slow for the download and a supercomputer with nvidia drivers that does not allow two process on the same GPU. 
For the first problem we can have a detection of network speed upon network downloading. 
For the second it should be possible to detect if two processes can run on the GPU. 
In both case I could resurrect my old branch where autogtp send time stats to the server per game, and add like a flag that tell the server "This client has slow network connection please send no matches" or "This client cannot share it GPU pleas no matches". 
@roy7 Do you have time to have a look at that old branch? (Last time I proposed it you were busy cleaning the server code before uploading it for the first time).

alreadydone:
@marcocalignano FYI this was the output from running two instances of leelaz on Big Red II: https://github.com/gcp/leela-zero/issues/505

WhiteHalmos:
And also my selfish desire to run resign analysis with no local changes. Sounds like you prefer automatic over command-line flag, which sounds good too.

PR #986, opened 2018-03-06
Autogtp: Add self play only option

WhiteHalmos:
Self play only allows clients who only want to run self play games to do
so. Especially important for slow/flaky networks and video card issues.


marcocalignano:
@gcp See my comment on #973

WhiteHalmos:
These are fair points. We can continue the discussion in #973 if you'd like. I will keep these changes to my local branch for resign analysis.

Issue #988, opened 2018-03-06
Windows support for training is broken


gcp:
#958 and #917 broke Windows support. The underlying problem is that python multiprocessing on Windows requires using pickeable objects, which those changes broke.

If there is no easy fix for this, they will be backed out.

ywrt:
I'll try and send through a fix in a few hours

ywrt:
NB: The reason for this is that windows apparently doesn't support fork, only spawn. And the spawn startup process is entirely different from the fork process, which results in it needing to pickle the Struct objects.

The Struct issue can probably be fixed by adding 
```
import copyreg
def pickle_struct(s):
    return Struct, (s.format,)
copyreg.pickle(Struct, pickle_struct)
```
somewhere near the top of chunkparser.py, but I suspect that will just uncover the issue that generators can't be pickled either.

PR #991, opened 2018-03-07
V2pipeline

ywrt:
Fix training to work on Windows again. (issue #988)

The python multiprocessing class doesn't support 'fork' on windows, only 'spawn',
and spawn requires that all the arguments be pickleable. This means that generators
and Structs can't be passed to workers.


gcp:
Thanks for the quick fix!

PR #993, opened 2018-03-08
--evals for multiple evals per node

roy7:
This will let you specify the # of network evals you'd like to do per node. ~I think it should be -almost- there but I just can't debug an issue. With --evals of 2 or higher, at the result.second += tmpresult.second I'll crash with "corrupted size vs prev_size". Or just for testing if I do result.second = tmpresult.second then I crash with "munmap_chunk".~

~Since the tempresult is made fresh in the loop and all I'm doing is adding to a float that is part of the std::pair, I can't understand where I'm misusing memory allocation. Might someone be able to spot the problem to finish this code up?~


roy7:
I used basic random header for std::shuffle but if someone has alternate suggestion that's fine.

src/Network.cpp
```diff
@@ -863,6 +864,40 @@ Network::Netresult Network::get_scored_moves(
...
+            }
+
+            result = get_scored_moves_internal(state, planes, r_list[0]);
+
+            for (int r = 1; r < rotation; ++r) {
+                Netresult tmpresult = get_scored_moves_internal(state, planes, r_list[r]);
+                result.second += tmpresult.second;
+
+                // Post-increment c for the extra PASS move
+                for (int c = 0; c < BOARD_SIZE * BOARD_SIZE; c++) {
```
WhiteHalmos: Found the bug. The output is not necessarily a vector of size BOARD_SQUARES. It skips intersections that already has stones.

roy7: Ahhh. So I need to use something more like a .length() call?

roy7:
Not sure why that change was needed, but I added it.

WhiteHalmos:
Oh, I guess it's not necessary since the state doesn't change. You're right, it's not needed. Somehow I thought that the illegal moves changes between calls, sorry. It's late here. 😴

WhiteHalmos:
Hmm... I'm getting a flat policy from this PR. Trying to debug...

roy7:
Should I add a way to generate a heatmap using the eval count? I'm not sure how you're seeing the policy.

WhiteHalmos:
I'm just running this:
```
make -j4 leelaz && echo -e "genmove b" | ./leelaz -w w -g -v100 --noponder -t1 -e2
NN eval=0.478179

  C4 ->       9 (V: 49.84%) (N:  7.79%) PV: C4 Q16 C16 Q4
  Q3 ->       9 (V: 49.50%) (N:  7.64%) PV: Q3 D4 D16 Q16
 R16 ->       9 (V: 49.46%) (N:  7.90%) PV: R16 D4 R4 D16 C3
 D17 ->       9 (V: 49.28%) (N:  7.67%) PV: D17 Q4 Q17 D4 C3
  R4 ->       8 (V: 49.43%) (N:  7.90%) PV: R4 D16 R16 D4
  Q4 ->       8 (V: 49.39%) (N:  8.10%) PV: Q4 D16 R16 D4
  D4 ->       8 (V: 49.37%) (N:  7.96%) PV: D4 Q16 C16 Q4
 Q16 ->       8 (V: 49.28%) (N:  8.10%) PV: Q16 D4 R4 D16
 C16 ->       8 (V: 49.14%) (N:  7.79%) PV: C16 Q4 C4 Q16
 Q17 ->       8 (V: 49.13%) (N:  7.64%) PV: Q17 D4 D17 Q4
 D16 ->       8 (V: 48.78%) (N:  7.96%) PV: D16 Q4 C4
  D3 ->       7 (V: 48.63%) (N:  7.67%) PV: D3 Q16 Q3 D16 R17
3.6 average depth, 6 max depth
63 non leaf nodes, 1.57 average children
100 visits, 35941 nodes, 99 playouts, 247 n/s

= C4
```

roy7:
Shouldn't the star points on an empty board all have identical N% though? And all 3-4 points be identical? Using -e 8 those moves are all identical in every way.

roy7:
Ok I tried bringing in your changes direct to my branch... let's see if it worked or if it's all a terrible mess. :)

roy7:
Ah yes, symmetric moves in opening have identical N % and as long as the symmetry holds, so does the identical values. Perfect. :)

gcp:
Either this makes the program stronger for some value, and I'll pull it, or it does not, and it should not be pulled.

Eddh:
It can be good for networks that are confused by the rotation of the board but bad for nets which are more rotation invariant so there is a point in this being an option.

WhiteHalmos:
The biggest problem is that it takes so long to go through all the rotations/symmetries. Here's **-v400** normal and -e8.
```
leelaz-next_7560_ v leelaz-multi-eval_7560_-e8 (75/100 games)                                                    
board size: 19   komi: 7.5  
leelaz-next_7560_              28 37.33%       12 31.58%     16 43.24%     41.12                                 
leelaz-multi-eval_7560_-e8     47 62.67%       21 56.76%     26 68.42%    302.72                                 
                                               33 44.00%     42 56.00%
```
Config:
```
competition_type = 'playoff'
LZ_ARGS = '/projects/lztest/{} -v400 --noponder -r10 -g -t1 -d -w {} {}'
def LZ(vals):
  return Player(LZ_ARGS.format(*vals))
players = dict()
p1 = 'leelaz-next_7560_'
p2 = 'leelaz-multi-eval_7560_-e8'
players[p1] = LZ(p1.split('_'))
players[p2] = LZ(p2.split('_'))
board_size = 19
komi = 7.5
matchups = [
  Matchup(p1, p2, alternating=True, number_of_games=100),
]
```
It's certainly stronger, no doubt about it, but so much slower.

Eddh:
I'd suggest only doing multiple symetries for highly visited node. Maybe even only first 2 depth

gcp:
>It's certainly stronger, no doubt about it, but so much slower.

Those are one and the same thing, i.e. it can be verified at fixed timecontrols.

>It can be good for networks that are confused by the rotation of the board but bad for nets which are more rotation invariant so there is a point in this being an option.

Then verify what is the case for our training procedure.

>I'd suggest only doing multiple symetries for highly visited node. Maybe even only first 2 depth

This is a more sensible approach, though I'd use visits, not depth.

WhiteHalmos:
@gcp That's a good idea. Any suggestions on how to run gomill with fixed time control? Or maybe I should just hack it into the binaries under test?

Eddh:
> Then verify what is the case for our training procedure.

I've seen a case where depending on the rotation, the net fc2b evaluated winrate at 1% or 80%, and the policy was similarly confused. And this case, though extreme, is not unique. I have seen at least 2 other cases like this. And roy7 tested on other nets and they are also confused by rotation on these problems. Basically it seems like this is the case for most stupid moves like self ataris. With this PR merged, if someone point out a game in which LZ does a stupid self-atari, you could just tell them to turn on this option. And I bet it will work for most cases.

Btw I think the name for this option should be "bagging" since that is what it was called in this paper http://www.lamsade.dauphine.fr/~cazenave/papers/resnet.pdf which used this method to increase prediction accuracy.

roy7:
I think of it as an option in the sense of --playouts or --visits. Maybe someone is willing to trade search speed for more accuracy. Since I intended to use this myself for RoyalZeroSlow, I figured I'd try to do it in a way to share it with others. Although if we end up not wanting it a command line option hopefully the rest can remain so it's easy for people to do experiments without reduplicating the effort (and my personal fork only needs to have the command line for my personal use). And I likely wouldn't have gotten it fixed without @WhiteHalmos's help. :)

roy7:
A feature to go along with this I didn't add yet (or look to see how easy it would be to add), if we're doing multiple rotated evals on a position, should we collect those hashes and store the averaged results in the NNCache for each of those rotations? We're doing the NN eval work anyway which is the slow part. So --evals 8 would add 8 hashes to NNCache.

jkiliani:
I think @Eddh is most likely right in his suggestion to only do this for nodes close to the root node. If done for every node, it would have to beat a tree with 8 times as many nodes, which I highly doubt it can.

Of course, in cases when the game tree limit is reached, calculating additional rotations would help...

roy7:
Another thought to reduce # of wasted extra evals is to measure the variance of evals between children of a node. If the children are all well behaved, the maybe you don't need extra evals. But if some children are saying the game is lost and another child is saying 80% win rate, maybe those children need extra evals to be sure it's real and not just certain rotations.

jkiliani:
I think this would definitely help... maybe do a second eval for nodes with a certain minimum number of visits, and then calculate the variance between the two (evaluation and policy priors). If it's low, don't evaluate the node further, but if the variance is high, a third or forth eval should be requested.

This would be a good way to help the search determine which situations the net is poorly trained on, and behave accordingly.

Eddh:
> I think of it as an option in the sense of --playouts or --visits. Maybe someone is willing to trade search speed for more accuracy.

I didn't even think of that aspect but that is a very strong argument for the existence of this option. Currently we are limited by max_tree_size for analysis, but this should allow for better analysis with the same number of nodes.

WhiteHalmos:
Let's hold off speculating special ways of re-evaluating nodes as that brings a whole new level of complexity, as the value net output is propagated up the tree and complex to reverse.

* Caching the average in all 8 rotations is a great idea. More speed for same results.
* Doing fewer than all 8 rotations is also error prone, especially as we probably only want to consider rotations for fewer nodes, they should be as correct as possible.
* Depth is a good and simple method to differentiate.

I think a better way to go about this is a flag to determine the depth before which we consider all the rotations, defaulting to 0. When I get a chance I will implement this and test it out with fixed time limits against next.

WhiteHalmos:
Hmm... expanding all 8 but only at a certain depth really messes with tree reuse.

roy7:
There's doesn't seem to be an easy way to get rotated hashes so for now I won't be adding that but others are welcome to. ;)

killerducky:
Here is a quick outline of my idea for how to do this: https://github.com/killerducky/leela-zero/tree/multieval2

Original code: A playout/visit means you go down the tree and do a network eval on a leaf. 

New code: A playout/visit can also stop partway down the tree if it reaches a node with many visits. Evaluate a different rotation of that node, and propagate that value up the tree like a normal playout.

Code is not thread safe, and NNCache is bypassed.

Eddh:
@killerducky I don't think this works because a visit that only increase accuracy on a node and doesn't add another one is less valuable than a normal visit. https://github.com/gcp/leela-zero/issues/956
So a visit that stops partway down the tree to add precision shouldn't count as a normal one for move selection and -v limit.

killerducky:
@Eddh the argument about some visits being inaccurate applies to the normal visits as well, for example when it investigates a low prior move the turns out to be a blunder. Any effort to try to weigh which visits should count for how much is separate from this. Also when the user says -v 3200, it really serves as a proxy for how much time to think, so I think keeping it as 3200 calls to the network makes sense.

WhiteHalmos:
As mentioned earlier, I implemented https://github.com/gcp/leela-zero/pull/993#issuecomment-371557945 in my branch here: https://github.com/WhiteHalmos/leela-zero/tree/multi-eval

Given the same amount of time, it didn't work so well, approx. 200 elo difference.
```
leelaz-next_b3a8_ v leelaz-multi-eval_b3a8_-e2 (105/400 games)                                     
board size: 19   komi: 7.5                                                                         
                             wins              black         white       avg cpu                   leelaz-next_b3a8_              79 75.24%       37 69.81%     42 80.77%    338.66                   leelaz-multi-eval_b3a8_-e2     26 24.76%       10 19.23%     16 30.19%    337.33                                                                  47 44.76%     58 55.24%
```
Config:
```
competition_type = 'playoff'
LZ_ARGS = '/projects/lztest/{} --noponder --timemanage off -r10 -g -t1 -d -w {} {}'
def LZ(vals):
  commands = ['time_settings 450 0 0']                                                             
  return Player(LZ_ARGS.format(*vals), startup_gtp_commands=commands)
players = dict()
p1 = 'leelaz-next_b3a8_'
p2 = 'leelaz-multi-eval_b3a8_-e2'
players[p1] = LZ(p1.split('_'))
players[p2] = LZ(p2.split('_'))
board_size = 19
komi = 7.5
matchups = [
]
```
Here's some typical output:
```
Thinking at most 3.7 seconds...                                                           [3/61267]
NN eval=0.500174                                                                                   
Playouts: 1533, Win: 50.24%, PV: G16 G15 G17 H15 K18 L18 E18 D15 K17 L17 D18 C16 C14 C18 F17       
                                                                                                    G16 ->    1543 (V: 51.48%) (N:  9.70%) PV: G16 G15 G17 F17 K17 J16 J17 L17 L18 M18 M17 L16 N18 K1$ M19 J18 P17 P18 O17 N16 O18 Q18 M16 M15 N17                                                        G17 ->     937 (V: 50.90%) (N: 14.74%) PV: G17 G16 K18 L18 L17 J18 M18 J17 O17 P17 N16 O15 O18 M1$ L16 M15 L15 L14 K15 C16
 J17 ->     838 (V: 48.32%) (N: 60.44%) PV: J17 K17 J16 J15 F18 C16 D15 F17 E18 D16 K18 L18 J18    
 J16 ->     112 (V: 49.57%) (N:  4.93%) PV: J16 J15 J17 K17 G16 H15 F17 G15 E16 G17 E17 G18 D16 F4 
 K17 ->      32 (V: 44.90%) (N:  4.49%) PV: K17 J17 J18 J16 L18 L17 K18 G17 M17 N18 L16            
  G5 ->      27 (V: 48.59%) (N:  1.77%) PV: G5 C16 D15 G17 D10                                     
 F17 ->       6 (V: 46.21%) (N:  0.68%) PV: F17 G17 G18 G16 E17                                     N16 ->       1 (V: 38.45%) (N:  0.49%) PV: N16  
11.1 average depth, 26 max depth                 
2298 non leaf nodes, 1.52 average children       
3497 visits, 977477 nodes, 2290 playouts, 617 n/s

Thinking at most 3.6 seconds...                  
NN eval=0.478715        
Playouts: 750, Win: 48.58%, PV: G15 G17 H15 K18 L18 E18 D15 D18 C16 K17 L17 C14 C18 F17             
 G15 ->    1006 (V: 48.98%) (N: 66.22%) PV: G15 G17 H15 K18 L18 E18 J18 D18 C17 E16 E17 F17 F15    
 G17 ->     334 (V: 48.37%) (N: 28.93%) PV: G17 G15 J17 F17 H18 E16 F18 F15 D11 C11 D10 G5 C10 H3 $
12 C13 C16              
  T7 ->       5 (V: 42.09%) (N:  1.76%) PV: T7 R7 G15                                              
 H15 ->       2 (V: 41.16%) (N:  0.98%) PV: H15 F17                                                
10.7 average depth, 18 max depth                 
881 non leaf nodes, 1.53 average children        
1348 visits, 376056 nodes, 1110 playouts, 303 n/s
```
So we see that it is about 2.5 times more visits without expanding the top two layers (root and root's child nodes) all 8 rotations. This is in line with our expectation of 2x the visits => 100-200 elo increase. I think a big problem is tree reuse not being as effective. I turned off time management since it would have the wrong n/s. 😕

roy7:
I've not tried it, but someone could play around with --evals 8 --playouts 1 for a pure policy network challenge.

Eddh:
I modified @killerducky's branch to make it backup values without using false visits. It's probably still not correct with multithread but seems to work well for -t1 https://github.com/Eddh/leela-zero/commits/multieval2

WhiteHalmos:
@Eddh Looks very good. Should eval before opt_update be weighted by the number of rotations we have already done? i.e. after 7 rotations, we should trust the old value more than the next single rotation so we get a true average?

I like the change to 100s.

Eddh:
@WhiteHalmos if I understand what you mean correctly, the true average is already done in do_next_rotation from @killerducky.

WhiteHalmos:
Ah, right. Thanks for fixing it.

WhiteHalmos:
BTW have you tested it with fixed timeout like my config above?

Eddh:
Not yet. I'll have to do a dual boot to be able to run ringmaster more effectively than in a VM. Don't know when I'll have time to do that xD. My computer would still do it very slowly at 800 playouts anyway. So if someone can run the tests I will be grateful.

WhiteHalmos:
Alright, I'm testing it now with the following config:
```
competition_type = 'playoff'
LZ_ARGS = '/projects/lztest/{} --noponder --timemanage off -r10 -g -t1 -d -w {} {}'
def LZ(vals):
  commands = ['time_settings 450 0 0']
  return Player(LZ_ARGS.format(*vals), startup_gtp_commands=commands)
players = dict()
p1 = 'leelaz-next_545c_'
p2 = 'leelaz-eddh_545c_'
players[p1] = LZ(p1.split('_'))
players[p2] = LZ(p2.split('_'))
board_size = 19
komi = 7.5
matchups = [
]
```

Eddh:
I realized that even though the evaluation is backed up in the tree, m_blackevals of the multiply rotated node isn't updated, so the change wouldn't be taken into account by the uct search of its direct parent node. Oh well it isn't that much of a difference, we will see if the results are already good without that or not. Thanks a lot for running the test @WhiteHalmos

WhiteHalmos:
Only 4 games, so amended and restarted. (it was 2:2)

WhiteHalmos:
@Eddh get your PR ready. I think even if we don't end up merging his, @roy7 will want to use it for his RoyalZero accounts since it'll get rid of heatmap weaknesses for his bots.
```
leelaz-next_545c_ v leelaz-eddh_545c_ (80/400 games)                      
board size: 19   komi: 7.5           
leelaz-next_545c_     36 45.00%       20 50.00%     16 40.00%    338.32   
leelaz-eddh_545c_     44 55.00%       24 60.00%     20 50.00%    337.81   
                                      44 55.00%     36 45.00%
```
Config:
```python
competition_type = 'playoff'
LZ_ARGS = '/projects/lztest/{} --noponder --timemanage off -r10 -g -t1 -d -q -w {} {}'
def LZ(vals):
  commands = ['time_settings 450 0 0']
  return Player(LZ_ARGS.format(*vals), startup_gtp_commands=commands)
players = dict()
p1 = 'leelaz-next_545c_'
p2 = 'leelaz-eddh_545c_'
players[p1] = LZ(p1.split('_'))
players[p2] = LZ(p2.split('_'))
board_size = 19
komi = 7.5
matchups = [
]
```

WhiteHalmos:
Although please fix these warnings. 😄 
```
/projects/leela-zero/src/UCTNode.cpp: In member function ‘float UCTNode::do_next_rotation(GameState&)’:
/projects/leela-zero/src/UCTNode.cpp:185:22: warning: unused variable ‘orig_score’ [-Wunused-variable]
                 auto orig_score = child->m_score;
                      ^~~~~~~~~~
/projects/leela-zero/src/UCTNode.cpp:182:14: warning: variable ‘found’ set but not used [-Wunused-but-set-variable]
         auto found = false;
```

roy7:
Should I close this PR and Eddh can make a new PR with his version directly? If we feel like Eddh's version is a good improvement (it does have a good win rate in testing). And I know @gcp wasn't a big fan of letting users control --evals directly. :) I could always keep the command line portion of things in a personal branch if I really wanted. But if Eddh's version can help prevent the late game self-atari blunders automatically and using a lot less excess evals than I was going to, I'm all for it.

WhiteHalmos:
Agreed, I think the automatic one that @killerducky and @eddh worked on has a higher chance of getting merged. Will keep the test running.

WhiteHalmos:
@roy7 Please update us if you do try this out with RoyalZero*.

roy7:
Although does it need fixed/tested for mutli-threading first?

If we're still using my MUTLI option underneath some of these improved versions I could remove the --evals from my PR and just leave the support for multiple evals, which other pulls could be based from.

Eddh:
Yeah I guess I need to take care of thread safety. Should there be a m_being_evaluated variable that I read and switch on inside a mutex block, indicating other thread to not evaluate the same not twice at the same time ?

Splee99:
@Eddh I just downloaded your branch. Is there any flag I should set during compilation? It still does the self-atari in the game between LZ and Leela 0.11 (game 10) as mentioned in:
https://www.reddit.com/r/cbaduk/comments/83guiu/lz_vs_leela_011_rematch_20_game_series/
What I did was play the game manually until just before the move K1 where LZ played self-atari. Then let LZ play white and I played black following the LZ's prediction (I'm not strong enough). Eventually LZ played K1 again. I know this branch is not particularly for the self-atari problem, but if it's not fixed we still need to do more things.

WhiteHalmos:
@Splee99 What is your playouts number?

Splee99:
@WhiteHalmos I set it to 30 seconds per move, about 6000 playouts per move.

WhiteHalmos:
I see... it takes over 7000 visits to see it (tested locally).

Splee99:
I tested 40 seconds per move with about 8500 playouts, it still plays K1.

gcp:
Closing this one. As explained, you need to demonstrate a strength improvement first (and not at fixed playouts obviously!) before this can be taken.

Eddh:
@gcp have you considered this aspect ? 

> Currently we are limited by max_tree_size for analysis, but this should allow for better analysis with the same number of nodes. If you want to use LZ for analysing a position and have all the time in the world, you would want to use this option.

WhiteHalmos:
```
leelaz-next_545c_ v leelaz-eddh_545c_ (209/400 games)   
board size: 19   komi: 7.5  
leelaz-next_545c_    109 52.15%       58  55.24%     51 49.04%    337.38                                        
leelaz-eddh_545c_    100 47.85%       53  50.96%     47 44.76%    336.97                                        
                                      111 53.11%     98 46.89%
```

Eddh:
Ok too bad. So the true solution will only come from the net. I still think @roy7 pr is good for analysis. Thanks you very much for the tests @WhiteHalmos

WhiteHalmos:
It's actually not a bad result. The conclusion I draw is that you and @killerducky's branch, despite about 30% slower in nodes per second than next, actually matches next in performance. This means that it is quite impressive. The full solution is to wait until the net gets rid of these weaknesses by itself, but that may (never) take long time to happen.

I will try to clean up that branch and make it more performant and see if that helps.

roy7:
@gcp If I redo my PR to not include the --evals option and only include MULTI support as a rotation mode, would that be acceptable? It won't change existing functionality at all but has the support in there for future experimentation by others.

@Eddh @WhiteHalmos Or anyone who is experimenting with extra rotations, I had the thought this morning instead of making it an average, use the max() of each rotation for each prior. Because the purpose of the priors isn't "this is the best move" it is "this is a move that a lot of search has gone to before based on the value network at that time". 

In other words, the exact MCTS %s aren't what is important. The new search we're doing right now will guide if we care about this move or not once we explore it. What's important is that we look at the important moves. So if one rotation says "Look at this move!" and the others don't, it should be on the list of moves to look at.

Thus, if we use multiple rotations and use the max() of each prior, any move that any rotation thought was important would get looked at early during our new search. Winrate would still be averaged of course.

PR #1003, opened 2018-03-09
Add support for training on mini-batches larger than GPU RAM size.

ywrt:
Add support for mini-batches larger than GPU RAM size.
    
    1. Minor bugfix to use correctly use WEIGHTS instead of REGULARIZATION_LOSSES.
    2. Refactor out statistics into a separate class.
    3. Refactor out timing into a separate class.
    4. Refactor out batch_norm
    5. Refactor out measuring loss
    6. Drop use of tf.Dataset: instead, feed batches directly into placeholders.  (Simplifies code)
    7. Run 'while True' loop inside tfprocess.py instead of parse.py
    8. Split out training into 'compute_gradient' + 'apply gradient'.
    9. Accumulate gradient until a full macrobatch has been processed before applying.

(NB: This is based on #959)
Should be fully backward compatible with previous iterations of training scripts.

### DIFF
--- a/training/tf/parse.py
+++ b/training/tf/parse.py
@@ -19,32 +19,23 @@
 
 from tfprocess import TFProcess
 from chunkparser import ChunkParser
-import binascii
+import argparse
 import glob
 import gzip
-import itertools
-import math
 import multiprocessing as mp
-import numpy as np
-import queue
 import random
 import shufflebuffer as sb
-import struct
 import sys
 import tensorflow as tf
 import time
-import threading
-
-# 16 planes, 1 side to move, 1 x 362 probs, 1 winner = 19 lines
-DATA_ITEM_LINES = 16 + 1 + 1 + 1
 
 # Sane values are from 4096 to 64 or so. The maximum depends on the amount
 # of RAM in your GPU and the network size. You need to adjust the learning rate
 # if you change this.
 BATCH_SIZE = 512
 
-# Use a random sample of 1/16th of the input data read. This helps
-# improve the spread of games in the shuffle buffer.
+# Use a random sample input data read. This helps improve the spread of
+# games in the shuffle buffer.
 DOWN_SAMPLE = 16
 
 def get_chunks(data_prefix):
@@ -114,28 +105,40 @@ def _parse_function(planes, probs, winner):
 
     return (planes, probs, winner)
 
-def main(args):
-    train_data_prefix = args.pop(0)
+def main():
+    parser = argparse.ArgumentParser(description='Train network from game data.')
+    parser.add_argument("trainpref", help='Training file prefix', nargs='?', type=str)
+    parser.add_argument("restorepref", help='Training snapshot prefix', nargs='?', type=str)
+    parser.add_argument("--train", '-t', help="Training file prefix", type=str)
+    parser.add_argument("--test", help="Test file prefix", type=str)
+    parser.add_argument("--restore", help="Prefix of tensorflow snapshot to restore from", type=str)
+    args = parser.parse_args()
 
-    chunks = get_chunks(train_data_prefix)
-    print("Found {0} chunks".format(len(chunks)))
+    train_data_prefix = args.train or args.trainpref
+    restore_prefix = args.restore or args.restorepref
 
-    if not chunks:
+    training = get_chunks(train_data_prefix)
+    if not args.test:
+        # Generate test by taking 10% of the training chunks.
+        random.shuffle(training)
+        training, test = split_chunks(training, 0.1)
+    else:
+        test = get_chunks(args.test)
+
+    if not training:
+        print("No data to train on!")
         return
 
-    # The following assumes positions from one game are not
-    # spread through chunks.
-    random.shuffle(chunks)
-    training, test = split_chunks(chunks, 0.1)
     print("Training with {0} chunks, validating on {1} chunks".format(
         len(training), len(test)))
 
     train_parser = ChunkParser(FileDataSrc(training),
-            shuffle_size=1<<19, sample=DOWN_SAMPLE, batch_size=BATCH_SIZE)
+            shuffle_size=1<<20, sample=DOWN_SAMPLE, batch_size=BATCH_SIZE)
     #benchmark(train_parser)
     dataset = tf.data.Dataset.from_generator(
         train_parser.parse, output_types=(tf.string, tf.string, tf.string))
     dataset = dataset.map(_parse_function)
+
     dataset = dataset.prefetch(4)
     train_iterator = dataset.make_one_shot_iterator()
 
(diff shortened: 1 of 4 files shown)


training/tf/parse.py
```diff
@@ -19,32 +19,27 @@
...
 
-# 16 planes, 1 side to move, 1 x 362 probs, 1 winner = 19 lines
-DATA_ITEM_LINES = 16 + 1 + 1 + 1
-
-# Sane values are from 4096 to 64 or so. The maximum depends on the amount
-# of RAM in your GPU and the network size. You need to adjust the learning rate
-# if you change this.
+# Sane values are from 4096 to 64 or so.
+# You need to adjust the learning rate if you change this. Should be
+# a multiple of RAM_BATCH_SIZE
```
gcp: I would add some remark here that there is no real benefit to making this larger than RAM_BATCH_SIZE. Many people seem to be extremely confused about this.

It's really only useful if you are trying to exactly replicate some other setup.

training/tf/tfprocess.py
```diff
@@ -278,66 +341,60 @@ def get_batchnorm_key(self):
...
+        # a unique scope that we can store, and use to look them back up
+        # later on.
+        scope = self.get_batchnorm_key()
+        self.weights.append(scope + "/batch_normalization/moving_mean:0")
+        self.weights.append(scope + "/batch_normalization/moving_variance:0")
+        with tf.variable_scope(scope):
+            return tf.layers.batch_normalization(
+                    net,
+                    epsilon=1e-5, axis=1, fused=True,
+                    center=True, scale=True,
```
gcp: You changed this and it's wrong now.

training/tf/tfprocess.py
```diff
@@ -16,17 +16,18 @@
...
 import tensorflow as tf
+import time
 
 def weight_variable(shape):
     """Xavier initialization"""
     stddev = np.sqrt(2.0 / (sum(shape)))
     initial = tf.truncated_normal(shape, stddev=stddev)
     weights = tf.Variable(initial)
-    tf.add_to_collection(tf.GraphKeys.REGULARIZATION_LOSSES, weights)
+    tf.add_to_collection(tf.GraphKeys.WEIGHTS, weights)
```
gcp: I don't understand why you did this renaming. It's now much less clear that bias_variable is also a WEIGHT yet not included in regularization.

zediir:
Is this change useful for multigpu processing?

training/tf/tfprocess.py
ywrt: nuts. that leaked from something else. will fix.

training/tf/tfprocess.py
ywrt: Looking at https://www.tensorflow.org/api_docs/python/tf/GraphKeys and https://www.tensorflow.org/api_docs/python/tf/contrib/layers/apply_regularization it seems that the intention is that regularizers take tf.GraphKeys.WEIGHTS as inputs and auto-add the resulting losses to tf.GraphKeys.REGULARIZATION_LOSSES.

In this case it doesn't seem to matter too much (as there's no other regularizers in the graph), but if there were, it would result in the output of those regularizers being fed as input into the l2 regularizer which I suspect would not be the intention.

ywrt:
> The large amount of changes at once makes this really hard to review and pull :(

Apologies for that :( 
How would you prefer me to do in future? I could:
a. do multiple small pull requests? (but chained changes are difficult to handle).
b. multiple small commits inside a single pull request? (I'm not super familar with github, but it doesn't seem to it easier to review?)
c. only have one small pull request outstanding at a time? (I would struggle with this as the review delay would mean my code base becomes very divergent, and there's little opportunity for you to provide feedback on the direction).

ywrt:
> Is this change useful for multigpu processing?
Not directly. It would make it easier to to train batches over multiple GPUs, but there's no direct support for that yet.

training/tf/parse.py
ywrt: done

gcp:
Multiple, small pull requests for sure. (and I might not take them in the order you expect...)

>I would struggle with this as the review delay would mean my code base becomes very divergent, and there's little opportunity for you to provide feedback on the direction

I'll comment on the patches I take, and may ask what the point of some of these is.

Right now these are near complete rewrites, where it's almost impossible to do a good review (it's very well researched that the larger the review request, the more useless it becomes because you can't spot the bugs).

And indeed, we have #1025...

Error323:
Note that V2 problems from #1025 are present in the _next_ branch, meaning it's in a broken state. I'm investigating currently, in the leela-chess version at https://github.com/glinscott/leela-chess/pull/104. The code is pretty much synchronized with the next branch that is. I'll report when:
1. I've found a/the problem
2. unit tested 
3. have proper results coming out of our V2

ywrt:
Could you confirm that https://github.com/gcp/leela-zero/pull/1034 fixes
the issue?

gcp:
I can't really confirm the v2 stuff works as expected, see issue #1061.

ywrt:
I think the issues noted in #1061 stem from a few places.
1. v2_write_training.py does a better job of keeping validate data separated from training data over restarts, so it's expected that the validation measure should get worse to some degree..
2. With the default settings (512 batch, 128 ram batch), the training measures are over (512/128)*1000 = 4000 ram batches, where the test is done over 800 ram batches, so the variance in the measurement will be ~ 2.4x higher. (This is an increase on the previous code which did 1000 v 800).
3. The validate data isn't shuffled to the same degree (only uses a 2^16 element shuffle buffer). Running with --sample=16 should obviate that, but running with --sample=1 would mean the buffers are under-sized.

What batch/ram batch/sample did you use for the run in #1061 ?

gcp:
>I'm assuming the v2 data was generated by v2_write_training.py with the default settings?

No, not at all. The v2 format is useless for me because it blows up on the on-disk size, so I'd have to dump everything again every time I try to upload data. Given that training is not CPU limited even with large drop factors, v1 is much more useful.

I tried to fix it like this:
https://github.com/gcp/leela-zero/commit/83900554335fbd2d59051de6fe106d49c5322c5e

ywrt:
FWIW: A long from-scratch run shows: 
screenshot-2018-3-20 tensorboard

This is done with strict separation between validate and train (the validate data has never been used for training the network).

ywrt:
Ahh, I see now: you're using the code currently in next, not the #1003 code! 

That totally makes more sense now :). Yes, the current code doesn't shuffle validate, and 8390055 is the right thing to do.

gcp:
>Ahh, I see now: you're using the code currently in next, not the #1003 code!

Yes, I needed to make sure that was working first (it was not really, twice...) before pulling in another huge blob of changes.

ywrt:
Understand. FWIW, #1003 has the shuffle for the test data which is why I was confused.

gcp:
This breaks net_to_model.

ywrt:
ack. I'll fix net_to_model.

Issue #994, opened 2018-03-08
Leelaz v12 release windows x64 crashed


jjoshua2:
Was playing CGOS all day and it crashed after running about 24 hours in the middle of just a regular game.

I clicked debug and see different errors:

The thread 0x293c has exited with code 0 (0x0).

Unhandled exception at 0x00007FFF8A1B0730 (ntdll.dll) in leelaz.exe: 0xC0000005: Access violation reading location 0x0000096403B89018. occurred

It looks like it entered into ntdll and did `call        RtlpFreeHeap (07FFF8A1ADCC0h)` which ended up calling `call        RtlpInsertFreeBlock` and erroring with with instruction:
` mov         edx,dword ptr [rdi-8]`

Is this some sort of double free edge case maybe?

The lczero.exe code that broke in the call:
```
00007FF7880F50B3  mov         rcx,qword ptr [7FF7881BB528h]  
00007FF7880F50BA  call        qword ptr [7FF78812D128h] 
 ```

jjoshua2:
If I have the leela-zero.pdb I can debug it better... I tried making one from the master branch with a release compile but it didn't work. Are these made with a debug build or does anyone have the actual pdb?

gcp:
>I tried making one from the master branch with a release compile but it didn't work.

What's the problem there?

jjoshua2:
The pdb says it can't find the right symbols. The exe is presumably fine

killerducky:
I found a potential problem, both jjoshua2 and I had crashes on CGOS with bots that ponder. When a bot wins a game, CGOS does not inform the bot until the start of the next round. This means the bot could ponder the same position, worst case, the entire round time of 30 minutes.

I think it's possible to overflow m_visits by doing 2 billion visits. I was able to create some positions with very few legal moves and calculated it would overflow in around 200 minutes on my machine. Maybe in some other cases it could be faster than CGOS round time of 30 minutes.

I guess we should put some code in ponder to prevent it from thinking forever?

killerducky:
Ok I looked at the logfile of my CGOS bot that crashed again. I think my system ran out of memory while pondering between rounds, and when the next round started it was very slow, maybe because the system was swapping.

Can we modify LZ to also look at available system memory when allocating nodes instead of a hard coded MAX_TREE_SIZE?

[...]

PR #1009, opened 2018-03-10
Fixes to prevent pondering from running forever.

killerducky:
Fixes to prevent pondering from running forever. Stop after m_maxvisits or m_maxplayouts. I think this is probably close to what a user expects when they use `-v` but leave pondering on. Without `-v` in normal cases it will stop on MAX_TREE_SIZE. But in some endgame cases the tree is not growing, so this change also prevents overflowing m_visits in nodes. See more here https://github.com/gcp/leela-zero/issues/994#issuecomment-371999812

I just picked numeric_limits::max/2, that should be more than enough I think. Another possibility is subtract something like 1000 from max just to prevent multiple threads from causing it to go past max.

This fixes one possible cause for #994. I think we need to do something more dynamic about MAX_TREE_SIZE to prevent using all memory and slowing a user's machine to a crawl. Anyone have an idea for that?

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -593,11 +593,11 @@ int UCTSearch::think(int color, passflag_t passflag) {
     Time elapsed;
     int elapsed_centis = Time::timediff_centis(start, elapsed);
     if (elapsed_centis+1 > 0) {
-        myprintf("%d visits, %d nodes, %d playouts, %d n/s\n\n",
+        myprintf("%d visits, %d nodes, %d playouts, %.0f n/s\n\n",
                  m_root->get_visits(),
                  static_cast<int>(m_nodes),
                  static_cast<int>(m_playouts),
-                 (m_playouts * 100) / (elapsed_centis+1));
+                 (m_playouts * 100.0) / (elapsed_centis+1));
     }
     int bestmove = get_best_move(passflag);
 
@@ -641,6 +641,7 @@ void UCTSearch::set_playout_limit(int playouts) {
                                       decltype(m_maxplayouts)>::value,
                   "Inconsistent types for playout amount.");
     if (playouts == 0) {
+        // Divide max by 2 to prevent overflow corner cases.
         m_maxplayouts = std::numeric_limits<decltype(m_maxplayouts)>::max()/2;
     } else {
         m_maxplayouts = playouts;
@@ -652,6 +653,7 @@ void UCTSearch::set_visit_limit(int visits) {
                                       decltype(m_maxvisits)>::value,
                   "Inconsistent types for visits amount.");
     if (visits == 0) {
+        // Divide max by 2 to prevent overflow corner cases.
         m_maxvisits = std::numeric_limits<decltype(m_maxvisits)>::max()/2;
     } else {
         m_maxvisits = visits;
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -615,13 +615,16 @@ void UCTSearch::ponder() {
     for (int i = 1; i < cpus; i++) {
         tg.add_task(UCTWorker(m_rootstate, this, m_root.get()));
     }
+    auto keeprunning = true;
     do {
         auto currstate = std::make_unique<GameState>(m_rootstate);
         auto result = play_simulation(*currstate, m_root.get());
         if (result.valid()) {
             increment_playouts();
         }
-    } while(!Utils::input_pending() && is_running());
+        keeprunning  = is_running();
+        keeprunning &= !stop_thinking(0, 1);
+    } while(!Utils::input_pending() && keeprunning);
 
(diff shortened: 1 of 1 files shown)


src/UCTSearch.cpp
```diff
@@ -615,13 +615,16 @@ void UCTSearch::ponder() {
...
+    auto keeprunning = true;
     do {
         auto currstate = std::make_unique<GameState>(m_rootstate);
         auto result = play_simulation(*currstate, m_root.get());
         if (result.valid()) {
             increment_playouts();
         }
-    } while(!Utils::input_pending() && is_running());
+        keeprunning  = is_running();
+        keeprunning &= !stop_thinking(0, 1);
```
gcp: Whoops, nice catch. It was definitely intended pondering should keep accurate --visits.

gcp: Do we still need the m_maxvisits check inside if_running() that if we call stop_thinking in the same code path?

The only ones who don't call it now are the search threads but they will abort when the main thread does. So I think we can split the responsibility up, and have is_running just check the flag.

src/UCTSearch.cpp
```diff
@@ -638,7 +641,7 @@ void UCTSearch::set_playout_limit(int playouts) {
                                       decltype(m_maxplayouts)>::value,
                   "Inconsistent types for playout amount.");
     if (playouts == 0) {
-        m_maxplayouts = std::numeric_limits<decltype(m_maxplayouts)>::max();
+        m_maxplayouts = std::numeric_limits<decltype(m_maxplayouts)>::max()/2;
```
gcp: This is a bit sad and hacky, but I don't see a cleaner solution. Doing the m_maxplayouts check in the same place as the increment is not very clean either, and still requiring making the code robust against the overflow.

Maybe add a comment about avoiding overflow.

Hersmunch: Another option might be to not increment m_playouts if it is already at the m_maxplayouts in UCTSearch::increment_playouts().
Both variables could be changed to unsigned if there really is a desire to not limit them to half.

Hersmunch: Also just noticed on line 600 there is `(m_playouts * 100)` which could still overflow. I suppose that overflow is not critical but it might cause some confusion.

killerducky: See below. 1) improved code. 2) test proposed change. 3) test that the old way would have broken. I guess it probably isn't necessary to worry about it still overflowing when elapsed_centis is less than 100.
```
        // Calculate in two steps to lower chances of overflow.
        auto nps = m_playouts / (elapsed_centis+1);
        nps *= 100;
        myprintf("%d visits, %d nodes, %d playouts, %d n/s\n\n",
                 m_root->get_visits(),
                 static_cast<int>(m_nodes),
                 static_cast<int>(m_playouts),
                 nps);
        nps = std::numeric_limits<decltype(m_maxvisits)>::max()/2 / (elapsed_centis+1);
        nps *= 100;
        myprintf("debug1 %d visits, %d nodes, %d playouts, %d n/s\n\n",
                 m_root->get_visits(),
                 static_cast<int>(m_nodes),
                 static_cast<int>(m_playouts),
                 nps);
        nps = (std::numeric_limits<decltype(m_maxvisits)>::max() * 100) / (elapsed_centis+1);
        myprintf("debug2 %d visits, %d nodes, %d playouts, %d n/s\n\n",
                 m_root->get_visits(),
                 static_cast<int>(m_nodes),
                 static_cast<int>(m_playouts),
                 nps);
```

Example output, note 2nd works and 3rd is broken.
```
13 visits, 4673 nodes, 12 playouts, 200 n/s

debug1 13 visits, 4673 nodes, 12 playouts, 715827816 n/s

debug2 13 visits, 4673 nodes, 12 playouts, -16 n/s
```

src/UCTSearch.cpp
killerducky: I think is_running was changed recently. Now it only checks the flag and also MAX_TREE_SIZE, so I think it is close to what you are saying. We could move the MAX_TREE_SIZE check into stop_thinking also. There might be some weird spinning of the workers if the main search is busy expanding a NN eval, I guess it would be ok. What do you think?

src/UCTSearch.cpp
killerducky: Oh this is easier. Also the other way was rounding to nearest 100 n/s.
```
        myprintf("%d visits, %d nodes, %d playouts, %.0f n/s\n\n",
                 m_root->get_visits(),
                 static_cast<int>(m_nodes),
                 static_cast<int>(m_playouts),
                 (m_playouts * 100.0) / (elapsed_centis+1));
```

src/UCTSearch.cpp
gcp: I'd make the change, I think it makes the code more sensible.

PR #1015, opened 2018-03-11
adopt LeelaTero for KGS (as option)

petgo3:
adopt LeelaTero for KGS (as option)


petgo3:
can't resolve, as far as i understand i have no permission

PR #1026, opened 2018-03-12
Multi-GPU thread scalability improvements

ihavnoid:
- Instead of using worker threads, maintain a pool of thread contexts
- This means that any thread can acquire any GPU, but reduces the number thread-to-thread communication

'netbench 2000' performance on my i7-8700 + GTX 1080 + GTX 1050 setup:

- Before this patch:
2t 1567n/s
3t 2579n/s
4t 2970n/s
5t 3381n/s
- After this patch
2t 1276n/s
3t 2637n/s
4t 3101n/s
5t 3438n/s

Probably will need some more comparisons on bigger machines.

### DIFF
--- a/src/OpenCLScheduler.cpp
+++ b/src/OpenCLScheduler.cpp
@@ -36,7 +36,6 @@ void OpenCLScheduler::initialize(const int channels) {
         m_thread_data_count = 0;
         // on a multi-gpu situation, we are going to maintain a thread data pool.
 
-        m_thread_data.resize(cfg_gpus.size());
         for (auto gpu : cfg_gpus) {
             // create opencl thread data explicitly here with the right 'gnum'.
             opencl_thread_data = std::make_unique<ThreadData>(gnum);
@@ -50,8 +49,9 @@ void OpenCLScheduler::initialize(const int channels) {
             // starting next GPU, let's not dump full list of GPUs
             silent = true;
 
-            m_thread_data[gnum].push_back(std::move(opencl_thread_data));
-            m_thread_data[gnum].emplace_back( new ThreadData(gnum) );
+            m_thread_data.emplace_back(new ThreadDataQueue());
+            m_thread_data[gnum]->push_back(std::move(opencl_thread_data));
+            m_thread_data[gnum]->emplace_back( new ThreadData(gnum) );
 
             m_thread_data_count += 2;
             gnum++;
@@ -90,7 +90,7 @@ void OpenCLScheduler::forward(const std::vector<net_t>& input,
         // while trying to avoid the most recently scheduled thread.
         for(auto i = size_t{0}; i<m_thread_data.size(); i++) {
             auto gnum = (i + m_thread_data_recently_scheduled + 1) % m_thread_data.size();
-            auto sz = m_thread_data[gnum].size();
+            auto sz = m_thread_data[gnum]->size();
             if (sz > max_gnum_cnt) {
                 max_gnum = gnum;
                 max_gnum_cnt = sz;
@@ -98,8 +98,8 @@ void OpenCLScheduler::forward(const std::vector<net_t>& input,
         }
        
         assert(max_gnum_cnt > 0); 
-        opencl_thread_data = std::move(m_thread_data[max_gnum].front());
-        m_thread_data[max_gnum].pop_front();
+        opencl_thread_data = std::move(m_thread_data[max_gnum]->front());
+        m_thread_data[max_gnum]->pop_front();
         m_thread_data_recently_scheduled = max_gnum;
         m_thread_data_count--;
     }
@@ -113,7 +113,7 @@ void OpenCLScheduler::forward(const std::vector<net_t>& input,
         std::lock_guard<std::mutex> lk(m_mutex);
 
         int gnum = opencl_thread_data->m_gpu_num;
-        m_thread_data[gnum].push_back(std::move(opencl_thread_data));
+        m_thread_data[gnum]->push_back(std::move(opencl_thread_data));
         m_thread_data_count++;
         m_cv.notify_one();
     }
--- a/src/OpenCLScheduler.cpp
+++ b/src/OpenCLScheduler.cpp
@@ -15,11 +15,12 @@
     You should have received a copy of the GNU General Public License
     along with Leela Zero.  If not, see <http://www.gnu.org/licenses/>.
 */
-#include <cassert>
 
 #include "config.h"
 
 #ifdef USE_OPENCL
+#include <cassert>
+
 #include "GTP.h"
 #include "Random.h"
 #include "OpenCLScheduler.h"
@@ -28,34 +29,33 @@ OpenCLScheduler opencl;
 
 void OpenCLScheduler::initialize(const int channels) {
     // multi-gpu?
-    if (!cfg_gpus.empty()) {
+    if (cfg_gpus.size() >= 2) {
         auto silent{false};
-        for(auto gpu : cfg_gpus) {
+        int gnum = 0;
+
+        // on a multi-gpu situation, we are going to maintain a thread data pool.
+        for (auto gpu : cfg_gpus) {
+            // create opencl thread data explicitly here with the right 'gnum'.
+            opencl_thread_data = std::make_unique<ThreadData>(gnum);
+
             auto opencl = std::make_unique<OpenCL>();
             auto net = std::make_unique<OpenCL_Network>(*opencl);
             opencl->initialize(channels, {gpu}, silent);
             m_opencl.push_back(std::move(opencl));
             m_networks.push_back(std::move(net));
 
-            // Clear thread data on every init call.  We don't know which GPU
-            // this thread will be eventually be assigned to
-            if (opencl_thread_data != nullptr) {
-                opencl_thread_data = nullptr;
-            }
-
             // starting next GPU, let's not dump full list of GPUs
             silent = true;
-        }
 
-        for(size_t gnum = 0; gnum < m_networks.size(); gnum++) {
-            // two instances for optimal scalability
+            m_thread_data.push_back(std::move(opencl_thread_data));
             m_thread_data.emplace_back( new ThreadData(gnum) );
-            m_thread_data.emplace_back( new ThreadData(gnum) );
+
+            gnum++;
         }
     } else {
         auto opencl = std::make_unique<OpenCL>();
         auto net = std::make_unique<OpenCL_Network>(*opencl);
-        opencl->initialize(channels, {});
+        opencl->initialize(channels, cfg_gpus);
 
         m_opencl.push_back(std::move(opencl));
         m_networks.push_back(std::move(net));
--- a/src/OpenCLScheduler.cpp
+++ b/src/OpenCLScheduler.cpp
@@ -15,6 +15,8 @@
     You should have received a copy of the GNU General Public License
     along with Leela Zero.  If not, see <http://www.gnu.org/licenses/>.
 */
+#include <cassert>
+
 #include "config.h"
 
 #ifdef USE_OPENCL
@@ -22,7 +24,6 @@
 #include "Random.h"
 #include "OpenCLScheduler.h"
 
-thread_local auto current_thread_gpu_num = size_t{0};
 OpenCLScheduler opencl;
 
 void OpenCLScheduler::initialize(const int channels) {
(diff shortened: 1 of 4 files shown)


bjiyxo:
I'll do the experiment tomorrow. BTW, it looks like the progress is little(from 3381n/s to 3438n/s), doesn't it?

ihavnoid:
Well, the idea (at least from my side) was that we saturate performance on a lower thread count, e g., look at 4 threads.  On even bigger setups this might start matter though.
I dont think we will ever saturate performance when lower than 2x the number of GPUs though.

fell111:
Would you please also run the benchmark for t1 one 1 GPU? I'm wondering whether this PR will have negative performance impact for self play.

ihavnoid:
No difference on my setup, at least.

bjiyxo:
Before this patch:
	1st move	2nd move
1t	1132n/s 	809n/s
2t 	1986n/s 	1426n/s
3t 	2708n/s 	1908n/s
4t 	3065n/s 	2209n/s
5t 	3521n/s		2642n/s
6t	4137n/s		3156n/s
7t	4889n/s		3679n/s
8t	5193n/s		3858n/s
9t	5302n/s		3818n/s
10t	5318n/s		3909n/s

After this patch
	1st move	2nd move
1t	1777n/s 	1296n/s
2t 	3670n/s 	2578n/s
3t 	4809n/s 	3625n/s
4t 	5985n/s 	5618n/s
5t 	6695n/s		6118n/s
6t	6557n/s		6223n/s
7t	3091n/s		2290n/s
8t	3150n/s		2286n/s
9t	3059n/s		2180n/s
10t	3163n/s		2152n/s

It looks like the new one is significantly better at 5~6 thread.  Nice job! @ihavnoid

src/OpenCLScheduler.cpp
```diff
@@ -38,22 +39,18 @@ void OpenCLScheduler::initialize(const int channels) {
 
             // Clear thread data on every init call.  We don't know which GPU
             // this thread will be eventually be assigned to
-            opencl_thread_data = ThreadData();
+            if (opencl_thread_data != nullptr) {
+                opencl_thread_data = nullptr;
```
gcp: Just call .reset()?

In what case could this code be needed?

src/OpenCLScheduler.cpp
```diff
@@ -15,14 +15,15 @@
     You should have received a copy of the GNU General Public License
     along with Leela Zero.  If not, see <http://www.gnu.org/licenses/>.
 */
+#include <cassert>
```
gcp: Move this below config.h, inside the #ifdef USE_OPENCL.

gcp:
Looks good to me but I'd want to test this for stability on AMD cards too.

ihavnoid:
Addressed the code review comments.  Well, for the first comment I just rewrote the code so that I don't need to destroy the initially created context, only to be recreated later on.

ihavnoid:
Well, I would hope there is somebody who owns multiple AMD GPUs but most of the people I saw had multiple NVIDIA GPUs.  Any volunteers...?

@gcp - are you concerned of threading bugs related to having multiple threads sharing and handing over OpenCL contexts?  Seems like you had some prior experience dealing with those kind of bugs?

gcp:
>Seems like you had some prior experience dealing with those kind of bugs?

https://community.amd.com/message/2824449

That one shouldn't affect us, luckily. But yeah, I don't trust the drivers too much. It should be testable on a single-GPU card if necessary with a small source patch.

CamWagner:
I tested this on the machine we used for Haylee vs Leela Zero, with 4x 1080 and up to 28 threads (Intel Xeon E5-2690 v4).

With netbench 50000, our best results were 5800 n/s on 7 threads without this patch, and 5500 n/s on 6 threads with this patch.

The machine should be powerful, I'm surprised it did not exceed 6500 in netbench. During the actual match we reached 11,000 n/s sometimes. But in the netbench testing we did slightly worse with this patch. If it is a slight improvement for everyone else, though, then it's worth it :)

roy7:
Once upon a time didn't netbench fail to use extra threads? Maybe I'm thinking of something else. Would --benchmark be better to test with?

ihavnoid:
Looking at the netbench code, it seems that it can result in some bad numbers if the thread scheduler is unfair and we have a lot of threads.  How it works is that it divides the evaluations by the number of threads, and then the evaluations are considered done when the last thread finishes its work.

This means it reports the slowest thread's runtime and then multiplies it by the number of threads.  I dont think fairness is one of the requirements here so I dont think it is right for netbench to penalze unfair thread schedulers either.

Probably we have to rely on something that isn't netbench, or fix the netbench scalability problem...

gcp:
Netbench could split up the work in smaller batches, instead of worksize / threads.

gcp:
```
index 3baf33f..6d37fdc 100644
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -93,18 +93,19 @@ static std::array<std::array<int, BOARD_SQUARES>, 8> rotate_nn_idx_table;
 
 void Network::benchmark(const GameState * const state, const int iterations) {
     const auto cpus = cfg_num_threads;
-    const auto iters_per_thread = (iterations + (cpus - 1)) / cpus;
+    const auto loops_per_batch = 8;
+    const auto batches = iterations / loops_per_batch;
 
     const Time start;
 
     ThreadGroup tg(thread_pool);
-    for (auto i = 0; i < cpus; i++) {
-        tg.add_task([iters_per_thread, state]() {
-            for (auto loop = 0; loop < iters_per_thread; loop++) {
+    for (auto batch = 0; batch < batches; batch++) {
+        tg.add_task([loops_per_batch, state]() {
+            for (auto loop = 0; loop < loops_per_batch; loop++) {
                 get_scored_moves(state, Ensemble::RANDOM_ROTATION, -1, true);
             }
         });
-    };
+    }
     tg.wait_all();
 
     const Time end;
```

I guess the problem here is that we keep queuing up all work. Doing this a bit smarter would require extending ThreadPool so we can block on a CPU freeing up.

ihavnoid:
I will come up with a PR for the netbench.  I was thinking of having a std::atomic counter shared by all threads and all threads stop looping when it hits the global threshold.

gcp:
Sounds OK. Given that nobody replied to you request for an AMD test, I'll just see if it doesn't do anything funny on my machine and assume it works otherwise.

But we may need to ask people to redo the tests to see if it does actually scale better?

Matuiss2:
I have AMD GPU

ihavnoid:
@Matiuss2 do you have 1 or 2 GPUs?  Can you try to compile and test this patch and see how it performs?

ihavnoid:
For the netbench accuracy patch, I created a PR: https://github.com/gcp/leela-zero/pull/1060

@bjiyxo and @CamWagner Can you also test the new patches?  PR 1060 tries to address the netbench benchmark itself's scalability issues.  For @bjiyxo I see that you didn't do a netbench to start with, so it's possible that it doesn't really matter though.

"before" : https://github.com/gcp/leela-zero/pull/1060 only
"after" : https://github.com/gcp/leela-zero/pull/1060 plus https://github.com/gcp/leela-zero/pull/1026

ihavnoid:
In my machine with a 6x128 net:

Before:
- 2t 1440n/s
- 3t 2643n/s
- 4t 3387n/s
- 5t 3555n/s
- 6t 3529n/s

After:
- 2t 1249n/s
- 3t 2883n/s
- 4t 3413n/s
- 5t 3536n/s
- 6t 3472n/s

Hmmm.... looks like some improvement on lower thread counts but not so easily noticeable?

bjiyxo:
@ihavnoid I'll test it tomorrow. @CamWagner I'd like to see the test among 2, 3 & 4 GPUs if possible.

CamWagner:
testing might have to wait until I'm less busy but I'll try to get those tests soon :)

Matuiss2:
Ah I only have 1 so I cant help sorry

bjiyxo:
"before" : #1060 only

2t 1539 n/s
3t 1691 n/s
4t 2810 n/s
5t 3231 n/s
6t 3609 n/s
7t 3362 n/s
8t 3347 n/s
9t 3305 n/s
10t 3394 n/s

"after" : #1060 plus #1026
2t 2235 n/s
3t 3013 n/s
4t 3747 n/s
5t 3182 n/s
6t 1497 n/s
7t 1496 n/s
8t 1463 n/s
9t 1583 n/s
10t 1469 n/s

I'm not sure what's wrong with netbench there, but I redo 20s/move and the result of after patch is 5t(7972 n/s) & 6t (8268 n/s).

ihavnoid:
The "after" result certainly looks puzzling.  Maybe I should have a look more carefully at the code...

ihavnoid:
I added another patch here which implements a more sensible load balancing on the scheduler, rather than just doing something naive.  This scheduling behavior actively picks the most underutilized GPU.

ihavnoid:
For more scalability tests, I rented a Google Cloud machine with 16 CPUs and 4 Tesla P100s.  Using latest 10x128 net.

Netbench:

Thread count | Before | After | Rate
-- | -- | -- | --
1 | 1085 | 1134 | 1.05
2 | 1896 | 2267 | 1.20
3 | 3015 | 3506 | 1.16
4 | 3685 | 4023 | 1.09
5 | 3938 | 3937 | 1.00
6 | 4045 | 3518 | 0.87
7 | 3602 | 3354 | 0.93
8 | 3783 | 3100 | 0.82
9 | 4184 | 3402 | 0.81
10 | 4326 | 4163 | 0.96
11 | 4073 | 4143 | 1.02
12 | 4185 | 4262 | 1.02
13 | 4319 | 4276 | 0.99
14 | 4275 | 4094 | 0.96
15 | 4296 | 4246 | 0.99
16 | 4215 | 4232 | 1.00

--benchmark

Thread count | Before | After
-- | -- | --
1 | 3200 visits, 1130633 nodes, 3199 playouts, 1073 n/s | 3200 visits, 1130633 nodes, 3199 playouts, 1198 n/s
2 | 3201 visits, 1131070 nodes, 3200 playouts, 1975 n/s | 3202 visits, 1131482 nodes, 3201 playouts, 2462 n/s
3 | 3202 visits, 1131404 nodes, 3201 playouts, 2964 n/s | 3204 visits, 1132317 nodes, 3203 playouts, 3559 n/s
4 | 3207 visits, 1133521 nodes, 3206 playouts, 3910 n/s | 3203 visits, 1131775 nodes, 3202 playouts, 4510 n/s
5 | 3413 visits, 1206175 nodes, 3412 playouts, 4549 n/s | 3204 visits, 1132461 nodes, 3203 playouts, 5005 n/s
6 | 3207 visits, 1133468 nodes, 3206 playouts, 4932 n/s | 3214 visits, 1135935 nodes, 3213 playouts, 4657 n/s
7 | 3208 visits, 1133588 nodes, 3207 playouts, 4334 n/s | 3382 visits, 1195288 nodes, 3381 playouts, 4900 n/s
8 | 3212 visits, 1135435 nodes, 3211 playouts, 4722 n/s | 3216 visits, 1136881 nodes, 3215 playouts, 5023 n/s
9 | 3402 visits, 1202757 nodes, 3401 playouts, 4199 n/s | 3218 visits, 1137903 nodes, 3217 playouts, 4021 n/s
10 | 3212 visits, 1136354 nodes, 3211 playouts, 4225 n/s | 3211 visits, 1135645 nodes, 3210 playouts, 3527 n/s
11 | 3223 visits, 1139845 nodes, 3222 playouts, 4239 n/s | 3216 visits, 1137534 nodes, 3215 playouts, 4070 n/s
12 | 3221 visits, 1139350 nodes, 3220 playouts, 4411 n/s | 3226 visits, 1140826 nodes, 3225 playouts, 3981 n/s
13 | 3218 visits, 1138167 nodes, 3217 playouts, 4021 n/s | 3214 visits, 1136905 nodes, 3213 playouts, 3780 n/s
14 | 3222 visits, 1139584 nodes, 3221 playouts, 4537 n/s | 3226 visits, 1141021 nodes, 3225 playouts, 3583 n/s
15 | 3387 visits, 1197487 nodes, 3386 playouts, 4180 n/s | 3327 visits, 1176551 nodes, 3326 playouts, 4056 n/s
16 | 3219 visits, 1138387 nodes, 3218 playouts, 3699 n/s | 3367 visits, 1190575 nodes, 3366 playouts, 4055 n/s

The new fix is peaking performance around 5~8 threads (4 GPUs x2) while having a little bit more performance on the peaks.  Not by a whole lot, though.

gcp:
If I see it correctly, the differences are small and the best performance result is on the existing code?

ihavnoid:
At this point I am not so sure if this is worth merging.  I think depending on circumstances it may help or it may hurt - IMHO it makes sense the thread scheduling will be less important as the network size gets larger - the thread scheduling overhead turns into a smaller portion compared to the rest of the problem.

Unless @gcp thinks this is a cleaner implementation (at least it creates less number of threads) I think we can just close this.  What do you think?

gcp:
I don't think this is necessarily cleaner than what we have (less threads but more complex code and context management), and I agree the benefit is not at all clear.

PR #1034, opened 2018-03-13
Bugfix: was reusing the same chunk repeatedly.

ywrt:
Copied the list-ref instead of the list itself resulting in reusing the same chunk over and over again.


ywrt:
#1032 is better. closing in favour of that.

PR #1037, opened 2018-03-14
Require --noponder for --visits

WhiteHalmos:
This matches behaviour for --playouts.

### DIFF
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -60,10 +60,11 @@ static void parse_commandline(int argc, char *argv[]) {
         ("threads,t", po::value<int>()->default_value(cfg_num_threads),
                       "Number of threads to use.")
         ("playouts,p", po::value<int>(),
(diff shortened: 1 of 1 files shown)


CamWagner:
This is fixed by https://github.com/gcp/leela-zero/pull/1009

roy7:
I sort of thought --visits working with pondering was a feature, not a bug.

CamWagner:
Same. I really like it, it makes responses much quicker.

WhiteHalmos:
If it's a feature, please close the PR.

I thought --visits and --playouts didn't have different meanings other than limiting strength by a different metric.

gcp:
--visits has a sensible meaning with pondering enabled, playouts does not really, so the behavior is intended.

WhiteHalmos:
Ah, with pondering enabled, --visits can end much earlier when it is the bot's turn. Makes sense now.

PR #1039, opened 2018-03-14
Add gtp heatmaps command

roy7:
Convenience function to show all heatmaps with one command.


gcp:
Pff, what about making heatmap take an "all" argument? And make it display the average of all maps by default?

roy7:
I thought about doing like heatmap_average and heatmap_max but didn't think you'd take them. My purpose on heatmaps was to easily look at questionable decisions (like self-atari) to see if it might be a rotational issue (some rotations see the danger, some don't, etc).

If you like I can remove heatmaps and add "heatmap all" "heatmap average" and "heatmap max". ("max" being the max() which as some point I want to toy with when using multiple evals, since the point of an eval isn't choosing the best move, it's to know which moves to spend search on. I feel like max would be sure to search every move that any rotation felt was important.)

gcp:
I'd rather have it take arguments than add a new comment for a trivial variation on an existing one, yes.

PR #1045, opened 2018-03-17
Next 709 fast board tests bb

barrybecker4:
I was able to run the existing google unit tests in my Ubuntu VM, and have started to add tests for FastBoard. To make testing of FastBoard easier, I split display_board into 2 methods. As serialize_board method which creates a string (for use with unit tests) and the same display_board method which now calls serialize_board to get the string to print. 

If it looks like I am on the right track, let me know, and I will continue adding more FastBoard tests to test all of the methods in that class.

### DIFF
--- a/src/FastBoard.cpp
+++ b/src/FastBoard.cpp
@@ -25,8 +25,6 @@
 #include <queue>
 #include <sstream>
 #include <string>
-#include <cstdarg>
-#include <cstdio>
 
 #include "Utils.h"
 #include "config.h"
--- a/src/FastBoard.cpp
+++ b/src/FastBoard.cpp
@@ -19,11 +19,14 @@
 #include "FastBoard.h"
 
 #include <cassert>
+#include <boost/format.hpp>
 #include <array>
 #include <iostream>
 #include <queue>
 #include <sstream>
 #include <string>
+#include <cstdarg>
+#include <cstdio>
 
 #include "Utils.h"
 #include "config.h"
@@ -277,48 +280,58 @@ float FastBoard::area_score(float komi) const {
     return black - white - komi;
 }
 
-void FastBoard::display_board(int lastmove) {
-    int boardsize = get_boardsize();
 
-    myprintf("\n   ");
-    print_columns();
-    for (int j = boardsize-1; j >= 0; j--) {
-        myprintf("%2d", j+1);
-        if (lastmove == get_vertex(0, j))
-            myprintf("(");
-        else
-            myprintf(" ");
-        for (int i = 0; i < boardsize; i++) {
-            if (get_square(i,j) == WHITE) {
-                myprintf("O");
-            } else if (get_square(i,j) == BLACK)  {
-                myprintf("X");
-            } else if (starpoint(boardsize, i, j)) {
-                myprintf("+");
-            } else {
-                myprintf(".");
-            }
-            if (lastmove == get_vertex(i, j)) myprintf(")");
-            else if (i != boardsize-1 && lastmove == get_vertex(i, j)+1) myprintf("(");
-            else myprintf(" ");
-        }
-        myprintf("%2d\n", j+1);
-    }
-    myprintf("   ");
-    print_columns();
-    myprintf("\n");
+void FastBoard::display_board(int lastmove) {
+    myprintf("%s", FastBoard::serialize_board(lastmove).c_str());
 }
 
-void FastBoard::print_columns() {
-    for (int i = 0; i < get_boardsize(); i++) {
-        if (i < 25) {
-            myprintf("%c ", (('a' + i < 'i') ? 'a' + i : 'a' + i + 1));
+std::string FastBoard::serialize_board(int lastmove) {
+    
+    int boardsize = get_boardsize();
+    std::ostringstream oss;
+
+    oss << "\n   ";
+    oss << get_columns();
+    for (int j = boardsize-1; j >= 0; j--) {
+        oss << boost::format("%2d") % (j + 1);
+        if (lastmove == get_vertex(0, j))
+            oss << "(";
+        else
+            oss << " ";
+        for (int i = 0; i < boardsize; i++) {
+            if (get_square(i,j) == WHITE) {
+                oss << "O";
+            } else if (get_square(i,j) == BLACK)  {
+                oss << "X";
+            } else if (starpoint(boardsize, i, j)) {
+                oss << "+";
+            } else {
+                oss << ".";
+            }
+            if (lastmove == get_vertex(i, j)) 
+                oss << ")";
+            else if (i != boardsize-1 && lastmove == get_vertex(i, j)+1) 
+                oss << "(";
+            else oss << " ";
         }
-        else {
-            myprintf("%c ", (('A' + (i - 25) < 'I') ? 'A' + (i - 25) : 'A' + (i - 25) + 1));
-        }
+        oss << boost::format("%2d\n") % (j + 1);
     }
-    myprintf("\n");
+    oss << "   ";
+    oss << get_columns();
+    oss << "\n";
+    return oss.str();
+}
+
+std::string FastBoard::get_columns() {
+    std::ostringstream oss;
+    for (int i = 0; i < get_boardsize(); i++) {
+        char c =  (i < 25) ?
+            (('a' + i < 'i') ? 'a' + i : 'a' + i + 1) :
+            (('A' + (i - 25) < 'I') ? 'A' + (i - 25) : 'A' + (i - 25) + 1);
+        oss << c << " ";
+    }
+    oss << "\n";
+    return oss.str();
 }
 
 void FastBoard::merge_strings(const int ip, const int aip) {
--- a/src/Utils.h
+++ b/src/Utils.h
@@ -44,13 +44,14 @@ namespace Utils {
 
     template<typename T>
     T rotl(const T x, const int k) {
-	    return (x << k) | (x >> (std::numeric_limits<T>::digits - k));
+        return (x << k) | (x >> (std::numeric_limits<T>::digits - k));
     }
 
     inline bool is7bit(int c) {
         return c >= 0 && c <= 127;
     }
 
+    /* Returns the smallest multiple of b that is larger than a */
     size_t ceilMultiple(size_t a, size_t b);
 }
 
--- a/src/Utils.h
+++ b/src/Utils.h
@@ -51,7 +51,7 @@ namespace Utils {
         return c >= 0 && c <= 127;
     }
 
-    /* Returns the smallest multiple of b that is larger than a */
+    /* Returns the smallest multiple of b that is larger than or equal to a */
     size_t ceilMultiple(size_t a, size_t b);
 }
 
--- a/src/tests/FastBoard_unittest.cpp
+++ b/src/tests/FastBoard_unittest.cpp
@@ -21,10 +21,8 @@
 #include <limits>
 #include <vector>
 
-#include "Utils.h"
 #include "FastBoard.h"
 
-using namespace Utils;
 
 TEST(FastBoardTest, Board3x3) {
     FastBoard b;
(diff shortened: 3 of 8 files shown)


src/Utils.h
```diff
@@ -44,13 +44,14 @@ namespace Utils {
...
     T rotl(const T x, const int k) {
-	    return (x << k) | (x >> (std::numeric_limits<T>::digits - k));
+        return (x << k) | (x >> (std::numeric_limits<T>::digits - k));
     }
 
     inline bool is7bit(int c) {
         return c >= 0 && c <= 127;
     }
 
+    /* Returns the smallest multiple of b that is larger than a */
```
gcp: Looking at the code, that's not true?

src/FastBoard.cpp
```diff
@@ -19,11 +19,14 @@
...
 
 #include <cassert>
+#include <boost/format.hpp>
 #include <array>
 #include <iostream>
 #include <queue>
 #include <sstream>
 #include <string>
+#include <cstdarg>
+#include <cstdio>
```
gcp: Why would you need these?

barrybecker4: They are not needed. I will remove them. I must have added them while experimenting with something, then forgot to remove.

src/Utils.h
barrybecker4: Based on the unit tests (and the name of the function), this is exactly what it does. 
Here are some sample tests:
 ```
EXPECT_EQ(ceilMultiple(13, 5), (size_t)15);
EXPECT_EQ(ceilMultiple(23, 5), (size_t)25);
EXPECT_EQ(ceilMultiple(99, 100), (size_t)100);
EXPECT_EQ(ceilMultiple(149, 10), (size_t)150);
```
Is there a better way to describe it?

Comment (gcp) [reply src/Utils.h]: What does ceilMultiple(5, 5) return?

Comment (barrybecker4) [reply src/Utils.h]: Its five. I added a test for it
EXPECT_EQ(ceilMultiple(5, 5), (size_t)5);
I will change the comment from 
`Returns the smallest multiple of b that is larger than a`
to
`Returns the smallest multiple of b that is larger than or equal to a`

**Comment by barrybecker4** [comment]
Thanks for the review. I just wanted to verify that I was on the right track. I will compete the rest of the tests for FastBoard shortly and add them to this request. Then we will also be able to close this issue, and I will open a separate one for FullBoard, then FastState, and so on.

**Comment by barrybecker4** [comment]
I made the 2 review changes, and added some more FastBoard tests to test all the methods that made sense to test. You can merge to next, if there are no more changes requests.

I noticed a few things while adding the tests for FastBoard.
I will open issues for any of the following that you agree are justified and implement them as time permits.
 - Either remove set/get_square (because it is unused) or make m_square private in FastBoard and allow access only with set/get_square. My preference is for the latter.
 - Refactor out FastBoardSerializer class from FastBoard. I think it would simplify the code to extract all the logic having to do with board serialization to a separate class. It would have methods for serialize, text_to_move, text_to_move_sgf and probably starpoint.
 - Not 100% sure, but I think there should be an update_board method in FastBoard that contains all the logic from FullBoard::update_board except the part that deals with the Zobrist hashing. Then the FullBoard::update_board method would just do the Zobrist update and call FastBoard::update_board. If we did this, then we could create sample FastBoards in the unit tests and the tests for count_pliberties, is_eye, is_suicide, and area_score would then make sense. As it is now, it really only makes sense to test those methods in the context of a FullBoard.

**Comment by barrybecker4** [comment]
As indicated by @WhiteHalmos, I was able to turn on assertions when running the tests by using:
```
cmake -DCMAKE_BUILD_TYPE=Debug ..
make leelaz
make tests
./tests
```
The the negative test that I had added to check that an assertion was triggered passed, when I ran locally. The problem is that the CI build does not run in Debug apparently, so that test did not pass when I pushed it. For now I commented that one negative test that relies on assertions being on.
I assume that we do want CI builds to use Debug because then we might hit assertions that are being missed now, and second, so that unit tests can verify that assertions are triggered when they are supposed to be.
How can we make the CI build run with Debug set?
The changes in this PR can be pulled now. I can uncomment that one test later when the CI build uses Debug.

File: README.md [hunk 3]
```diff
@@ -127,7 +127,7 @@ source and remove the line that says "#define USE_OPENCL".
 
     # Use stand alone directory to keep source dir clean
     mkdir build && cd build
-    cmake ..
+    cmake -DCMAKE_BUILD_TYPE=Debug ..
```
gcp: Why would people want to run a debug build? Remove this. Not everyone is a developer.

src/tests/FastBoard_unittest.cpp
```diff
@@ -0,0 +1,342 @@
...
+    FastBoard b = create_filled_9x9();
+    
+    EXPECT_EQ(2, b.count_pliberties(b.get_vertex(0, 0)));
+    EXPECT_EQ(4, b.count_pliberties(b.get_vertex(1, 2)));
+    EXPECT_EQ(4, b.count_pliberties(b.get_vertex(4, 3)));
+    EXPECT_EQ(4, b.count_pliberties(b.get_vertex(4, 4))); 
+    EXPECT_EQ(4, b.count_pliberties(b.get_vertex(5, 4)));
+}
+
+// Results will make more sense in FullBuard test
```
gcp: Typos: FullBoard.

src/tests/FastBoard_unittest.cpp
```diff
@@ -0,0 +1,342 @@
...
+    EXPECT_EQ("tt", b.move_to_text_sgf(FastBoard::RESIGN));
+}
+
+TEST(FastBoardTest, GetStoneList) {
+    FastBoard emptyBoard;
+    emptyBoard.reset_board(3);
+    EXPECT_EQ("", emptyBoard.get_stone_list());
+    
+    FastBoard b = create_filled_5x5();
+    EXPECT_EQ("A4 B2 C2 C3 C4 C5 D2 D3", b.get_stone_list());
```
gcp: What if the order of the stones changes? Why would that be wrong?

src/tests/FastBoard_unittest.cpp
```diff
@@ -0,0 +1,342 @@
...
+    EXPECT_EQ("resign", b.move_to_text(FastBoard::RESIGN));
+}
+
+TEST(FastBoardTest, MoveToTextSgf) {
+    FastBoard b = create_filled_3x3();
+    EXPECT_EQ("bc", b.move_to_text_sgf(b.get_vertex(1, 0)));
+    EXPECT_EQ("ab", b.move_to_text_sgf(b.get_vertex(0, 1)));
+    EXPECT_EQ("ca", b.move_to_text_sgf(b.get_vertex(2, 2)));
+    EXPECT_EQ("tt", b.move_to_text_sgf(FastBoard::PASS));
+    EXPECT_EQ("tt", b.move_to_text_sgf(FastBoard::RESIGN));
```
gcp: This test (resign=tt) is probably not very meaningful.

src/tests/FastBoard_unittest.cpp
```diff
@@ -0,0 +1,342 @@
...
+        " 2 . . X . . . . . . . . . . . . . . . .  2\n"
+        " 1 . . . . . . . . . . . . . . . . . . .  1\n"
+        "   a b c d e f g h j k l m n o p q r s t \n\n";
+    EXPECT_EQ(expected, b.serialize_board());
+}
+
+TEST(FastBoardTest, GetVertexOn19x19) {
+    FastBoard b;
+    b.reset_board(19);
+    EXPECT_EQ(22, b.get_vertex(0, 0));
```
gcp: Given that these vertex coordinates are internal to Leela, it's not clear to me what the use of this test is.

Wouldn't it make more sense to combine both tests and just see if you can round-trip the coordinates, rather than check if they are equal to arbitrary numbers?

src/tests/FastBoard_unittest.cpp
```diff
@@ -0,0 +1,342 @@
...
+    EXPECT_EQ(4, b.count_pliberties(b.get_vertex(5, 4)));
+}
+
+// Results will make more sense in FullBuard test
+TEST(FastBoardTest, IsSuicideWhenNotForBlack) {
+    FastBoard b;
+    b.reset_board(5);  
+    b.set_square(2, 2, FastBoard::WHITE);
+    EXPECT_EQ(false, b.is_suicide(b.get_vertex(1, 1), FastBoard::BLACK));
+    EXPECT_EQ(false, b.is_suicide(b.get_vertex(2, 1), FastBoard::BLACK));
```
gcp: Wouldn't testing the is_suicide=True case be useful?

src/tests/FastBoard_unittest.cpp
```diff
@@ -0,0 +1,342 @@
...
+}
+
+// Results will make more sense in FullBuard test
+TEST(FastBoardTest, IsSuicideWhenForBlackInAllWhiteField) {
+    FastBoard b = create_5x5_all_white_field();
+
+    EXPECT_EQ(false, b.is_suicide(b.get_vertex(1, 1), FastBoard::BLACK));
```
gcp: Same question, they're all = false tests.

src/tests/FastBoard_unittest.cpp
```diff
@@ -0,0 +1,342 @@
...
+    FastBoard b = create_filled_5x5();
+    EXPECT_EQ("A4 B2 C2 C3 C4 C5 D2 D3", b.get_stone_list());
+    
+    FastBoard whiteFieldBoard = create_5x5_all_white_field();
+    EXPECT_EQ("A3 B3 C1 C2 C3 C4 C5 D3 D5 E4", whiteFieldBoard.get_stone_list());
+}
+
+TEST(FastBoardTest, StarPoint9x9) {
+    
+    EXPECT_EQ(true, FastBoard::starpoint(9, 2, 2));
```
gcp: This test probably makes more sense if one uses SGF or GTP text coordinates rather than the numbers.

i.e. check that on 19x19 "D4" is indeed a star point, rather than the (2,2) numbers which make little sense outside Leela.

README.md
barrybecker4: The Debug option was added so that assertions would be enabled when running the tests. Without that option, any tests that verified assertions or inadvertently triggered them would fail. Is there a way to have assertions turned on just when running tests? I think we also want them on when doing the CI build. Even though most people are not developers, it might be nice to have assertions/debug on while they run so that if any assertions are hit while running, they will be reported, and whatever problems they indicate, fixed. I'll remove it if you insist, but I at least wanted to offer this opinion.

src/tests/FastBoard_unittest.cpp
barrybecker4: It is true that if the implementation changed the order of stones in get_stone_list, then the test would fail, and it probably should not. But if the order of the stones does not matter, maybe that method should return a Set instead of a string. Then checking varying order would be easier. As it is now it is more of a regression test rather than a unit test. It would have to be updated if the order changed. Possible resolutions:
   1) leave it as is and accept that it is only a regression test.
   2) remove it and not test this method.
   3) add a get_stone_set that returns a set of stones and is unite tested, and a separate get_stone_list method that calls get_stone_set and converts the set to a string. Unit test the get_stone_set method.
   4) Convert the space delimited string to a set and then compare it with a set expectation. While this would be most correct, it would also mean adding quite a bit of test logic.
   5) Some other suggestion
   My preference is for 1).

src/tests/FastBoard_unittest.cpp
barrybecker4: Yes, it would be more usful, but since update_board is not part of FastBoard, there is no way to do it!  is_suicide will always be false unless we can move update_board to FastBoard. Then it could be properly tested here. As it is now, it can only be tested in the context of a FullBoard (which I have done on a separate branch).

src/tests/FastBoard_unittest.cpp
barrybecker4: If there were a text_to_move method, or if starpoint had a signature that accepted a string like "D4", then I would agree with you. But since there is not, this is difficult to do. I wouldn't want to add an extra method to FastBoard that is used only for testing purposes, and I don't want to add logic to the unit test code to do the conversion because then it would need testing too. Is it OK to leave this as is, or can you think of a nice way to do what you suggest?

src/tests/FastBoard_unittest.cpp
barrybecker4: Same reply as previous comment. I think we should move the update_board logic (excluding the part having to do with Zobrist) to FastBoard.

src/tests/FastBoard_unittest.cpp
barrybecker4: Yes, that is a good suggestion. The test now verifies that get_xy(get_vertex) is consistent.

src/tests/FastBoard_unittest.cpp
barrybecker4: I removed resign == tt case, although I didn't really see the harm in having it.

src/tests/FastBoard_unittest.cpp
barrybecker4: Fixed the typos.

barrybecker4:
Thanks for the review. I made some of the suggested changes, but have some reservations about the ones I did not make yet. Please see comments above.

Regarding my 3 proposals for further improvements above, I have a few more comments:

You did not respond the first one about making m_square private, so I will assume you are OK with it.

For item 2, the reason to separate out a FastBoardSerializer class is so that then each class would have just a [single responsibility](https://en.wikipedia.org/wiki/Single_responsibility_principle) and be easier to test, and easier to understand for anyone reading the code. When looking at the FastBoard class you would not need to look at the serialization logic and whatever dependencies it had. The FastBoardSerializer would just focus on the printing of a board and one could focus on that instead of wondering if there were dependencies on any other methods that would be in the same file if they were combined as they are now. Maybe I will make a PR for this so you can see what I mean.

For item 3, the reason to have the update_board method in FastBoard is so that all the tests for things like is_eye, is_suicide, and count_pliberties, which currently make no sense, would then make sense. In other words, all those methods could be tested independent from the Zobrist hash. FullBoard could also test those things, but at least you would know they worked within the context of only FastBoard. As it is now, the FastBoard tests for is_eye, is_suicide, count_pliberties are kind of meaningless because they are verifying incorrect answers. For example, you cannot check for is_suicide or is_eye beig true because they are never true unless FastBoard has the logic in update_board (which updates neighbours, etc). If you agree, I can do this change in a separate PR.

gcp:
Endlessly arguing about the tests is not a good use of my time.

PR #1048, opened 2018-03-17
Clean up Network.

TFiFiE:
Mostly const correctness and auto maximization.


gcp:
This also makes the passed-by-value parts of function arguments const, which, while not wrong, has not much to do with const correctness and is somewhat arguable (and is often argued in C++ circles...).

That said, other parts of the code are woefully inconsistent in that anway, so I might as well pull this.

Issue #1007, opened 2018-03-10
self-atari


pw31:
Leela Zero sometimes plays self-ataris with confidence, e.g. this game 
https://online-go.com/game/11990416
where she claimed a 97.5% win-probability against Bifron [6d] on OGS
(with nicely increasing trend since 10 moves or so) just before she played 
118. n9 which puts a 15-stone-chain of hers in self-atari, then resigned 
immediately after W captured next move. 
I know this has been noticed before, e.g. in a game against AQ, but can't remember 
which issue it was.
I can't imagine this has anything to do with the network weights, can it?  Seems
just like a bug to me. How could one not include the capture of a 15-stone chain
into account as a move candidate in the search?

TFiFiE:
#936

Eddh:
It has a lot to do with the network weights. The net simply has blind spots such as this sometimes. We are not including manually the capture of these 15 groups chain move because that would not be the "zero" approach of AlphagoZero but using human knowledge.

pw31:
@TFiFiE : thanks for reminding me
@Eddh : Is this because such self-atari moves rarely happen in training games, so LZ doesn't "learn" that her stones can be captured?  Does it have to do with the bocksize? I haven't seen self-ataris of small groups, only of big groups. Thanks

Eddh:
@pw31 Yes they rarely happen, and yes it is generally thought that bigger blocksize can deal better with liberties of large groups because more blocksize can make information travel farther. So it's probably a combination of both factor.

roy7:
Full game log from Leela's side: http://termbin.com/jwb4

If I loadsgf 11990416.sgf 120, and look at the heatmaps:

0 .25
0 .37
0 .23
0 .29
4 .48
1 .39
1 .35
0 .23

Those are the heatmap values and the win rates. 3 random rotations see the capture move as barely visible compared to other moves on the map, 7 of them give it a weight that rounds to 0 so it's so small it is basically invisible. As far as Leela can tell, the capture will never happen.

This isn't the case of an illegal move becoming legal either. So maybe it's just something Leela still needs to learn.

Splee99:
After studied this issue for some time, now I understand this is indeed related to the weights. As neutral network increases in size, we hardly know what happens exactly inside the residential layers. It doesn't take into account of explicit valuation of taking stones out of board. In this case taking 15 stones is not that worth according to the network at the first glance, but with increased payouts, it may think differently depending on the tree size.

[...]

Issue #980, opened 2018-03-05
UCTSearch Tree memory optimizations


sethtroisi:
There have been several attempts to reduce the memory usage of the UCTSearch tree.
I'm not sure what the correct approach is anymore and wanted to vet the other developer's opinions.

At a high level there have been three approaches

1. **expandable** originally written by @ywrt (https://github.com/gcp/leela-zero/pull/574) and shepherded by @sethtroisi (https://github.com/gcp/leela-zero/pull/887)
2. **edges** written by @WhiteHalmos (https://github.com/gcp/leela-zero/pull/878)
3. **trimmed** suggested by @GCP (https://github.com/gcp/leela-zero/issues/956#issuecomment-369914185) and coded by @sethtroisi ([source](https://github.com/gcp/leela-zero/compare/next...sethtroisi:v__small_node_simple?expand=1))

All three approaches reduce memory usage (trimmed by ~80%, expandable by ~70%, edges by ~30%) but with some downsides (trimmed requires an occasional 2nd pass to re-add children, expandable is highly complex bug-prone, edges is only the smallest memory reduction) more on this latter.

I think I'd benefit (as well as everyone else) in defining what the goals are right now
1. Allowing deeper search (AKA much larger MAX_TREE_SIZE) for long time control games / analysis
2. Decreasing self-play client requirements
3. Ability to experiment with various UCT / MCTS ideas

---
**Description of the changes**

**expandable nodes**: Currently when a leaf node is expanded we create one child UCTNode (~60 bytes) for each legal child move (~250 average). **expandable** change store the children's data (6 bytes of move + score) in a vector until `uct_select_child` tries to select a child for play_simulation and then expands them on demand. The compact memory format is close to the max theoretical saving and strongly enables Goal 1 and 2. The cost here is the code is highly complex and will slow any future development and experimentation in UCTNode as well as possibly introducing race-conditions for multi-threaded search.

**edges**: Instead of expanding each child to a full UCTNode (~60 bytes) expand a partial data structure (~16 bytes) and only do the rest of the expansion on demand. This change is much less risky than expandable because expansion is more cleanly handled. The downside is the saving maybe isn't large enough to justify the change when the other options exist.

[...]

PR #1049, opened 2018-03-18
Ignore nodes with extremely low policy priors

WhiteHalmos:
Skip creating child nodes with <0.001 policy prior during expansion.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -92,6 +92,7 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
 
     auto legal_sum = 0.0f;
     for (const auto& node : raw_netlist.first) {
+        if (node.first < 0.001f) continue;
         auto vertex = node.second;
         if (state.is_move_legal(to_move, vertex)) {
             nodelist.emplace_back(node);
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -90,9 +90,14 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
 
     std::vector<Network::scored_node> nodelist;
 
+    auto min_psa = 0.0f;
+    if (!cfg_noise && cfg_random_cnt == 0) {
+        // Trimming should only be used for matches with no randomness.
+        min_psa = 0.001f;
+    }
     auto legal_sum = 0.0f;
     for (const auto& node : raw_netlist.first) {
-        if (node.first < 0.001f) continue;
+        if (node.first < min_psa) continue;
         auto vertex = node.second;
         if (state.is_move_legal(to_move, vertex)) {
             nodelist.emplace_back(node);


WhiteHalmos:
```
leelaz-next_ed26 v leelaz-trim_ed26 (233/400 games)
board size: 19   komi: 7.5
leelaz-next_ed26     92 39.48%       40  34.19%     52  44.83%    341.48
leelaz-trim_ed26    141 60.52%       64  55.17%     77  65.81%    339.31
                                     104 44.64%     129 55.36%
```
Ringmaster config:
```python
competition_type = 'playoff'
LZ_ARGS = '/projects/lztest/{} --timemanage off --noponder -r10 -g -t1 -d -w {}'
def LZ(vals):
  commands = ['time_settings 450 0 0']
  return Player(LZ_ARGS.format(*vals), startup_gtp_commands=commands)
players = dict()
p1 = 'leelaz-next_ed26'
p2 = 'leelaz-trim-math_ed26'
players[p1] = LZ(p1.split('_'))
players[p2] = LZ(p2.split('_'))
board_size = 19
komi = 7.5
matchups = [
]
```
```
>>> sprt.SPRT(141,92,1,195,225,0.05,0.05)
'H1'
```

Eddh:
Interesting. In theory this should be relevant only in cases where there is a big drop in winrate because of fpu reduction. This is a bigger difference than I expected from this.

src/UCTNode.cpp
```diff
@@ -92,6 +92,7 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
 
     auto legal_sum = 0.0f;
     for (const auto& node : raw_netlist.first) {
+        if (node.first < 0.001f) continue;
```
TFiFiE: As a magic number, this should at the least be captured in a constant or, preferably, be made a configurable parameter (to be exposed by USE_TUNER).

killerducky:
Maybe this is due to the increased speed of e.g. uct_select_child on CPU? But if you run with `-t 2` or higher GPU will be the bottleneck and CPU speed will not matter at all. Can you try that? Also why disable timemanage?

killerducky:
Also I think this should only be done for match games, not self-play. Perhaps use the same trick as timemanage and overload the `-m` switch.

Eddh:
Why shouldn't it be done for self-play ? the noise can easily take care of exploration and those low policy moves are almost never explored at root level anyway.

Mardak:
From my rotation explorations in https://github.com/gcp/leela-zero/issues/1007#issuecomment-373833483 and https://github.com/gcp/leela-zero/issues/877#issuecomment-373255981, there have often been the correct capture/punish-LZ move that had N < 0.1%.

For example from the self-atari issue, here's the L11 move 120 to capture LZ with 1000 visits for each of the 8 rotations with 5b90bd32:
```
seed 22: L11 ->     472 (V: 100.00%) (N:  0.05%) PV: L11 R8 F9 F10 E9 E10 D10 D11 C11 D9 C10
seed:  6 L11 ->     483 (V: 100.00%) (N:  0.05%) PV: L11 R8 F9 F10 E9 E10 D10 D11 D9 C11 E11
seed:  2 L11 ->     491 (V: 100.00%) (N:  0.15%) PV: L11 R8 F9 F10 E9 E10 D10 D11 C11 D9 C10 G8
seed: 11 L11 ->     448 (V: 100.00%) (N:  0.08%) PV: L11 R8 F9 E11 D9 C11 F11 F10 E10 G9 F8 G8
seed:  3 L11 ->     497 (V: 100.00%) (N:  0.15%) PV: L11 R8 F9 E11 D10 D11 C11 C10 C9 B10
seed:  4 L11 ->     496 (V: 100.00%) (N:  0.17%) PV: L11 R8 F9 F10 E9 E10 D10 D11 D9 C11 E11 E12
seed: 12 L11 ->     489 (V: 100.00%) (N:  0.13%) PV: L11 R8 F9 F10 E10 E9 F8 E11 D10 D11 F11
seed:  1 L11 ->     490 (V: 100.00%) (N:  0.02%) PV: L11 R8 F9 F10 E10 E9 F8 E11 D10 D11 F11 C10 D9
```

With this PR, half of the possible rotations would no longer have found the capture.

Would that means it would make it harder to learn that this move should have a higher policy prior?

Although at a high level, basically this change magnifies what the net knows or fails to know. So if it doesn't know about self-atari in certain rotations, then it will more likely fail, but also if it thinks a certain line of play is better, it won't waste visits on things it learned to not play anyway.

Eddh:
Oh and the dirichlet noise is applied after so yeah it shouldn't be used in selfplay.

WhiteHalmos:
@Eddh Opps! Updated SPRT test, thanks for catching that. I thought the results looked wonky!

@killerducky Happy to have more people help test this. My machine is tied up right now with the currrent test.

@Mardak Agreed that this is not good for self-play.

src/UCTNode.cpp
WhiteHalmos: Agreed, but I'd like others to easily test this and for more to weigh in before making this more complicated than necessary. Here's a branch with more math to back it up: https://github.com/WhiteHalmos/leela-zero/commit/1d8ec2d8e0c11d3a57850a00dced1d1b5cba979c

WhiteHalmos:
@petgo3 Assuming LZ is resource constrained on KGS, this very minor change might be worth trying for an extra boost.

Eddh:
What do you think about applying Dirichlet noise before trimming ? Would that allow for using this PR in self-play games ?

ihavnoid:
My general feeling is that the threshold 0.001 seems a bit arbitrary.  If we are doing ten million playouts, for example, there probably is some value to try a couple of moves for the lowest probabilities just in case something is wrong with the network.  On a ten thousand playout setting, skipping anything lower than 0.001 might make sense.

WhiteHalmos:
@ihavnoid See this for a more mathematically sound approach: https://github.com/WhiteHalmos/leela-zero/commit/1d8ec2d8e0c11d3a57850a00dced1d1b5cba979c

fell111:
It's better to introduce a new parameter to set the threshold. Then we can use the same binary for testing. For self play, no threshold will be added. And in matches/other games, we could tune this parameter to get better winrate.

gcp:
I'd like to understand why this makes a difference. We discussed this a lot during #980 and related threads, but it's something I'd have enabled when we surpassed half our memory budget.

If the strength increase is not simply due to the speedup, it means the current UCT behavior is too exploratory.

>It's better to introduce a new parameter to set the threshold. 

Nope. You can do this while tuning this value, but relying on the user to set magic values is incredibly hostile.

gcp:
>@ihavnoid See this for a more mathematically sound approach: WhiteHalmos/leela-zero@1d8ec2d

You can estimate this off of the time control to make it better, but then you still have the critical question: what are you going to do in "infinite analysis"?

You will have problems if you control the tree search by determining these values before starting your search, that's why I think this is better tackled in UCT itself.

gcp:
Hmm, actually...currently MAX_TREE_SIZE is at most 100'000'000, which means the maximum amount of playouts will be around 280k, which means that independent of PUCT, the maximum ratio between best PSA and worst PSA would be 0.0019 in order for us to discover it.

I guess a dynamic formula that caps at no less from that number would work, i.e. it is guaranteed to not make the program worse? But the question is, would you still see the strength improvement you saw now?

ihavnoid:
@gcp - I would argue that the MAX_TREE_SIZE parameter should be command-line controllable so that the user can select reasonable values depending on their setup.  Maybe we can set the cutoff threshold based on whatever user-provided value this is.

If possible I would think that we would want the MAX_TREE_SIZE to be dynamically detected but unfortunately I failed to find any portable method that can do that (e.g., measure the process's memory upper bound in a reliable way)

Eddh:
"If it's mathematically safe, why not prune the root after setting Dirichlet noise?"
Unless I'm misunderstanding something, there is still a potential problem with the first 30 moves, since even 0.0001 policy moves can be picked sometimes. But honestly I think the noise can take care of that well enough and the impact would be very small.

gcp:
>Unless I'm misunderstanding something, there is still a potential problem with the first 30 moves, since even 0.0001 policy moves can be picked sometimes

You are misunderstanding something. It's not possible to select a move that the search did not explore.

Eddh:
Ahh I see nevermind then.

gcp:
>I would argue that the MAX_TREE_SIZE parameter should be command-line controllable so that the user can select reasonable values depending on their setup.

No way. Allowing the user to set the maximum memory usage, that makes more sense, though.

But we are so wasteful with this right now, and this patch will change the behavior a lot, that it you get back to the original problem I raised above. You literally even risk that the program will end up playing WEAKER if you give it MORE memory, because it will drop the value of this magic parameter that doesn't just control memory but also has an effect on the strength.

I repeat: the only reason I'd consider taking it right now is because the *current* code has this limit, so it can't make things worse. But the approach is very deeply problematic because of the above.

WhiteHalmos:
@gcp Agreed with all your points. This PR was more intended to get the idea of improving UCT search through better pruning. I have more math to back up https://github.com/WhiteHalmos/leela-zero/tree/trim_math but it needs lots more testing.

Updated results. Makes me realize that I need to consistently test with 400 games since the winrate kept dropping from 65% in ~100 games to 60% in ~200 games to now 56% in ~400 games.
```
leelaz-next_ed26 v leelaz-trim_ed26 (400/400 games)     
board size: 19   komi: 7.5  
leelaz-next_ed26    176 44.00%       78  39.00%     98  49.00%    341.08                                        
leelaz-trim_ed26    224 56.00%       102 51.00%     122 61.00%    338.98                                        
                                     180 45.00%     220 55.00%
```
Literally one loss away from SPRT significance. Strange that 56% after 400 games isn't SPRT significant on (0, 35).
```python
>>> -400 * math.log(1/.56-1)/math.log(10)
41.89414020800522
>>> sprt.SPRT(224,1,176,0,35,0.05,0.05)
''
>>> sprt.SPRT(224,1,175,0,35,0.05,0.05)
'H1'
```

gcp:
SPRT is stricter because it assumes you will potentially play an infinite amount of games and cherry pick the PASS result.

WhiteHalmos:
Closing this PR as it has achieved its purpose: Demonstrating that there is performance improvements to be had along with huge memory savings by trimming the UCT search tree.

gcp:
I tried to implement a "conservative" version of this, i.e. that gives identical results but saves memory, by abusing MAX_TREE_SIZE, but it's tricky.

Assume we are on the second visit of a node. On the first visit we got a 0% winrate after expanding the first node.

Now, a node with an arbitrarily small policy prior can be selected. It's independent of max_playouts or anything. It does not matter that this node will not be part of the final selection, it still influences in the tree shape and returned score.

PR #1052, opened 2018-03-18
Optimised various parts of Network

Hersmunch:
Main benefit is from `winograd_transform_in()` by reading the input into a zero padded array for each channel rather than each block (of each channel). I'll see if I can do something similar for opencl (outside of this pr and I may ask for help if I get stuck).
CPU only:
`$ time ./leelaz -w ed26f634 -t 1 --noponder --benchmark -p 10000`
was:
`10001 visits, 3506140 nodes, 10000 playouts, 53 n/s`
`real    3m11.195s`
now:
`10001 visits, 3506140 nodes, 10000 playouts, 59 n/s`
`real    2m51.633s`
With GPU and `-p 100000`:
was:
`100001 visits, 34117003 nodes, 100000 playouts, 491 n/s`
`real    3m27.845s`
now:
`100001 visits, 34117003 nodes, 100000 playouts, 492 n/s`
`real    3m26.939s`

### DIFF
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -680,18 +680,14 @@ void innerproduct(const std::vector<float>& input,
                 &input[0], 1,
                 0.0f, &output[0], 1);
 
-    if (ReLU) {
-        const auto lambda_ReLU = [](auto val) { return (val > 0.0f) ?
-                                                        val : 0.0f; };
-        for (unsigned int o = 0; o < outputs; o++) {
-            auto val = biases[o] + output[o];
+    const auto lambda_ReLU = [](const auto val) { return (val > 0.0f) ?
+                                                          val : 0.0f; };
+    for (unsigned int o = 0; o < outputs; o++) {
+        auto val = biases[o] + output[o];
+        if (ReLU) {
             val = lambda_ReLU(val);
-            output[o] = val;
-        }    
-    } else {
-        for (unsigned int o = 0; o < outputs; o++) {
-            output[o] += biases[o];
         }
+        output[o] = val;
     }
 }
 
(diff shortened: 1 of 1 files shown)


gcp:
Needs rebasing.

src/Network.cpp
```diff
@@ -704,26 +702,31 @@ void batchnorm(const size_t channels,
...
-    for (auto c = size_t{0}; c < channels; ++c) {
-        const auto mean = means[c];
-        const auto scale_stddiv = stddivs[c];
-
-        if (eltwise == nullptr) {
-            // Classical BN
-            auto arr = &data[c * spatial_size];
+    if (eltwise == nullptr) {
+        // Classical BN
+        const auto lambda_ReLU = [](auto stddiv, auto val) { return (val > 0.0f) ?
```
gcp: This isn't obviously an optimization to me, because it removes the possibility to vectorize the stddiv multiplication.

src/Network.cpp
```diff
@@ -704,26 +702,31 @@ void batchnorm(const size_t channels,
...
-                                            val : 0.0f; };
-
-    for (auto c = size_t{0}; c < channels; ++c) {
-        const auto mean = means[c];
-        const auto scale_stddiv = stddivs[c];
-
-        if (eltwise == nullptr) {
-            // Classical BN
-            auto arr = &data[c * spatial_size];
+    if (eltwise == nullptr) {
```
gcp: Ugh, duplicating the code is ugly.

src/Network.cpp
```diff
@@ -685,15 +680,18 @@ void innerproduct(const std::vector<float>& input,
                 &input[0], 1,
                 0.0f, &output[0], 1);
 
-    const auto lambda_ReLU = [](auto val) { return (val > 0.0f) ?
-                                            val : 0.0f; };
-
-    for (unsigned int o = 0; o < outputs; o++) {
-        auto val = biases[o] + output[o];
-        if (ReLU) {
+    if (ReLU) {
```
gcp: What's the point of pulling this out if this is a template argument in the first place? You now duplicate the code in the source, whereas the compiler has everything it needs to know to eliminate the redundant code?

Hersmunch:
Thanks for going through it. I'll endeavor to keep commits short and sweet in the future.

I've reverted the changes you commented on. I went back through each of them and it seems my original testing was insufficient. Doing multiple benchmark runs shows that the changes I saw were below the level of the noise in my results.

My dayjob is using C on embedded micros so I'm used to trying to avoid as many branches or operations as possible :) I appreciate your feedback and I hoping it will help me find the right balance..

PR #1053, opened 2018-03-18
autogtp: clean up networks when they are not needed any more

marcocalignano:
`autogtp` now saves all networks in a subdirectory `networks`. If the option `--erase` (`-e`) is active then the old best network and the old networks used for matches are deleted so that in that directory there are at most 3 networks.
If you use the stop-resume option (`q+Enter`) to update to this version you have to manually copy the network used by the saved games. To do this look into the files `storefile*` and check the lines starting with 
`network` `firstNet` `secondNet` and copy them in the new directory networks.
If you are under Linux you can use the following command to find the networks you need:
`grep -e network -e firstNet -e secondNet storefile*  | awk '{print $2}' | sort | uniq`

### DIFF
--- a/autogtp/Job.cpp
+++ b/autogtp/Job.cpp
@@ -67,7 +67,7 @@ Job(gpu, parent)
 
 Result ProductionJob::execute(){
     Result res(Result::Error);
-    Game game(m_network, m_option);
+    Game game("networks/" + m_network, m_option);
     if (!game.gameStart(m_leelazMinVersion)) {
         return res;
     }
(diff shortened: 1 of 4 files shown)


akdtg:
With this commit the self-gaming SGF contains "networks" instead of 8 chars hash:

PB[Leela Zero 0.12 networks]PW[Leela Zero 0.12 1ccb7342]

marcocalignano:
I thought I fixed that!

marcocalignano:
Oh no this is what leelaz write in the sgf itself I will have to add a new regexp to fix this!

marcocalignano:
@roy7 this game I uploaded has a corrupted sgf format can you fix it or remove it: The problem is int the palayers name.
Game data  219c847b128e9318d1b92e3f2c3a477fab9c1271795147f06bb4bac1f2962e6c

marcocalignano:
sorry but Game data 76f40e6e2e20895392f378957a59f6cd73113cbeaf031bdc7425c0bc0b54b0cd and Game data 98cab135059cb94f70a490ff97a6fda84a64d3c7214975246c937bdd984d69ad  also have a problem (getting close ... )

roy7:
Those games have been deleted.

PR #1059, opened 2018-03-19
Make gather_features not depend on zeroed planes.

TFiFiE:


gcp:
Not sure what the point of this is?

gcp:
I meant it can't be performance because the current code zeroes it out. If it's safety/readability, then I believe it's cleaner to just zero out the planes in gather_features, or assert at the start that the planes vector is empty.

TFiFiE:
I think this is a clean and efficient change to get rid of an unnecessary precondition, while still allowing the option of reusing an already allocated vector: The function doesn't have to clear the input first, it doesn't introduce a conceptually unjustified asymmetry between set and unset bits and it removes branches in fill_input_plane_pair (which might even be a speedup).

gcp:
>The function doesn't have to clear the input first

There are no users that reuse a vector, so the resize ends up doing just that.

>it doesn't introduce a conceptually unjustified asymmetry between set and unset bits

I don't even...you're arguing against zero-initializing a feature map? You now have to make sure every bit is now either set or unset. I don't think that's an improvement in readability, and it makes things more error prone.

>it removes branches in fill_input_plane_pair (which might even be a speedup).

That part could even be separated from the other changes if it was beneficial.

I'm not pulling this without any user that could get a benefit out of it.

TFiFiE:
I only just noticed that TimeStep step (of which planes is a data member) in Training::record() gets cleared before. Then why not just let gather_features return a new NNPlanes then? I was planning to get rid of as many out parameters in Network.cpp anyway.

PR #1064, opened 2018-03-20
Implement stochastic weight averaging

Ttl:
Implements weight averaging method in: https://arxiv.org/abs/1803.05407. Outputs file called `leelaz-<number of nets in average>-<training steps>.txt` during the training. average_weights.py can be used also as a stand-alone program to average existing weights.

Average weight file is not restored when restoring from the model.

### DIFF
--- a/training/tf/average_weights.py
+++ b/training/tf/average_weights.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python3
+#
+#    This file is part of Leela Zero.
+#    Copyright (C) 2017 Henrik Forsten
+#
+#    Leela Zero is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation, either version 3 of the License, or
+#    (at your option) any later version.
+#
+#    Leela Zero is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with Leela Zero.  If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+import numpy as np
+
+def swa(inputs, output, weights=None):
+    """ Average weights of the weight files.
+
+    inputs : List of filenames to use as inputs
+    output : String of output filename
+    weights : List of numbers to use for weighting the inputs
+    """
+
+    out_weights = []
+
+    if weights == None:
+        weights = [1.0]*len(inputs)
+
+    if len(weights) != len(inputs):
+        raise ValueError("Number of weights doesn't match number of input files")
+
+    # Normalize weights
+    weights = [float(w)/sum(weights) for w in weights]
+
+    for count, filename in enumerate(inputs):
+        with open(filename, 'r') as f:
+            weights_in = []
+            for line in f:
+                weights_in.append(weights[count] * np.array(list(map(float, line.split(' ')))))
+            if count == 0:
+                out_weights = weights_in
+            else:
+                if len(out_weights) != len(weights_in):
+                    raise ValueError("Nets have different sizes")
+                for e, w in enumerate(weights_in):
+                    if len(w) != len(out_weights[e]):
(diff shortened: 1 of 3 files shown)


Eddh:
Very nice. Even in the worse case where this is bad in the long term for the pipeline (which is still an open question), this will still allow for producing better nets, even from past 5x64 or 6x128 weights, which could be useful for running a strong LZ on light hardware :)

gcp:
I already did my own implementation of this, but the test match was still running when you made this pull :-)

I think your implementation is wrong, or at least not according to the paper (but the difference may not matter much in practice). After the SWE run, you're supposed to recalculate the batchnorm weights.

What I wonder though, is whether you don't expect the correct batchnorm weights to also be the averages of the networks you've averaged. I didn't check back if recalculating the batchnorm weights improves the result, but if you have a set of training data it's a relatively "free" thing to do.

Ttl:
In my earlier tests averaging without correcting batch norm worked fine. It would be interesting to know if recalculating the batch norms does anything. I would guess that with constant learning rate and relatively low amount of training steps it can't really deviate that far anyway.

Eddh:
I just found that the authors answered a few questions on reddit https://www.reddit.com/r/MachineLearning/comments/84irfc/r_averaging_weights_leads_to_wider_optima_and/

isty2e:
@Eddh I am not sure about the "learning rate should be higher than it is now", since if our model is near the optimum, we don't want it to jump around the weight space with large steps, especially considering "after some predefined number of epochs" part. Of course the effect of different learning rates can be tested easily so it would be better to judge based on experiments.

Eddh:
We don't want it to jump around to much, but we don't want it to be too static either. They clearly state that it should not converge, and in their paper they did an analysis on the constant learning rate, and it was not optimal for it to be too low or too high. They recommended some intermediate value between the maximum and the minimum learning rate used in conventional annealing schedule. Since we are near the minimum, I think for this technique to work best, the learning rate should be higher. How much higher tho that's hard to tell. As you say experiments will tell us about that.

For the "after some predefined number of epochs" part, I think that refers specifically to training from scratch (random weights), so not really to our current context.

isty2e:
Yes, either a too low or too high learning rate is certainly suboptimal. My understanding is that this technique works best if the oscillating networks are averaged. So networks from initial steps, far from the optimum, will not positively contribute presumably, which is in accordance with "after some predefined number of epochs".

The question is that what the optimal learning rate is, and whether it would be higher or lower than the current one. IMHO, there is no particular reason to believe either way. We are indeed close to the minimum, but considering that the training set is constantly changing, there seems like no concrete reason for both arguments. We will be able to see the results from various experiments soon, though.

gcp:
>In my earlier tests averaging without correcting batch norm worked fine. It would be interesting to know if recalculating the batch norms does anything. I would guess that with constant learning rate and relatively low amount of training steps it can't really deviate that far anyway.

For what it's worth, I did this by using net_to_model, and then setting learning_rate=0. This will generate a new network where only the batchnorm weights are "trained". IIRC the batchnorm weights are some exponential average of 0.999 so a few thousand batches would correct any divergence.

Ttl:
I implemented the batch norm recalculation. It's kind of inelegant since it reads and writes the weights on disk, but it works. Having separate graphs would be more elegant, but I think it has problems with requiring twice as much GPU RAM.

Recalculating is currently kind of expensive since it does 1000 extra steps for every output net. Might be a good idea to only recalculate them for nets that are uploaded in the training.

ywrt:
Sorry, only just saw this. Note that it has a few small issues:

1. When self.swa_max_n networks have been written, it will just keep over-writing the last one instead of keeping the most recent 16. (which I assume is what it was supposed to do?)

2. By default, the batch_norm layers have [momentum=0.99](https://www.tensorflow.org/api_docs/python/tf/layers/batch_normalization), so training only 200 examples will retain ~13% of the previous values.

3. The code that attempts to recalculate the batch norms doesn't actually do anything, because the update ops aren't being run.

>  Having separate graphs would be more elegant, but I think it has problems with requiring twice as much GPU RAM.

Note that the graph itself uses ~ zero GPU ram. By far the majority of the RAM is used to hold batch gradients. (roughly: For a batch size of N, there are N copies of the trainable variables in the graph).

Issue #1067, opened 2018-03-21
LZ is not allowed to resign when the engine is limited by the low number of visits


akdtg:
UCTSearch::should_resign() does not allow to resign if the search procedure has not produced the certain number of visits:

```
    const auto visits = m_root->get_visits();
    if (visits < std::min(500, cfg_max_playouts))  {
        // low visits
        return false;
    }
```

Usually, this is not an issue when the playouts are limited, but when the engine is limited by the number of visits and the search procedure has not produced 500 visits then LZ is playing until the is no other option but a passing move.

PR #1068, opened 2018-03-21
Fix for permission to resign when the engine is limited by the number of visits

akdtg:
Fix issue #1067

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -247,7 +247,7 @@ bool UCTSearch::should_resign(passflag_t passflag, float bestscore) {
     }
 
     const auto visits = m_root->get_visits();
-    if (visits < std::min(500, cfg_max_playouts))  {
+    if (visits < std::min(500, std::min(cfg_max_visits, cfg_max_playouts)))  {
         // low visits
         return false;
     }
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -25,6 +25,7 @@
 #include <limits>
 #include <memory>
 #include <type_traits>
+#include <algorithm>
 
 #include "FastBoard.h"
 #include "FastState.h"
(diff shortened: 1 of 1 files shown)


src/UCTSearch.cpp
```diff
@@ -247,7 +247,7 @@ bool UCTSearch::should_resign(passflag_t passflag, float bestscore) {
     }
 
     const auto visits = m_root->get_visits();
-    if (visits < std::min(500, cfg_max_playouts))  {
+    if (visits < std::min(500, std::min(cfg_max_visits, cfg_max_playouts)))  {
```
TFiFiE: It might be more elegant to make it a single `std::min` taking an initializer_list of three arguments:
`std::min({500, cfg_max_visits, cfg_max_playouts})`
This does come with the unfortunate gotcha that both GCC and Clang will generate unhelpful error messages if`<algorithm>` isn't included now.

roy7:
Keep in mind that timemanage will often abort search when you're at about half the playouts if all playouts are going into a single move. So with, say, --visits 500 the min() test for 500 and max_visits will be 500, but almost no move ever will make it to 500 a single time.

Should it perhaps use half of the max visits/playouts as the min() to be more timemanage compatible? Or should timemanage perhaps not fire if we're in a situation where we'd be resigning?

akdtg:
@roy7, this fix is only for respecting of the max_visits in the same way as max_playouts, the influence of timemanage is an another question. During the tests I simply switched timemanage off.

src/UCTSearch.cpp
gcp: Sounds like cleaner code to me, so I like your suggestion. Just add <algorithm> to the includes.

gcp:
>Should it perhaps use half of the max visits/playouts as the min() to be more timemanage compatible?

Sounds like a good idea to me. You can also wonder if this check is needed at all in Leela Zero?

src/UCTSearch.cpp
akdtg: Done!

akdtg:
Honestly, I wonder if we need this check too, so I just modified it in a safe way.

src/UCTSearch.cpp
```diff
@@ -25,6 +25,7 @@
 #include <limits>
 #include <memory>
 #include <type_traits>
+#include <algorithm>
```
TFiFiE: The `#include`s are in alphabetical order.

gcp:
If nobody can think of a reason to have this check it should probably go. It may date back to there being Monte Carlo code (which needs some samples for an accurate winrate).

PR #1072, opened 2018-03-21
Remove out-parameters from network functions.

TFiFiE:
As these functions are never called with reused vectors, this change should be a no-brainer in terms of readability, safety and usability.

### DIFF
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -668,12 +668,11 @@ void convolve(const size_t outputs,
 template<unsigned int inputs,
          unsigned int outputs,
          bool ReLU,
-         size_t W, size_t B>
-void innerproduct(const std::vector<float>& input,
-                  const std::array<float, W>& weights,
-                  const std::array<float, B>& biases,
-                  std::vector<float>& output) {
-    assert(B == outputs);
+         size_t W>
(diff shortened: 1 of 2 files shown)


src/Network.cpp
```diff
@@ -912,14 +909,13 @@ Network::Netresult Network::get_scored_moves_internal(
 
     // Get the moves
     batchnorm<BOARD_SQUARES>(OUTPUTS_POLICY, policy_data, bn_pol_w1.data(), bn_pol_w2.data());
-    innerproduct<OUTPUTS_POLICY * BOARD_SQUARES, BOARD_SQUARES + 1, false>(policy_data, ip_pol_w, ip_pol_b, policy_out);
-    softmax(policy_out, softmax_data, cfg_softmax_temp);
-    auto& outputs = softmax_data;
+    const auto policy_out = innerproduct<OUTPUTS_POLICY * BOARD_SQUARES, BOARD_SQUARES + 1, false>(policy_data, ip_pol_w, ip_pol_b);
+    const auto outputs = softmax<BOARD_SQUARES + 1>(policy_out, cfg_softmax_temp);
```
gcp: I think this no longer needs to be a template function given that policy_out is sized correctly?

PR #1074, opened 2018-03-22
autogtp: fix sgf player names

marcocalignano:
This fixes a flow of leela zero.
If the weights file at command line is given with a relative path Leelaz will write the first 8 chars of the path and not of the file itself. This fixes it for autogtp generated files but actually I believe is a Leelaz flow.


gcp:
Wouldn't we rather fix the leelaz bug then?

marcocalignano:
I guess it would be better but I didn't look at the code.
I also think that this autogtp fix would work later even when Leelaz is fixed.

marcocalignano:
@gcp Ok then wait a min with this fix maybe I have a fix for leelaz

PR #1075, opened 2018-03-22
Leelaz: extract the right SGF player name

marcocalignano:
If the weigths parameter given at command line contains a path the real file name is extracted

### DIFF
--- a/src/SGFTree.cpp
+++ b/src/SGFTree.cpp
@@ -401,7 +401,13 @@ std::string SGFTree::state_to_string(GameState& pstate, int compcolor) {
     auto leela_name = std::string{PROGRAM_NAME};
     leela_name.append(" " + std::string(PROGRAM_VERSION));
     if (!cfg_weightsfile.empty()) {
-        leela_name.append(" " + cfg_weightsfile.substr(0, 8));
+        size_t pos = cfg_weightsfile.find_last_of("\\/");
+        if (std::string::npos == pos) {
+            pos = 0;
+        } else {
+            ++pos;
+        }
+        leela_name.append(" " + cfg_weightsfile.substr(pos, 8));
     }
 
     if (compcolor == FastBoard::WHITE) {


WhiteHalmos:
Does this work on both Windows and Linux?

src/SGFTree.cpp
```diff
@@ -401,7 +401,13 @@ std::string SGFTree::state_to_string(GameState& pstate, int compcolor) {
     auto leela_name = std::string{PROGRAM_NAME};
     leela_name.append(" " + std::string(PROGRAM_VERSION));
     if (!cfg_weightsfile.empty()) {
-        leela_name.append(" " + cfg_weightsfile.substr(0, 8));
+        size_t pos = cfg_weightsfile.find_last_of("\\/");
```
gcp: I remember we had a discussion about using ::basename here (from boost, because it's only in C++17). I'm not sure why that wasn't added, maybe because it requires an additional library to link with and so more project file changes?

The problem with the code above is that it's valid for a Unix path to have a \ in it.

marcocalignano: I found a similar code solution for windows and according to http://www.cplusplus.com/reference/string/string/find_last_of/ should work. But I guess that it is an easy test to do for anyone that has a windows machine, get my next branch compile leelaz start it and save the sgf file from the console command line and look at the name of the players.

gcp: The counterexample I gave is on Unix, not Windows.

It's legal to have a weights file named "leela\swa\1.txt".

Maybe we can just ignore that case though :-)

marcocalignano:
The PR says merged, but I do not see the commit when I fetch the `next` branch. Di I miss something?

Mardak:
Looks like this was merged on top of https://github.com/gcp/leela-zero/commit/adcb35ed1f721cfb973d507b3b1c3192aac6bd93 for #1064, but https://github.com/gcp/leela-zero/commit/94f48bb10b20d2267fcd3b9b0531f00c51907909 probably got force pushed instead causing https://github.com/gcp/leela-zero/commit/615d1bca62831d1a3bde75300f1596aa8f4906cb to disappear.

marcocalignano:
@gcp Do You need me to make another PR?

gcp:
I'll fix it up.

PR #1076, opened 2018-03-22
Trim the search tree during the search.

gcp:
In the master, do a check whether we are 3/4ths through our memory
budget. If so, attempt to do a depth first pruning of the search tree.

We try to get a lock on a node and check if there are any other CPUs
searching inside it. If so, we give up. If not, we prune every child
that has less than a given number of visits. These can be re-expanded by
the search later on. We find the right threshold by just doubling it
until the number of nodes in the tree drops below half our memory budget.

With a lot of threads active, we might not necessarily be able to prune
enough, especially with low maximum tree sizes, so we will still abort
if we reach the full memory budget.

This reworks the virtual loss code to accurately reflect the number of
active CPUs in a subtree.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -148,11 +148,11 @@ int UCTNode::get_move() const {
 }
 
 void UCTNode::virtual_loss() {
-    m_virtual_loss += VIRTUAL_LOSS_COUNT;
+    m_active_cpus++;
 }
 
 void UCTNode::virtual_loss_undo() {
-    m_virtual_loss -= VIRTUAL_LOSS_COUNT;
+    m_active_cpus--;
 }
 
 void UCTNode::update(float eval) {
@@ -180,7 +180,7 @@ float UCTNode::get_eval(int tomove) const {
     // Due to the use of atomic updates and virtual losses, it is
     // possible for the visit count to change underneath us. Make sure
     // to return a consistent result to the caller by caching the values.
-    auto virtual_loss = int{m_virtual_loss};
+    auto virtual_loss = m_active_cpus.load() * VIRTUAL_LOSS_COUNT;
     auto visits = get_visits() + virtual_loss;
     assert(visits > 0);
     auto blackeval = get_blackevals();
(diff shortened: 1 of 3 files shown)


gcp:
This need some testing, i.e. checking whether it is actually better than stopping the search when we hit MAX_TREE_SIZE.

There's going to be a zone between MAX_TREE_SIZE / 2 and MAX_TREE_SIZE where it prunes when the old code wouldn't, so a slight regression there would be expected.

WhiteHalmos:
I think the active CPU change is good regardless of trimming.

gcp:
Did I not just literally say it must be tested whether it's actually an improvement over stopping?

The key thing that prevents problems like you describe is that it does a strong global culling (of about 1/4th the total memory budget). This is likely to also cull some low visit entries near the critical path, which will have to be re-expanded. But because the cull is so large, they can get enough new node budget to survive the next cull (if indeed critical) and culling mostly removes low visit entries that didn't accumulate many visits between the last cull. The total visit information in those parents stays. This gives some LRU behavior. 

For extremely long searches, you can get situations where the freed up memory budget isn't enough to compete with the total visits of the tree part that is retained in memory. But because the trimming behavior of doubling thresholds until it reaches the memory target, I think it requires extreme/pathological cases to really break. 

But that is all theory. It must be tested.

If this doesn't work though, we'll need a depth first version of the search algorithm with a combined LRU/MFU cache, if we want an infinitely scaling search with limited memory. I don't see how to achieve it with a best first (UCT) implementation.

ihavnoid:
@gcp yes there seemed to be some race condition that I typed when there was no comment...

Basically what you are trying to do seems that on the next expansion after the culling we will expand the more promising side of the tree while 'forgetting' the less promising tree.  I am not totally sure but I think maybe we can make it work with heuristics (e.g., bias the culling to happen more on the less promising subtrees) but probably it's something to think of as the next step...

gcp:
Correct. It will probably stop scaling up strength at some point, but I'm hoping that point is pretty far away.

It seems depth-first reformulations of UCT do exist, although the algorithms don't exactly give identical  results. More invasive changes than this, of course.

gcp:
Pull request #1082 has actual data and shows that while this works, the conceptually simpler idea there works better.

PR #1078, opened 2018-03-22
Fix net_to_model.py to work with tfprocess changes.

ywrt:


gcp:
This didn't find the problem, see https://github.com/gcp/leela-zero/commit/2a5f2814b17068576b8b4cb937e5eb1ad85dd0bd#commitcomment-28232378.

I merged the SWA code in next, so remember to rebase.

PR #1079, opened 2018-03-22
Mutlieval internal use only

roy7:
Same pull but with --evals removed.

This could be adapted for "'heatmap average" and "heatmap max" since we'd already have a multi-eval feature in place.


gcp:
This code by itself serves no purpose.

roy7:
I was going to do this before doing 'heatmap average', but I'll just combine those together then. :)

PR #1083, opened 2018-03-23
Disable FPU reductions at a noised root.

gcp:
The Dirichlet noise that is added to self-play games was originally
tuned for a regular UCT search in the AGZ paper. With FPU reduction,
the possibility to discover a noised move is greatly reduced. So
we should not use FPU reduction at the place where we add noise, i.e. at
the root.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -209,9 +209,9 @@ void UCTNode::accumulate_eval(float eval) {
     atomic_add(m_blackevals, (double)eval);
 }
 
-UCTNode* UCTNode::uct_select_child(int color) {
+UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     UCTNode* best = nullptr;
-    auto best_value = -1000.0;
+    auto best_value = std::numeric_limits<double>::lowest();
 
     LOCK(get_mutex(), lock);
 
@@ -228,7 +228,13 @@ UCTNode* UCTNode::uct_select_child(int color) {
     }
 
     auto numerator = std::sqrt((double)parentvisits);
-    auto fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy);
+    auto fpu_reduction = 0.0f;
+    // Lower the expected eval for moves that are likely not the best.
+    // Do not do this if we have introduced noise at this node exactly
+    // to explore more.
+    if (!is_root || !cfg_noise) {
+        fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy);
+    }
     // Estimated eval for unknown nodes = original parent NN eval - reduction
     auto fpu_eval = get_net_eval(color) - fpu_reduction;
 
@@ -245,7 +251,7 @@ UCTNode* UCTNode::uct_select_child(int color) {
         auto denom = 1.0 + child->get_visits();
         auto puct = cfg_puct * psa * (numerator / denom);
         auto value = winrate + puct;
-        assert(value > -1000.0);
+        assert(value > std::numeric_limits<double>::lowest());
 
         if (value > best_value) {
             best_value = value;
--- a/src/UCTNode.h
+++ b/src/UCTNode.h
@@ -49,7 +49,7 @@ public:
     const std::vector<node_ptr_t>& get_children() const;
     void sort_children(int color);
     UCTNode& get_best_root_child(int color);
-    UCTNode* uct_select_child(int color);
+    UCTNode* uct_select_child(int color, bool is_root);
 
     size_t count_nodes() const;
     SMP::Mutex& get_mutex();
@@ -65,8 +65,6 @@ public:
     void set_score(float score);
     float get_eval(int tomove) const;
     float get_net_eval(int tomove) const;
-    double get_blackevals() const;
-    void accumulate_eval(float eval);
     void virtual_loss(void);
     void virtual_loss_undo(void);
     void update(float eval);
@@ -88,6 +86,8 @@ private:
     };
     void link_nodelist(std::atomic<int>& nodecount,
                        std::vector<Network::scored_node>& nodelist);
+    double get_blackevals() const;
+    void accumulate_eval(float eval);
 
     // Note : This class is very size-sensitive as we are going to create
     // tens of millions of instances of these.  Please put extra caution
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -142,7 +142,7 @@ SearchResult UCTSearch::play_simulation(GameState & currstate,
     }
 
     if (node->has_children() && !result.valid()) {
-        auto next = node->uct_select_child(color);
+        auto next = node->uct_select_child(color, node == m_root.get());
 
         if (next != nullptr) {
             auto move = next->get_move();


gcp:
Difference this makes at the root. Note that most positions seem less extreme. Also note that combined with -m 30, this will make the self-play games WAY more random and they will have WAY more blunders.

```
no noise, FPU
 
 R16 ->     334 (V: 47.02%) (N:  8.99%) PV: R16 D4 C3 D3 C4 C5 B5 B6 C6 D5 B7 B4
 Q17 ->     315 (V: 47.01%) (N:  8.56%) PV: Q17 D4 C3 D3 C4 C5 B5 B6 C6 D5 B7 B4
  D3 ->     314 (V: 46.96%) (N:  8.83%) PV: D3 Q16 R17 Q17 R16 R15 S15 S14 R14 Q15 S13 S16
  R4 ->     311 (V: 47.05%) (N:  8.18%) PV: R4 D16 C17 D17 C16 C15 B15 B14 C14 D15 B13
  Q3 ->     311 (V: 46.96%) (N:  8.82%) PV: Q3 D16 C17 C16 D17 E17 E18 F18 F17 E16 G18 D18
  C4 ->     309 (V: 46.98%) (N:  8.59%) PV: C4 Q16 R17 R16 Q17 P17 P18 O18 O17 P16 N18
 C16 ->     302 (V: 47.03%) (N:  8.07%) PV: C16 Q4 R3 Q3 R4 R5 S5 S6 R6 Q5 S7 S4
 D17 ->     287 (V: 47.03%) (N:  7.63%) PV: D17 Q4 R3 Q3 R4 R5 S5 S6 R6 Q5 S7 S4 T6 S3
 Q16 ->     193 (V: 46.62%) (N:  6.90%) PV: Q16 D4 C3 C4 D3 E3 E2 F2 F3 E4
  Q4 ->     190 (V: 46.62%) (N:  6.79%) PV: Q4 D16 C17 C16 D17 E17 E18 F18 F17
  D4 ->     167 (V: 46.60%) (N:  6.07%) PV: D4 Q16 R17 Q17 R16 R15 S15 S14 R14
 D16 ->     166 (V: 46.59%) (N:  6.01%) PV: D16 Q4 R3 Q3 R4 R5 S5 S6 R6
5.4 average depth, 15 max depth
2057 non leaf nodes, 1.56 average children
3200 visits, 1144299 nodes, 3199 playouts, 1403 n/s


noise, FPU

 R16 ->     333 (V: 47.05%) (N:  6.79%) PV: R16 D4 C3 C4 D3 E3 E2 F2 F3 E4 G2 D2
  C4 ->     330 (V: 47.08%) (N:  6.51%) PV: C4 Q16 R17 R16 Q17 P17 P18 O18 O17 P16 N18
  Q3 ->     328 (V: 47.05%) (N:  6.67%) PV: Q3 D16 C17 D17 C16 C15 B15 B14 C14 D15 B13 B16 A14
 Q17 ->     327 (V: 47.07%) (N:  6.48%) PV: Q17 D4 C3 C4 D3 E3 E2 F2 F3 E4 G2
  D3 ->     316 (V: 47.02%) (N:  6.72%) PV: D3 Q16 R17 R16 Q17 P17 P18 O18 O17 P16 N18 Q18
  R4 ->     302 (V: 47.04%) (N:  6.22%) PV: R4 D16 C17 D17 C16 C15 B15 B14 C14 D15 B13
 C16 ->     301 (V: 47.06%) (N:  6.12%) PV: C16 Q4 R3 Q3 R4 R5 S5 S6 R6 Q5 S7
 D17 ->     291 (V: 47.07%) (N:  5.80%) PV: D17 Q4 R3 Q3 R4 R5 S5 S6 R6 Q5 S7 S4 T6
  Q4 ->     184 (V: 46.70%) (N:  5.20%) PV: Q4 D16 C17 D17 C16 C15 B15 B14 C14 D15
 Q16 ->     168 (V: 46.57%) (N:  5.23%) PV: Q16 D4 C3 C4 D3 E3 E2 F2
 D16 ->     161 (V: 46.70%) (N:  4.55%) PV: D16 Q4 R3 R4 Q3 P3 P2 O2
  D4 ->     158 (V: 46.65%) (N:  4.63%) PV: D4 Q16 R17 R16 Q17 P17 P18 O18 O17
5.4 average depth, 14 max depth
2056 non leaf nodes, 1.56 average children
3200 visits, 1144259 nodes, 3199 playouts, 1467 n/s

noise, no FPU

  D3 ->     302 (V: 47.12%) (N:  6.72%) PV: D3 Q16 R17 Q17 R16 R15 S15 S14 R14 Q15 S13 S16
  C4 ->     286 (V: 47.10%) (N:  6.51%) PV: C4 Q16 R17 R16 Q17 P17 P18 O18 O17 P16 N18
 R16 ->     284 (V: 47.04%) (N:  6.79%) PV: R16 D4 C3 C4 D3 E3 E2 F2 F3 E4 G2 D2
  Q3 ->     283 (V: 47.06%) (N:  6.67%) PV: Q3 D16 C17 C16 D17 E17 E18 F18 F17 E16 G18
  R4 ->     279 (V: 47.12%) (N:  6.22%) PV: R4 D16 C17 C16 D17 E17 E18 F18 F17 E16 G18
 Q17 ->     276 (V: 47.07%) (N:  6.48%) PV: Q17 D4 C3 C4 D3 E3 E2 F2 F3 E4 G2 D2
 C16 ->     264 (V: 47.08%) (N:  6.12%) PV: C16 Q4 R3 Q3 R4 R5 S5 S6 R6 Q5 S7
 D17 ->     245 (V: 47.06%) (N:  5.80%) PV: D17 Q4 R3 R4 Q3 P3 P2 O2 O3 P4 N2
  Q4 ->     154 (V: 46.61%) (N:  5.20%) PV: Q4 D16 C17 D17 C16 C15 B15 B14
 Q16 ->     154 (V: 46.59%) (N:  5.23%) PV: Q16 D4 C3 C4 D3 E3 E2 F2 F3
 D16 ->     140 (V: 46.66%) (N:  4.55%) PV: D16 Q4 R3 R4 Q3 P3 P2 O2
  D4 ->     140 (V: 46.64%) (N:  4.63%) PV: D4 Q16 R17 R16 Q17 P17 P18 O18
  P4 ->       4 (V: 46.22%) (N:  0.14%) PV: P4 D16 C17 C16
  R3 ->       3 (V: 46.74%) (N:  0.11%) PV: R3 D16 C17
  E4 ->       3 (V: 46.61%) (N:  0.10%) PV: E4 Q16 R17
 R17 ->       3 (V: 46.60%) (N:  0.12%) PV: R17 D4 C3
  Q5 ->       3 (V: 46.59%) (N:  0.10%) PV: Q5 D16 C17
 C17 ->       3 (V: 46.58%) (N:  0.10%) PV: C17 Q4 R3
 E17 ->       3 (V: 46.56%) (N:  0.10%) PV: E17 Q4 R3
  D5 ->       3 (V: 46.55%) (N:  0.11%) PV: D5 Q16 R17
  C3 ->       3 (V: 46.53%) (N:  0.12%) PV: C3 Q16 R17
 Q15 ->       3 (V: 46.42%) (N:  0.13%) PV: Q15 D4 C3
 P16 ->       3 (V: 46.34%) (N:  0.12%) PV: P16 D4 C3
  O3 ->       3 (V: 45.95%) (N:  0.15%) PV: O3 R4 D3
 D15 ->       2 (V: 46.07%) (N:  0.11%) PV: D15 Q4
  P5 ->       2 (V: 45.33%) (N:  0.11%) PV: P5 D4
  F3 ->       2 (V: 45.19%) (N:  0.11%) PV: F3 C4
 P15 ->       2 (V: 44.93%) (N:  0.10%) PV: P15 Q4
  E5 ->       2 (V: 44.93%) (N:  0.11%) PV: E5 D16
 C14 ->       2 (V: 44.58%) (N:  0.13%) PV: C14 D17
  Q2 ->       2 (V: 44.19%) (N:  0.13%) PV: Q2 D16
 G10 ->       2 (V: 43.89%) (N:  0.09%) PV: G10 D4
 F12 ->       2 (V: 43.73%) (N:  0.10%) PV: F12 D16
 R18 ->       2 (V: 43.00%) (N:  0.09%) PV: R18 D4
  P2 ->       2 (V: 42.99%) (N:  0.12%) PV: P2 D16
 B17 ->       2 (V: 42.90%) (N:  0.10%) PV: B17 Q4
 M14 ->       2 (V: 42.60%) (N:  0.12%) PV: M14 Q16
 C18 ->       2 (V: 42.44%) (N:  0.10%) PV: C18 Q4
  F5 ->       2 (V: 41.90%) (N:  0.14%) PV: F5 D4
  K6 ->       2 (V: 41.83%) (N:  0.09%) PV: K6 D4
  O9 ->       2 (V: 40.68%) (N:  0.09%) PV: O9 Q4
 E15 ->       1 (V: 46.39%) (N:  0.06%) PV: E15 
  S3 ->       1 (V: 46.34%) (N:  0.07%) PV: S3 
  J6 ->       1 (V: 46.28%) (N:  0.08%) PV: J6 
  M6 ->       1 (V: 46.26%) (N:  0.07%) PV: M6 
 N11 ->       1 (V: 46.23%) (N:  0.07%) PV: N11 
 J13 ->       1 (V: 46.23%) (N:  0.08%) PV: J13 
  J7 ->       1 (V: 46.23%) (N:  0.07%) PV: J7 
 O11 ->       1 (V: 46.12%) (N:  0.08%) PV: O11 
  F9 ->       1 (V: 46.12%) (N:  0.08%) PV: F9 
 K12 ->       1 (V: 46.12%) (N:  0.07%) PV: K12 
 Q14 ->       1 (V: 46.05%) (N:  0.08%) PV: Q14 
  O4 ->       1 (V: 46.05%) (N:  0.08%) PV: O4 
 R14 ->       1 (V: 46.04%) (N:  0.09%) PV: R14 
  J8 ->       1 (V: 46.01%) (N:  0.06%) PV: J8 
 P18 ->       1 (V: 45.93%) (N:  0.07%) PV: P18 
 S15 ->       1 (V: 45.93%) (N:  0.09%) PV: S15 
 M12 ->       1 (V: 45.93%) (N:  0.06%) PV: M12 
  M8 ->       1 (V: 45.93%) (N:  0.06%) PV: M8 
```

jkiliani:
Agreed that this will increase blunders for the first 30 moves, but may well be worth it in the end, and certainly seems safer. Thanks for coding this, I'll do some tests as well.

A pity that FPU reduction looks a bust for Leela Chess at the moment in general. Every new net I tested there looked worse with FPU reduction than the last.

jkiliani:
I think this heatmap is actually quite telling. It's probably the effect of FPU reduction that we gradually trained any starting moves except the 4-4 and 4-3 points out of the network, from repeated feedback where other moves did not get any visits.

Eddh:
I'm a bit surprised, I thought the effect of the noise was stronger.

WhiteHalmos:
@Eddh The problem with relying on that noise is that with self-play's super tree-reuse (including opponent's tree), by the time a move is played it already had lots of visits, possibly even almost all its visits with FPU-reduction and no noise applied to the priors. Thus even though the noise is added, it will only affect the *new* visits after tree reuse.

gcp:
>I'm a bit surprised, I thought the effect of the noise was stronger.

The distribution is such that it occasionally peaks. But the occasionally is important :-)

jkiliani:
About the tree-reuse: Would it make sense to up the mixing factor of noise when a very large part of the tree is reused?

gcp:
It might make sense but without restarting everything multiple times you're not going to be able to find out the right factor.

Note that FPU reductions might well work with noise if you just increase the factor. But because of the above, I don't think it's *practical* for us.

Eddh:
@WhiteHalmos I agree. Personally I think black and white players should be kept separate, which would mitigate that by a lot and help game diversity, but well it would be slower and there is no data on it.

gcp:
>I think this move distribution is actually quite telling. It's probably the effect of FPU reduction that we gradually trained any starting moves except the 4-4 and 4-3 points out of the network, from repeated feedback where other moves did not get any visits.

Huh? Are you misreading the labels?

"no noise no FPU" would look very similar to "noise, no FPU", so what you say above is not true, at least not from the output given.

jkiliani:
I may have misinterpreted... can you tell me which net you used so I can check the heat map with no noise?

gcp:
>Thus even though the noise is added, it will only affect the new visits after tree reuse.

Note that this is OK, because it can start low policy moves in a positive feedback loop. Also note that noise limits the maximum policy score of moves to 75% (in practice), so there's always a few playouts to accumulate and get a new move going.

The remark in the other thread made me realize that FPU reductions mean that the noise parameters are likely not correct.

jkiliani:
> You can see in the example above that without FPU, a few very low policy moves can get a few (even if very few!) extra iterations. Not much, but enough to start boosting their score. Without FPU reductions, they keep getting pulled towards 0.

That is what I meant by the remark about 4-4 and 4-3. I think this happened to for example the 3-3 and 5-3 points already in the month since 0.12 was released.

gcp:
It also shows the very annoying behavior of MCTS to go depth-first when the score improves compared to the root, and bread-first when the score drops. FPU reductions fixed that largely :-/

Eddh:
Yeah that's why I think something like the average q value of child nodes or even the q value of parent would be a better starting point than current.

On the bright side the current behavior is more human like since we simulate emotions of panic inside the tree search.

gcp:
I think we tested q value of parent but it was worse?

jkiliani:
I think with the root-noise feedback fixed, FPU reduction as it currently is should be safe enough. Seems @amj was right in his hesitation to pull this for Minigo...

Eddh:
It was worse than fpu reduction but it was probably better than parent's net_eval (not tested directly but the winrate wasn't as bad against fpu reduction, and it was tested on weak nets and it was better if I remember right).

gcp:
>Seems @amj was right in his hesitation to implement it for Minigo...

There were a number of people arguing against it because it lowers exploration, but that's by itself is a terrible argument because so does changing PUCT, and the AGZ PUCT wasn't published. It should not matter as long as there's a mechanism to discover new moves.

I jumped on this because you very correctly pointed out the interaction at the root, i.e. that it "breaks" the mechanism because suddenly the used noise parameters are not necessarily close to optimal any more.

gcp:
Running a little test to see the strength impact.

killerducky:
@gcp what do you think about disabling tree reuse in self play so the noise is applied to root before playouts are done? NNCache should make it so performance is not degraded too badly. 

An alternative is to keep the tree but zero out m_visits/m_blackeval for either the whole tree or just the root.

gcp:
i.e. the whole discussion about tree reuse/noise everywhere etc is basically about how much playouts you want a noised move (or some noised moves) to be able to accumulate. As long as it's not 0 the procedure would work fine.

This whole pull is here because FPU *does* pull the noised moves to 0 much more (when using AGZ noise params).

Eddh:
Honestly I think it's very unclear if AGZ used tree reuse in self play games. If I'm in their place and I'm designing a system which learns of the MCTS distribution, I would make it have 1600 simulations per move, and that's all. Else the distribution would be very inconsistent depending on the preceding moves. Would the NN starts to look at previous moves to try to guess how much tree reuse was going on in order to better predict the distribution ?

Anyway, looking at the section talking about the training pipeline, the wording seems to me like they used specifially 1600 simulations per move, nothing else. 

> using 1,600 simulations of MCTS to select each move (this requires approximately 0.4s per search)

If you use tree reuse, you don't use 1600 simulations to select each move, you potentially use a lot more, that were played out previously.

gcp:
>can you tell me which net you used so I can check the heat map with no noise?

FWIW this was 1ccb.

gcp:
>Honestly I think it's very unclear if AGZ used tree reuse in self play games. If I'm in their place and I'm designing a system which learns of the MCTS distribution, I would make it have 1600 simulations per move, and that's all.

But that's demonstrably very inefficient and losing work, so why do that?

"The search tree is reused at subsequent time-steps: the child node corresponding to the played action becomes the new root node; the subtree below this child is retained along with all its statistics, while the remainder of the tree is discarded."

"using 1,600 simulations for each MCTS" doesn't seem to imply anywhere that you don't do them on the reused tree.

jkiliani:
So do you think it's more likely they actually went with 1600 visits, or that they really did 1600 new playouts per move? Agreed that this confirms they had tree reuse, I wasn't so sure since they have the compute to waste...

Eddh:
I explain why : 
> Else the distribution would be very inconsistent depending on the preceding moves. Would the NN starts to look at previous moves to try to guess how much tree reuse was going on in order to better predict the distribution ?

Apart from that, DM obviously are less concerned about efficiency than us since they can throw 5000 TPU at their problems. They didn't even bother with transpositon tables for the 20 block version. It seems to indicate they went for the "make it work, then make it fast" way.

Also, I don't really get how they would tune their noise value if the result is also inconsistent depending on how much the tree has been reused. And it has effects on the shape of the distribution which will be flatter when the tree is "fresh" because of dirichlet noise. I'm not an expert but does that mean the net will potentially try to make flatter predictions if it can figure out that previous moves led to not much tree reuse ? It seems possible to me. And it seems bad if part of the nets are trying to figure that out instead of truly finding good moves.

The parts you are quoting are in a different  section of the AGZ paper which potentially refers more to the true matches with long time settings than the self play games.

gcp:
>So do you think it's more likely they actually went with 1600 visits, or that they really did 1600 new playouts per move?

I have no idea. The paper isn't very clear on that.

>Apart from that, DM obviously are less concerned about efficiency than us since they can throw 5000 TPU at their problems.

They did implement resigning etc. And they implemented tree reuse.

>Also, I don't really get how they would tune their noise value if the result is also inconsistent depending on how much the tree has been reused.

It's not even clear how to tune it even if you don't have this inconsistency. If they just repeated preliminary experiments until they found a good value, the inconsistency doesn't really matter.

Eddh:
They probably didn't use visits. Again :  

> using 1,600 simulations of MCTS to select each move (this requires approximately 0.4s per search)

Why mention it requires 0.4s per search if it is not a somewhat consistent value ? If it requires 0.01s when the tree is reused and 2 seconds when it is not it is not very useful to mention the average value, and they say "approximately 0.4", which is specifically around 0.4 and not far away.

gcp:
I agree. "1600 simulations of MCTS" also wouldn't be true in that case.

WhiteHalmos:
I am starting some resign analysis with current next and this PR.

odeint:
So will there be a new binary release due to this? Is it enough if a few `next`-clients are in the population to nudge the learning the right way?

gcp:
It's probably good to do a new release soon, also because there are some nice fixes in "next". But it could use some "bake" time.

Indeed because a lot of clients are using next anyway, the more diverse data will start appearing in the training window.

jkiliani:
Technical question: I just tried to port the modified FPU reduction to LCZero, but it won't compile:
```
UCTSearch.cpp:82:66: error: no member named 'get' in 'UCTNode'
        auto next = node->uct_select_child(color, node == m_root.get());
                                                          ~~~~~~ ^
1 error generated.
```
Obviously I overlooked something, seems the .get() is undefined. Generally, UCTNode.cpp and UCTSearch.cpp look very different between Leela Zero and Leela Chess, seems there were a lot of improvements to LZ's search code made after it was ported. What do I have to fix here?

alreadydone:
New jobs on Big Red II will be running 'next'. It takes 12 hours for old jobs to finish though. Back to ~180 GPUs today, after a busy week following the maintenance.

gcp:
```
leelaz-next v leelaz-no-noise (179/1000 games)
board size: 19   komi: 7.5
leelaz-next         60 33.52%       26 28.89%     34 38.20%    315.13
leelaz-no-noise    119 66.48%       55 61.80%     64 71.11%    302.60
                                    81 45.25%     98 54.75%
```

About 100 Elo loss from this. I think most of this isn't so much the loss of FPU reductions at the root, but "-m 30" meaning that on every move there is now a higher probability of playing a bad move.

I think we have to take this hit though. It's not like the network is making a lot of progress or that the larger ones are clearly better.

Eddh:
I've thought about something. I've tried so much to have something that doesn't panic when winrate drop, but in the bigger context of mcts, a node where winrate is dropping, causing "search panic", will explore a lot of bad moves, which will lower its q value very fast, discouraging the exploration. If you do the sensible thing and exploit the best available node, the value of the node will not drop as fast, and it will likely still be considered by its parent. So in a way, this panicking behavior help converge on nodes where winrate is rising or stagnating instead of dropping. So everything I did trying to stop that was bound to have a hard time being successful. That's probably why even simple parent_eval, was equivalent or better than everything I've tried in terms of winrate.

There is probably some kind of balance to be found between searching for the best move, and discouraging exploration of "winrate drop" nodes. I think I'll avoid trying to search that. The root cause of the problem is the backup operator. I'm not sure what to do about that yet. So in the meantime i will probably go back to trying simpler things based on parent's net_eval. I'll do my tests on LZChess since I can do more test games and verify if the base ideas work.

PR #1087, opened 2018-03-23
Only allocate for untrimmed children.

TFiFiE:


src/UCTNode.cpp
```diff
@@ -140,9 +140,17 @@ void UCTNode::link_nodelist(std::atomic<int>& nodecount,
         } else {
             min_psa = max_psa * 0.001f;
         }
-    }
 
-    m_children.reserve(nodelist.size());
+        auto children = 0;
+        for (const auto& node : nodelist) {
+            if (node.first >= min_psa) {
```
gcp: Pff, having to add this loop just to presize correctly rather sucks. What about only keeping the else branch here? Or doing a swap trick afterwards if min_psa > 0.0f?

PR #1089, opened 2018-03-23
Add 'average' option to heatmap gtp command

roy7:
Also uses alias 'avg', ~~and implements the code for multiple eval.~~ (Full 8 evals used by 'heatmap average'.)

~~Renamed the ensemble type to MULTI_AVG with the thought that I or someone else may eventually add MULTI_MAX as an eval option, to provide the max() of each action's value instead of the average. To allow for "heatmap max".~~


src/GTP.cpp
```diff
@@ -516,15 +516,27 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
 
         if (!cmdstream.fail()) {
-            auto vec = Network::get_scored_moves(
-                &game, Network::Ensemble::DIRECT, rotation, true);
-            Network::show_heatmap(&game, vec, false);
+            if (rotation == "all") {
+                for (int r=0; r<8; r++) {
+                    auto vec = Network::get_scored_moves(
+                        &game, Network::Ensemble::DIRECT, r, true);
+                    Network::show_heatmap(&game, vec, false);
```
gcp: These two lines are almost the same in all branches, and can be pulled out.

src/Network.cpp
```diff
@@ -860,9 +861,40 @@ Network::Netresult Network::get_scored_moves(
...
+    } else if (ensemble == MULTI_AVG && rotation > 1) {
+        // MULTI_AVG rotation is the # of random rotations to provide
+        assert(rotation >= 1 && rotation <= 8);
+        std::vector<int> r_list{0, 1, 2, 3, 4, 5, 6, 7};
+        // Need to account for extra 2 in m_squaresize.
+        float policy[(BOARD_SIZE + 2) * (BOARD_SIZE + 1)] = {0};
+
+        if (rotation < 8) {
+            std::random_device rd;
+            std::mt19937 g(rd());
```
gcp: This is completely unneeded, we already do a lot of work to correctly initialize a suitable RNG. IIRC you can see how to use it with shuffle in Training.cpp.

src/Network.cpp
```diff
@@ -860,9 +861,40 @@ Network::Netresult Network::get_scored_moves(
     if (ensemble == DIRECT) {
         assert(rotation >= 0 && rotation <= 7);
         result = get_scored_moves_internal(state, planes, rotation);
+    } else if (ensemble == MULTI_AVG && rotation > 1) {
+        // MULTI_AVG rotation is the # of random rotations to provide
+        assert(rotation >= 1 && rotation <= 8);
+        std::vector<int> r_list{0, 1, 2, 3, 4, 5, 6, 7};
+        // Need to account for extra 2 in m_squaresize.
+        float policy[(BOARD_SIZE + 2) * (BOARD_SIZE + 1)] = {0};
```
gcp: Use a std::vector or (at worst) std::array. There should already be constants for the policy array size.

src/Network.cpp
```diff
@@ -860,9 +861,40 @@ Network::Netresult Network::get_scored_moves(
     if (ensemble == DIRECT) {
         assert(rotation >= 0 && rotation <= 7);
         result = get_scored_moves_internal(state, planes, rotation);
+    } else if (ensemble == MULTI_AVG && rotation > 1) {
+        // MULTI_AVG rotation is the # of random rotations to provide
+        assert(rotation >= 1 && rotation <= 8);
+        std::vector<int> r_list{0, 1, 2, 3, 4, 5, 6, 7};
```
gcp: Not clear why this is needed?

src/Network.cpp
```diff
@@ -860,9 +861,40 @@ Network::Netresult Network::get_scored_moves(
...
+
+            for (auto res : tmpresult.first) {
+                // +1 to account for PASS
+                policy[res.second + 1] += res.first;
+            }
+        }
+
+        for (size_t c = 0; c < result.first.size(); c++) {
+            result.first[c].first += policy[result.first[c].second + 1];
+            result.first[c].first /= rotation;
```
gcp: Dividing by rotation looks suspect.

src/Network.cpp
roy7: If we're going to average 2-7 rotations we need to choose them randomly without replacement. I thought the simple way to handle that was shuffle a list of index numbers 0-7 then walk that list in order for the # of rotations requested. Is there a better way I overlooked? I didn't want 3 rotations to always be 0,1,2.

roy7: Although I did always assume 'heatmap' by itself choose a random one that stuck around, and wasn't tied to 'heatmap 0'.

src/GTP.cpp
roy7: I'll see what I can do here. In one branch we need to show 8 heatmaps so I do need that loop. In the other 3 of the 4 branches the show_heatmap command is the same. Each of the 4 branches uses a different get_scored_moves options.

roy7: I re-worked this. I hope it's what you had in mind. :)

src/Network.cpp
roy7: Or did you mean since we're only doing rotation=8 for 'heatmap average' we don't need support for 2-7 and should delete that until someone needs it one day?

src/Network.cpp
roy7: The closest constant I've found so far from FastBoard.h is `MAXSQ = ((BOARD_SIZE + 2) * (BOARD_SIZE + 2))`. I'll keep looking.

In get_scored_moves_internal I see `std::vector<float> policy_data(OUTPUTS_POLICY * width * height)` which is much larger since OUTPUTS_POLICY is 2

roy7: Strange how the simple change to std::vector has broken this in confusing ways. Will try to clean it up, sorry about pushing a commit with non-working code. My lack of C++ experience is showing.

src/Network.cpp
Hersmunch: As far as I can see it looks like it is taking the average of the random rotations but it isn't clear from the name of the variable.

roy7: I did use the variable two ways. For DIRECT rotation is the rotation #. For MULTI_AVG rotation is the # of rotations. Maybe that was silly of me.

WhiteHalmos: I think it would be clearer if we just used 8 everywhere. 😊

src/Network.cpp
gcp: No it just wasn't clear to me what the array was there for. Looks like std::sample is only in C++17 and it's not in boost, so it indeed seems the best way to do this.

src/Network.cpp
gcp: Yeah same variable different meaning is pretty evil :-)

roy7: Should I make a new variable earlier on with a different name for readability?

src/GTP.cpp
```diff
@@ -516,18 +516,30 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
+            if (cmdstream.fail()) {
+                vec = Network::get_scored_moves(
+                    &game, Network::Ensemble::DIRECT, 0, true);
+            } else if (rotation == "average" || rotation == "avg") {
+                vec = Network::get_scored_moves(
+                    &game, Network::Ensemble::MULTI_AVG, 8, true);
+            } else {
+                vec = Network::get_scored_moves(
+                    &game, Network::Ensemble::DIRECT, std::stoi(rotation), true);
+            }
```
TFiFiE: Can we get some invalid-argument handling here? Right now, anything that's not a number between 0 and 8 crashes the program, either with an uncaught exception or a segmentation fault.

roy7: It's always been that way. I thought about trying to handle it but wasn't sure on the best way to do so. Any suggestions?

TFiFiE: I don't like ```std::stoi``` throwing an exception, as it makes handling both nonintegral as well as out-of-bounds arguments unwieldy. I'd say attempt the conversion through a ```std::stringstream```, as the validity test can then be done with a single ```if``` statement.

src/Network.cpp
```diff
@@ -860,9 +860,38 @@ Network::Netresult Network::get_scored_moves(
...
+        }
+
+        result.second /= rotation;
     } else {
-        assert(ensemble == RANDOM_ROTATION);
-        assert(rotation == -1);
+        assert((ensemble == RANDOM_ROTATION && rotation == -1) ||
+               (ensemble == MULTI_AVG && rotation == 1));
```
gcp: This looks wrong to me: MULTI_AVG is handled in the other branch, so if ensemble is MULTI_AVG here something is seriously broken.

src/GTP.cpp
```diff
@@ -516,18 +516,30 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
+            for (int r=0; r<8; r++) {
+                vec = Network::get_scored_moves(
+                    &game, Network::Ensemble::DIRECT, r, true);
+                Network::show_heatmap(&game, vec, false);
+            }
         } else {
-            auto vec = Network::get_scored_moves(
-                &game, Network::Ensemble::DIRECT, 0, true);
+            if (cmdstream.fail()) {
+                vec = Network::get_scored_moves(
```
gcp: Just set rotation = 0 and merge this with the else branch?

src/Network.cpp
gcp: I'd use num_rotations or rotations_cnt or similar when "rotation" is really "number of rotations".

src/GTP.cpp
roy7: Done

src/Network.cpp
Mardak: It's true that MULTI_AVG with 1 rotation is effectively picking out a single rotation just like RANDOM_ROTATION. But what's the intent of allowing a subset of random rotation / `num_rotations` to be averaged? Why not always just average all 8? This would simplify the MULTI_AVG branch if `num_rotations` is assumed to be 8.

Mardak:
Not directly changed in this PR, but turns out running the heatmap command influences the next `genmove` because the result gets cached:
https://github.com/gcp/leela-zero/blob/97c2f8137a3ea24938116bfbb2b0ff05c83903f0/src/Network.cpp#L870-L871

This means after running `heatmap avg`, the root node will use the averaged policy scored moves when exploring. For example when I analyzed https://online-go.com/game/11952774 with `0f6a8c5f #102`, the N ranged from 0.67 to 65.93% as shown running with the seeds below https://github.com/gcp/leela-zero/issues/877#issuecomment-373255981

```
expect: T7, move: 129, who: b, weight: 0f6a8c5f, visit: 10, no heatmap
rotation: 0, seed: 22  T7 ->       8 (V: 100.00%) (N: 25.80%) PV: T7 B10 C10 B9 G13
rotation: 1, seed:  6  B9 ->       7 (V:  2.86%) (N: 43.94%) PV: B9 B8 C9 E8 B4 B5
rotation: 2, seed:  2  T7 ->       8 (V: 100.00%) (N: 19.96%) PV: T7 B10 C10 B9 G13
rotation: 3, seed: 11  T7 ->       8 (V: 100.00%) (N: 18.21%) PV: T7 B10 C10 B9 G13
rotation: 4, seed:  3  T7 ->       8 (V: 100.00%) (N: 13.37%) PV: T7 B10 C10 B9 G13 F12
rotation: 5, seed:  4  B9 ->       9 (V:  2.20%) (N: 49.14%) PV: B9 B8 C9 E8 D8 D7
rotation: 6, seed: 12  T7 ->       9 (V: 100.00%) (N: 65.93%) PV: T7 B10 C10 B9 G13 F12 F11
rotation: 7, seed:  1  T7 ->       8 (V: 100.00%) (N: 17.64%) PV: T7 B10 C10 B9 B16 B17
```
And here's the result when doing `heatmap avg` just before `genmove b`:
```
expect: T7, move: 129, who: b, weight: 0f6a8c5f, visit: 10, heatmap avg
rotation: 0, seed: 22  T7 ->       8 (V: 100.00%) (N: 20.93%) PV: T7 B10 C10 B9 G13 F12 H14 H15
rotation: 1, seed:  6  T7 ->       8 (V: 100.00%) (N: 20.93%) PV: T7 B10 C10 B9 B16
rotation: 2, seed:  2  T7 ->       8 (V: 100.00%) (N: 20.93%) PV: T7 B10 C10 B9 G13
rotation: 3, seed: 11  T7 ->       8 (V: 100.00%) (N: 20.93%) PV: T7 B10 C10 B9 G13 F12
rotation: 4, seed:  3  T7 ->       8 (V: 100.00%) (N: 20.93%) PV: T7 B10 C10 B9 B16 B17
rotation: 5, seed:  4  T7 ->       8 (V: 100.00%) (N: 20.93%) PV: T7 B10 C10 B9 G13
rotation: 6, seed: 12  T7 ->       8 (V: 100.00%) (N: 20.93%) PV: T7 B10 C10 B9 G13 F12
rotation: 7, seed:  1  T7 ->       8 (V: 100.00%) (N: 20.93%) PV: T7 B10 C10 B9 G13
```

So @roy7, that's a simple way to get RoyalZero to analyze the next move using the average policy. Although notice that the PVs are different because only the policy of the root was averaged. (Although maybe that won't be entirely what you want anyway as typically RoyalZero's "one bad rotation" failure is from not realizing the opponent could capture, etc.)

Not sure if this behavior is intended as one would perhaps expect a "display-only" heatmap command to not influence other things.

src/Network.cpp
roy7: I guess I just didn't want to tear the support for any # of rotations out since I bothered to write it originally lol. :)

roy7: I updated this to use a num_rotations variable and so forth to try and make it more clear. I didn't seem to update this review automatically, but some changes have been made. Let me know if you need anything further?

roy7:
Hmm I missed that something in Network.cpp changed which broke this. Will fix it.

roy7:
Added some partial fixes. NetResult struct changed and it needs some rewrite in my PR in a section of code I didn't originally write myself, so trying to figure it out.

roy7:
@WhiteHalmos I might need you to swoop in and fix this all again, I think you did this area of the code originally, haha.

src/Network.cpp
```diff
@@ -862,9 +862,41 @@ Network::Netresult Network::get_scored_moves(
...
+
+        result = get_scored_moves_internal(planes, r_list[0]);
+
+        for (int r = 1; r < num_rotations; ++r) {
+            Netresult tmpresult = get_scored_moves_internal(planes, r_list[r]);
+            result.winrate += tmpresult.winrate;
+            result.policy_pass += tmpresult.policy_pass;
+
+            for (auto res : tmpresult.first) {
+                policy[res.second] += res.first;
```
Mardak: Pretty sure you'll need to do something like below as the original vertex as `second` is stored now an implicit vector index.
```c++
for (auto idx = size_t{0}; idx < BOARD_SQUARES; idx++) {
  policy[idx] += tmpresult.policy[idx];
}
```

src/Network.cpp
```diff
@@ -862,9 +862,41 @@ Network::Netresult Network::get_scored_moves(
...
         assert(rotation >= 0 && rotation <= 7);
         result = get_scored_moves_internal(planes, rotation);
+    } else if (ensemble == MULTI_AVG && rotation > 1) {
+        // MULTI_AVG rotation is the # of random rotations to provide
+        int num_rotations = rotation;
+
+        assert(num_rotations >= 1 && num_rotations <= 8);
+        std::vector<int> r_list{0, 1, 2, 3, 4, 5, 6, 7};
+        // Need to account for extra 2 in m_squaresize.
+        std::vector<float> policy((BOARD_SIZE + 2) * (BOARD_SIZE + 1));
```
Mardak: I don't think a separate `policy` is needed to accumulate as you could just use the one from the first rotation that you store as `result` (so use `result.policy`) ?

roy7:
Thanks to @Mardak I think this is ported and functional again.

src/Network.cpp
roy7: Fixed!

roy7:
@gcp If the official recommendation is I drop variable # of evals support and hard code it for 8 I'll do that, just let me know.

src/GTP.cpp
```diff
@@ -515,18 +515,27 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
+            }
         } else {
-            auto vec = Network::get_scored_moves(
-                &game, Network::Ensemble::DIRECT, 0, true);
+            if (!cmdstream.fail() && (rotation == "average" || rotation == "avg")) {
+                vec = Network::get_scored_moves(
+                    &game, Network::Ensemble::MULTI_AVG, 8, true);
+            } else {
+                vec = Network::get_scored_moves(
+                    &game, Network::Ensemble::DIRECT, cmdstream.fail() ? 0 : std::stoi(rotation), true);
```
gcp: I'd move the ternary operator out of the function call for readability.

gcp: Also this is really the only case that must handle cmdstream.fail(), so move that case out of this if and move it to the top? Then all further can assume that it did not fail.

src/Network.cpp
```diff
@@ -862,9 +862,38 @@ Network::Netresult Network::get_scored_moves(
     if (ensemble == DIRECT) {
         assert(rotation >= 0 && rotation <= 7);
         result = get_scored_moves_internal(planes, rotation);
+    } else if (ensemble == MULTI_AVG && rotation > 1) {
+        // MULTI_AVG rotation is the # of random rotations to provide
+        int num_rotations = rotation;
```
gcp: There doesn't seem to be a use case for averaging multiple (instead of all) rotations, so you can simplify the code by removing support for this. That also avoids the awkward overload of "rotation" having multiple meanings.

src/Network.cpp
```diff
@@ -862,9 +862,38 @@ Network::Netresult Network::get_scored_moves(
...
+
+        assert(num_rotations >= 1 && num_rotations <= 8);
+        std::vector<int> r_list{0, 1, 2, 3, 4, 5, 6, 7};
+
+        if (num_rotations < 8) {
+            std::shuffle(r_list.begin(), r_list.end(), Random::get_Rng());
```
gcp: This kind of thing also becomes unneeded.

src/Network.cpp
gcp: I'm with @Mardak on this one in that supporting multiple rotations averaged complicates the code for no gain, so I'd just get rid of that.

gcp:
github is a bit confused and seems to think your conflict fixes were addressing the review remarks.

roy7:
@gcp So to confirm, remove support for #s 2-7 and just hard code it as 8 all the time?

src/Network.cpp
```diff
@@ -891,11 +891,28 @@ Network::Netresult Network::get_scored_moves(
...
+
+        for (auto idx = size_t{0}; idx < BOARD_SQUARES; idx++) {
+            result.policy[idx] /= 8;
+        }
+
+        result.winrate /= 8;
+        result.policy_pass /= 8;
     } else {
-        assert(ensemble == RANDOM_SYMMETRY);
+        assert(ensemble == RANDOM_ROTATION);
```
Mardak: This looks like a merge conflict fail. This branch probably should be left unchanged

PR #1090, opened 2018-03-24
Clean up the SWA code somewhat.

ywrt:
1. Bugfix: Batch norms weren't being updated.
2. Bugfix: Use 400 batches when recomputing batch norms to reduce error is ~2% (from 13%).
3. Remove the need for external files: SWA is now accumulated in RAM.
4. Change TF restore to be optimistic. This allows restoring from snapshots
that don't have the SWA variable (and vice versa).

NB: I've verified by hand that the batch norms vars are being updated. If I get a little spare time later I'll try and add some unit tests for that.

### DIFF
--- a/training/tf/tfprocess.py
+++ b/training/tf/tfprocess.py
@@ -21,7 +21,6 @@ import numpy as np
 import time
 import tensorflow as tf
 from shutil import copyfile
-from average_weights import swa
 
 def weight_variable(shape):
     """Xavier initialization"""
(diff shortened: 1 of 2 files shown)


Ttl:
Much cleaner implementation than I did. It really shows that I'm not too familiar with Tensorflow.

When calculating the error in batch norm weights, you need to also take into account that the previous averaged network already had the correct batch norms and the new network being added only slightly changes the weights. In 8000 steps the network weights don't also have that much time to change and the weights are very close to the previous output nets. If the error is 10% in the beginning 200 steps should be enough to bring it to 1%.

training/tf/tfprocess.py
```diff
@@ -434,38 +441,57 @@ def construct_net(self, planes):
...
+        swa_path = path + "-swa-" + str(num) + "-" + str(steps) + ".txt"
 
+        # save the current network.
+        self.snap_save()
+        # Copy the swa weights into the current network.
+        self.session.run(self.swa_load_op)
+        if self.swa_recalc_bn:
             print("Recalculating SWA batch normalization")
-            for _ in range(200):
+            self.update_ops = tf.get_collection(tf.GraphKeys.UPDATE_OPS)
```
Ttl: Does this need to be redefined? It's already defined once in init_net?

ywrt:
> When calculating the error in batch norm weights, you need to also take into account that the previous averaged network already had the correct batch norms and the new network being added only slightly changes the weights.

Now that I think about this, I'm of the opinion that recalculating the batch norms is the wrong thing to do.

My logic: 
1. Assuming that learning is converged, then the weights (inc unsmoothed batch norms) are [sampling](https://arxiv.org/abs/1704.04289) from a Gaussian, with assumed standard deviations of *s1*.
2. The batch norms are smoothed with an IIR, alpha=1/100. These form an estimator for the true value of the norms, with std dev = *s1*/sqrt(100) = *s1*/10
3. SWA samples the weights every 8000 batches, and is again smoothed with an IIR, alpha=1/16. This is likewise an estimator for the norms, with std dev = *s1*/sqrt(1600) = s1/40.

But: If we now run batches through the batch norm layer, those batches will be smoothed with alpha=1/100 again, meaning that the variance (error) in the estimate will increase!

NB: Of course, this makes the (bold!) assumption that the Gaussian we're sampling from is invariant, and that we're taking unbiased samples. But given that those are the same assumptions the SWA paper makes, it seems reasonable?

If we actually want the error to reduce, then we need to make sure that SWA batch norm recalculation is run with a strong filter. e.g. at least alpha=1/1600.

But in that case, if SWA is sampling every 8000 batches, then why not run the original filter with a stronger alpha (i.e batch norm momentum higher than 0.99) with would give us the same result at lower cost?

It doesn't seem to ever make sense to do the batch norm recalculation under the SWA assumptions?

Or to put it another way: If this actually helps, it's strongly implies that the SWA assumptions don't hold. That is, that the Gaussian is not invariant, and/or the samples are biased in some fashion.

training/tf/tfprocess.py
ywrt: thanks :) fixed.

PR #1092, opened 2018-03-24
UCTNode memory optimization

ihavnoid:
Since the last attempt was closed without merging, this time I am proposing a less intrusive (but possibly a bit controversial, IMHO) approach.  What I did was replaced the std::vector inside the UCTNode with a plain-old array of pointer.  This saves about 16 bytes from sizeof(UCTNode) which can mean a quite a bit of memory footprint.  This makes future code maintenance a bit more confusing at the cost of lower memory footprint.

Memory footprint while generating first move with 1 million playouts:
- 12 threads, 2 GPUs
- long enough time limit, large enough playout limit to hit 1M playout per move

Before change : saturate at 8.95GB
After change : saturate at 7.49GB

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -142,20 +142,19 @@ void UCTNode::link_nodelist(std::atomic<int>& nodecount,
         }
     }
 
-    m_children.reserve(nodelist.size());
+    m_children = std::make_unique<node_ptr_t[]>(nodelist.size());
+    auto cnt = size_t{0};
     for (const auto& node : nodelist) {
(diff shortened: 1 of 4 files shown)


gcp:
Tiny gain isn't worth the resulting ugliness, I'm afraid.

ihavnoid:
Yes after submitting the patch I figured that we need something bigger, and I started coding something similar to the previous attempt.  Will submit patch once code seems healthy.

ihavnoid:
(When I meant previous attempt, I meant #887)

PR #1097, opened 2018-03-25
UCTNode lazy initialization for memory footprint reduction (attempt 3 of "Reduce the memory consumption of nodes")

ihavnoid:
- When we expand children nodes, we don't actually create the UCTNode but only keep a proxy of it.
- The actual child node is created (called 'expanded' here) only if it is selected from uct_select_child
  - (One more case - for ease of implementation, the root's children are always expanded)
- As of this commit, we saturate around 3.45GB memory footprint when we hit MAX_TREE_SIZE (down from 8.9GB)

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -280,7 +280,7 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     }
 
     assert(best != nullptr);
-    best->expand();
+    best->inflate();
     return best->get();
 }
 
@@ -318,7 +318,7 @@ UCTNode& UCTNode::get_best_root_child(int color) {
 
     auto ret = std::max_element(begin(m_children), end(m_children),
                                 NodeComp(color));
-    ret->expand();
+    ret->inflate();
     return *(ret->get());
 }
 
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -145,16 +145,14 @@ void UCTNode::link_nodelist(std::atomic<int>& nodecount,
     m_children.reserve(nodelist.size());
     for (const auto& node : nodelist) {
         if (node.first < min_psa) continue;
-        m_children.emplace_back(
-            std::make_unique<UCTNode>(node.second, node.first)
-        );
+        m_children.emplace_back(node.second, node.first);
     }
 
     nodecount += m_children.size();
     m_has_children = true;
 }
 
-const std::vector<UCTNode::node_ptr_t>& UCTNode::get_children() const {
+const std::vector<UCTNodePointer>& UCTNode::get_children() const {
     return m_children;
 }
 
@@ -226,7 +224,7 @@ void UCTNode::accumulate_eval(float eval) {
 }
 
 UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
-    UCTNode* best = nullptr;
+    UCTNodePointer * best = nullptr;
     auto best_value = std::numeric_limits<double>::lowest();
 
     LOCK(get_mutex(), lock);
@@ -235,10 +233,10 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     auto total_visited_policy = 0.0f;
     auto parentvisits = size_t{0};
     for (const auto& child : m_children) {
-        if (child->valid()) {
-            parentvisits += child->get_visits();
-            if (child->get_visits() > 0) {
-                total_visited_policy += child->get_score();
+        if (child.valid()) {
+            parentvisits += child.get_visits();
+            if (child.get_visits() > 0) {
+                total_visited_policy += child.get_score();
             }
         }
     }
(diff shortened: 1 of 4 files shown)


src/UCTNode.h
```diff
@@ -24,20 +24,97 @@
...
+
+// WARNING : expand() is not thread-safe and hence has to be protected by an external lock.
+
+class UCTNodePointer {
+private:
+    // the raw storage used here.
+    // if bit 0 is 0, m_data is the actual pointer.
+    // if bit 0 is 1, bit [31:16] is the vertex value, bit [63:32] is the score.
+    // (I really wanted to use C-style bit fields and unions, but those aren't portable)
+    mutable uint64_t m_data = 1;
```
Hersmunch: Is the reason for the use of bit-fields and unions not being portable because this "is expanded" flag's bit location is implementation-specific and/or are there other reasons?
Would storing the flag "is expanded" elsewhere (e.g. packed in a bit-field with `m_is_expanding`) allow your desired approach to be used here?

ihavnoid: Yeah, the reason was that bit positions are not portable.  I couldn't find a suitable library on standard C++ nor it seemed worth bringing in another external dependency for such a small amount of code.  Any suggestions to make it prettier are welcome :)

Hersmunch: Thanks for confirming my understanding. 
I can't tell from your reply whether you have considered:
> Would storing the flag "is expanded" elsewhere (e.g. packed in a bit-field with m_is_expanding) allow your desired approach to be used here?

src/UCTNode.cpp
```diff
@@ -311,16 +310,20 @@ UCTNode& UCTNode::get_best_root_child(int color) {
     LOCK(get_mutex(), lock);
     assert(!m_children.empty());
 
-    return *(std::max_element(begin(m_children), end(m_children),
-                              NodeComp(color))->get());
+    auto ret = std::max_element(begin(m_children), end(m_children),
+                              NodeComp(color));
```
Hersmunch: Indentation (sorry I missed these the first time around).

src/UCTNode.cpp
```diff
@@ -311,16 +310,20 @@ UCTNode& UCTNode::get_best_root_child(int color) {
...
+    return *(ret->get());
 }
 
 size_t UCTNode::count_nodes() const {
     auto nodecount = size_t{0};
     if (m_has_children) {
         nodecount += m_children.size();
         for (auto& child : m_children) {
-            nodecount += child->count_nodes();
+            if(child.get_visits() > 0) {
```
Hersmunch: Missing space

src/UCTNode.h
ihavnoid: Elsewhere... means where?  Adding a separate field ("bool m_is_expanding") on the UCTNodePointer will add another byte of memory on the eight byte field, making it a 9-byte structure (which probably will turn into 12 or 16 on modern compilers).

I wasn't considering adding more memory footprint simply because if I wanted to, I could avoid using the union mess altogether and implement it as sizeof(UCTNodePointer) == 16 instead of 8 (and spending more memory).

ihavnoid:
I pushed some code fixing the issues @Hersmunch  brought up.

src/UCTNode.h
Hersmunch: I should have made it clearer than I meant within the `UCTNode` class, next to the existing `m_is_expanding` that is currently taking up a whole byte for a single bit. Something like:
```
    ...
    // Is someone adding scores to this node?
    // We don't need to unset this.
    bool m_is_expanded : 1;
    bool m_is_expanding : 1;
    SMP::Mutex m_nodemutex;
    ...
```

ihavnoid:
Pushed another fix - forgot to test pondering, and fixed a pondering segfault bug.

gcp:
>+// WARNING : expand() is not thread-safe and hence has to be protected by an external lock.

This could be pretty tricky? I think right now the code assumes that it's safe to inspect all kinds of child variables, specifically all the atomics, because it (can) assume(s) that any expanded node isn't going to have the m_children array change.

src/UCTNode.h
```diff
@@ -24,20 +24,97 @@
...
+        return (m_data & 1ULL) == 0;
+    } 
+    UCTNode * _readptr() const {
+        assert(_expanded());
+        return reinterpret_cast<UCTNode*>(m_data);
+    }
+
+    std::int16_t _vertex() const {
+        assert(!_expanded());
+        std::uint16_t x = static_cast<std::uint16_t>( (m_data >> 16) & 0xffff);
```
gcp: auto x =

Do you need the -> uint -> int cast sequence? (You need it in the other direction to tackle the sign bit, but not here I think)

src/UCTNode.h
```diff
@@ -24,20 +24,97 @@
...
+        return static_cast<std::int16_t>(x);
+    }
+    float _score() const {
+        static_assert(sizeof(float) == sizeof(std::uint32_t), "This code assumes floats are 32-bit");
```
gcp: Why not compare to 4 directly?

src/UCTNode.h
```diff
@@ -24,20 +24,97 @@
...
+        assert(!_expanded());
+        std::uint32_t x = static_cast<std::uint32_t>( (m_data >> 32) & 0xffffffffLL);
```
gcp: auto x =

I think you don't need the LL suffix on the mask.

src/UCTNode.h
```diff
@@ -24,20 +24,97 @@
...
+        float ret;
+        std::memcpy(&ret, &x, sizeof(ret));
```
gcp: You could cast &m_data to char*, and offset that pointer, avoiding the above conversion.

ihavnoid:
Regarding thread safety - the reason we dont need this to be lock protected boils down to:
- The parent's m_children has to be lock-protected.  Otherwise we have zero chance surviving even without this patch.
- Before being expanded, this is essentially read only.
- After being expanded, this is identical to previous implementation.

I will address rest of the comments in the following day or two.

src/UCTNode.h
ihavnoid: Wouldn't the offset be platform dependent, depending on the Indian?  This was done to be platform neutral...

src/UCTNodeRoot.cpp
```diff
@@ -141,12 +141,14 @@ UCTNode* UCTNode::get_nopass_child(FastState& state) const {
...
+// Used to find new root in UCTSearch.
+// BEWARE : caller is now responsible of freeing the return value
+UCTNode* UCTNode::find_child(const int move) {
     if (m_has_children) {
         for (auto& child : m_children) {
-            if (child->get_move() == move) {
-                return std::move(child);
+            if (child.get_move() == move) {
+                child.expand(); // no guarantee that this is a non-expanded node
+                return child.release();
```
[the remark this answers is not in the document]
ihavnoid: I guess you meant just constructing a std::unique_ptr<UCTNode>, correct?

ihavnoid:
One more thing - I figured out that we are using the term "expand" on other places.  I'll rather rename it on this PR to be "deflate".  Any better word is welcome.

src/UCTNode.h
ihavnoid: Wanted to be explicit that I was intended to copy this to a std::uint32_t but well, doesn't really matter.  I'll put a 4 instead.

src/UCTNode.h
ihavnoid: Yeah I think we don't need it.  We don't even need the 0xffff mask too so I am removing that too.

src/UCTNode.h
ihavnoid: Neither we need the whole mask thingy here - I'll throw away the mask altogether.

ihavnoid:
Addressed as much comments as possible, plus renamed expand() to inflate() to avoid confusion.  Except this comment....

> You could cast &m_data to char*, and offset that pointer, avoiding the above conversion.

src/UCTNodeRoot.cpp
```diff
@@ -141,16 +141,25 @@ UCTNode* UCTNode::get_nopass_child(FastState& state) const {
     return nullptr;
 }
 
-// Used to find new root in UCTSearch
-UCTNode::node_ptr_t UCTNode::find_child(const int move) {
+// Used to find new root in UCTSearch.
+// BEWARE : caller is now responsible of freeing the return value
```
gcp: The BEWARE is pretty weird if it returns a unique_ptr!

src/UCTNode.h
```diff
@@ -113,7 +195,71 @@ class UCTNode {
 
     // Tree data
     std::atomic<bool> m_has_children{false};
-    std::vector<node_ptr_t> m_children;
+    std::vector<UCTNodePointer> m_children;
 };
 
+
+inline UCTNodePointer::~UCTNodePointer() {
```
gcp: Is there any particular reason why these are all inline in the header? I'd prefer all the UCTNodePointer things in a separate .h/.cpp in fact.

src/UCTNodeRoot.cpp
ihavnoid: Stale comment - this comment was there when this was a raw pointer return.  Will remove.

src/UCTNode.h
ihavnoid: Well, it's an old habit dealing with older (e.g., gcc 4.7) compilers.  Experience is that when it's only small snippets and it's a critical piece of code, just putting it into headers yield much better performance.  If you think modern linkers are much better at handling global optimization I see no reason to keep them in the header though.  If you think performance won't change much then I have no reason that we keep in the header.

ihavnoid:
By the way, there seems to be at least one thing that we might want to change... and it's merged.  Should I post another PR?

Issue #785, opened 2018-01-30
Ideas from Minigo


gcp:
https://github.com/tensorflow/minigo

The TensorFlow team did their own re-implementation of AlphaGo Zero, apparently meant to demonstrate TensorFlow and how to run machine learning apps on the Google Cloud. I had a peek at how they resolved some things which were considered ambiguous in the AlphaGo Zero paper:

They use FPU=parent, not FPU=0 (in -1...1 range):
https://github.com/tensorflow/minigo/blob/188fb197fdafbe9664a32142373b1cbd1459bc67/mcts.py#L204

They added specific knowledge about passes to avoid the early pass wars (boo, cheating!): 
https://github.com/tensorflow/minigo/commit/a8af0d8280b4b80a53ed92e59428d1990cbb3a46

They store the probabilities that were used to pick the move (not the original search probabilities)
https://github.com/tensorflow/minigo/blob/188fb197fdafbe9664a32142373b1cbd1459bc67/strategies.py#L135
https://github.com/tensorflow/minigo/blob/188fb197fdafbe9664a32142373b1cbd1459bc67/strategies.py#L233

but it's not quite T=0 vs T=1:
https://github.com/tensorflow/minigo/blob/a3e98185c192636389aff113e674a2cf8e0d9488/mcts.py#L223

gcp:
I think this is a great idea, I'd pick C++ over Go though.

https://github.com/tensorflow/minigo/issues/14

gcp:
Scale/Center layers used for the batchnorm during residual:
https://github.com/tensorflow/minigo/blob/188fb197fdafbe9664a32142373b1cbd1459bc67/dual_net.py#L213

But not for the head:
https://github.com/tensorflow/minigo/blob/188fb197fdafbe9664a32142373b1cbd1459bc67/dual_net.py#L237

(Scale is entirely redundant, but Center may not be - it's one of the things I'd fix in a v2 weights format)

I think it also has the problem that the convolution bias weights are entirely redundant.

gcp:
They also batch the MCTS search by 8:
https://github.com/tensorflow/minigo/blob/188fb197fdafbe9664a32142373b1cbd1459bc67/strategies.py#L177

I'm guessing the loss in search efficiency is worth it compared to any speedups they get due to the backend being TF+cuDNN.

Ttl:
[...]

PR #1099, opened 2018-03-25
Batch norm beta support

Ttl:
Backwards compatible batch norm beta support. Replaces zero biases in network file with `beta*sqrt(var + e)` so that the output is correct in leelaz which adds biases before the batch norm.

In my previous test (https://github.com/gcp/leela-zero/issues/785#issuecomment-362274170) it seemed like adding beta gave a very small decrease in loss. Maybe the effect would be bigger when starting from scratch or with bigger learning rate? Even if the strength increase is minor it should be worth it since this doesn't require any additional work in leelaz.

Tested that restoring and saving gives back the input weights. Training gave reasonable loss and output weights and the network works in leelaz. However I don't have resources to do a bigger training run to check how this affects strength.

### DIFF
--- a/training/tf/tfprocess.py
+++ b/training/tf/tfprocess.py
@@ -38,12 +38,6 @@ def bias_variable(shape):
     initial = tf.constant(0.0, shape=shape)
     return tf.Variable(initial)
 
-# No point in learning bias weights as they are cancelled
-# out by the BatchNorm layers's mean adjustment.
-def bn_bias_variable(shape):
-    initial = tf.constant(0.0, shape=shape)
-    return tf.Variable(initial, trainable=False)
-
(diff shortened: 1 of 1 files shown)


gcp:
I think this trick is going to confuse the hell out of everyone that doesn't know the history. Perhaps this trick with keeping the weight format backwards compatible should be explained somewhere.

gcp:
This broke net_to_model with ```AttributeError: 'Tensor' object has no attribute 'assign'``` and has been backed out.

Ttl:
It seems that there is a conflict with the new SWA code. I have a stupid fix for it but I'll see if I can have something better that would prevent similar issues in future.

PR #1106, opened 2018-03-25
Allow minimum_leelaz_version from server, for a future required update

roy7:


gcp:
@marcocalignano please review

roy7:
@gcp @marcocalignano @Mardak Since everyone is already on 0.13 or newer, is this safe to swap on server side? Or should I wait for the mandatory use of 0.15?

Want to reduce the number of "leelaz_version" bug reports we'll get this time around ahead of time if I can. ;)

Mardak:
This landed on March 26th -- 2 days after 0.13 release https://github.com/gcp/leela-zero/releases/tag/v0.13

We can remove "leelaz_version" from the server if we don't expect any 0.13 clients or older. I would expect there to still be a good portion of clients still running 0.13.

roy7:
Ok I'll change this on server once 0.15 is mandatory.

0.13 release prompted bug reports that it was saying "leelaz_version" was 0.12, which is what led to this change.

marcocalignano:
@roy7 What you want to change is an option in autogtp. So you can first force autogtp v16 and then you can change in json the name of  `leela_version`

Mardak:
@roy7 as @marcocalignano noted in https://github.com/gcp/leela-zero/pull/1106#issuecomment-387996648, we need to force autogtp 16 and let that propagate kicking off those clients. And **then** switch to minimum_leelaz_version once we're confident there are no longer autogtp clients before 16. Otherwise old clients think they're good because the server is not requiring any version in a way that the old clients understand.

https://github.com/gcp/leela-zero-server/issues/138

PR #1125, opened 2018-03-30
net2net: Add option to increase the number of input planes

Ttl:
The additional weights connecting to the new input planes are filled with normal distribution noise with 0 mean and standard deviation given by the --noise argument. The new planes are added after the existing planes.

The default noise might be little too small but I'm not changing it for this feature. When visualizing the new input plane weights using the script in https://github.com/glinscott/leela-chess/issues/199 gives the following plot:

add_inputs

The network above has 5 new planes.

@alreadydone could you test this?


alreadydone:
Thanks! Will test ...

PR #1126, opened 2018-03-30
Add weight file format dependent note

rchs819:
Leela Zero Server started parsing weight file format and display `filters` and `blocks` on the website.

I've added a dependent note in the comment section ensuring when weight file format is changed, a PR is also needed on gcp/leela-zero-server

### DIFF
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -190,6 +190,10 @@ std::vector<float> Network::zeropad_U(const std::vector<float>& U,
     return Upad;
 }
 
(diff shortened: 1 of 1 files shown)


kris-computer-go:
That's great.
Can we also change earlier networks with filters x blocks instead of TBD

It would be nice as well to put a new column with internal ELO rating.

gcp:
I don't think this is needed. Would be good to make sure the server only tries to parse v1 files though.

PR #1130, opened 2018-03-31
Tidy Network gathering of features into input vector 

Hersmunch:
Changed to produce a vector of feature directly rather than going via the NNPlanes representation. Simplified application of rotations by storing board vertices instead of NN plane vertices.

Also makes it easier for a future change to produce rotations of training data, if ever desired.

### DIFF
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -347,7 +347,7 @@ int main (int argc, char *argv[]) {
     auto komi = 7.5f;
     maingame->init_game(BOARD_SIZE, komi);
 
-    Network::init_rotation_table(*maingame);
+    Network::init_symmetry_table(*maingame);
 
     if (cfg_benchmark) {
         cfg_quiet = false;
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -347,6 +347,8 @@ int main (int argc, char *argv[]) {
     auto komi = 7.5f;
     maingame->init_game(BOARD_SIZE, komi);
 
+    Network::init_rotation_table(*maingame);
+
     if (cfg_benchmark) {
         cfg_quiet = false;
         benchmark(*maingame);
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -856,17 +856,14 @@ Network::Netresult Network::get_scored_moves(
         }
     }
 
-    NNPlanes planes;
-    gather_features(state, planes);
-
     if (ensemble == DIRECT) {
         assert(rotation >= 0 && rotation <= 7);
-        result = get_scored_moves_internal(state, planes, rotation);
+        result = get_scored_moves_internal(state, rotation);
     } else {
         assert(ensemble == RANDOM_ROTATION);
         assert(rotation == -1);
         const auto rand_rot = Random::get_Rng().randfix<8>();
-        result = get_scored_moves_internal(state, planes, rand_rot);
+        result = get_scored_moves_internal(state, rand_rot);
     }
 
     // Insert result into cache.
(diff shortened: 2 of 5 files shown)


src/Leela.cpp
```diff
@@ -347,6 +347,8 @@ int main (int argc, char *argv[]) {
     auto komi = 7.5f;
     maingame->init_game(BOARD_SIZE, komi);
 
+    Network::init_rotation_table(*maingame);
```
gcp: Why not let Network call this? Why does main have to know about it?

src/Network.cpp
```diff
@@ -1000,81 +991,88 @@ void Network::show_heatmap(const FastState* const state,
...
-    auto x = vertex % BOARD_SIZE;
-    auto y = vertex / BOARD_SIZE;
-    int newx;
-    int newy;
-
-    if (symmetry >= 4) {
-        std::swap(x, y);
-        symmetry -= 4;
+int Network::get_board_vertex(const GameState& state, 
+                              const int x, const int y, int rotation) {
```
gcp: I much prefer the vertex rotation (err, symmetry) and "obtain board contents" be kept split. This has been a source of subtle bugs early on and it could use unit tests, which merging is going to make more problematic.

src/Network.cpp
```diff
@@ -1000,81 +991,88 @@ void Network::show_heatmap(const FastState* const state,
...
-                    black[idx] = true;
-                } else {
-                    white[idx] = true;
-                }
-            }
-            idx++;
+void Network::gather_features(const GameState* const state, NNPlanes& planes) {
+    
+    std::vector<net_t>features_vector;
+    gather_features_vector(state, features_vector, 0);
```
gcp: If you're going to do it this way, it might as well return the vector. (And this function might, too)

src/Network.cpp
```diff
@@ -1000,81 +991,88 @@ void Network::show_heatmap(const FastState* const state,
...
-    int newy;
-
-    if (symmetry >= 4) {
-        std::swap(x, y);
-        symmetry -= 4;
+int Network::get_board_vertex(const GameState& state, 
+                              const int x, const int y, int rotation) {
+    assert(x >= 0 && x < BOARD_SIZE);
+    assert(y >= 0 && y < BOARD_SIZE);
+    assert(rotation >= 0 && rotation <= 7);
```
gcp: Although the original code is inconsistent too, calling it a "rotation" is mathematically false, which is pretty annoying given that this code is calculating...symmetries. I'd prefer we fix this naming, not make it worse.

src/Network.cpp
Hersmunch: Understood, thanks for the warning about potential subtle bugs. Are you ok with the "obtaining of board vertex" to still be included within the symmetry table initialisation?

src/Network.cpp
Hersmunch: So should the `BoardPlane` and `NNPlanes` types be moved to Training.cpp for use when streaming and dumping the training data or are you thinking that it would be simpler to remove the use of `bitset`s and to iterate through a vector in those cases instead?

src/Leela.cpp
Hersmunch: @gcp (since this comment has been hidden by github)

src/Network.cpp
Hersmunch: I haven't done anything about this for now. Would you mind if I did further tidy up to Training.cpp in a separate follow-up PR?

Hersmunch:
I'll go away and simplify this and raise two separate PRs; one for renaming to symmetry and the other for using a vector for the features but without any rotation/symmetry changes.

PR #1082, opened 2018-03-23
Trim low policy nodes when node budget is low

WhiteHalmos:

PR #1134, opened 2018-04-01
Allow incremental node expansions.

TFiFiE:
This addresses the issue raised in https://github.com/gcp/leela-zero/pull/1082#issuecomment-375688136. It may not appear to be necessary in order to protect the strength of the engine for subsequent moves in a game, but one could imagine this becoming a problem when it's used for analysis.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -54,9 +54,9 @@ SMP::Mutex& UCTNode::get_mutex() {
 bool UCTNode::create_children(std::atomic<int>& nodecount,
                               GameState& state,
                               float& eval,
-                              float mem_full) {
+                              float min_psa_ratio) {
     // check whether somebody beat us to it (atomic)
-    if (has_children()) {
+    if (!expandable(min_psa_ratio)) {
         return false;
     }
     // acquire the lock
@@ -66,7 +66,7 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
         return false;
     }
     // check whether somebody beat us to it (after taking the lock)
-    if (has_children()) {
+    if (!expandable(min_psa_ratio)) {
         return false;
     }
     // Someone else is running the expansion
@@ -113,13 +113,16 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
         }
     }
 
-    link_nodelist(nodecount, nodelist, mem_full);
+    link_nodelist(nodecount, nodelist, min_psa_ratio);
+    m_is_expanding = false;
     return true;
 }
 
 void UCTNode::link_nodelist(std::atomic<int>& nodecount,
                             std::vector<Network::scored_node>& nodelist,
-                            float mem_full) {
+                            float min_psa_ratio) {
+    assert(min_psa_ratio < m_min_psa_ratio_children);
+
     if (nodelist.empty()) {
         return;
     }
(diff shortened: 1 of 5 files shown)


src/UCTSearch.cpp
```diff
@@ -535,12 +551,16 @@ int UCTSearch::think(int color, passflag_t passflag) {
     // create a sorted list off legal moves (make sure we
     // play something legal and decent even in time trouble)
     float root_eval;
-    if (!m_root->has_children()) {
+    const auto had_children = m_root->has_children();
+    if (m_root->expandable()) {
```
Hersmunch: Do we care that if a node is expanded multiple times that it may be done so based on a different rotation each time? This would only happen if the cache had pushed the original result out.
I guess there is also a slim possibility that a node's result is pushed out of the cache but its children's results aren't. Hmm, would that have any interesting effects?

src/UCTNode.cpp
```diff
@@ -113,13 +113,16 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
         }
     }
 
-    link_nodelist(nodecount, nodelist, mem_full);
+    link_nodelist(nodecount, nodelist, min_psa_ratio);
+    m_is_expanding = false;
```
TFiFiE: I guess this could theoretically have been a source for race conditions, if it were not for the fact that this atomic store uses the strictest memory order, `memory_order_seq_cst`.

src/UCTSearch.cpp
gcp: >This would only happen if the cache is full and had pushed the original result out.

It should be noted that this code only does something if we have a very large tree, so this may be very likely to happen.

src/UCTSearch.cpp
```diff
@@ -121,22 +121,38 @@ void UCTSearch::update_root() {
...
         if (currstate.get_passes() >= 2) {
             auto score = currstate.final_score();
             result = SearchResult::from_score(score);
         } else if (m_nodes < MAX_TREE_SIZE) {
-            auto mem_full_pct = m_nodes / static_cast<float>(MAX_TREE_SIZE);
             float eval;
             auto success =
-                node->create_children(m_nodes, currstate, eval, mem_full_pct);
+                node->create_children(m_nodes, currstate, eval, get_min_psa_ratio());
             if (success) {
```
gcp: I think this may be breaking the search? You will return an eval result multiple times from the same node. I think we observed this makes the program weaker.

src/UCTNode.h
```diff
@@ -112,7 +113,7 @@ class UCTNode {
     SMP::Mutex m_nodemutex;
 
     // Tree data
-    std::atomic<bool> m_has_children{false};
+    std::atomic<float> m_min_psa_ratio_children{2.0f};
```
gcp: Does changing this from bool to float blow up the node size? If so, you might be able to extend m_status to have a PARTIALLY_EXPANDED state and use that instead.

TFiFiE:
Oops, `m_is_expanding` is not an atomic at all. In fact, I fail to see the point of its existence in the first place and why the mutex lock couldn't just be extended for the remainder of `link_nodelist`.

gcp:
>why the mutex lock couldn't just be extended for the remainder of link_nodelist

Because that means the mutex is held longer.

src/UCTSearch.cpp
alreadydone: https://github.com/gcp/leela-zero/commit/bae6d40fbefcef970c57939d0766ebdd5baa4f70#diff-28216771f77a09f4e0782ad5404fdd28R152
The interesting effect is that there is no guarantee that all children will be recovered (even for the root node!) and you may not be reserving enough for the vector when a different symmetry is chosen at re-expansion, because the policy is different. We need to store the symmetry (requires 3 bits ...) to be used when calling get_scored_moves (now get_output) in create_children at re-expansion. Otherwise, we need to move elements m_children to corresponding spots of the nodelist, delete elements with policy <new_min_psa and without visits from the nodelist, then use the nodelist as the new m_children, which is a bit too complicated.

PR #1140, opened 2018-04-03
Add UCTNodePointer.cpp to Makefile.

TFiFiE:
Why didn't the CI services catch this?


bood:
Reason is CI uses CMakefile rather than Makefile

PR #1151, opened 2018-04-05
Output verbosity

Cabu:
Allow different levels of output verbosity:
0: same as --quiet
1: Thinking progress
2: Some initalisation informations and thinking progress
3: All initialisation informations, thinking progress, and pondering progress


gcp:
Definitely won't merge this as is. This should probably be done by integrating proper logging with logging levels, not wrapping every output statement in an if.

Again, what's the use case?

PR #1153, opened 2018-04-05
Deep ponder

Cabu:
Allow deeper pondering throught the parameter --ponder N
-1: like --noponder
0: Infinite pondering like without specifying -v and -p
N: Multiply Visits and Playouts by N while pondering


Cabu:
There is a default initialisation error. Creating a new pull request.

PR #1154, opened 2018-04-05
Deeper ponder

Cabu:
Allow deeper pondering throught the parameter --ponder N
-1: like --noponder
0: Infinite pondering like without specifying Visits and Playouts
1: (Default) Respect Visits and Playouts
N: Multiply Visits and Playouts by N while pondering

### DIFF
--- a/src/FastBoard.cpp
+++ b/src/FastBoard.cpp
@@ -408,8 +408,12 @@ std::string FastBoard::move_to_text(int move) const {
     column--;
     row--;
 
-    assert(move == FastBoard::PASS || move == FastBoard::RESIGN || (row >= 0 && row < m_boardsize));
-    assert(move == FastBoard::PASS || move == FastBoard::RESIGN || (column >= 0 && column < m_boardsize));
+    assert(move == FastBoard::PASS
+           || move == FastBoard::RESIGN
+           || (row >= 0 && row < m_boardsize));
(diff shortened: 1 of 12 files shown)


Cabu:
My use case is to limit the programe through --visits and --playout to have it play 'quickly'. But I still like it to use all the time it could (infinite pondering) while I play.

gcp:
>to have it play 'quickly'

Use time_settings.

Cabu:
Even if I set time_settings at the beginning of a games session, it is overridden by OGS and KGS servers at the beginning of each game. As a result time_settings could solve my problem but doesn't :(

gcp:
Then file a bug against the OGS and KGS interfaces? Adding these options to LZ seems senseless.

Cabu:
Setting the time_settings is part of the definition of a new game. How should the engine know how much time it have to play a game if not by using time_settings/kgs-time>_settings ? It is not a OGS/KGS bug, but part of the GTP protocol when starting a new game.

gcp:
> How should the engine know how much time it have to play a game if not by using time_settings/kgs-time>_settings ?

It shouldn't, that's what that command is for, which is what I'm saying: if you want it to play fast, set a fast timecontrol.

WhiteHalmos:
@Cabu you can always keep local modifications specific to KGS and OGS.

PR #1203, opened 2018-04-14
Netresult memory size optimization

ihavnoid:
Instead of keeping a list of std::pair<float,int>, just keep a flat array of floats.  This reduces per-entry NNCache size by half, hence allowing bigger cache sizes for larger playouts.

### DIFF
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -861,12 +861,12 @@ Network::Netresult Network::get_scored_moves(
 
     if (ensemble == DIRECT) {
         assert(rotation >= 0 && rotation <= 7);
-        result = get_scored_moves_internal(state, planes, rotation);
+        result = get_scored_moves_internal(planes, rotation);
     } else {
         assert(ensemble == RANDOM_ROTATION);
         assert(rotation == -1);
         const auto rand_rot = Random::get_Rng().randfix<8>();
-        result = get_scored_moves_internal(state, planes, rand_rot);
+        result = get_scored_moves_internal(planes, rand_rot);
     }
 
     // Insert result into cache.
@@ -876,7 +876,7 @@ Network::Netresult Network::get_scored_moves(
 }
 
 Network::Netresult Network::get_scored_moves_internal(
-    const GameState* const state, const NNPlanes & planes, const int rotation) {
+    const NNPlanes & planes, const int rotation) {
     assert(rotation >= 0 && rotation <= 7);
     assert(INPUT_CHANNELS == planes.size());
     constexpr auto width = BOARD_SIZE;
@@ -930,45 +930,33 @@ Network::Netresult Network::get_scored_moves_internal(
     // Sigmoid
     const auto winrate_sig = (1.0f + std::tanh(winrate_out[0])) / 2.0f;
 
-    std::vector<scored_node> result;
+    Netresult result;
+
     for (auto idx = size_t{0}; idx < outputs.size(); idx++) {
         if (idx < BOARD_SQUARES) {
             const auto rot_idx = rotate_nn_idx_table[rotation][idx];
-            const auto x = rot_idx % BOARD_SIZE;
-            const auto y = rot_idx / BOARD_SIZE;
-            const auto rot_vtx = state->board.get_vertex(x, y);
-            if (state->board.get_square(rot_vtx) == FastBoard::EMPTY) {
-                result.emplace_back(outputs[idx], rot_vtx);
-            }
+            result[rot_idx] = outputs[idx];
         } else {
-            result.emplace_back(outputs[idx], FastBoard::PASS);
+            result[idx] = outputs[idx];
         }
     }
+    result[BOARD_SQUARES+1] = winrate_sig;
 
-    return std::make_pair(result, winrate_sig);
+    return result;
 }
 
 void Network::show_heatmap(const FastState* const state,
                            const Netresult& result,
                            const bool topmoves) {
-    auto moves = result.first;
     std::vector<std::string> display_map;
     std::string line;
 
     for (unsigned int y = 0; y < BOARD_SIZE; y++) {
         for (unsigned int x = 0; x < BOARD_SIZE; x++) {
-            const auto vtx = state->board.get_vertex(x, y);
-
-            const auto item = std::find_if(moves.cbegin(), moves.cend(),
-                [&vtx](scored_node const& test_item) {
-                return test_item.second == vtx;
-            });
-
             auto score = 0;
-            // Non-empty squares won't be scored
-            if (item != moves.cend()) {
-                score = int(item->first * 1000);
-                assert(vtx == item->second);
+            auto vertex = state->board.get_vertex(x, y);
+            if (state->board.get_square(vertex) == FastBoard::EMPTY) {
+                score = result[y * BOARD_SIZE + x] * 1000;
             }
 
             line += boost::str(boost::format("%3d ") % score);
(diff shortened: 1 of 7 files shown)


src/Network.cpp
```diff
@@ -930,45 +930,33 @@ Network::Netresult Network::get_scored_moves_internal(
...
-                [&vtx](scored_node const& test_item) {
-                return test_item.second == vtx;
-            });
-
             auto score = 0;
-            // Non-empty squares won't be scored
-            if (item != moves.cend()) {
-                score = int(item->first * 1000);
-                assert(vtx == item->second);
+            auto vertex = state->board.get_vertex(x, y);
```
Hersmunch: const was removed

src/Network.cpp
```diff
@@ -981,12 +969,22 @@ void Network::show_heatmap(const FastState* const state,
...
-    assert(result.first.back().second == FastBoard::PASS);
-    const auto pass_score = int(result.first.back().first * 1000);
+    const auto pass_score = int(result[BOARD_SQUARES] * 1000);
     myprintf("pass: %d\n", pass_score);
-    myprintf("winrate: %f\n", result.second);
+    myprintf("winrate: %f\n", result[BOARD_SQUARES+1]);
 
     if (topmoves) {
+        std::vector<std::pair<float,int>> moves;
+        for (int i=0; i < BOARD_SQUARES; i++) {
```
Hersmunch: auto

src/Network.cpp
```diff
@@ -981,12 +969,22 @@ void Network::show_heatmap(const FastState* const state,
...
+            auto x = i % BOARD_SIZE;
+            auto y = i / BOARD_SIZE;
+            auto vertex = state->board.get_vertex(x, y);
```
Hersmunch: Can all be const

src/UCTNode.cpp
```diff
@@ -81,22 +81,31 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
...
-        if (state.is_move_legal(to_move, vertex)) {
-            nodelist.emplace_back(node);
-            legal_sum += node.first;
+    for (auto i = 0; i < BOARD_SQUARES+1; i++) {
+        if (i < BOARD_SQUARES) {
+            const auto x = i % BOARD_SIZE;
+            const auto y = i / BOARD_SIZE;
+            const auto vertex = state.board.get_vertex(x, y); 
+            if (state.is_move_legal(to_move, vertex)) {
+                auto p = std::make_pair(raw_netlist[i], vertex);
```
Hersmunch: Can be const. Same for similar line a few lines down

ihavnoid:
Addressed code review comments.

src/UCTNode.h
```diff
@@ -88,7 +88,7 @@ class UCTNode {
         ACTIVE
     };
     void link_nodelist(std::atomic<int>& nodecount,
-                       std::vector<Network::scored_node>& nodelist,
+                       std::vector<std::pair<float,int>>& nodelist,
```
gcp: Why remove the typedef?

src/Network.cpp
```diff
@@ -930,45 +930,33 @@ Network::Netresult Network::get_scored_moves_internal(
...
-            if (state->board.get_square(rot_vtx) == FastBoard::EMPTY) {
-                result.emplace_back(outputs[idx], rot_vtx);
-            }
+            result[rot_idx] = outputs[idx];
         } else {
-            result.emplace_back(outputs[idx], FastBoard::PASS);
+            result[idx] = outputs[idx];
         }
     }
+    result[BOARD_SQUARES+1] = winrate_sig;
```
gcp: This is very ugly. At least make it a pair of an array and a float, or define a type/class/struct for the result.

src/Network.h
```diff
@@ -39,8 +39,9 @@ class Network {
     };
     using BoardPlane = std::bitset<BOARD_SQUARES>;
     using NNPlanes = std::vector<BoardPlane>;
-    using scored_node = std::pair<float, int>;
-    using Netresult = std::pair<std::vector<scored_node>, float>;
+
+    // {19x19 board positions, pass, winrate}
+    using Netresult = std::array<float, BOARD_SQUARES+2>;
```
gcp: Rather use a custom struct/class or at worst a tuple.

ihavnoid:
Addressed @gcp 's comments, although I am not sure it is satisfactory.  The thing is, because the loop index itself is the location of the move, I ended up using the index itself anyway and hence the 'foreach' style of loop iteration wasn't that useful anyway.

One additional change I made was that the policy is not raw std::array, it's now a std::vector sized of BOARD_SQUARE - this is because there are many locations doing std::move, and using a std::vector instead of std::array will avoid many deep copy operations.

src/Network.cpp
```diff
@@ -981,12 +967,23 @@ void Network::show_heatmap(const FastState* const state,
...
     }
-    assert(result.first.back().second == FastBoard::PASS);
-    const auto pass_score = int(result.first.back().first * 1000);
+    const auto pass_score = int(result.policy_pass * 1000);
     myprintf("pass: %d\n", pass_score);
-    myprintf("winrate: %f\n", result.second);
+    myprintf("winrate: %f\n", result.winrate);
 
     if (topmoves) {
+        using ScoreVertexPair = std::pair<float,int>;
```
gcp: You don't need to redefine this?

src/UCTNode.cpp
```diff
@@ -81,24 +81,32 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
...
-        auto vertex = node.second;
+    for (auto i = 0; i < BOARD_SQUARES; i++) {
+        const auto x = i % BOARD_SIZE;
+        const auto y = i / BOARD_SIZE;
+        const auto vertex = state.board.get_vertex(x, y); 
         if (state.is_move_legal(to_move, vertex)) {
-            nodelist.emplace_back(node);
-            legal_sum += node.first;
+            const auto p = std::make_pair(raw_netlist.policy[i], vertex);
+            nodelist.emplace_back(p);
```
gcp: I think you can just do emplace_back(raw_netlist.policy[i], vertex)?

src/UCTNode.cpp
```diff
@@ -81,24 +81,32 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
...
+            legal_sum += raw_netlist.policy[i];
         }
     }
+    {
+        const auto p = std::make_pair(raw_netlist.policy_pass, FastBoard::PASS);
+        nodelist.emplace_back(p);
```
gcp: Same.

src/Network.h
```diff
@@ -39,8 +39,13 @@ class Network {
     };
     using BoardPlane = std::bitset<BOARD_SQUARES>;
     using NNPlanes = std::vector<BoardPlane>;
-    using scored_node = std::pair<float, int>;
-    using Netresult = std::pair<std::vector<scored_node>, float>;
+
+    // {19x19 board positions, pass, winrate}
+    struct Netresult {
+        std::vector<float> policy = std::vector<float>(BOARD_SQUARES);
```
gcp: Gah, sucks how uniform initialization and initializer lists interact :(

gcp:
>The thing is, because the loop index itself is the location of the move, I ended up using the index itself anyway and hence the 'foreach' style of loop iteration wasn't that useful anyway.

I understand, it was more about stuffing the extra info into magic locations. The struct is much cleaner.

src/Network.cpp
gcp: I'd reuse the typedef (even if the Vertex is not really the same), and put it wherever is most appropriate.

ihavnoid:
Pushed a commit addressing the third batch of comments.

PR #1206, opened 2018-04-14
Fixes segfault in dump_stats()

Ka-zam:
Fixes segfault

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -219,6 +219,7 @@ void UCTSearch::dump_stats(FastState & state, UCTNode & parent) {
     for (const auto& node : parent.get_children()) {
         // Always display at least two moves. In the case there is
         // only one move searched the user could get an idea why.
+        node.inflate();
         if (++movecount > 2 && !node->get_visits()) break;
 
         std::string move = state.move_to_text(node->get_move());


Ka-zam:
I'm a bit skeptical to this PR now. It does work and gets rid of the segfault. However, since it's explicitly stated that inflate() needs a lock perhaps it shouldn't be merged until the atomic UCTNodePointer has been merged as well. It's not any worse I guess.

ihavnoid:
The PR itself is okay (dump_stats() itself is only called when there is a single thread, so no lock needed) but the original intent was that it doesn't even need to do so.  All children are assumed to be already inflated before calling dump_stats().  The main problem is *why dump_stats() is calling a non-inflated list of children*.

If you see UCTSearch::ponder(), there is this code which already inflates all children:
```
    // There are a lot of special cases where the root node assumes all
    // childen are inflated.
    m_root->inflate_all_children();
```
and I think I know why : this is a case where pondering is done on a completely empty tree.  Hence m_root had no children yet because there were zero simulations.  Later on when pondering was run, it started to have children, and they were left there non-inflated.

The non-ponder UCTSearch::think() has this piece of code:
```
    // create a sorted list off legal moves (make sure we
    // play something legal and decent even in time trouble)
    float root_eval;
    const auto had_children = m_root->has_children();
    if (m_root->expandable()) {
        m_root->create_children(m_nodes, m_rootstate, root_eval);
    }
    if (had_children) {
        root_eval = m_root->get_eval(color);
    } else {
        m_root->update(root_eval);
        root_eval = (color == FastBoard::BLACK ? root_eval : 1.0f - root_eval);
    }
```

I think the solution is one of these:
- Use @Ka-zam 's patch as-is : it should solve the problem but if somebody else assumes root's children are always inflated they will get the same surprise
- Put another  `m_root->inflate_all_children()` right after the main loop : I don't think there will be code assuming the root's children are always inflated within the main loop, so it should be okay, but not perfect.
- do something similar to what we did on think() (always create root's children) but I am not sure about all the possible side effects here.  Maybe this is the fundamental fix for dodging all sort of corner cases, but I am not sure.

@gcp what do you think?

Ka-zam:
I think the third alternative makes most sense. Just blindly inflating things to pass the assert is a bit crude.

gcp:
Making sure the ponder root has children by expanding it sounds fine yes.

PR #1217, opened 2018-04-18
Create autoenv.sh

cngoodboy:


gcp:
This seems like it would only work on some specific Ubuntu versions.

PR #1224, opened 2018-04-19
SMP : Lock destructor should not unlock mutex if lock is not owned

ihavnoid:
If a thread explicitly calls unlock() then the destructor may try to unlock it again.
Check if this Lock instance owns the lock, and attempt unlocking only if it owns it.

### DIFF
--- a/src/SMP.cpp
+++ b/src/SMP.cpp
@@ -18,6 +18,7 @@
 
 #include "SMP.h"
 
+#include <cassert>
 #include <thread>
 
 SMP::Mutex::Mutex() {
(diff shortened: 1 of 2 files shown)


gcp:
Good catch.

src/SMP.cpp
```diff
@@ -26,19 +27,31 @@ SMP::Mutex::Mutex() {
...
 }
 
 void SMP::Lock::unlock() {
-    m_mutex->m_lock.store(false, std::memory_order_release);
+    assert(m_selflock);
+    auto prevval = m_mutex->m_lock.exchange(false, std::memory_order_release);
+
+    // if this fails it means we are unlocking an unlocked lock
+    assert(prevval);
+    m_selflock = false;
```
oni-link: How can we be sure that `m_selflock` has the right value if one thread just unlocked the mutex and another thread locked the mutex before this line was reached?

gcp: m_selflock is only modified by the mutex itself, so it can never be modified concurrently.

PR #1233, opened 2018-04-21
Encode Netresult value to be 16-bit unsigned integer

ihavnoid:
Since all values of the NN outputs are from 0\~1, encode them to be 0\~65535,
and keep them as a 16-bit integer to save space on NNCache.
Bump up the max cache size to be 300k entries

### DIFF
--- a/msvc/VS2015/leela-zero.vcxproj
+++ b/msvc/VS2015/leela-zero.vcxproj
@@ -108,6 +108,7 @@
     <ClInclude Include="..\..\src\Im2Col.h" />
     <ClInclude Include="..\..\src\KoState.h" />
     <ClInclude Include="..\..\src\Network.h" />
+    <ClInclude Include="..\..\src\Netresult.h" />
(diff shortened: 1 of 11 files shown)


gcp:
Lots of complexity for very little gain. I see no reason to pull this.

gcp:
It probably works given that we tested fp16 at some point, but the gain is too nebulous to worry about having to figure out any discrepancy caused by rounding later on.

Also, more generally speaking, I'd like a more fundamental solution to the problem of reusing search. It's sad the TTable didn't seem to entirely work as wanted with MCTS, because the current construction of NNCache and tree reuse is solving the same thing twice. (The code for tree reuse is already much more complicated than I'd like, and it has caused bugs)

Maybe something like the NNCache but having it own subtrees? Only keeping nodes with at least xxx visits below? Not easy to come up with something good here. I'm just very reluctant to increase the complexity of solutions that seem like they head in the wrong direction. At least the current NNCache code is simple.

PR #1240, opened 2018-04-23
'Real time' pondering output for GUI like Lizzie

Cabu:
Generate more output during the pondering phase to let GUI display that information in a more 'real time' fashion.

This should reduce the number of fork of the engine just for the sake of having more information out of it to display in GUI.

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -670,6 +670,9 @@ int UCTSearch::think(int color, passflag_t passflag) {
 }
 
 void UCTSearch::ponder() {
+    // set up timing info
+    Time start;
+
     update_root();
 
     m_run = true;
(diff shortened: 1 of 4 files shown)


featurecat:
Thanks. The ~begin and ~end outputs aren't FULLY necessary but we'll see what gcp thinks. something like newlines might be fine. of course, I like the ~begin flags 😁

src/UCTSearch.cpp
```diff
@@ -684,12 +687,30 @@ void UCTSearch::ponder() {
...
+        Time elapsed;
+        int elapsed_centis = Time::timediff_centis(start, elapsed);
+
+        // output some stats every few seconds
+        // check if we should still search
+        if (elapsed_centis - last_update > (cfg_rt_log ? 10 : 250)) {
+            last_update = elapsed_centis;
+
+            if (cfg_rt_log) {
+                myprintf("~begin\n");
```
gcp: Random marker? Nah.

src/UCTSearch.cpp
```diff
@@ -684,12 +687,30 @@ void UCTSearch::ponder() {
...
+                dump_stats(m_rootstate, *m_root);
```
gcp: This isn't threadsafe.

gcp:
I would suggest we change the engine to just dump the same output as normal while pondering (this might confuse some existing tools, but that should be possible to deal with?), and make the dump interval configurable?

gcp:
The main problem of this patch seems to be that dump_stats is very much not threadsafe (particularly due to sort_root_children).

gcp:
I would advise designing this more like UCI works: the engine advertises a "dump_move_stats" command in list_commands, which the GUI uses to probe that capability.

That command could then either be send during the search (trickier to implement), or the GUI could send "dump_move_stats 10" to enable periodic dumping every xx centiseconds.

Whatever you do do not add command line options! That's terrible design for these kind of features because you can't steer them without reloading the entire engine.

src/UCTSearch.cpp
gcp: You need to check if taking the root node lock early on in dump_stats still allows the engine to work. You'll get a performance dip every time you dump, but that's unavoidable.

gcp: If that doesn't work, you rewrite the "root moves" dumping part of dump_stats into a separate function and call that instead. You can't use the sort_root_moves code there though, you'll have to write some code that gets the top x moves and remember them.

Cabu:
Does https://github.com/Cabu/leela-zero/tree/Lizzie2 could be the way to go?

gcp:
If you want to handle the command while the search is running it needs special handling. That's why I said that this option is "trickier to implement".

Cabu:
For Lizzie use, the think function of the engine is not used, only the ponder function is used to generate the datas to display on the board. This make handling the command while the search is running is not a problem as ponder stop when e new command is send and can be re-lunch like for the time_left command (that what I have done in the new pull request https://github.com/gcp/leela-zero/pull/1242).

PR #1242, opened 2018-04-23
'Real time' pondering output for GUI like Lizzie

Cabu:
Add a GTP command to generate more output during the pondering phase to let GUI display that information in a more 'real time' fashion.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -70,6 +70,7 @@ std::string cfg_weightsfile;
 std::string cfg_logfile;
 FILE* cfg_logfile_handle;
 bool cfg_quiet;
+int cfg_move_stats;
 std::string cfg_options_str;
 bool cfg_benchmark;
 
@@ -101,6 +102,7 @@ void GTP::setup_default_parameters() {
     cfg_random_cnt = 0;
     cfg_dumbpass = false;
     cfg_logfile_handle = nullptr;
+    cfg_move_stats = 0;
     cfg_quiet = false;
     cfg_benchmark = false;
 
@@ -143,6 +145,7 @@ const std::string GTP::s_commands[] = {
     "kgs-time_settings",
     "kgs-game_over",
     "heatmap",
+    "dump_move_stats"
     ""
 };
 
@@ -789,6 +792,33 @@ bool GTP::execute(GameState & game, std::string xinput) {
         if (!cmdstream.fail()) {
             gtp_printf(id, "");
         } else {
+            gtp_fail_printf(id, "syntax not understood");
+        }
+
+        return true;
+    } else if (command.find("dump_move_stats") == 0) {
+        std::istringstream cmdstream(command);
+        std::string tmp;
+        int period;
+
+        // tmp will eat dump_move_stats
+        cmdstream >> tmp >> period;
+
+        if (!cmdstream.fail()) {
+            if (period >= 0) {
+                cfg_move_stats = period;
+                gtp_printf(id, "");
+            } else {
+                gtp_fail_printf(id, "invalid period");
+            }
+            if (cfg_allow_pondering) {
+                // Could be send anytime even while pondering
+                if (!game.has_resigned()) {
+                    search->ponder();
+                }
+            }
+        }
+        else {
             gtp_fail_printf(id, "syntax not understood");
         }
 
--- a/src/GTP.h
+++ b/src/GTP.h
@@ -53,6 +53,7 @@ extern std::string cfg_logfile;
 extern std::string cfg_weightsfile;
 extern FILE* cfg_logfile_handle;
 extern bool cfg_quiet;
+extern int cfg_move_stats;
 extern std::string cfg_options_str;
 extern bool cfg_benchmark;
 
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -670,6 +670,9 @@ int UCTSearch::think(int color, passflag_t passflag) {
 }
 
 void UCTSearch::ponder() {
+    // set up timing info
+    Time start;
+
     update_root();
 
     m_run = true;
@@ -684,12 +687,27 @@ void UCTSearch::ponder() {
         tg.add_task(UCTWorker(m_rootstate, this, m_root.get()));
     }
     auto keeprunning = true;
+    int last_update = 0;
     do {
         auto currstate = std::make_unique<GameState>(m_rootstate);
         auto result = play_simulation(*currstate, m_root.get());
         if (result.valid()) {
             increment_playouts();
         }
+
+        Time elapsed;
+        int elapsed_centis = Time::timediff_centis(start, elapsed);
+
+        // output some stats every few seconds
+        // check if we should still search
+        if (cfg_move_stats > 0 && elapsed_centis - last_update > cfg_move_stats) {
+            last_update = elapsed_centis;
+
+            myprintf("~begin-move-stats\n");
+            dump_stats(m_rootstate, *m_root);
+            myprintf("~end-move-stats\n");
+        }
+
         keeprunning  = is_running();
         keeprunning &= !stop_thinking(0, 1);
     } while(!Utils::input_pending() && keeprunning);


src/GTP.cpp
```diff
@@ -792,6 +795,33 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
+
+        if (!cmdstream.fail()) {
+            if (period >= 0) {
+                cfg_move_stats = period;
+                gtp_printf(id, "");
+            } else {
+                gtp_fail_printf(id, "invalid period");
+            }
+            if (cfg_allow_pondering) {
+                // Could be send anytime even while pondering
```
gcp: This will cause the search to be stopped and restarted every time while you send the command. It probably doesn't even work in a non-pondering search?

Cabu: Yes, that's true. But for now lizzie only need that. If another GUI request something else, it will be time to rethink the thing with another merge request :)

gcp:
I will probably implement this myself in a cleaner manner, also to address things like #1244.

It's a good time now to extend GTP in a manner similar to UCI, and to try to avoid the mistakes of earlier protocols...

featurecat:
@gcp thanks. It will help so much with the custom Leelaz compilation issues that some users are having with Lizzie :)

dwt:
As I'm too interested in this - what is the state here? What is your plan? How do you plan to extend GTP for this? Is there a standard on how extensions are to be done?

roy7:
GoGUI has a sort of formalized extension setup that is supported by clients like Sabaki. It can do a lot of neat things. If possible, it'd be nice for the extensions Lizzie needs to be compatible with how gogui does things.

dwt:
Maybe I'm misunderstanding what you wrote, but it sounds like the `info` command would send information unasked in between gtp answers to the frontend.

Wouldn't it be much easier to build / process if the GUI could just send the `leelaz_pondering_info` (or whatever it is called) command whenever it wants to display something to get the latest info from the engine? That way the GTP protocol query / reply cycle isn't broken and also the GUI app can easily decide itself how often it wants to refresh this info.

Or is there a clear reason why asynchronously pushing this info (in band none the less) is a much better idea?

Apart from that: Having a GTP command to to set the other command line values sounds really useful to me.

yishn:
@dwt The problem is that GTP commands should be replied in sequential order according the the specifications. That means if the controller requests a move and wants to retrieve pondering info via GTP command while the engine is thinking, then the engine (according to spec) should reply with the move first, then the pondering info. Needless to say, this way, real time UI updates are impossible.

But I do agree that by writing `info` lines into stdout will break the GTP command/reply cycle. Why not include them in stderr? That way, controllers, that don't support `info` lines, will ignore them.

kuba97531:
Another option other than stderr is to specify that the output lines which start with some special character - say '#' should be ignored.

dwt:
@yishn: Stderr actually sounds like a decent workaround to avoid confusion with GTP. Regarding the info command, requiring immediate replies would actually work fine here, as we want the pondering information, i.e. the information when the engine is _not_ currently trying to make a move.

In general, bringing out this information on another file descriptor (like stderr, but possibly another pipe) would also be nice cleanup of the protocol. I'm thinking something like `leelaz-pondering-info /path/to/file $frequency`

gcp:
>But I do agree that by writing info lines into stdout will break the GTP command/reply cycle. Why not include them in stderr? That way, controllers, that don't support info lines, will simply ignore them.

Because right now stderr is "anything goes" and a GTP client isn't even expected to listen to it. 

>But I do agree that by writing info lines into stdout will break the GTP command/reply cycle.

The proposal above doesn't change the command/reply cycle.

The command will be ```genmove b```

and the reply will be 

```
info move G3 visits 50 winrate 52.25 policy 60.1 pv G3 H3
info move H3 visits 30 winrate 51.15 policy 25.3 pv H3 G3
info move G3 visits 100 winrate 53.25 policy 60.1 pv G3 H3 R16
info move H3 visits 60 winrate 52.15 policy 25.3 pv H3 G3 R16
G3
=
```

gcp:
>In general, bringing out this information on another file descriptor (like stderr, but possibly another pipe) would also be nice cleanup of the protocol. I'm thinking something like leelaz-pondering-info /path/to/file $frequency

I see absolutely no upside to this proposal at all. Not only does it making portability more of an issue, it complicates clients on both sides, for no reason whatsoever.

gcp:
>Wouldn't it be much easier to build / process if the GUI could just send the leelaz_pondering_info (or whatever it is called) command whenever it wants to display something to get the latest info from the engine?

The original XBoard/WinBoard protocol worked like this. https://www.gnu.org/software/xboard/engine-intf.html See 12. Analyze mode. ```.   Send a search status update (optional); see below.```

From experience, it's a huge nuisance to work this way. Suddenly, the engine not only has to parse the input asynchronously, it also has to deal with the GUI/protocol state in the middle of the search. Even worse, it's being requested to give an update on search status at a moment it may not make much sense for it, or nothing of relevance has changed.

The engine knows when its search state has changed in a meaningful manner, so it makes a lot more sense for it to push that info into the GUI. This design mistake in XBoard/CECP was fixed in UCI.

(The problem is not so bad in MCTS actually, but let's not built protocols on assumptions of which algorithm will drive the engine. This can change.)

gcp:
>But I do agree that by writing info lines into stdout will break the GTP command/reply cycle

Hmm, for genmove there won't be a problem, but there will be an issue for pondering, right? Because pondering isn't included in the GTP spec, right now Leela just assumes that after a search, she can ponder, or at worst, after an update of ```time_left```, she can (restart) pondering.

So, you'd do a

```time_left b 100 10```

and get a ```=```

But a bit later, you'd start getting a sequence of

```
info move blah 
info move blah 
info move blah 
info move blah
```

Which *does* break the command/response cycle, unlike with genmove.

I'm not sure if there can be a good solution here that maintains the cycle given that pondering isn't handled in GTP. Given that a) the GUI will only get this if it enabled the capability via setoption b) the output is guaranteed to be prefixed by "info", it looks manageable to me to integrate into a GTP controller, but I'm curious if @yishn sees something I missed.

Sending the output on stderr would allow maintaining the command/reply cycle on stdin, but this seems like an illusion to me: don't you run into synchronization problems now trying to figure out which "info" command on one channel corresponded to which GTP board state on the other? I really have problems believing this would be a good solution.

**Comment by Ka-zam** [comment]
Wouldn't it be better to move this functionality to a libleelaz instead? That way GTP can be out of the picture and pondering information simply polled when needed.

**Comment by gcp** [comment]
>Wouldn't it be better to move this functionality to a libleelaz instead?

I think this is a bad idea on several levels:

* Then the GUI needs to link to the engine, which has catastrophic licenses implications.
* Instead of one protocol that is reasonably well supported, now you need to define an API that no-one has ever used on either side.
* If am not even sure why you think that moving it to a library changes anything about what I described above wrt push/pull and races. It's the same thing with using multiple pipes, etc, you are just making things more complicated and moving the problem around.

**Comment by dwt** [comment]
@gcp: The library approach would allow registering a callback to get updates on pondering information. This is roughly equivalent to getting pondering info on another file descriptor. (But API, who wants an API when you can have a standard ASCII protocol...)

Studying the GTP protocol it seems that it is missing a command to explicitly start / stop pondering, something the UCI protocol supports. So maybe that is the best option?

To my understanding in UCI `go ponder infinite` would start pondering for infinite time, until it is stopped by `stop`. Maybe we should adopt a similar design?

I.e. we have something like this:

```
lz-ponder infinite
= ok
info: ....
info: ....
info: ....
info: ....
info: ....
lz-ponder stop
= ok
```

This would adopt the `uci` convention to prefix engine to gui communication with info, in a style that we still have to design (if there is no usable precedent in any of the other engines)

Looking at how GTP is implemented it seems that this should not clash with the protocol implementation. The protocol requires that an answer to any GTP command is prefixed by `=` but can be any number of lines (e.g. `showboard`, `list_commands`)

Outputting currently evaluated best move sequences during `genmove` should follow the `info` format to make it easier to parse.

Is this missing something?

**Comment by gcp** [comment]
>Looking at how GTP is implemented it seems that this should not clash with the protocol implementation. The protocol requires that an answer to any GTP command is prefixed by = but can be any number of lines (e.g. showboard, list_commands)

Oops, you are right here. But this means the example I gave with genmove b (without pondering) is wrong!

This means stderr is the only option to remain compatible. So I'll have to use that. If GTP_ShowRefutations and/or GTP_ShowCurrline are enabled, then give the "info xxxx" output on stderr. But this brings back the synchronization problem if there are two pipes. Maybe if the GUI considers it a potential issue (it may not really be one if it's just stats output), it can use the GTP id feature when giving the genmove command, and I can have Leela respond "info id foo xxxx" to reflect the GTP command the output corresponds to.

>But API, who wants an API when you can have a standard ASCII protocol...

Who wants standards when we can invent everything anew without apparently understanding the reasoning that went behind the original design, thereby ending up with something clearly worse.

**Comment by kuba97531** [comment]
> This means stderr is the only option to remain compatible. 

No, it hass all problems that you have mentioned. I suggest we use comments.

From specification:

2.9 Comments
Comments can be included in the command stream. All text between a hash sign (#) and the following newline is considered as comments and should be discarded on input.

**Comment by Ka-zam** [comment]
My original thought was to avoid pipes and have the GUI register a callback so that the engine could publish pondering info at will.

But, OK. A unique API is maybe not the best idea. But now we're back to stderr again which was rejected to start with?

**Comment by kuba97531** [comment]
Comments are only for inputs :(. Sorry for confusion.

**Comment by gcp** [comment]
>But now we're back to stderr again which was rejected to start with?

I'm not a fan if there are alternatives because:

a) You'll now get some intermixing between free form and protocol specified output on stderr. I think that's manageable because the "info blah" parts are GUI requested (so they won't appear on non-compatible GUIs), and a compatible GUI can filter out everything that doesn't start with ```info ``.

b) There are some potential synchronization issues when dealing with both streams, but they might be manageable by using the synchronization mechanisms already specified by GTP.

That said, the only alternative I see is that a GUI that enables "GTP_ShowRefutations" or "GTP_ShowCurrline" accepts that there might be ```info xxx``` output in between the output it'd normally expect from GTP.

Both of these work. Using stderr has the advantage that the interface design would stay more closely GTP compatible, and I got the feeling from @yishn that he perfers this method.

**Comment by Ka-zam** [comment]
All it takes is one message on stderr that's not terminated with \n and all of a sudden there's a huge mess. I noticed that beignet throws warnings on stderr today when asked for certain worksizes in the self tuning. Properly terminated though.

Named pipes? But I don't know how portable that is under other OS's apart from Unices.

**Comment by yishn** [comment]
@gcp To remain GTP compatible is nice, but I can't deny that in this case it would be better to have those `info` lines in stdout. But these should probably be turned off by default and opt-in only via a GTP command (or option). These commands could have names that explicitly state that they break GTP protocol, like a special kind of prefix.

**Comment by gcp** [comment]
Thank you very much for the feedback, this gives me something to work with.

This feature would be disabled unless the GUI explicitly enables them via the list_options + setoption combo that was listed above.

**Comment by marcocalignano** [comment]
But why not create a new command like `info` and the interested GUI can just send that command after the `genmove` command? Like that the output would be in stdout, and would berequested only from the GUI that needs that.

**Comment by gcp** [comment]
@marcocalignano it's already explained above why this is problematic...

**Comment by bittsitt** [comment]
> Oops, you are right here. But this means the example I gave with genmove b (without pondering) is wrong!
> 
> This means stderr is the only option to remain compatible.

@gcp You lost me here.. Isn't there just two scenarios we need to support: Pondering a move (used when analyzing e.g. a pro game) and pondering until a decision is made and then play move.(?)

Couldnt we add a  lz-ponder command like @dwt suggested and a lz-genmovewithponder:
lz-genmovewithponder
= info ...
info ...
G3

Reponses can be multi-line and terminated with \n\n. There is no requirement that a response is completed before sending a new command (that I can tell). Both commands are aborted if engine receives other command (which could be stop or move or whatever) (at which point engine will transmit \n\n before responding to new command).

Other guis will not be confused because they will not issue these commands..

**Comment by gcp** [comment]
>Couldnt we add a lz-ponder command

No, because existing GUIs won't send it, and it would make Leela appear weaker. Besides, the post you are replying about *specifically* talks about the *non*-pondering case.

**Comment by gcp** [comment]
>Pondering a move (used when analyzing e.g. a pro game)

This concept doesn't exist in GTP. You either genmove for the current move, or you play out a move and genmove for the opponent.

**Comment by bittsitt** [comment]
Sorry, I'm probably using the wrong terminology (or I'm still missing something). 

> No, because existing GUIs won't send it, and it would make Leela appear weaker.

So first of all, the current implementation of GTP is kept, so that existing GUIS will not notice a difference and continue to use play and genmove etc as today. We extend GTP for GUIs that want to support the extensions.

> This concept doesn't exist in GTP. You either genmove for the current move, or you play out a move and genmove for the opponent.

I'm suggesting that the GUI only sends the play command (use case 1). Nothing in GTP says that genmove and play has to alternate. After sending play, the GUI (if it wants to analyze a move) sends a  lz-analyze-and-display-winrates-playouts-and-variations-until-I-say-stop command. When LZ receives this it will reply with winrates, playout and n moves for it's m best variations in a multi-line command until it receives stop. In other words something like this:

```
play G3
=
play G4
= 
# At this points, user wants to look at move suggested by LZ
lz-analyze-and-display-winrate infinite
= info G6 156 23% D4 D6 A9 A7 A8 A6 C8|G5 22.5% D2 C3
info G6 156 24% D4 D6 B3 A7 A8 A6 C8|G5 22.6% D2 C4
info G6 156 24% D4 D6 B3 A7 A8 A6 C8|G5 22.6% D2 C4
# User moves to next
play G5
# This stops the analysis
=
# User wants to look at moves from LZ again
lz-analyze-and-display-winrate infinite
= info G6 156 23% D4 D6 A9 A7 A8 A6 C8|G5 22.5% D2 C3
info G6 156 24% D4 D6 B3 A7 A8 A6 C8|G5 22.6% D2 C4
info G6 156 24% D4 D6 B3 A7 A8 A6 C8|G5 22.6% D2 C4
```
If it is beneficial to let LZ search in the background while the user is moving back and forth (sending play and undo commands), that can be an option that is enabled with a lz-allow-background-analysis command.

**Comment by yishn** [comment]
@bittsitt Wow, I like this idea! Not only the controller can know exactly when to expect the info lines, this way the GTP will be completely valid as well.

**Comment by gcp** [comment]
The problem I see with the proposal is that you can't show the engine thinking while you are playing? Like, what do you do on genmove?

**Comment by bittsitt** [comment]
You have a special genmove that outputs the same information:

```
play G4
=
lz-genmove-while-displaying-winrate white
= info G6 156 23% D4 D6 A9 A7 A8 A6 C8|G5 22.5% D2 C3
info G6 189 24% D4 D6 B3 A7 A8 A6 C8|G5 80 22.6% D2 C4
G6
play G7
=
= info G8 156 23% D4 D6 A9 A7 A8 A6 C8|G5 22.5% D2 C3
info G8 172 24% D4 D6 B3 A7 A8 A6 C8|G5 22.6% D2 C4
G8
```
If you don't want output, you use genmove, if you want output, you use this instead.

The info line format is supposed to be:
```
info <best move> <playouts> <winrate> <n best variation moves> | <2nd best move> <playouts> etc 
```
So you get a new info move at some interval (configurable?)

**Comment by bittsitt** [comment]
And if you want to see what the engine is thinking while you are thinking:
```
play G4
=
lz-genmove-while-displaying-winrate white
= info G6 156 23% D4 D6 A9 A7 A8 A6 C8|G5 22.5% D2 C3
info G6 189 24% D4 D6 B3 A7 A8 A6 C8|G5 80 22.6% D2 C4
G6
lz-analyze-and-display-winrate infinite
= info G6 156 23% D4 D6 A9 A7 A8 A6 C8|G5 22.5% D2 C3
info G6 156 24% D4 D6 B3 A7 A8 A6 C8|G5 22.6% D2 C4
play G7
lz-genmove-while-displaying-winrate white
= info G9 156 23% D4 D6 A9 A7 A8 A6 C8|G5 22.5% D2 C3
<etc>
```

gcp:
Are there no races with this model?

If you send a play to the engine, and you get an info back at the same moment, how do you know which move it refers to?

bittsitt:
If you send play, the gui has to wait for end of response (\n\n) and then response to the play command (=) before he will get winrates for the next move. So in the protocol itself, there should not be a problem.

But yes, the engine needs to be careful with race conditions so that he doesn't send end of response in the middle of info data e.g.. You might need to synchronize this some way if there are different threads handling this.

I think this is a problem you will have in a stderr solution as well.

gcp:
I think the races probably aren't a problem in LZ itself. So this could work.

gcp:
Going to close this, let's continue with #1388.

PR #1244, opened 2018-04-23
Make maximum tree size a configurable setting.

TFiFiE:


killerducky:
I would prefer a UCI option, `Hash` seems appropriate.

http://wbec-ridderkerk.nl/html/UCIProtocol.html
Hash, type is spin the value in MB for memory for hash tables can be changed,
this should be answered with the first "setoptions" command at program boot
if the engine has sent the appropriate "option name Hash" command,
which should be supported by all engines!
So the engine should use a very small hash first as default.

ihavnoid:
@gcp I see why you want to have it a GTP command and I agree it's a good idea.  However, by the same logic, shouldn't we also have things like number of playouts also be GTP-programmable?

Regarding memory usage, the only way (I think) we can have it reflect memory usage accurately is to count the number of inflated nodes, and then calculate the tree size as something like nodes * (sizeof(UCTNode) + sizeof(Network::UCTNodePointer) * 360) although the final constant 300 may be a bit questionable.  What do you think?

If you are not into changing the metric, we can empirically determine the memory footprint by trying random tree sizes and see the trend from it.  Not accurate but possibly good enough...

gcp:
>shouldn't we also have things like number of playouts also be GTP-programmable?

Yes! I need to get around to doing an example implementation of a few of those together with the stuff Lizzie needs.

Something like list_options, working similary to "uci" in chess engine. After which you can do setoption name blah value foo.

gcp:
> Not accurate but possibly good enough...

That should be fine. I don't think users are going to complain if they ask for 4G and we use 5G. But if they want 4G, do you know offhand how many nodes that is? I don't!

gcp:
Discussion about option setting support is in #1242, closing this as it won't be merged as is.

Issue #1185, opened 2018-04-11
Time management bug?


Marcin1960:
I used big net, without --timemanage option. My hardware is weak, so trying to give more time is critical.

At my surprise my bot having 20 second for a move was leaving early almost each move, after 10 second or 7 seconds and losing game after game. Guys told me that the best move was already found by timemanage so nothing can be done about.

OK, I put "--timemanage off". Sure enough all 20 sec were used and my bot started to win!

The conclusion - timemanage evaluation (when to leave early), does not work!

Marcin1960:
BTW, I get this strange discrepancy in evaluation output:

Thinking at most 18.5 seconds...
NN eval=0.335862
Playouts: 251, Win: 66.52%, PV: O17 N16 P19 S15 S14 S16 ....
Playouts: 488, Win: 66.61%, PV: O17 N16 P19 S15 S14 S16 ...

When "Win:" gets into 90s%, NN is in single digits. Is it correct?

Marcin1960:
It happened each time, out of 10 games maybe.

Not always big evidence is necessary, sometimes logical reasoning should suffice. If I use big net like 192x15 on my PC I can get 2000 playouts in 20 seconds with "--timemanage off".

Otherwise, I get about 1000 playouts most of the moves. Is it possible that so few playouts are sufficient?
When I gave 2 minutes per move with 20x256 net, the engine was leaving early most of the moves after about 60 seconds.

gcp:
The point of timemanage is:

1) To stop thinking when it's 100% sure the best move can't change.
2) To leave more time for the next moves.

So "logical reasoning" says this must be invalid. I might be wrong, but you need data, not vague anecdotes.

Marcin1960:
What type of data do you need?

killerducky:
Marcin1960, I recommend you just run with `--timemanage off`. I believe you are running with very little main time, and byoyomi period of 20 seconds. In this mode the advantage of using (almost) all 20 seconds is to push more moves into the tree for reuse in the next move. The disadvantage is your opponent can think more during that time also, and your human opponent might be annoying you spend extra timing thinking on "obvious" replies. Note that many strong humans use their entire 20 seconds even if they decide on their move very early on.

Also note that with large main times it's more complicated what you should do. I think I saw another issue discussing changing how much time is allocated per move in that case.

[...]

PR #1252, opened 2018-04-24
Don't use time management if there's no benefit.

gcp:
Don't move quickly if we cannot save up time anyway. Can be overruled by
setting --timemanage forced.

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -519,22 +519,34 @@ size_t UCTSearch::prune_noncontenders(int elapsed_centis, int time_for_move) {
 }
 
 bool UCTSearch::have_alternate_moves(int elapsed_centis, int time_for_move) {
-    if (cfg_timemanage != TimeManagement::ON) {
+    if (cfg_timemanage == TimeManagement::OFF) {
         return true;
     }
     auto pruned = prune_noncontenders(elapsed_centis, time_for_move);
-    if (pruned == m_root->get_children().size() - 1) {
-        // In a timed search we will essentially always exit because
-        // the remaining time is too short to let another move win, so
-        // avoid spamming this message every move. We'll print it if we
-        // save at least half a second.
-        if (time_for_move - elapsed_centis > 50) {
-            myprintf("%.1fs left, stopping early.\n",
-                     (time_for_move - elapsed_centis) / 100.0f);
-        }
-        return false;
+    if (pruned < m_root->get_children().size() - 1) {
+        return true;
     }
-    return true;
+    // If we cannot save up time anyway, use all of it. This
+    // behavior can be overruled by setting "fast" time management,
+    // which will cause Leela to quickly respond to obvious/forced moves.
+    // That comes at the cost of some playing strength as she now cannot
+    // think ahead about her next moves in the remaining time.
+    auto my_color = m_rootstate.get_to_move();
+    auto tc = m_rootstate.get_timecontrol();
+    if (!tc.can_accumulate_time(my_color)) {
+        if (cfg_timemanage != TimeManagement::FAST) {
+            return true;
+        }
+    }
+    // In a timed search we will essentially always exit because
+    // the remaining time is too short to let another move win, so
+    // avoid spamming this message every move. We'll print it if we
+    // save at least half a second.
+    if (time_for_move - elapsed_centis > 50) {
+        myprintf("%.1fs left, stopping early.\n",
+                    (time_for_move - elapsed_centis) / 100.0f);
+    }
+    return false;
 }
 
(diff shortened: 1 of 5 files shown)


gcp:
Comments and reviews welcome!

roy7:
In case you didn't see my comment elsewhere, an idea to try and benefit from timemanage in byoyomi is to focus all search into the most visited move if it's too late to change moves. I normal time we'd just move, but in byoyomi we keeps searching. But why search a better #2 move we don't have time to let become #1? Put all search into #1.

(Works against the spirit of max_root_lcb, but dealing with /next as it stands today, I think this helps allocate search better than no time manage at all during byoyomi.)

src/UCTSearch.cpp
```diff
@@ -519,22 +519,31 @@ size_t UCTSearch::prune_noncontenders(int elapsed_centis, int time_for_move) {
...
     auto pruned = prune_noncontenders(elapsed_centis, time_for_move);
-    if (pruned == m_root->get_children().size() - 1) {
-        // In a timed search we will essentially always exit because
-        // the remaining time is too short to let another move win, so
-        // avoid spamming this message every move. We'll print it if we
-        // save at least half a second.
-        if (time_for_move - elapsed_centis > 50) {
-            myprintf("%.1fs left, stopping early.\n",
-                     (time_for_move - elapsed_centis) / 100.0f);
+    if (pruned > m_root->get_children().size() - 1) {
```
Hersmunch: typo - should be <

gcp:
>is to focus all search into the most visited move if it's too late to change moves.

This code does *not* disable prune_noncontenders. It just disables stopping the search.

src/UCTSearch.cpp
gcp: Well, <= or < size(), I guess.

TFiFiE:
I object to the semantics that have "not wasting time" (e.g. by using up the entire byoyomi period when in it) equaling "time management being turned off"; this is just bound to confuse things. I believe that a straightforward meaning of "time management" is that the portion of the available search time that is used is not fixed and can vary, **with the engine still being smart enough not to let "free" search time go to waste**. If you're going to have this new option (which I consider to be of dubious value), at least call it something like "wasteful" rather than "forced".

src/UCTSearch.cpp
gcp: <= or < size(), I think.

gcp:
>call it something like "wasteful" rather than "forced".

Is it wasting the users' time or the opportunity to find a better move?

I don't think this naming is any better.

src/UCTSearch.cpp
Hersmunch: Wouldn't that mean that all children need to be pruned to stop? That can't happen because the child with the most visits will always have enough to survive the pruning.

gcp:
Maybe something like FAST? Conveys what will happen (and what the downside could be)

src/UCTSearch.cpp
```diff
@@ -519,22 +519,31 @@ size_t UCTSearch::prune_noncontenders(int elapsed_centis, int time_for_move) {
...
     auto pruned = prune_noncontenders(elapsed_centis, time_for_move);
-    if (pruned == m_root->get_children().size() - 1) {
-        // In a timed search we will essentially always exit because
-        // the remaining time is too short to let another move win, so
-        // avoid spamming this message every move. We'll print it if we
-        // save at least half a second.
-        if (time_for_move - elapsed_centis > 50) {
-            myprintf("%.1fs left, stopping early.\n",
-                     (time_for_move - elapsed_centis) / 100.0f);
+    if (pruned < m_root->get_children().size()) {
```
Hersmunch: (In case the notification of my other reply disappeared)
Wouldn't this mean that all children need to be pruned to stop? That can't happen because the child with the most visits will always have enough to survive the pruning.

TFiFiE:
I still can't see why anyone would ever want to use this option. Why have the time management be dynamic but still dumb in this one respect? Why not just use a different time control then?

gcp:
It was requested in the original issue #1185. Playing faster when there's an obvious move just feels more pleasant and humanlike, even if sub-optimal for strength.

killerducky:
One more thought, if both you and your opponent are pondering, it's possible that playing faster even in Japanese Byo-yomi is better. Imagine your opponent has a tough decision and plays 10 time-suji forcing moves in a row. Thinking for your entire period to play obvious answers is giving your opponent an advantage. 

I don't have data proving the general case, but mostly I prefer it just because I find it annoying to wait the whole period.

bjiyxo:
Haha, I made a mistake. I will test with same time now.

gcp:
> I will test with same time now.

Needs to be byo yomi time to take effect. If it does, it can't really be worse.

PR #1262, opened 2018-04-25
fp16 datatypes for OpenCL evaluation

ihavnoid:
Revived the USE_HALF.

The datatypes are fp16 in global memory.  The actual computations are still fp32 because many consumer-grade cards are still not so good at fp16 yet.

Constant number of playouts (4000)
```
50/50 games played

               A     B
A leelaz-base       24-26
B leelaz-fp16 26-24

leelaz-base v leelaz-fp16 (50/50 games)
board size: 19   komi: 7.5
leelaz-base     24 48.00%       12 48.00%     12 48.00%    505.46
leelaz-fp16     26 52.00%       13 52.00%     13 52.00%    507.16
                                25 50.00%     25 50.00%

player b: Leela Zero:0.13
player w: Leela Zero:0.13
```

Constant time (10 seconds, 2x GTX1080, 12 threads)
```
50/50 games played

               A     B
A leelaz-base       19-31
B leelaz-fp16 31-19

leelaz-base v leelaz-fp16 (50/50 games)
board size: 19   komi: 7.5
leelaz-base     19 38.00%       7  28.00%     12 48.00%   1480.88
leelaz-fp16     31 62.00%       13 52.00%     18 72.00%   1496.12
                                20 40.00%     30 60.00%

player b: Leela Zero:0.13
player w: Leela Zero:0.13
```

The last time fp16 was merged it was a compile-time option which was broken multiple times while developing other code.  Hence I am not sure if this should be merge or not...  Any thoughts?

### DIFF
--- a/src/Tuner.cpp
+++ b/src/Tuner.cpp
@@ -45,16 +45,31 @@
 #include <cblas.h>
 #endif
 
+#ifdef USE_HALF
+const auto TUNER_FILE_LOCAL = std::string("leelaz_opencl_tuning_half");
+constexpr auto MAX_ERROR = 1e-2f;
+#else
 const auto TUNER_FILE_LOCAL = std::string("leelaz_opencl_tuning");
 constexpr auto MAX_ERROR = 1e-4f;
+#endif
 
 using namespace Utils;
 
-static void sgemmBatched_ref(const std::vector<float>& a,
-                             const std::vector<float>& b,
-                             std::vector<float>& c,
+static void sgemmBatched_ref(const std::vector<net_t>& a,
+                             const std::vector<net_t>& b,
+                             std::vector<net_t>& c,
                              const int m, const int n, const int k,
                              const int batch_size) {
+    std::vector<float> ar(a.size());
+    std::vector<float> br(b.size());
+    std::vector<float> cr(c.size());
+
+    for(auto i=size_t{0}; i<a.size(); i++) {
+        ar[i] = a[i];
+    }
+    for(auto i=size_t{0}; i<b.size(); i++) {
+        br[i] = b[i];
+    }
     for (auto batch = 0; batch < batch_size; batch++) {
         auto offset_u = batch * m * k;
         auto offset_v = batch * n * k;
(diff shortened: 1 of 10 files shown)


gcp:
>The last time fp16 was merged it was a compile-time option which was broken multiple times while developing other code. Hence I am not sure if this should be merge or not... Any thoughts?

We can probably compile (even if not test) it in CI?

src/Tuner.cpp
```diff
@@ -45,16 +45,31 @@
...
+                             const std::vector<net_t>& b,
+                             std::vector<net_t>& c,
                              const int m, const int n, const int k,
                              const int batch_size) {
+    std::vector<float> ar(a.size());
+    std::vector<float> br(b.size());
+    std::vector<float> cr(c.size());
+
+    for(auto i=size_t{0}; i<a.size(); i++) {
+        ar[i] = a[i];
```
gcp: Wouldn't std::copy work?

src/Tuner.cpp
```diff
@@ -45,16 +45,31 @@
 #include <cblas.h>
 #endif
 
+#ifdef USE_HALF
+const auto TUNER_FILE_LOCAL = std::string("leelaz_opencl_tuning_half");
+constexpr auto MAX_ERROR = 1e-2f;
```
gcp: Ouch! Guess it's more than enough to detect driver breakage regardless.

You could consider using the same file and just storing it under the "XgemmBatchedHalf" key in the file.

gcp:
You're missing a file:
fatal error: half/half.hpp: No such file or directory

gcp:
I have fixes for most of these, I'll take a try to incorporate it into CI.

It's a pretty big speedup even on my non fp16 card. Given that there appears to be no regression in play, we should perhaps consider defaulting this.

ihavnoid:
@gcp - Thanks to get the rest of the required fixes in there.  I was a bit concerned because I am going through a family emergency which is resulting in myself unable to work on any personal SW development for a couple of month.

I would like to give you a big thank you for creating such a project that I could learn a lot!

PR #1266, opened 2018-04-26
Make CMakeLists.txt file consistently formatted 

MKolman:
In large code bases consistency is important. This PR changes all CMake commands to be upper case with the exception of the command `if`.

If these kind of style PRs are welcomed, I would like to submit more to fix the styling of files in `src/`. Eventually I wish to bring in a script like Google's `cpplint.py` as a part of CI to force style checks in all CI runs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,27 +12,27 @@
 # You should have received a copy of the GNU General Public License
 # along with Leela Zero.  If not, see <http://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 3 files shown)


gcp:
In general I like this. But I'm finding a few references saying that modern CMake style generally prefers lower case. So I'll ask at the risk of making you pull your hair out, could you make the opposite change?

MKolman:
Yeah, that's no problem. I only used upper case, because that was the style you used. Will fix it in a moment.

MKolman:
Changed all to lower case.

PR #1274, opened 2018-04-27
Suppress Clang warning in half.cpp.

TFiFiE:
When using `-DUSE_HALF`, Clang will otherwise give me the warning "`class 'numeric_limits' was previously declared as a struct [-Wmismatched-tags]`".


gcp:
This file is from an upstream, so I'd like to avoid patches to it if possible.

I'll change the CMake file to add -Wno-mismatched-tags.

TFiFiE:
What about `src/Makefile`?

Issue #1084, opened 2018-03-23
First and second move symmetries decrease exploration efficiency


Vandertic:
I just wanted to remark this fact in view of a second run of the training from scratch. Maybe I am wrong, but it seems to me that this could be a nasty problem.

When LZ builds the tree it doesn't know about symmetries. Suppose at some point in the training the policy network gives a high probability (say 80%) to the four hoshis: then UCT search must explore all of them separately, 20% of playouts each, without averaging knowledge between the four branches and without going as deep as it could in each of these branches.

Moreover, suppose the second most probable positions are komokus (4-3) (say 16% probability). They will have their probability split in 8 (2% each!) and so will be explored much less.

With limited number of playouts/visits, first moves with 8 symmetries will have shallower exploration than moves with 4 symmetries and much lower **visit number**.

So it is not at all surprising that the first chosen move converges to P(hoshi)=100%.

I would recommend to write special code for the first (and second) move, in order for LZ to exploit symmetries (just in this crucial moment of the game).

For the first move it would be enough to allow only moves in the NNE eighth of the goban. (Which is also tradition and etiquette.)

Am I misunderstanding something or this could be a real issue?

Eddh:
I've already thought about it. It is certainly a bias away from "duplicated" moves, but I don't think it is really problematic. If the duplicate moves are really better than the unique one, then they will get higher winrate and get picked more and the unique one will tend to disappear.

gcp:
Already discussed before, not interesting, not worth bothering about.

As Eddh explained, if the moves were actually better, then the split policy prior does not matter. We already went from 4-4 to 4-3 in this regard.

PR #1275, opened 2018-04-27
Look for symmetrical position in cache.

TFiFiE:
As suggested in https://github.com/gcp/leela-zero/issues/1084#issuecomment-375668610.

I took the liberty of changing the benchmark position to one in which this doesn't cause a speedup, because it's not typical for the game as a whole.

### DIFF
--- a/src/FastBoard.h
+++ b/src/FastBoard.h
@@ -72,7 +72,7 @@ public:
     int get_boardsize(void) const;
     square_t get_square(int x, int y) const;
     square_t get_square(int vertex) const ;
-    int get_vertex(int i, int j) const;
+    int get_vertex(int x, int y) const;
     void set_square(int x, int y, square_t content);
     void set_square(int vertex, square_t content);
     std::pair<int, int> get_xy(int vertex) const;
@@ -124,6 +124,8 @@ protected:
 
     int m_boardsize;
     int m_squaresize;
+
+    std::array<std::array<unsigned short, MAXSQ>, 8> m_symmetry_idx;
 
     int calc_reach_color(int color) const;
 
--- a/src/FastBoard.h
+++ b/src/FastBoard.h
@@ -27,6 +27,8 @@
 #include <utility>
 #include <vector>
 
+#include "Utils.h"
+
 class FastBoard {
     friend class FastState;
 public:
@@ -125,7 +127,7 @@ protected:
     int m_boardsize;
     int m_squaresize;
 
-    std::array<std::array<unsigned short, MAXSQ>, 8> m_symmetry_idx;
+    std::array<std::array<unsigned short, MAXSQ>, NUM_SYMMETRIES> m_symmetry_idx;
 
     int calc_reach_color(int color) const;
 
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -91,7 +91,7 @@ static std::array<float, 1> ip2_val_b;
 static bool value_head_not_stm;
 
 // Symmetry helper
-static std::array<std::array<int, BOARD_SQUARES>, 8> symmetry_nn_idx_table;
+static std::array<std::array<int, BOARD_SQUARES>, NUM_SYMMETRIES> symmetry_nn_idx_table;
 
 void Network::benchmark(const GameState* const state, const int iterations) {
     const auto cpus = cfg_num_threads;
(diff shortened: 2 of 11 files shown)


gcp:
How does the performance look?

TFiFiE:
This gives a speed boost for the first three moves or so and any performance hit afterwards is too insignificant for me to detect.

gcp:
The other PR broken this, but I'll need a bit to review this anyway, wrt calc_hash changes, performance impact and the need for the rotation table (which relates to performance impact again).

gcp:
@TFiFiE can you do a me a favor and rebase this?

src/FastBoard.h
```diff
@@ -125,6 +125,8 @@ class FastBoard {
     int m_boardsize;
     int m_squaresize;
 
+    std::array<std::array<unsigned short, MAXSQ>, 8> m_symmetry_idx;
```
Hersmunch: How much of a pain would it be to make it so there is only one of these tables? Or is there a reason why it makes sense to have one for each object?

TFiFiE: It's tied to the board size, which is variable for this class.

Hersmunch: Of course. Can totally see it now, thanks.

gcp: If we can live with an extra indirection you can make it a static class owned vector (indexed by board size) and just check if that is initialized the moment you initialize FastBoard.

Alternatively, do the calculation on the fly...if necessary limit the symmetry check to the first 10-20 moves or so?

That's some of the things I want to look at.

src/Network.cpp
```diff
@@ -868,8 +870,19 @@ Network::Netresult Network::get_scored_moves(
 
     if (!skip_cache) {
         // See if we already have this in the cache.
-        if (NNCache::get_NNCache().lookup(state->board.get_hash(), result)) {
-            return result;
+        for (auto symmetry = 0; symmetry < 8; ++symmetry) {
```
Hersmunch: Name clash now that `rotation` parameter has been renamed. Sorry about that..

AncalagonX:
I have been running this in my custom LZ build (essentially mirrors LZ_next + LCB_max_root) for a week now with Lizzie over many, many 9dan vs. LZ games.

I have noticed no downsides or performance decreases. In the beginning, it massively boosts reported visit counts in all four corners as expected.

Although I cannot personally confirm one way or the other whether the implementation has any bugs in aggregating symmetrical moves or truly results in better move selection, I can say that: 1) I have noticed no errors, misplays, or bizarrities from enabling this PR, which presumably is in effect for the entire game. 2) I have noticed no decrease in n/sec overall, although for this point I have only conducted the bare minimum of testing.

If there are no identifiable problems in how this is implemented, this PR could be enabled for selfplay and provide a mild boost to early-game search strength.

Though **do note that** -v 3200 might interfere with the greatly boosted visit counts from this PR—and so perhaps a special allowance of extra visits for early game symmetrical positions should be given, or else a "true visits" value should be tallied separately from the reported "true+symmetrical position visits" search results for the purposes of correctly enforcing a full search with 3200 real visits. Otherwise, using the visit counts directly from this PR might result in early-game searches erroneously ending search calculations hundreds or thousands of visits earlier than the 3200 mark.

This is a simple issue to address in any case, but I cannot fully interpret the code above on my own so perhaps this effect has already been accounted for.

gcp:
This doesn't interfere with visit counting AFAIK, just makes it faster.

I basically only want to see if we can get the extra array out of FastBoard, and if getting rid of it would ( to keep performance parity) require limiting this to the first moves of the game.

TFiFiE:
Why would you want such a change if only one `FastBoard` object is typically constructed and `FastBoard::reset_board` isn't called all the time?

AncalagonX:
I've only conducted rather shallow testing on the issue, but I see no performance impacts in long games with this enabled.

That said, it's pretty clear that there is negligible benefit beyond the first x number of moves. I'd be happy with a max of 10 moves, just judging by gut feelings. If we are limiting this, considerations may need to be made for symmetrical play involving pre-placed handicap stones depending on how LZ counts which move it's on internally.

gcp:
>if only one FastBoard object is typically constructed

Because it might be quite useful to construct multiple, for example if asked to list dead stones.

gcp:
I guess this should be disabled during training? Else we can't take advantage of rotational asymmetry in the opening.

gcp:
I reworked this a bit in #1275.

gcp:
I can't measure a performance difference, and the other pull addresses the self-play games, so going to close this.

TFiFiE:
@gcp: Can you reopen this?

PR #1279, opened 2018-04-28
Renamed rotation to symmetry

Hersmunch:


TFiFiE:
Conflicts with #1275.

PR #1281, opened 2018-04-28
Limit the NNCache size when -v used

### LINKED ISSUE #1193: NNCache is maximum size when not using -p
When `-p` isn't used `cfg_max_playouts` is left defaulted to:
`    cfg_max_playouts = std::numeric_limits<decltype(cfg_max_playouts)>::max();`
Which means when the NNCache size is set in `init_global_objects()`
`    NNCache::get_NNCache().set_size_from_playouts(cfg_max_playouts);`
```
void NNCache::set_size_from_playouts(int max_playouts) {
    // cache hits are generally from last several moves so setting cache
    // size based on playouts increases the hit rate while balancing memory
    // usage for low playout instances. 50'000 cache entries is ~250 MB
    auto max_size = std::min(50'000, std::max(6'000, 3 * max_playouts));
    NNCache::get_NNCache().resize(max_size);
}
```
will result in:
* the multiply by 3 wrapping to 0x7FFFFFD
* the cache being maximum allowed size

From the comment this doesn't sound like the intention. Of course reducing the size probably will make games take a bit longer..
I know they are not directly comparable but in this case would a simple min(cfg_max_playouts, cfg_max_visits) suffice? Should the maximum size be used when neither `-p` or `-v` are specified, regardless of the time settings? Though I suppose if time is short, a larger cache would be more helpful.

For reference I ran a few games with `-v 3200` and observed the cache being full around move 83ish. With a cache size forced to be 6000 it is full on move 4.

Hersmunch:
Fixes #1193.
Uses the -v max visits as an overestimate for the number of playouts per move.
Also avoid potential overflow in the size calculation.


src/Leela.cpp
```diff
@@ -305,7 +305,9 @@ void init_global_objects() {
     // improves reproducibility across platforms.
     Random::get_Rng().seedrandom(cfg_rng_seed);
 
-    NNCache::get_NNCache().set_size_from_playouts(cfg_max_playouts);
+    // When visits are limited ensure cache size is still limited.
+    auto playouts = std::min(cfg_max_playouts, cfg_max_visits);
```
gcp: I believe it's currently legal to set -v 3200 -p 0 and have this mean 3200 visits with no playout limit. At some point AutoGTP and the server made use of this because -p 1600 was defaulted somewhere in AutoGTP and needed to be overruled.

This already went wrong here, but this change would make the problem worse.

src/NNCache.cpp
```diff
@@ -77,7 +77,7 @@ void NNCache::set_size_from_playouts(int max_playouts) {
     // cache hits are generally from last several moves so setting cache
     // size based on playouts increases the hit rate while balancing memory
     // usage for low playout instances. 150'000 cache entries is ~225 MB
-    auto max_size = std::min(150'000, std::max(6'000, 3 * max_playouts));
+    auto max_size = 3 * std::min(50'000, std::max(2'000, max_playouts));
```
gcp: Not clear why this is being changed. It's awkward now because the 150`000 from the comment is now implicit.

src/Leela.cpp
Hersmunch: Thanks for pointing that out, I hadn't spotted it. 
I'll change to how `-p 0` is handled so it is defaulted to int max when the arg is parsed. Then the two places that use `cfg_max_playouts` can be changed to avoid potential overflow as they each need.

src/NNCache.cpp
Hersmunch: It is avoiding the potential overflow when `-p` isn't used and the `cfg_max_playouts` is left defaulted to int max (or `-p` is used and anything at least int max / 3 is specified). It seemed awkward to test the int max / 3 limit only to then limit the result again. Thinking about it again, I agree that it is clearer to have the max limit written in code to match the comment, in terms of as cache entries rather than as playouts, so I'll just do it the "awkward" way.

Hersmunch:
Note that before this PR when `-p 0` is specified by AutoGTP that the cache size will be set to the minimum of 6000 entries. With this PR it will be set to 3*3200 entries.

src/Leela.cpp
Hersmunch: Fixed elsewhere.

src/NNCache.cpp
```diff
@@ -77,7 +77,13 @@ void NNCache::set_size_from_playouts(int max_playouts) {
     // cache hits are generally from last several moves so setting cache
     // size based on playouts increases the hit rate while balancing memory
     // usage for low playout instances. 150'000 cache entries is ~225 MB
-    auto max_size = std::min(150'000, std::max(6'000, 3 * max_playouts));
+    constexpr auto num_cache_moves = 3;
+    auto max_playouts_per_move =
+        std::min(max_playouts,
+            std::numeric_limits<decltype(max_playouts)>::max() /
+            num_cache_moves);
```
gcp: I think, following the reasoning, this should be / 2*num_cache_moves (because /2 is the real max playouts). But it makes no difference as it's capped below anyway.

PR #1287, opened 2018-04-29
Break up Dockerfile into multiple stages

MKolman:
As a general rule each Dockerfile should be responsible for one function. In that manner the original Dockerfile was split into 4:
* Dockerfiles/Dockerfile.base
  * Installs all dependencies and copies the source into the container
  * All other images should use this one as their base image
* Dockerfiles/Dockerfile.gpu
  * Builds `leelaz` binary for GPU
* Dockerfiles/Dockerfile.gpu-half
  * Builds `leelaz` binary for GPU with 16 bit memory
* Dockerfiles/Dockerfile.cpu
  * Builds `leelaz` binary for CPU
* Dockerfiles/Dockerfile.tests
  * Builds and runs `tests` on CPU

# Travis CI stages
Better utilization of Travis CI stages. This way it is easy to see which component fail and where. There are three stages:
* Build (all three builds run in parallel)
  * Build for GPU
  * Build for GPU-half
  * Build for CPU
* Build and run tests
* Check cpplint

The following image shows Travis interface when testing with the new config:
screenshot from 2018-04-29 11-12-23


gcp:
Looks good to me, @bood please review?

bood:
Only concern is put test in a different docker file will cost more time to compile since cpu version is compiled twice. Not worth it. Putting build and test together is good enough.

MKolman:
On Travis tests in the same stage are run in parallel (so currently all 3 build jobs are run in parallel), but stages are run sequentially (build > tests > style). Would it be enough to move the job from `tests` stage into `build`? This way it can be run in parallel without waiting for two sequential docker builds.

gcp:
I think that sounds OK and in line with what @bood suggested.

PR #1289, opened 2018-04-29
Fix 3 linting errors.

MKolman:
Fix and enable linting of the following errors:
- whitespace/indent,
- whitespace/parens,
- legal/copyright.

# Input
Please ensure that added Copyright information is as you desire.

Can you please write what is your desired size of pull request for style fixes i.e. should I do more or less line changes in a single pull request? Basically it's a question of reviewability vs. how fast can everything be fixed.


gcp:
This seems about right, the problem is more about lints where I don't necessarily agree with the Google style guide (like here, the private/public indents in a class). I might change my mind because consistency is nice.

gcp:
Would it be a lot of work to undo the "2 space indent"? I'm OK with everything else?

gcp:
I "fixed up" (haha!) cpplint a bit.

Issue #1259, opened 2018-04-25
Time Control: Save time on first few moves


bood:
Currently, each move is given remaining_time/moves_remaining in main time. See:
https://github.com/gcp/leela-zero/blob/next/src/TimeControl.cpp#L186-L190
and
https://github.com/gcp/leela-zero/blob/next/src/TimeControl.cpp#L220-L227

This seems to be not optimal for Go in the first few moves. For example, in a 15 minutes match with time management on, first move would be given 15*60/(19*19/9) = ~22 seconds. Noticing several moves have almost the same visits so time management cannot help here.

Some hack code could be written like giving first 5 - 10 moves a max thinking time. But I wonder if there is a general way to optimize this a bit and save time for later use.

```
Leela: genmove b
Thinking at most 22.5 seconds...
NN eval=0.476264
Playouts: 87, Win: 47.78%, PV: C16 Q4 R3 Q3 R4
Playouts: 188, Win: 47.79%, PV: C16 Q4 R3 R4 Q3 P3
Playouts: 298, Win: 47.75%, PV: D17 Q4 R3 R4 Q3 P3 P2
Playouts: 411, Win: 47.77%, PV: D3 Q16 R17 R16 Q17 P17 P18 O18
Playouts: 565, Win: 47.77%, PV: Q17 D4 C3 C4 D3 E3 E2 F2
Playouts: 690, Win: 47.78%, PV: Q17 D4 C3 C4 D3 E3 E2 F2
Playouts: 814, Win: 47.78%, PV: D3 Q16 R17 R16 Q17 P17 P18 O18
Playouts: 964, Win: 47.78%, PV: D17 Q4 R3 R4 Q3 P3 P2 O2 O3

 D17 ->     129 (V: 47.77%) (N:  9.83%) (LCB: 34.20%) (UCB: 61.58%) PV: D17 Q4 R3 Q3 R4 R5 S5 S6 R6
  D3 ->     127 (V: 47.81%) (N:  9.51%) (LCB: 34.13%) (UCB: 61.72%) PV: D3 Q16 R17 R16 Q17 P17 P18 O18 O17
 R16 ->     127 (V: 47.79%) (N:  9.63%) (LCB: 34.12%) (UCB: 61.70%) PV: R16 D4 C3 D3 C4 C5 B5 B6 C6
[...]

PR #1312, opened 2018-05-03
Support Facebook's ELF Open Go networks.

gcp:
Although the weights for those networks can be converted to be Leela
Zero compatible, their value output is for black instead of the side to
move. As this can't be fixed up in the network itself, add support for
a new network format version to cover those.

Note that the training code isn't changed, so training such a network
further is currently not possible.

See issue #1311.


src/Network.cpp
```diff
@@ -325,11 +330,18 @@ std::pair<int, int> Network::load_network_file(const std::string& filename) {
...
             return {0, 0};
         } else {
-            assert(format_version == FORMAT_VERSION);
+            // Version 2 networks are identical to v1, except
+            // that they return the score for black instead of
+            // the player to move. This is used by ELF Open Go.
+            if (format_version == 2) {
+                value_head_not_stm = true;
+            } else {
+                value_head_not_stm = true;
```
Ttl: This should probably be false for v1?

gcp: Uh. Yeah.

gcp:
@Ttl if we're going to define a weight format v2...I'm not sure if we should change things to fix up the whole conv_bias / batchnorm scale/center situation.

Do you have any thoughts on that?

I guess that trying to train the ELF network further may reveal that we have no other choice? In that case, taking this quick fix here seems OK.

gcp:
First results at -v800:

```
leelaz-18e6 v leelaz-elf (9/1000 games)
board size: 19   komi: 7.5
leelaz-18e6      3 33.33%       2 40.00%     1 25.00%     81.10
leelaz-elf       6 66.67%       3 75.00%     3 60.00%    127.91
                                5 55.56%     4 44.44%
```

Doesn't look like this is going 198-2. Is there some breakage remaining?

killerducky:
I reckon it's due to differences in the search algorithm (exact method of init to parent Q is different, FPU reduction, etc).

roy7:
> I reckon it's due to differences in the search algorithm (exact method of init to parent Q is different, FPU reduction, etc).

We might want to see how they init parent Q and do FPU and so on if it makes such a big difference in win rate. Those are changes LZ should be able to fold right in.

jkiliani:
I think the discrepancy looks kind of large for only differences in the search algorithm. Can we compare direct network inference output from OpenGo to that from Leela Zero in an identical position?

gcp:
>I reckon it's due to differences in the search algorithm (exact method of init to parent Q is different, FPU reduction, etc).

It would be quite something for that to explain such a huge performance difference.

> Can we compare direct network inference output from OpenGo to that from Leela Zero in an identical position?

I think that's the right idea, yes.

Mardak:
If you could also bump AUTOGTP_VERSION in this same PR, people on `next` could start getting generating self-play if https://github.com/gcp/leela-zero-server/pull/77 is merged.

https://github.com/gcp/leela-zero/blob/97c2f8137a3ea24938116bfbb2b0ff05c83903f0/autogtp/Management.h#L33

gcp:
@Mardak I am not sure why you want to increment AutoGTP's version when this is a change in the leelaz binary. That won't work correctly.

There's one outstanding large Elo gainer that I want to merge: #1259. After that I think we're good for a new Leela Zero release anyway, but it might be good to take a close look and there isn't anything missing in ELF support.

Mardak:
The AutoGTP version bump is to differentiate it from clients that only support network version 1. The server could respond with required_client_version (leelaz version), but the server doesn't provide an "alternative" game for the old client to play instead.

roy7:
@gcp The reason is so we can use the AutoGTP version at the time of /get-task to only send 50% of self play using the ELF network to clients that support it.

gcp:
But why check the AutoGTP version (which does not tell you anything about what the client can go!) instead of checking the engine version?

Mardak:
Because autogtp doesn't tell the server what's the client version. ;)

gcp:
Ah, I think I understand. The client isn't sending the version he has, he's only checking if what he receives ... right.

gcp:
I'll just bump the versions on release though. No need to jump the gun here while we're debugging support still.

jkiliani:
Just a hypothesis, but was it ever tested how the network inference error from OpenCL compared to that from CuDNN? For a highly optimised net, could even those errors still within the current OpenCL error tolerance cause a significant strength loss?

gcp:
>For a highly optimised net, could even those errors still within the current OpenCL error tolerance cause a significant strength loss?

We don't even see strength loss for the USE_HALF mode which requires way looser bounds, so no, I don't believe this can be the case. On top of that, FB said they used fp16 mode in their match, which makes what you say even less likely. I don't think rounding can explain this.

hydrogenpi:
Even if cannot get ELF native working by tomorrow, if the 1000 sgf games are going to be published then even comparing against that might yield some general idea what is wrong.

hydrogenpi:
Perhaps ELF would be much relatively stronger at 1600 po or more, but that would take you much more time to get to 1000 games or statistically results. Good to know score is still going up. Hope it can be optimized to get anywhere close to what the published 198:2 rate

yuandong-tian:
Note that one potential bias about parallel evaluation is that a stronger bot might resign faster in the bad situations, so the results will appear to be closer until all games are finished.

bood:
@yuandong-tian Maybe I am asking too much but is it possible for you to run a match between ELF and modified LZ with converted 20x224 to ensure the convert is correct? with same playouts of course e.g. 1600

bubblesld:
I must do something wrong. I updated three changed files. Then I have the following errors when making leelaz. Can someone help me?

======================================================
In file included from Network.cpp:47:0:
UCTNode.h:89:36: error: ‘scored_node’ is not a member of ‘Network’
                        std::vector<Network::scored_node>& nodelist,
                                    ^
UCTNode.h:89:36: error: ‘scored_node’ is not a member of ‘Network’
UCTNode.h:89:56: error: template argument 1 is invalid
                        std::vector<Network::scored_node>& nodelist,
                                                        ^
UCTNode.h:89:56: error: template argument 2 is invalid
Makefile:61: recipe for target 'Network.o' failed
make[1]: *** [Network.o] Error 1

ryouiki:
I'm testing 18e6 vs elf-v0 with -v 400. Elf lost 5 out of 58 games.
4 of 5 losing games were **ladder failure**.

qucheng:
Hi, our match are using 80k rollouts. Ladder failures are almost eliminated in this case. With small rollouts, you also need to reduce batchsize and mcts_batch_per_rollout to get the best performance.

bood:
> With small rollouts, you also need to reduce batchsize and mcts_batch_per_rollout to get the best performance.

It is kind of vague how should one reduce it knowing little about what they mean. Any recommendation config for rollout 1600, 10000 respectively?

diadorak:
According to a few low playout games played on my machine using the ELF network, it seems the ladder issue of Elf is much more serious than that of the current network of LZ.

jkiliani:
@yuandong-tian Sorry if this was asked before, but what was the average number of playouts used for training the ELF models? The ladder weakness at low playouts combined with the excellent strength at high playouts suggests that training games were played using considerably more tree search than Leela Zero uses in training, is that correct? If this was the case, then the answer to the performance discrepancy may simply be that ELF is optimised for much higher playout counts than Leela Zero, and that our MCTS parameters (PUCT, FPU reduction) may also not be optimal under these circumstances.

diadorak:
@qucheng @yuandong-tian Do you allow early passes? The win rate drops to 0% when white passes for the first move.

marcocalignano:
I could get the leelaz version while I do the tuning and then send it along with each get_work request. 
But if I do this it will be late at night (UTC), when I get home.

gcp:
>In this way we could also have a test phase in next

What would you want to test?

marcocalignano:
That the server receive the right version from the client and distribute the right amount of different games, that autogtp won't make something wrong with the received data. I'd like to test all this before we make a release.

gcp:
OK, will bump next. I have seen no objections to the mixed ELF games proposal, so I think we're good there.

herazul:
Will you queue the ELF_LZ net for test matchs ? Would be cool to have it on the graph and see %WR result with 400 games !

gcp:
I think I will do that after I make a new release. Which will take no more than a few days, probably.

herazul:
Cool, good to know !

marcocalignano:
@gcp I meant to bump leelaz version and I will do the changes to tell the server the leelaz version.

marcocalignano:
I think that now Who compiles next has to be very careful to which version autogtp uses (as extern binary).

gcp:
Yes, which I was warning against this in the posts that are default-hidden above. I thought you were asking the same thing.

I don't want to bump Leela Zero to 0.14 because there are more changes incoming.

marcocalignano:
@gcp So we can do the following: 
I can code the part where autogtp sends the leela version, you do all your changes and take your time, @roy7 prepares the code to read the leelaz version at the get_work request, but **DO NOT ACTIVATE IT** and, when you are ready to release, we can then bump leelaz version up, activate the server and test a couple of days still in `next` before the real release.
So if the server is not activate autogtp is still secure.

Mardak:
@marcocalignano if you're going to extract leelaz version and send it to the server, you can just append it with some separator, e.g., `…/get-task/15/0.14` (And there's no need to actually bump AUTOGTP_VERSION in this case for this particular change as implicitly, only the newer version knows to send leelaz version, but feel free to bump anyway.)

gcp:
FWIW once the test in #1326 finishes and it's not worse, I am ready to release. So if you want changes in AutoGTP, let me know and I can hold off?

Mardak:
Bumping both versions is probably good enough for most people especially the released binaries. Problems arise if someone compiles autogtp without updating leelaz AND the server doesn't require lz 0.14 (assuming we want a transition before requiring an update). Although the failure when someone doesn't compile the new leelaz: "Weights file is the wrong version." which would be a pretty clear indicator that something needs to be done.

marcocalignano:
So I am home now and I guess I do not have to implement anything since you trust the people. 
I guess the server can force the min leelaz version required to get a ELF network.

gcp:
Right.

The idea is to check for AutoGTP v16 and assume that there's no-one that was on next, compiles everything by themselves, and recompiled autogtp but forgot to update leelaz, and then can't figure out why there's a weights format error.

I think that's OK.

We indeed can't really schedule an ELF match unless we force the clients to update (which would be sad, since there's no important self-play changes).

herazul:
Well adding some ELF games in the mix is a kinda big self play change =)

gcp:
Yeah but those will just get sent to capable clients. For matches that code isn't there yet, and it's probably more complicated to hack in.

marcocalignano:
The only thing to note that if a client uses autogtp with the `-e` (erase) to delete the unused network. Every time it receive the ELF network it will delete the current best network and the next get work it might do the opposite (delete the elf network and download the current net) which gives a lot of unnecessary traffic. So I just suggest to everyone to use the option `-e` to turn it off untill we find a solution.

roy7:
For those interested, Aquabot ran LZ with ELF weights against my RoyalZero. I got crushed.

https://online-go.com/game/12738753

hydrogenpi:
@roy how is that possible? LZ with ELF lost to Aquabot or Aquabot was running LZ with ELF against your regular LZ? Big difference and not sure which one you mean.

hydrogenpi:
Okay nevermind I confirmed. He wants to use LZELF to rematch against your SLOW version while capping his at 4800 v

hydrogenpi:
Impresiveh https://online-go.com/game/12739700

PR #1326, opened 2018-05-04
Accelerate the first N moves.

gcp:
* Make fastmove calculation flexible with board size.

See issue #1259.

### DIFF
--- a/src/TimeControl.cpp
+++ b/src/TimeControl.cpp
@@ -33,10 +33,10 @@ TimeControl::TimeControl(int boardsize, int maintime, int byotime,
     : m_maintime(maintime),
       m_byotime(byotime),
       m_byostones(byostones),
-      m_byoperiods(byoperiods) {
+      m_byoperiods(byoperiods),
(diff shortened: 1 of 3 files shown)


gcp:
```
leelaz-next v leelaz-bood (35/1000 games)
board size: 19   komi: 7.5
leelaz-next     20 57.14%       8  44.44%     12 70.59%    198.70
leelaz-bood     15 42.86%       5  29.41%     10 55.56%    205.76
                                13 37.14%     22 62.86%
```

Still very few games, but looks OK so far. From my side this is the last change that I wanted for a 0.14 release.

gcp:
Quick 9x9 check also suggests I didn't completely break that. So, good to go.

TFiFiE:
Why not make it a gradual formula? The ratio of the dividing move number to the number of intersections looks pretty constant.

gcp:
Basically say "move fast until 1/6th of the board is filled"? Sure, sounds reasonable.

PR #1352, opened 2018-05-06
Tidy gathering of input vector

Hersmunch:
Network now gathers inputs directly into a vector rather than going via a bitset. The bitset conversion is now only done for the selected move when recording training data. 2nd attempt of #1130 (without other changes).
A couple of questions:
1. My 2nd commit in this avoids `get_scored_moves_internal()` having to be passed the `state` but it looks weird to me that input data has the symmetry applied outside of the function but the outputs are converted back within the function. So:
a. also pull applying symmetry to `result.policy[]` out into `get_scored_moves()`
b. put applying of symmetry to inputs back into `get_scored_moves()`
c. "who cares? you're worrying about nothing"
2. Should `fill_planes()` be a method of `TimeStep` rather than `Training`? I'm not that experienced with OO (I program in C mostly) but what I have done doesn't look too out of place with what is currently in Training.cpp

### DIFF
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -1044,9 +1044,9 @@ void Network::show_heatmap(const FastState* const state,
     }
 }
 
-void Network::get_input_moves(const GameState* const state, 
-                              std::vector<net_t>& input_data, 
-                              const int symmetry, 
+void Network::get_input_moves(const GameState* const state,
+                              std::vector<net_t>& input_data,
+                              const int symmetry,
                               const int color)
 {
     assert(symmetry >= 0 && symmetry <= 7);
@@ -1068,7 +1068,7 @@ void Network::get_input_moves(const GameState* const state,
 }
 
 void Network::gather_features_vector(const GameState* const state,
-                                     std::vector<net_t>& input_data, 
+                                     std::vector<net_t>& input_data,
                                      const int symmetry) {
     const auto to_move = state->get_to_move();
     const auto blacks_move = to_move == FastBoard::BLACK;
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -885,15 +885,12 @@ Network::Netresult Network::get_scored_moves(
         }
     }
 
-    NNPlanes planes;
-    gather_features(state, planes);
-
     if (ensemble == DIRECT) {
         assert(symmetry >= 0 && symmetry <= 7);
-        result = get_scored_moves_internal(planes, symmetry);
+        result = get_scored_moves_internal(state, symmetry);
     } else if (ensemble == AVERAGE) {
-        for (auto r = 0; r < 8; ++r) {
-            auto tmpresult = get_scored_moves_internal(planes, r);
+        for (auto sym = 0; sym < 8; ++sym) {
+            auto tmpresult = get_scored_moves_internal(state, sym);
             result.winrate += tmpresult.winrate / 8.0f;
             result.policy_pass += tmpresult.policy_pass / 8.0f;
 
@@ -905,7 +902,7 @@ Network::Netresult Network::get_scored_moves(
         assert(ensemble == RANDOM_SYMMETRY);
         assert(symmetry == -1);
         const auto rand_sym = Random::get_Rng().randfix<8>();
-        result = get_scored_moves_internal(planes, rand_sym);
+        result = get_scored_moves_internal(state, rand_sym);
     }
 
     // v2 format (ELF Open Go) returns black value, not stm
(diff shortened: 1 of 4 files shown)


src/Network.cpp
```diff
@@ -885,19 +885,18 @@ Network::Netresult Network::get_scored_moves(
...
-        result = get_scored_moves_internal(planes, symmetry);
     } else {
         assert(ensemble == RANDOM_SYMMETRY);
         assert(symmetry == -1);
-        const auto rand_sym = Random::get_Rng().randfix<8>();
-        result = get_scored_moves_internal(planes, rand_sym);
+        symmetry = Random::get_Rng().randfix<8>();
     }
 
+    std::vector<net_t> input_data;
```
gcp: Can use auto here, i.e. auto input_data = std::vector<net_t>{};

src/Training.cpp
```diff
@@ -139,11 +139,26 @@ void Training::clear_training() {
     Training::m_data.clear();
 }
 
+void Training::fill_planes(const GameState* const state,
+                           TimeStep::NNPlanes& planes) {
```
gcp: Let this return planes instead of passing it by ref.

src/Training.cpp
```diff
@@ -139,11 +139,26 @@ void Training::clear_training() {
...
+
+    std::vector<net_t>features_vector;
```
gcp: auto

src/Network.cpp
```diff
@@ -1044,50 +1034,43 @@ void Network::show_heatmap(const FastState* const state,
...
 
-    const auto moves = std::min<size_t>(state->get_movenum() + 1, INPUT_MOVES);
-    // Go back in time, fill history boards
-    for (auto h = size_t{0}; h < moves; h++) {
-        // collect white, black occupation planes
-        fill_input_plane_pair(state->get_past_board(h),
-                              planes[black_offset + h],
-                              planes[white_offset + h]);
-    }
+    input_data.insert(input_data.cend(), BOARD_SQUARES, net_t(blacks_move));
```
gcp: cend(input_data)

src/Network.h
```diff
@@ -56,7 +53,7 @@ class Network {
 
     static Netresult get_scored_moves(const GameState* const state,
                                       const Ensemble ensemble,
-                                      const int symmetry = -1,
+                                      int symmetry = -1,
                                       const bool skip_cache = false);
     // File format version
```
gcp: This comment is a leftover, might want to axe it.

src/Network.cpp
```diff
@@ -1044,50 +1034,43 @@ void Network::show_heatmap(const FastState* const state,
...
+        for (auto idx = 0; idx < BOARD_SQUARES; idx++) {
+            const auto sym_idx = symmetry_nn_idx_table[symmetry][idx];
+            const auto x = sym_idx % BOARD_SIZE;
+            const auto y = sym_idx / BOARD_SIZE;
+            const auto color_is_at_vtx = (color == board.get_square(x, y));
+            input_data.emplace_back(net_t(color_is_at_vtx));
         }
     }
+    for (auto c = moves; c < INPUT_MOVES; c++) {
+        input_data.insert(input_data.cend(), BOARD_SQUARES, net_t(false));
```
gcp: cend(input_data)

gcp:
As for 1, it seems more logical that whoever applies the symmetry and reverses it. 
For 2 I don't think I care.

Hersmunch:
I went with 1b and reverted to keep the amount of change smaller. Thanks for your help with the other bits.

src/Training.cpp
```diff
@@ -139,11 +139,25 @@ void Training::clear_training() {
     Training::m_data.clear();
 }
 
+TimeStep::NNPlanes Training::get_planes(const GameState* const state) {
+    auto input_data = std::vector<net_t>{};
+    Network::gather_features(state, input_data, 0);
```
gcp: I guess gather_features could also just return the result. (Don't have to do this in this patch, I just noticed)

src/Network.cpp
```diff
@@ -1054,50 +1043,48 @@ void Network::show_heatmap(const FastState* const state,
...
+        }
+    }
+    for (auto c = moves; c < INPUT_MOVES; c++) {
+        input_data.insert(cend(input_data), BOARD_SQUARES, net_t(false));
+        color_not_to_move_data.insert(cend(color_not_to_move_data),
+            BOARD_SQUARES, net_t(false));
     }
+
+    input_data.insert(cend(input_data),
+        std::make_move_iterator(cbegin(color_not_to_move_data)),
```
gcp: Given that the planes are single-dimensional floating point vectors, I don't see how you're actually going to have any possibility of moving here. You end up copying all the data from your temporary array back to the original one.

To avoid this, I think it's better to calculate the right offset for color_not_to_move, resize input_data suitably in the beginning, and use the offsets for directly storing instead of emplace_back. 

Your previous version avoided the copy so I think this now too clever.

src/Training.cpp
Hersmunch: I thought about that too but decided not to as this would involve data being copied from the stack on return. I'm happy to change it if you prefer it though.

src/Network.cpp
Hersmunch: Yeah, I'm still learning :) I'll put it back the way it used to be done. Thanks for the useful info again..

src/Training.cpp
gcp: Look up Return Value Optimization.

Issue #1355, opened 2018-05-06
FPU interaction with randomize


Ttl:
AGZ and Facebook's ELF Go both used FPU equivalent to 0.5 in LZ, but LZ uses evaluation of parent. This difference causes big changes with the randomize option.

In this game played with ELF network: http://zero.sjeng.org/view/8a7ffd2112234f5310e70f376179f23a7259aea3a945a38a9bc9ec085a6513b2?viewer=wgo, black was leading with winrate of 99.9% but played a pass at move 159 losing the game.

Evaluated moves with different FPUs for move 159:

<details>
 <summary>FPU=parent</summary>
<p>

  K4 ->      54 (V: 99.80%) (N:  1.57%) PV: K4 N11 N10 O11 Q13 P13 Q14 P14 F17 G18 F18 F19
 O12 ->      44 (V: 99.83%) (N:  1.24%) PV: O12 O11 N11 O13 P11 N13 P13 P14 Q12 O15
 N12 ->      35 (V: 99.85%) (N:  0.99%) PV: N12 N13 O13 M13 O12 O15 K4 H4 H3 G4 G3 G6 F17 G18 F18 F19
 T12 ->      33 (V: 99.75%) (N:  1.01%) PV: T12 N11 N10 O11 K4 H4 H3 G4 G3
 N13 ->      30 (V: 99.82%) (N:  0.86%) PV: N13 O15 K4 O5 H5 H6 G5 D12 F17 G18 F18 F19
 M12 ->      29 (V: 99.81%) (N:  0.83%) PV: M12 N13 M13 K14 O13 O12 N12 O15
 M13 ->      29 (V: 99.81%) (N:  0.84%) PV: M13 L13 L12 N13 M12 O13 K14 O15 K13
 K14 ->      29 (V: 99.80%) (N:  0.85%) PV: K14 K15 K4 N11 N10 O11 M11 M12 H5 L12
  K2 ->      28 (V: 99.80%) (N:  0.83%) PV: K2 N11 N10 O11 M11 M12 Q13 P13 Q14 P14 F17
 N10 ->      28 (V: 99.74%) (N:  0.86%) PV: N10 N12 K4 L12 M12 M11 N11 M13 O12 O13 O11 N13
 M11 ->      27 (V: 99.82%) (N:  0.79%) PV: M11 N12 K4 H4 H3 G4 G3 G6 F17 G18
[...]

PR #1376, opened 2018-05-08
Exclude low visit moves from random selection.

gcp:
Right now the randomized move selection will still pick moves that only
got a single visit. That means that no matter how badly they are
evaluated, we still have a fair chance of playing them. This means our
learning cycle can't avoid some really bad blunders.

There's some indication that using FPU=parent causes these blunders to
stay happening when ahead, as opposed to the AGZ method of using
FPU=0.5, which ends up filtering them. Because we really want to keep
FPU=parent, filter moves with only a single visit to get a similar
effect, i.e. to not pick moves with a really bad evalution.

This defaults the minimum visits to 1, and also adds a configurable
temperature to move selection, defaulted at t=1.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -56,6 +56,8 @@ int cfg_lagbuffer_cs;
 int cfg_resignpct;
 int cfg_noise;
 int cfg_random_cnt;
+int cfg_random_min_visits;
+float cfg_random_temp;
 std::uint64_t cfg_rng_seed;
 bool cfg_dumbpass;
 #ifdef USE_OPENCL
@@ -99,6 +101,8 @@ void GTP::setup_default_parameters() {
     cfg_resignpct = -1;
     cfg_noise = false;
     cfg_random_cnt = 0;
+    cfg_random_min_visits = 1;
+    cfg_random_temp = 1.0f;
     cfg_dumbpass = false;
     cfg_logfile_handle = nullptr;
     cfg_quiet = false;
--- a/src/GTP.h
+++ b/src/GTP.h
@@ -39,6 +39,8 @@ extern int cfg_lagbuffer_cs;
 extern int cfg_resignpct;
 extern int cfg_noise;
 extern int cfg_random_cnt;
+extern int cfg_random_min_visits;
+extern float cfg_random_temp;
 extern std::uint64_t cfg_rng_seed;
 extern bool cfg_dumbpass;
 #ifdef USE_OPENCL
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -52,8 +52,8 @@ static void license_blurb() {
 static void parse_commandline(int argc, char *argv[]) {
     namespace po = boost::program_options;
     // Declare the supported options.
-    po::options_description v_desc("Allowed options");
-    v_desc.add_options()
+    po::options_description gen_desc("Generic options");
+    gen_desc.add_options()
         ("help,h", "Show commandline options.")
         ("gtp,g", "Enable GTP mode.")
         ("threads,t", po::value<int>()->default_value(cfg_num_threads),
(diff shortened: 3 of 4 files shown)


jkiliani:
Looks very nice, but why leave randomcnt in its current form in the code? Randomtemp looks much better, do you want to preserve randomcnt for downward compatibility only?

gcp:
>Randomtemp looks much better, do you want to preserve randomcnt for downward compatibility only?

Yes, for one this avoids changing AutoGTP. If it turns out there are more bad interactions, this also gives us some possibility to lower temperature in the late game.

roy7:
Have you considered using FPU=min(parent, .50)? So we use .50 AGZ style when winning, but use .50 when losing (search unknown children more).

gcp:
>Have you considered using FPU=min(parent, .50)?

Would need testing but unless it turns out to be better, I'd never use FPU=50%.

Eddh:
For what it's worth I tested on 2 different nets in LeelaChess and it was good against the base dynamic eval (with bugged FPU reduction).
What drove me to test this was a maybe wrong hypothesis that Deepmind didn't lie about FPU, but just didn't say the whole truth. And in the paper they say that they initialize Q to 0. I thought, maybe it was indeed initialized to 0, but then changed afterward. And in that case it might make sense to drag down FPU if the parent is bad, while the behavior above 0.5 doesn't really seem to need a correction.

gcp:
OK, I'll run a test of that as well then.

jkiliani:
About FPU=min(parent, .50), would this effectively replace the current FPU reduction? Also, can this change be implemented without another puct tuning run?

roy7:
When we set FPU to parent we're actually just using the parent's raw nn_eval right? Not the result of search under that node at the time the parent was chosen as our prior move?

Eddh:
No FPU reduction could be used at the same time. You'd first apply FPU reduction then do min(fpu_reduced, .50). That's what I did in my tests in LeelaChess anyway, with the bugged virtual loss. So for a test in LeelaZero I would just do that.

jkiliani:
@roy7 We're using the parent's raw net eval at the moment, yes. But using the current eval (without VL of course) seemed to be stronger for Leela Chess, with an optimal FPU reduction that was considerably smaller than that which worked for static parent FPU.

gcp:
>About FPU=min(parent, .50), would this effectively replace the current FPU reduction? 

No.

>Also, can this change be implemented without another puct tuning run?

If it's at least as good, I'd think so? If it fixes #1355 without losing strength, it looks preferable to my pull here because it fixes the issue by bringing us closer to A(G)Z.

>should dynamic parent eval instead of the current static parent eval not also be on the table?

Probably. No-one has bothered to test this though.

jkiliani:
> If it's at least as good, I'd think so? If it fixes #1355 without losing strength, it looks preferable to my pull here because it fixes the issue by bringing us closer to A(G)Z. Not to mention it's a literal 1-liner!

It may fix #1355 but could easily have unforeseen consequences: In any somewhat advantageous situation, this will act as an effective FPU reduction on the root, bringing back the suppression of Dirichlet noise associated with that prior to https://github.com/gcp/leela-zero/commit/0c8cd6aa3bc88d119100192191a6b7e148cb5126. Being a one-liner just means it would be easy to do, not necessarily a good idea.

If we decide that with the fixed Alpha, reduced noise effectiveness is an acceptable price in some situations, it seems like a valid option.

Eddh:
Yep but it would explain how AGZ played moves that didn't make any sense when it was winning. Haylee has commented that LeelaZero didn't seem to play bad moves when ahead. While AGZ routinely lost dozens of point without caring. Which fit with what we could expect from a FPU at 0.5 or maxed at 0.5.

jkiliani:
OK, so why should we follow AGZ in this when our procedure actually seems to work better, and we just decided to deviate from AGZ regarding temperature in favour of AlphaZero instead? I looked at policy priors of ELF-net in #1355, and all of them, even those of good moves, were extremely low (maximum 1.77%). This indicates to me that the network was extrapolating to a situation it was very unfamiliar with. By using temperature, such situations should have much better priors with Leela Zero networks, so we would not really need the 0.5 cap on FPU especially with the expected side effects...

gcp:
> In any somewhat advantageous situation, this will act as an effective FPU reduction on the root, bringing back the suppression of Dirichlet noise associated with that prior to 0c8cd6a.

Not at all. The problem pointed out in that commit is that the parameters for the noise were tuned to the AGZ method, not a method with FPU reductions. So it would improve the situation mentioned there by making a closer match at the root - when ahead - , not worsen it.

>Yep but it would explain how AGZ played moves that didn't make any sense when it was winning.

This is not an argument in favor.

>when our procedure actually seems to work better

Because it should to lead to ELF nets, which we already have? So we're looking for something that makes t=1 work, without reverting FPU reductions or FPU=parent, if possible. (Note that FPU adjustment is mostly helpful when scores go down)

jkiliani:
If you want to give it a try and don't have concerns regarding noise effectiveness, why not I guess... hopefully it won't change playing style when ahead too much. You might still consider removing dumbpass for the ELF games, since there's really nothing to be gained by leaving it on. It's not like this net could ever learn to pass correctly...

gcp:
Thing is I'm not sure if FPU=0.5 is effective enough when you're not strongly ahead. i.e. in the starting position it'll still accumulate a visit on P1, L1, T1, etc.

jkiliani:
Why is that a problem, if we're excluding 1-visit moves as well?

gcp:
I was thinking of the FPU=min(...) thing as an alternative to this pull, not something additional to try.

jkiliani:
I think this pull is better than FPU capped at 0.5, since it only allows moves to be played that either have a policy score or, if found by Dirichlet noise, an eval score that's not really bad.

Adjusting FPU seems considerably more risky since it's a more invasive change into the the node selection algorithm. Doing it in addition sounds OK, if it doesn't regress...

gcp:
To make it clear, my intent is to address the fact that in some situations we definitely blunder more than AZ/AGZ would have, as shown in #1355.

#1355 shows that using FPU=0.5 is better in this regard, in case you are ahead, because it will end up filtering bad moves, just like FPU reduction would have. So FPU=min(...) keeps the advantage of FPU reductions when losing filters bad moves when winning, is simple, and close to AGZ.

But looking at what it does in various positions, it'll definitely blunder more in equal positions, and it does reduce noise effectiveness a bit (though AGZ would have had the same problem). I guess you could say that 0.5 being magic is indeed a bad problem with that technique compared to this pull.

jkiliani:
I think a major issue of any FPU different from parent eval at the root node is the policy feedback for low policy moves. If you reduce the FPU at root level, those low policy moves may not get visits at all without the help of noise, or too late to get a larger visit share than their prior even if their evaluation is good. That would prevent good moves found by noise from increasing their policy scores through a feedback cycle, which does not sound like a good idea even if it mainly affects situations where the side to move has an advantage (causing FPU=min(...) to trigger).

gcp:
FWIW, I did a test by playing a match with -n and -m 999 enabled between FPU=min() and v!=1, and it's a 200 Elo advantage for v!=1. So this increases the quality of selfplay greatly, while still keeping it quite diverse.

PR #1386, opened 2018-05-09
Add multi GPU training support

godmoves:
Add multi-GPU support for training. 
I test the code on a Ubuntu 17.10 with two 1080ti and a similar code is used to train @bjiyxo's 20 block net.
But you will need to use `net_to_model` to convert a net to a model that can be used by the code as the way of getting variables is different.

### DIFF
--- a/training/tf/tfprocess.py
+++ b/training/tf/tfprocess.py
@@ -17,12 +17,11 @@
 #    along with Leela Zero.  If not, see <http://www.gnu.org/licenses/>.
 
 import math
+import numpy as np
 import os
+import tensorflow as tf
 import time
 import unittest
-
-import numpy as np
-import tensorflow as tf
 
 
 def weight_variable(name, shape):
@@ -32,7 +31,6 @@ def weight_variable(name, shape):
     weights = tf.get_variable(name, initializer=initial)
     tf.add_to_collection(tf.GraphKeys.WEIGHTS, weights)
     return weights
-
 
 # Bias weights for layers not followed by BatchNorm
 # We do not regularlize biases, so they are not
@@ -46,7 +44,6 @@ def bias_variable(name, shape):
 def conv2d(x, W):
     return tf.nn.conv2d(x, W, data_format='NCHW',
                         strides=[1, 1, 1, 1], padding='SAME')
-
 
 # Restore session from checkpoint. It silently ignore mis-matches
 # between the checkpoint and the graph. Specifically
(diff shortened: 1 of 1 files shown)


gcp:
@ywrt is this something you could review?

training/tf/tfprocess.py
```diff
@@ -65,49 +71,63 @@ def optimistic_restore(session, save_file, graph=tf.get_default_graph()):
...
+            ["{}={:g}".format(k, np.mean(v or [0])) for k, v in self.s.items()])
+
     def clear(self):
         self.s = {}
+
     def summaries(self, tags):
         return [tf.Summary.Value(
-            tag=k, simple_value=self.mean(v)) for k,v in tags.items()]
+            tag=k, simple_value=self.mean(v)) for k, v in tags.items()]
+
```
ywrt: In general, it's better to move all the style changes to a separate patch.
(makes things much easier to review)

training/tf/tfprocess.py
```diff
@@ -65,49 +71,63 @@ def optimistic_restore(session, save_file, graph=tf.get_default_graph()):
...
 
+
 class TFProcess:
     def __init__(self):
         # Network structure
         self.RESIDUAL_FILTERS = 128
         self.RESIDUAL_BLOCKS = 6
 
+        # Set number of GPUs for training
+        self.gpus_num = 2
```
ywrt: The default should be 1, so as not to break existing scripts et al.

training/tf/tfprocess.py
```diff
@@ -149,77 +169,95 @@ def init(self, batch_size, macrobatch=1, logbase='leelalogs'):
...
+            learning_rate=0.05, momentum=0.9, use_nesterov=True)
+
+        # Construct net here.
+        tower_grads = []
+        tower_loss = []
+        tower_policy_loss = []
+        tower_mse_loss = []
+        tower_reg_term = []
+        tower_y_conv = []
+        counter = 0
```
ywrt: You don't need this variable. It's the same as 'i' in the range(self.gpus_num) loop.

training/tf/tfprocess.py
```diff
@@ -149,77 +169,95 @@ def init(self, batch_size, macrobatch=1, logbase='leelalogs'):
...
+        self.sy_ = tf.split(probs, self.gpus_num)
+        self.sz_ = tf.split(winner, self.gpus_num)
         self.batch_norm_count = 0
-        self.y_conv, self.z_conv = self.construct_net(self.x)
+        self.reuse_var = None
 
-        if self.swa_enabled == True:
+        # You need to change the learning rate here if you are training
+        # from a self-play training set, for example start with 0.005 instead.
+        opt_op = tf.train.MomentumOptimizer(
```
ywrt: nit: this isn't an operation, so it probably shouldn't be called 'opt_op'. leaving it as the original 'opt' name would be better.

training/tf/tfprocess.py
```diff
@@ -16,32 +16,38 @@
...
+import math
 import os
-import random
-import tensorflow as tf
 import time
 import unittest
 
-def weight_variable(shape):
+import numpy as np
+import tensorflow as tf
```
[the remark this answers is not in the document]
godmoves: PEP8 suggests that:
>Imports should be grouped in the following order:
1.standard library imports
2.related third party imports
3.local application/library specific imports
You should put a blank line between each group of imports.

I am not sure about it, maybe I should use the original order and move all the style changes to a separate patch : )

training/tf/tfprocess.py
godmoves: Yes, this is a change for my test, I will fix it.

godmoves:
@ywrt Thank you for your review. Your advice is really useful to me.
Some fixes are made to make sure only necessary changes are included in this pr and I will move all style fixes into another pr : )

gcp:
Thanks for the patch and the review.

PR #1388, opened 2018-05-09
Extend GTP to add support for displaying real time winrates and variations 

bittsitt:
This is an alternative implementation to #1242 and implements the solution discussed in that thread.

I have added two gtp commands:
```
lz-analyze <interval in centis>
lz-genmove-analyze <color> <interval in centis>
```

lz-analyze will return a multi-line response starting with =\n (or =id\n), before outputting one new info line every <interval in centis>. Each info line has the following format:
```
info <variation 1 info> | <variation 2 info> | .. | <variation n info>\n
```
The variation info has the following format
```
<Move> <playouts> <winrate> <list of moves in variation>
```
Example:
```
info D7 2234 99.94 D7 E6 C4 C3 C5 B3 E7 F7 F6 G6 F5 E4 F8 G7 G8 J7 H8 K6 J8 K8 | C5 260 99.88 C5 C4 D7 D5 E7 C7 C8 B7 B8 B5 R14 Q13 | N3 133 99.94 N3 O3 N2 N1 D7 E7 D5 E6 C4 | S3 120 99.91 S3 C5 R14 R17 Q17 Q18 R16 | C4 105 97.04 C4 C5 B5 D5 B3 C3 B4 C7 D3 E3 C2 R12 R14 P12 | N2 92 99.96 N2 O1 D7 E7 D8 C5 E6 D5 F7 | P5 90 99.94 P5 P4 D7 E7 D8 C5 E6 D5 F7 R14 R15 | R14 44 99.91 R14 C5 G3 G2 H3 H2 F4 | S2 21 99.92 S2 C5 R14 R17 R16 Q17 P17 P18 | B5 20 99.91 B5 C5 D7 B6 C7 B4\n
```
Once a (any) command is received from the gui/controller, the response is terminated with a \n on a separate line (as defined in multi-line responses in GTP protocol), and no more info lines are output.

lz-genmove-analyze works the same way, except when it finds a move to play (using the same criteria as a regular genmove), it outputs the following line:
```
play <move>
```
So the output will typically look like this (info lines truncated for simplicity):
```
info D7 2234 .....
info D7 3344 .....
play D7
info D6 1200 ...
info D6 2200 ...
info D6 3400 ...
etc
```

Response is terminated when LZ receives a (any) command the same way as the other command.

This code has been tested against (a modified version of) Lizzie.

Please let me know if you would like any changes to this implementation.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -373,6 +373,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
                 return 1;
             }
             if (analysis_output) {
+                // Start of multi-line response
                 cfg_analyze_interval_centis = interval;
                 if (id != -1) gtp_printf_raw("=%d\n", id);
                 else gtp_printf_raw("=\n");
@@ -380,6 +381,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
             // start thinking
             {
                 game.set_to_move(who);
+                // Outputs winrate and pvs for lz-genmove_analyze
                 int move = search->think(who);
                 game.play_move(move);
 
@@ -393,10 +395,12 @@ bool GTP::execute(GameState & game, std::string xinput) {
             if (cfg_allow_pondering) {
                 // now start pondering
                 if (!game.has_resigned()) {
+                    // Outputs winrate and pvs through gtp for lz-genmove_analyze
                     search->ponder();
                 }
             }
             if (analysis_output) {
+                // Terminate multi-line response
                 gtp_printf_raw("\n");
             }
         } else {
@@ -417,13 +421,16 @@ bool GTP::execute(GameState & game, std::string xinput) {
             gtp_fail_printf(id, "syntax not understood");
             return true;
         }
+        // Start multi-line response
         if (id != -1) gtp_printf_raw("=%d\n", id);
         else gtp_printf_raw("=\n");
         // now start pondering
         if (!game.has_resigned()) {
+            // Outputs winrate and pvs through gtp
             search->ponder();
         }
         cfg_analyze_interval_centis = 0;
+        // Terminate multi-line response
         gtp_printf_raw("\n");
         return true;
     } else if (command.find("kgs-genmove_cleanup") == 0) {
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -74,6 +74,7 @@ FILE* cfg_logfile_handle;
 bool cfg_quiet;
 std::string cfg_options_str;
 bool cfg_benchmark;
+int cfg_analyze_interval_centis;
 
 void GTP::setup_default_parameters() {
     cfg_gtp_mode = false;
@@ -107,6 +108,7 @@ void GTP::setup_default_parameters() {
     cfg_logfile_handle = nullptr;
     cfg_quiet = false;
     cfg_benchmark = false;
+    cfg_analyze_interval_centis = 0;
 
     // C++11 doesn't guarantee *anything* about how random this is,
     // and in MinGW it isn't random at all. But we can mix it in, which
@@ -147,6 +149,8 @@ const std::string GTP::s_commands[] = {
     "kgs-time_settings",
     "kgs-game_over",
     "heatmap",
+    "lz-analyze",
+    "lz-genmove_analyze",
     ""
 };
 
@@ -345,12 +349,18 @@ bool GTP::execute(GameState & game, std::string xinput) {
             }
         }
         return true;
-    } else if (command.find("genmove") == 0) {
+    } else if (command.find("genmove") == 0 || command.find("lz-genmove_analyze") == 0 ) {
+        bool analysis_output = command.find("lz-genmove_analyze") == 0;
+        int interval;
+
         std::istringstream cmdstream(command);
         std::string tmp;
 
         cmdstream >> tmp;  // eat genmove
         cmdstream >> tmp;
+        if (analysis_output) {
+            cmdstream >> interval;
+        }
 
         if (!cmdstream.fail()) {
             int who;
@@ -362,6 +372,11 @@ bool GTP::execute(GameState & game, std::string xinput) {
                 gtp_fail_printf(id, "syntax error");
                 return 1;
             }
+            if (analysis_output) {
+                cfg_analyze_interval_centis = interval;
+                if (id != -1) gtp_printf_raw("=%d\n", id);
+                else gtp_printf_raw("=\n");
+            }
             // start thinking
             {
                 game.set_to_move(who);
@@ -369,7 +384,11 @@ bool GTP::execute(GameState & game, std::string xinput) {
                 game.play_move(move);
 
                 std::string vertex = game.move_to_text(move);
-                gtp_printf(id, "%s", vertex.c_str());
+                if (!analysis_output) {
+                    gtp_printf(id, "%s", vertex.c_str());
+                } else {
+                    gtp_printf_raw("play %s\n", vertex.c_str());
+                }
             }
             if (cfg_allow_pondering) {
                 // now start pondering
@@ -377,9 +396,35 @@ bool GTP::execute(GameState & game, std::string xinput) {
                     search->ponder();
                 }
             }
+            if (analysis_output) {
+                gtp_printf_raw("\n");
+            }
         } else {
             gtp_fail_printf(id, "syntax not understood");
         }
+        analysis_output = false;
+        return true;
+    } else if (command.find("lz-analyze") == 0) {
+        std::istringstream cmdstream(command);
+        std::string tmp;
+        int interval;
+        
+        cmdstream >> tmp; // eat lz-analyze
+        cmdstream >> interval;
+        if (!cmdstream.fail()) {
+            cfg_analyze_interval_centis = interval;
+        } else {
+            gtp_fail_printf(id, "syntax not understood");
+            return true;
+        }
+        if (id != -1) gtp_printf_raw("=%d\n", id);
+        else gtp_printf_raw("=\n");
+        // now start pondering
+        if (!game.has_resigned()) {
+            search->ponder();
+        }
+        cfg_analyze_interval_centis = 0;
+        gtp_printf_raw("\n");
         return true;
     } else if (command.find("kgs-genmove_cleanup") == 0) {
         std::istringstream cmdstream(command);
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -350,8 +350,8 @@ bool GTP::execute(GameState & game, std::string xinput) {
         }
         return true;
     } else if (command.find("genmove") == 0 || command.find("lz-genmove_analyze") == 0 ) {
-        bool analysis_output = command.find("lz-genmove_analyze") == 0;
-        int interval;
+        auto analysis_output = command.find("lz-genmove_analyze") == 0;
+        auto interval = 0;
 
         std::istringstream cmdstream(command);
         std::string tmp;
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -349,7 +349,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
             }
         }
         return true;
-    } else if (command.find("genmove") == 0 || command.find("lz-genmove_analyze") == 0 ) {
+    } else if (command.find("genmove") == 0 || command.find("lz-genmove_analyze") == 0) {
         auto analysis_output = command.find("lz-genmove_analyze") == 0;
         auto interval = 0;
 
@@ -408,7 +408,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
         std::istringstream cmdstream(command);
         std::string tmp;
         int interval;
-        
+
         cmdstream >> tmp; // eat lz-analyze
         cmdstream >> interval;
         if (!cmdstream.fail()) {
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -241,21 +241,21 @@ void UCTSearch::output_analysis(FastState & state, UCTNode & parent) {
     }
 
     const int color = state.get_to_move();
-    
+
     // sort children, put best move on top
     parent.sort_children(color);
-    
+
     if (parent.get_first_child()->first_visit()) {
         return;
     }
-    
+
     int movecount = 0;
     std::string separator = "info";
     for (const auto& node : parent.get_children()) {
         // Always display at least two moves. In the case there is
         // only one move searched the user could get an idea why.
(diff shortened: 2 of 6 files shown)


src/Utils.cpp
```diff
@@ -122,6 +122,17 @@ void Utils::gtp_printf(int id, const char *fmt, ...) {
     va_end(ap);
 }
 
+void Utils::gtp_printf_raw(const char *fmt, ...) {
```
gcp: Probably many of the functions in this file should be rewritten to call this instead of duplicating it.

src/GTP.cpp
```diff
@@ -345,12 +349,18 @@ bool GTP::execute(GameState & game, std::string xinput) {
             }
         }
         return true;
-    } else if (command.find("genmove") == 0) {
+    } else if (command.find("genmove") == 0 || command.find("lz-genmove_analyze") == 0 ) {
+        bool analysis_output = command.find("lz-genmove_analyze") == 0;
```
gcp: auto 

(I know some of the existing code doesn't use it, but avoid adding new)

src/GTP.cpp
```diff
@@ -362,24 +372,59 @@ bool GTP::execute(GameState & game, std::string xinput) {
                 gtp_fail_printf(id, "syntax error");
                 return 1;
             }
+            if (analysis_output) {
+                cfg_analyze_interval_centis = interval;
+                if (id != -1) gtp_printf_raw("=%d\n", id);
```
gcp: Isn't this the same as the regular gtp_printf calls? Except you're dealing with the id manually now, which is what those helper functions were for?

src/GTP.cpp
```diff
@@ -362,24 +372,59 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
+                }
             }
             if (cfg_allow_pondering) {
                 // now start pondering
                 if (!game.has_resigned()) {
                     search->ponder();
                 }
             }
+            if (analysis_output) {
+                gtp_printf_raw("\n");
```
gcp: Can't this be merged with the previous call?

Also, it looks suspicious that your output depends on when the ponder search finishes?

Or is the idea that if the GUI issues lz-genmove_analyze, it should expect potential ponder output after the play?

src/UCTSearch.cpp
```diff
@@ -670,13 +710,23 @@ void UCTSearch::ponder() {
     for (int i = 1; i < cfg_num_threads; i++) {
         tg.add_task(UCTWorker(m_rootstate, this, m_root.get()));
     }
+    Time start;
     auto keeprunning = true;
+    int last_output = 0;
```
gcp: auto

src/UCTSearch.cpp
```diff
@@ -602,6 +635,7 @@ int UCTSearch::think(int color, passflag_t passflag) {
 
     bool keeprunning = true;
     int last_update = 0;
+    int last_output = 0;
```
gcp: make all of these auto

src/UCTSearch.cpp
```diff
@@ -235,6 +235,39 @@ void UCTSearch::dump_stats(FastState & state, UCTNode & parent) {
...
 
+void UCTSearch::output_analysis(FastState & state, UCTNode & parent) {
+    if (!parent.has_children()) {
+        return;
+    }
+
+    const int color = state.get_to_move();
+    
+    // sort children, put best move on top
+    parent.sort_children(color);
```
gcp: It don't think it is safe to call this during the search. You need to loop over the children, collect the statistics, and put them in the right order in your local copy.

Have you thought about whether the order the program gives the moves in should be meaningful? I think in the current proposal it cannot, so you can skip the sorting, and avoid most of this complication.

src/UCTSearch.cpp
```diff
@@ -235,6 +235,39 @@ void UCTSearch::dump_stats(FastState & state, UCTNode & parent) {
...
+    for (const auto& node : parent.get_children()) {
+        // Always display at least two moves. In the case there is
+        // only one move searched the user could get an idea why.
+        if (++movecount > 2 && !node->get_visits()) break;
+        
+        std::string move = state.move_to_text(node->get_move());
+        FastState tmpstate = state;
+        tmpstate.play_move(node->get_move());
+        std::string pv = move + " " + get_pv(tmpstate, *node);
+        gtp_printf_raw("%s%s %d %.2f %s", separator.c_str(), move.c_str(),node->get_visits(),
```
gcp: I would really put tags in here like UCI does, i.e. "info move d4 visits 100 winrate 5000" instead of relying on the position or order. Else it gets really brittle if there's ever a need to display more info about a move, or if we don't know some of the info. With tags, the GUI can just parse and display those it understands.

(percentage in permilles)

src/UCTSearch.cpp
```diff
@@ -235,6 +235,39 @@ void UCTSearch::dump_stats(FastState & state, UCTNode & parent) {
...
+                 node->get_visits() ? node->get_eval(color)*100.0f : 0.0f, pv.c_str());
+        separator = " | ";
```
gcp: With tags this could go then.

src/GTP.cpp
bittsitt: search->ponder() will print to the gtp stream as well (that is where all the info lines come from), adding to the multi-line response. That's also why you can't use the regular gtp_ calls, since they add \n\n which means end of response. But we don't want end of response until search->ponder returns. That is why there is a lone \n there.

src/Utils.cpp
bittsitt: Agreed

src/UCTSearch.cpp
bittsitt: You could leave it to the controller/gui to find the best move (or rather the one with most playouts), but it is of course easier if the best move is the first (and what lizzie is getting right now). 

They also get sorted in dump_stats, is that different in some way that makes it safe there?

src/UCTSearch.cpp
bittsitt: I like that. info move d4 visits 100 winrate 5000 variation d4 d6 d2 c2
Now I've grouped all the alternatives on one line using the | as separator. Should we just use start again with the info tag, or just the move tag, or a different separator all together?

info move d4 visits 100 winrate 5000 pv d4 d6 d2 c2 info move d2 visits 60 winrate 4900 pv d3 d7 e3 e4
or just 
info move d4 visits 100 winrate 5000 pv d4 d6 d2 c2 move d2 visits 60 winrate 4900 pv d3 d7 e3 e4
(without the extra info)?
(What is pv short for btw? Play variation?)

dfannius: PV = Principal Variation. (This is the standard term and abbreviation in computer chess as well.)

src/GTP.cpp
bittsitt: The difference is that this is a multi-line response. The helper functions add \n\n at the end, and that is no good.

src/UCTSearch.cpp
bittsitt: Nevermind, dump_stats is not called during search.

I was about to just delete it, but then I noticed that the loop later in that function needs a sorted list and handling the case where you always output at least two variations (even if the second one has 0 visits) gets a wee bit complicated if it is not sorted (basically you would have to sort manually the same way sort_children does, first on score, then on color AFAICT).

I did also notice however that sort_children gets the mutex before sorting. Isn't that enough to make it safe even while searching?

I've fixed the other things you pointed out, only this remains.

src/GTP.cpp
gcp: Might be good to add some minimal documentation about this in the code. Just pointing out that this will terminate a multi-line response by adding the 2nd newline to the newline in the last analysis output.

src/UCTSearch.cpp
```diff
@@ -235,6 +235,39 @@ void UCTSearch::dump_stats(FastState & state, UCTNode & parent) {
...
 
+void UCTSearch::output_analysis(FastState & state, UCTNode & parent) {
+    if (!parent.has_children()) {
+        return;
+    }
+
+    const int color = state.get_to_move();
+
+    // sort children, put best move on top
+    parent.sort_children(color);
```
gcp: Not threadsafe to call during the search, see previous review. I think you can get rid of this, but you need to modify the loop below to not break early if there are no visits, just skip the current move (because moves will no longer be in visits or prior order).

src/UCTSearch.cpp
```diff
@@ -235,6 +235,39 @@ void UCTSearch::dump_stats(FastState & state, UCTNode & parent) {
...
+    if (parent.get_first_child()->first_visit()) {
+        return;
+    }
+
+    int movecount = 0;
+    std::string separator = "info";
+    for (const auto& node : parent.get_children()) {
+        // Always display at least two moves. In the case there is
+        // only one move searched the user could get an idea why.
+        if (++movecount > 2 && !node->get_visits()) break;
```
gcp: If there is a visual analysis GUI I guess this is less important, so the "at least 2" part can go.

src/UCTSearch.cpp
bittsitt: Ok, so the mutex lock in sort_children doesn't make it completely thread safe?

src/UCTSearch.cpp
bittsitt: Ok, agree.

src/UCTSearch.cpp
gcp: The problem is that there is code that assumes it can iterate over m_children and probe or update any information that is in the atomic of the node structure (without taking locks), because the m_children array is not supposed to change once the node is expanded and/or the search is multi-threaded.

I just reviewed the code, and it looks like those users may be down to prune_noncontenders, actually. But it's good to keep the above invariant if we ever run into locking bottlenecking us. It's pretty annoying if you have to lock a node just to be able to gather statistics on the children.

gcp: Uh, actually this function would itself be a user of that invariant too now. As long as the above invariant holds, it doesn't need to keep the root node locked while it's printing statistics. Which is good.

bittsitt: Ok, so conclusion is don't sort.

bittsitt:
I think I have responded to all the change requests now (thanks for the feedback by the way), let me know if there was something I missed or if there are any more changes you would like. 

Please also let me know if you want me to squash some of the commits (especially sorry for the whitespace commit).

jkiliani:
Out of curiosity, are the real time winrates cleared of virtual losses? I noticed with my (slow) computer that Lizzie winrate outputs when playing a move that was not considered usually start low, due to virtual loss, and recover when sufficient root visits accumulate to compensate this...

jkiliani:
Yes, that looks exactly what virtual loss will do to the winrates.

bittsitt:
Sounds like a separate issue/PR?

jkiliani:
I pushed a PR for the VL removal with #1412 now. Let's see what @gcp's preference is regarding boolean option or function clone...

gcp:
This dumps all moves on a single line. I prefer it if every move gets its own line (UCI style). Also makes parsing easier. So I'll change it that way.

oni-link:
Moves in selfplay games are not displayed anymore after updating to 'next':
```
Got new job: selfplay
net: 90560d926216c9842add40913a023fd59fde761a64014c48d8246648ddd272a3.
Engine has started.
Infinite thinking time set.
1 (B ) 2 (W ) 3 (B ) 4 (W ) 5 (B ) 6 (W ) 7 (B ) 8 (W ) 9 (B ) 10 (W ) 11 (B ) 12 (W ) 13 (B ) 14 (W ) 15 (B ) 16 (W ) 17 (B ) 18 (W ) 19 (B ) 20 (W ) 21 (B ) 22 (W ) 23 (B ) 24 (W ) 25 (B ) 26 (W ) 27 (B ) 28 (W ) 29 (B ) 30 (W ) 31 (B ) 32 (W ) 33 (B ) 34 (W ) 35 (B ) 36 (W ) 37 (B ) 38 (W ) 39 (B ) 40 (W ) 41 (B ) 42 (W ) 43 (B ) 44 (W ) 45 (B ) 46 (W ) 47 (B ) 48 (W ) 49 (B ) 50 (W ) 51 (B ) 52 (W ) 53 (B ) 54 (W ) 55 (B ) 56 (W ) 57 (B ) 58 (W ) 59 (B ) 60 (W ) 61 (B ) 62 (W ) 63 (B ) 64 (W ) 65 (B ) 66 (W ) 67 (B ) 68 (W ) 69 (B ) 70 (W ) 71 (B ) 72 (W ) 73 (B ) 74 (W ) 75 (B ) 76 (W ) 77 (B ) 78 (W ) 79 (B ) 80 (W ) 81 (B ) 82 (W ) 83 (B ) 84 (W ) 85 (B ) 86 (W ) 87 (B ) 88 (W ) 89 (B ) 90 (W ) 91 (B ) 92 (W ) 93 (B ) 94 (W ) 95 (B ) 96 (W ) 97 (B ) 98 (W ) 99 (B ) 100 (W ) 101 (B ) 102 (W ) 103 (B ) 104 (W ) 105 (B ) 106 (W ) 107 (B ) 108 (W ) 109 (B ) 110 (W ) 111 (B ) 112 (W ) 113 (B ) 114 (W ) 115 (B ) 116 (W ) 117 (B ) 118 (W ) 119 (B ) 120 (W ) 121 (B ) 122 (W ) 123 (B ) 124 (W ) 125 (B ) 126 (W ) 127 (B ) 128 (W ) 129 (B ) 130 (W ) 131 (B ) 132 (W ) 133 (B ) 134 (W ) 135 (B ) 136 (W ) 137 (B ) 138 (W ) 139 (B ) 140 (W ) 141 (B ) 142 (W ) 143 (B ) 144 (W ) 145 (B ) 146 (W ) 147 (B ) 148 (W ) 149 (B ) 150 (W ) 151 (B ) 152 (W ) 153 (B ) 154 (W ) 155 (B ) 156 (W ) 157 (B ) 158 (W ) 159 (B ) 160 (W ) 161 (B ) 162 (W ) 163 (B ) 164 (W ) 165 (B ) 166 (W ) 167 (B ) 168 (W ) 169 (B ) 170 (W ) 171 (B ) 172 (W ) 173 (B ) 174 (W ) 175 (B ) 176 (W ) 177 (B ) 178 (W ) 179 (B ) 180 (W ) 181 (B ) 182 (W ) 183 (B ) 184 (W ) 

```

alreadydone:
@marcocalignano

gcp:
Probably not @marcocalignano's fault, leela should have stayed GTP compatible.

gcp:
She no longer outputs the move after genmove. Looking why...

gcp:
Oh right, you can't forward ```...`` varargs, you need something that takes a va_list.

PR #1397, opened 2018-05-10
Tree reduction

ThorAvaTahr:
This pull requests proposes an optimization to reduce the size of the tree, reducing the number of reevaluations of previously visited nodes. I argue that gain in performance outweighs the negligible chance we miss the optimal move. This was confirmed in an interrupted experiment at 9 wins vs two losses on our best 5x64 network.

After reviewing the search algorithm I wondered about the cost/benefit of keeping the entire search result in the tree (especially the width of the tree: at each node the entire set of valid moves is expanded).

Off course we start pruning the tree once we approach MAX_TREE_SIZE, skipping children with low probability. However this will have the effect that we re-evaluate previously visited nodes for which we know we have pruned some children. It looked somewhat wasteful to me, so I was thinking of some alternatives. 

Thinking of that in fact made me wonder how valuable it was to keep all 300+ moves in every step of the tree. I would say if we just keep the top 90 (max) moves, we would be not likely miss the best move. (And even if the best move would be below the top 90 moves for our policy, that would probably still mean we would miss it).

So I have run an experiment on our best 5x64 network. Unfortunately the experiment was interrupted due to some company policy forcing a restart for updates. The last observed score was 9 wins vs 2 losses on Tree_reduction vs master branch (options: -g -v 3201 --noponder -t 1 -q -d -r 0 -w).

I reflect that the current pruning mechanism results in loss of performance near the end of the game, where we have to revisit nodes to re-add pruned children. This is enhanced by the fact that it seems to me we keep entire tree throughout the game. Is that true? It is nice for uses, such as lizzie, to move back and not have to rebuild the tree again, however, for important matches we may need an option to delete the old part of the tree (say -5 moves back or so, allowing at least a couple of takebacks).

Anyway in conclusion this simple change appears to benefit the performance.

### DIFF
--- a/msvc/VS2015/leela-zero.vcxproj
+++ b/msvc/VS2015/leela-zero.vcxproj
@@ -93,6 +93,7 @@
     <ClCompile Include="..\..\src\Training.cpp" />
     <ClCompile Include="..\..\src\Tuner.cpp" />
     <ClCompile Include="..\..\src\UCTNode.cpp" />
+    <ClCompile Include="..\..\src\UCTNodePointer.cpp" />
     <ClCompile Include="..\..\src\UCTNodeRoot.cpp" />
     <ClCompile Include="..\..\src\UCTSearch.cpp" />
     <ClCompile Include="..\..\src\Utils.cpp" />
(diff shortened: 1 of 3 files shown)


gcp:
There are many variations you can do on this idea, see for example the source in the very first commit:
https://github.com/gcp/leela-zero/blob/2f7463d2cfba1b4617b3bd73bbdf3e1f52382429/UCTNode.cpp#L246

https://github.com/gcp/leela-zero/blob/2f7463d2cfba1b4617b3bd73bbdf3e1f52382429/UCTNode.cpp#L279

But you need much more testing with this, at least until you get a statistically significant result. And you need to be very careful you do not regress performance at very high visit counts where low prior moves can be found, as suddenly the program turns out to not be able to find the best move any more regardless of search time.

This kind of change would only be done as an option, and even so it would need way more data than you have here.

>This is enhanced by the fact that it seems to me we keep entire tree throughout the game. Is that true?

No, not at all.

PR #1415, opened 2018-05-13
Update README.md

zinwalin:
add instructions for compiling autoatp on macOS.

### DIFF
--- a/autogtp/README.md
+++ b/autogtp/README.md
@@ -16,6 +16,15 @@ the SGF and training data at the end of the game.
     sudo apt install qt5-default qt5-qmake curl
     qmake -qt5
     make
+## Example of compliling - macOS
+
+	brew install qt5
+	brew link qt5 --force
+
+then go to autoatp folder, and execute:
+
+	qmake
+	make
 
 ## Compiling under Visual Studio - Windows
 
--- a/autogtp/README.md
+++ b/autogtp/README.md
@@ -16,7 +16,7 @@ the SGF and training data at the end of the game.
     sudo apt install qt5-default qt5-qmake curl
     qmake -qt5
     make
-## Example of compliling - macOS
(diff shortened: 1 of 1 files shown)


autogtp/README.md
```diff
@@ -16,6 +16,15 @@ the SGF and training data at the end of the game.
     sudo apt install qt5-default qt5-qmake curl
     qmake -qt5
     make
+## Example of compliling - macOS
+
+	brew install qt5
+	brew link qt5 --force
+
+then go to autoatp folder, and execute:
```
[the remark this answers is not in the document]
afalturki: I meant this line:

then go to autoatp folder, and execute:

It's inferred that the person needs to execute the command inside the autogtp folder.

autogtp/README.md
```diff
@@ -16,6 +16,13 @@ the SGF and training data at the end of the game.
     sudo apt install qt5-default qt5-qmake curl
     qmake -qt5
     make
+    
+## Example of compiling - macOS
+
+	brew install qt5
+	brew link qt5 --force
+	qmake
```
Mardak: The brew output shows
```
==> Caveats
This formula is keg-only, which means it was not symlinked into /usr/local,
because Qt 5 has CMake issues when linked.
```
Not clear if we should be having people force link it. Maybe just `/usr/local/opt/qt/bin//qmake` instead?

gcp: I believe the better fix is:

```
export PATH=/usr/local/opt/qt5/bin:$PATH
```

Before doing the cmake .. to generate the build files. (If you already did, you have to remove everything or the above won't work).

gcp: Oh, this is for compiling autogtp directly. Yeah, then what @Mardak says is preferable.

PR #1429, opened 2018-05-15
autogtp: use compressed weights networks

marcocalignano:


marcocalignano:
I am not sure that we need this PR but you can tell me!
The problem is that I have to decompress the network anyway the first time I download it to check the SHA signature and then I delete the decompress file. 
I guess that using .gz file take longer for leelaz to start while gunzip them would waste time only the first time.
Anyway before goes in next I would like that someone compiles and tests this in windows.

autogtp/Management.cpp
```diff
@@ -472,17 +472,26 @@ Order Management::getWork(bool tuning) {
 bool Management::networkExists(const QString &name) {
     QString realHash = name;
     realHash.remove(0,9);
+    realHash.chop(3);
     if (QFileInfo::exists(name)) {
-        QFile f(name);
+#ifdef WIN32
+        QProcess::execute("gzip.exe -k -d -q " + name);
```
gcp: Dawww :(

marcocalignano:
@gcp I have the next PR ready do you want I put this change in a different branch (since is not that urgent) and I PR the next?

autogtp/Management.cpp
Mardak: Is it necessary to check that the uncompressed network matches? It isn't really for security as the client trusts the server to provide the right file. A server calculated gziphash is still trusting the server. And even before, the server could have just told the client to download some other "network" that had the appropriate hash in the filename.

roy7:
@mardak I think the concern is if the download doesn't complete or is corrupted. It should be easy enough for us to calculate those on the server side and provide the hash of the gzip file to the client.

autogtp/Management.cpp
Mardak: But even with those types of network / file failures, it would likely result in leelaz complaining "Weights file is the wrong version."

gcp: File size check may be risky as all files are approximately the same size.

>it would likely result in leelaz complaining "Weights file is the wrong version." or "Failed to parse weight file. Error on line 2."

Yes, but AutoGTP won't know the underlying reason then.

gcp:
>@gcp I have the next PR ready do you want I put this change in a different branch (since is not that urgent) and I PR the next?

If it's unrelated to this, make a separate PR.

marcocalignano:
I think you should wait for the discussion on this https://github.com/gcp/leela-zero-server/issues/129.
If they do the change we can save the unzip part for the verification and I can make another commit to update this one.

Mardak:
> Also the hash check can perhaps be changed to use a streaming interface and thus avoid the ungzip.

If the streaming approach is not too complex, I would think that is preferable to adding more state on the server and to transfer with 3 additional json properties: gziphash, blackgziphash, whitegziphash (where both the server and autogtp need to handle the extra data and if the server ever needs to communicate more network hashes to autogtp). The additional gzip hash data is also just derived data, which I believe is generally undesired as it introduces ways for data to become inconsistent unless the benefits outweigh the complexity.

l1t1:
instead of use compressed plain text, i prefer use binary format, but it need another parser

gcp:
@marcocalignano @Mardak What should we do with this?

marcocalignano:
This PR works as it is but like you said would be better to get the hash of the zipped file not to unzip it just to check it. I would not merge it till the server can deliver the hash of the zipped file.

Mardak:
If we aren't merging until the server provides the hash, sounds like this PR should be closed for now as it would need to be updated anyway?

What's the proposed API to provide the gzip hash for match tasks? Or alternative to having a separate field is to include the checksum as part of every "network" field not too different from Link Fingerprints 😉  https://tools.ietf.org/html/draft-lee-uri-linkfingerprints-00
e.g., `{ white_hash: "abcdef#sha256:123456", black_hash: "fedcba#sha256:098765" }`

I suppose the client/server details could be hashed out in https://github.com/gcp/leela-zero-server/issues/129

marcocalignano:
I will make another PR soon as soon as I test the match task!

PR #1430, opened 2018-05-15
autogtp: send leelaz version embedded in the URL used to ask for new job

marcocalignano:


marcocalignano:
Working on this PR I have notice that when autogtp runs leelaz for the tuning process leelaz runs with 2 threads. Is it ok that the tuning is done with two threads while the normal games are done with 1 thread?

Issue #1431, opened 2018-05-16
Mutex stuff


OmnipotentEntity:
I was poking about in the code and I noticed a few things that were a bit weird.

In UTCNode.h `m_nodemutex` is not marked as `mutable`.  Typically, (at least in past projects I've worked on) the mutex is marked as `mutable` so that it can be used without a `const_cast` in a `const` function.  The reasoning is because locking and unlocking a mutex should always be a logically `const` operation, because it does not affect the external behavior of the class.

The mutex is not standard.  I was curious as to the reasoning behind this.  (As `std::mutex` is a thing now.)

Finally, I looked at the implementation of the mutex, and the implementation is TS rather than TTS, is this an intentional design decision because you expect this mutex to experience very short locks with low levels of contention?  Because if this isn't the case, then TTS lowers memory contention.

gcp:
The mutex isn't standard because in many C++ libraries a mutex is a very heavyweight object and we have one per node. At some point hashed locking was proposed as a solution for this, but IMHO that's just worse in terms of complexity than per node spinlocks. I think I didn't use atomic_flag because it's harder to add some debug checks into that, compared to a custom class.

Note that other parts of the code use regular mutexes.

As for mutable, the reason is that you don't need to take the lock if you want to inspect most variables. They're atomics. So if you're going to take the lock, it's typically going to be because you're doing more invasive changes.

roy7:
Part of where this came up is when I added a function to calculate a value and return it and I needed a lock to be sure it was thread safe. No problem. But I was using this value in another function elsewhere that I was trying to lock, so that caused a deadlock situation. So I removed one of the locks but I was also getting conflicts that I can't lock within a 'const'. So I removed the const as well.

The const makes since I'm not changing anything, but I need the lock since I'm reading multiple values that could change.

But I wonder... could I have simply make a single std::atomic struct object? So no lock required and I'd get all of the values I needed at once instead of having multiple separate atomic variables?

gcp:
>a single std::atomic struct object? So no lock required 

No, there's only a few things that are atomic (processor dependent), and the others are emulated with a lock anyway.

gcp:
>I need the lock since I'm reading multiple values that could change.

Do you have a more concrete example?

roy7:
I added `std::atomic<double> m_variance{0.0}` to UCTNode.h. I use that incrementally as we search so I can get real variance at any time without actually storing all of the search results. I replaced update() with:

```Js
[...]

PR #1432, opened 2018-05-16
Mutex tts

OmnipotentEntity:
This pull request changes the mutex from TS to TTS behavior.  This addresses one of https://github.com/gcp/leela-zero/issues/1431

From the Wikipedia page on Test and Test-and-Set:

> In computer science, the test-and-set CPU instruction is used to implement mutual exclusion in multiprocessor environments. Although a correct lock can be implemented with test-and-set, it can lead to resource contention in a busy lock (caused by bus locking and cache invalidation when test-and-set operation needs to access memory atomically).
> 
> To lower the overhead a more elaborate locking protocol test and test-and-set is used. The main idea is not to spin in test-and-set but increase the likelihood of successful test-and-set by using the following entry protocol to the lock:

```
boolean locked := false // shared lock variable
procedure EnterCritical() {
  do {
    while (locked == true) skip // spin until lock seems free
  } while TestAndSet(locked) // actual atomic locking
}
```

And from the page on test-and-set:

> The four major evaluation metrics for locks in general are uncontended lock-acquisition latency, bus traffic, fairness, and storage.[7]
> 
> Test-and-set scores low on two of them, namely, high bus traffic and unfairness.
> 
> When processor P1 has obtained a lock and processor P2 is also waiting for the lock, P2 will keep incurring bus transactions in attempts to acquire the lock. When a processor has obtained a lock, all other processors which also wish to obtain the same lock keep trying to obtain the lock by initiating bus transactions repeatedly until they get hold of the lock. This increases the bus traffic requirement of test-and-set significantly. This slows down all other traffic from cache and coherence misses. It slows down the overall section, since the traffic is saturated by failed lock acquisition attempts. Test-and-test-and-set is an improvement over TSL since it does not initiate lock acquisition requests continuously.
> 
> When we consider fairness, we consider if a processor gets a fair chance of acquiring the lock when it is set free. In an extreme situation the processor might starve i.e. it might not be able to acquire the lock for an extended period of time even though it has become free during that time.
> 
> Storage overhead for TSL is next to nothing since only one lock is required. Uncontended latency is also low since only one atomic instruction and branch are needed.

### DIFF
--- a/src/SMP.cpp
+++ b/src/SMP.cpp
@@ -32,7 +32,10 @@ SMP::Lock::Lock(Mutex & m) {
 
 void SMP::Lock::lock() {
     assert(!m_owns_lock);
-    while (m_mutex->m_lock.exchange(true, std::memory_order_acquire) == true);
+    // Test and Test-and-Set reduces memory contention
+    do {
+      while (m_mutex->m_lock);
+    } while (m_mutex->m_lock.exchange(true, std::memory_order_acquire) == true);
     m_owns_lock = true;
 }
 
--- a/src/SMP.cpp
+++ b/src/SMP.cpp
@@ -33,9 +33,11 @@ SMP::Lock::Lock(Mutex & m) {
 void SMP::Lock::lock() {
     assert(!m_owns_lock);
     // Test and Test-and-Set reduces memory contention
-    do {
-      while (m_mutex->m_lock);
-    } while (m_mutex->m_lock.exchange(true, std::memory_order_acquire) == true);
+    // However, just trying to Test-and-Set first improves performance in almost
+    // all cases
+    while (m_mutex->m_lock.exchange(true, std::memory_order_acquire)) {
+      while (m_mutex->m_lock.load(std::memory_order_relaxed));
+    }
     m_owns_lock = true;
 }
 


gcp:
I'm a bit wary of trusting a wikipedia article on this. The problem is what you expect the ```while (m_mutex->m_lock);``` line to do. Because m_lock is an atomic, I think by default you're getting a load with sequential consistency barriers enforced, which contrasts with the original implementation that only requires an acquire barrier.

You're also getting both barriers in the not contended case, which is the important one.

OmnipotentEntity:
Thanks for the feedback.  I've looked into it and run some tests.  And I believe that I've come up with a better implementation than both TS and TTS.

My test code is located here: https://github.com/OmnipotentEntity/mutex-test

My analysis is:

Threads | TS | TTS | TS_TTS | std
-- | -- | -- | -- | --
1 | 0.023972874 | 0.026053339 | 0.023923787 | 0.029143106
2 | 0.116962586 | 0.134221597 | 0.107408329 | 0.15834735
3 | 0.23190067 | 0.21676568 | 0.20705445 | 0.23094961
4 | 0.37241236 | 0.29830395 | 0.27341388 | 0.40447911
5 | 0.53444316 | 0.39559964 | 0.38254134 | 0.54423071
6 | 0.76065617 | 0.56397305 | 0.50434288 | 0.63581543
7 | 1.01450452 | 0.72960247 | 0.64581788 | 0.69388562
8 | 1.3041033 | 0.90854151 | 0.80155509 | 0.76378738
9 | 1.6494017 | 1.08732991 | 0.94580802 | 0.85084046
10 | 2.0444408 | 1.2777898 | 1.12159761 | 0.94789737

Issue #1361, opened 2018-05-07
Sigmoid function problem with PUCT formula


jkiliani:
I looked into the strange play I've seen lately from the ELF network whenever it was at a significant disadvantage, and observed that at same node count, the search using the ELF net is a lot deeper in a balanced position compared to a highly unbalanced one (e.g. handicap starting position):

<details>
  <summary>Starting position ELF output</summary>

```
Leela: go
Thinking at most 36.0 seconds...
NN eval=0.496108
Playouts: 23, Win: 36.09%, PV: Q4 D4 Q16 D17
Playouts: 47, Win: 41.71%, PV: Q4 D4 Q16 D17 C15 C16 D15
Playouts: 72, Win: 44.00%, PV: D16 D4 Q16 R4 P3 Q3 P4 R6
Playouts: 103, Win: 45.43%, PV: Q4 D4 Q16 D17 C15 C16 D15 F17 C6 F3
Playouts: 130, Win: 46.04%, PV: Q4 D4 Q16 D17 C15 C16 D15 F17 C6 F3 C4 C3
Playouts: 156, Win: 46.45%, PV: D16 Q16 D4 Q3 R5 R4 Q5 O3 R14 O17 R16 R17 S17
Playouts: 188, Win: 46.91%, PV: D16 D4 Q16 R4 P3 Q3 P4 R6 F3 C6 D3 C3 C2 E3 D2
Playouts: 216, Win: 47.06%, PV: Q4 D4 Q16 D17 C15 C16 D15 F17 C6 F3 C4 C3 B3 C5 B4 D5 D3 B5
Playouts: 244, Win: 47.21%, PV: Q4 D4 Q16 D17 C15 C16 D15 F17 C6 F3 C4 C3 B3 C5 B4 D5 D3 B5 E3
Playouts: 275, Win: 47.38%, PV: Q4 D4 Q16 D17 C15 C16 D15 F17 C6 F3 C4 C3 B3 C5 B4 D5 D3 B5 E3 F4
Playouts: 313, Win: 47.53%, PV: Q4 D4 Q16 D17 C15 C16 D15 F17 C6 F3 C4 C3 B3 C5 B4 D5 D3 B5 E3 F4 C10
Playouts: 344, Win: 47.61%, PV: Q4 D4 Q16 D17 C15 C16 D15 F17 C6 F3 C4 C3 B3 C5 B4 D5 D3 B5 E3 F4 C10

[...]

Issue #1313, opened 2018-05-03
Handicap training


Dorus:
There was some talk about training a handicap NN in the other issue, however it's already quite long and handi games are offtopic there too. So want to continue that part of the discussion elsewhere. 

https://github.com/gcp/leela-zero/issues/1311#issuecomment-386282213
> herazul commented 9 hours ago
>
> I think the most interesting thing by far would be to train  a network that can handle several stones handicaps and weird komi, but i  have no clue how. Seems like Golaxy kinda managed to do it, but i have  no idea how.

https://github.com/gcp/leela-zero/issues/1311#issuecomment-386283469
>  gcp commented 9 hours ago 
>
> Yes, this is something that would work. If it is the best way to do it, I do not know. Training with (high) handicap is trickier, as you somehow need to  ensure the games are not too lopsided. So you need imbalanced playouts,  or different temperature, or something.

https://github.com/gcp/leela-zero/issues/1311#issuecomment-386287315
>  Eddh commented 9 hours ago •
>
> For a next run I agree that the most interesting things to try would be variable komi and handicap. Following @bittsitt idea, it should be possible to have 1 input plane for komi 7.5, 1 input plane for komi 6.5, 1 input plane for komi 0.5 etc ?
> 
> For handicap games, would it be possible to have more random conditions for the game, in addition to the random komi ? Placement of random stones on the board, somewhat randomized playout count, etc. Randomized training condition have already been used in neural network training https://www.youtube.com/watch?v=9EN_HoEk3KY&t=23m36s so that the neural net can better adapt to new conditions.
> 
> Also, so that the neural net can take into account the end score rather than just win/loss, is it possible to score a games as 1.15 if it is won by resign, but also 1.15 if it is won with a score that is more than 15. That way the system would want to resign if it is losing by more than 15 points difference, but carry on otherwise, which would be similar to human play.
> 
> I am aware that all the pretty ideas will need work and expertise to be implemented at some point though.

[...]

PR #1435, opened 2018-05-16
[Experimental] enhance tree search in low winrate situations

alreadydone:
Idea: use the existing NN (without additional training) more effectively in low winrate situations by modifying the tree search. 

#### 1. make c_puct proportional to winrate
When winrate is low, Makes tree search more sensitive to slight change in winrate and not so sensitive to policy prior. Potentially benefits handicapped play; getting from 0.1% to 1% would be a big deal. (Note that the opponent has high (close to 100%) winrate so c_puct on his/her turn is twice as usual (~50% winrate).) fpu reduction is also made proportional to winrate though I'm not sure whether this is the right thing to do.

Suggested way of testing: 
+ match /next with higher playouts against /next with lower playouts but with handicap stones (using the same network)
+ match my patch with higher playouts against /next with lower playouts but with handicap stones
+ compare to see if my patch fares better
+ **added**: when matched against /next with equal playouts in even games, the performance should not be worse; it's designed to work exactly the same way when winrate is 50%

I'll test myself when I manage to compile. ~~I tried to build in VS2015 through university's cloud service but keep getting 
` fatal error C1033: cannot open program database '\\Client\G$\leela-zero-next\msvc\VS2015\x64\Debug\vc140.idb' `~~ (moving files to local storage on the cloud fixed it)

#### 2. simulate a weaker opponent in tree search by effectively reducing playouts (work in progress)

When backing up the winrate of a newly expanded node to its ancestors, ignore the opponent's moves with a certain probability (depending on current winrate). More precisely, the Q-value of the opponent's last move will be updated with say 90% probability, the second-to-last move with 81% probability, the third-to-last 72.9%, and so on. (Namely the backup process has a 10% chance of terminating at each of the opponent's moves.) The probability should depend on the winrate (~~of that node or~~of the root node) or the strength difference with the actual opponent, but I haven't come up with a formula. I think I need to add ~~an argument (whether it's the opponent's turn) and~~ (let's only do partial backup for white since white is the weaker side in normal handicap games) an output (whether backup has terminated) to `play_simulation` to achieve what I want; I'll do it ~~tomorrow~~ later.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -274,7 +274,7 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     if (!is_root || !cfg_noise) {
         fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy) * pure_eval / 0.5;
     }
-    // Estimated eval for unknown nodes = original parent NN eval - reduction
+    // Estimated eval for unknown nodes = current parent winrate - reduction
     auto fpu_eval = pure_eval - fpu_reduction;
 
     auto best = static_cast<UCTNodePointer*>(nullptr);
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -75,6 +75,7 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
     }
     // We'll be the one queueing this node for expansion, stop others
     m_is_expanding = true;
+    lock.unlock();
 
     const auto raw_netlist = Network::get_scored_moves(
         &state, Network::Ensemble::RANDOM_SYMMETRY);
@@ -87,7 +88,6 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
         m_net_eval = 1.0f - m_net_eval;
     }
     update(m_net_eval);
-    lock.unlock();
     eval = m_net_eval;
 
     std::vector<Network::ScoreVertexPair> nodelist;
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -269,11 +269,13 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     // Lower the expected eval for moves that are likely not the best.
     // Do not do this if we have introduced noise at this node exactly
     // to explore more.
+    
+    auto pure_eval = get_pure_eval(color); 
     if (!is_root || !cfg_noise) {
-        fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy);
+        fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy) * pure_eval / 0.5;
     }
     // Estimated eval for unknown nodes = original parent NN eval - reduction
-    auto fpu_eval = get_net_eval(color) - fpu_reduction;
+    auto fpu_eval = pure_eval - fpu_reduction;
 
     auto best = static_cast<UCTNodePointer*>(nullptr);
     auto best_value = std::numeric_limits<double>::lowest();
@@ -289,7 +291,7 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
         }
         auto psa = child.get_score();
         auto denom = 1.0 + child.get_visits();
-        auto puct = cfg_puct * psa * (numerator / denom);
+        auto puct = cfg_puct * psa * (numerator / denom) * pure_eval / 0.5;
         auto value = winrate + puct;
         assert(value > std::numeric_limits<double>::lowest());
 
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -75,7 +75,6 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
     }
     // We'll be the one queueing this node for expansion, stop others
     m_is_expanding = true;
-    lock.unlock();
 
     const auto raw_netlist = Network::get_scored_moves(
         &state, Network::Ensemble::RANDOM_SYMMETRY);
@@ -87,6 +86,8 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
     if (state.board.white_to_move()) {
         m_net_eval = 1.0f - m_net_eval;
     }
+    update(m_net_eval);
+    lock.unlock();
     eval = m_net_eval;
 
     std::vector<Network::ScoreVertexPair> nodelist;
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -168,6 +168,7 @@ SearchResult UCTSearch::play_simulation(GameState & currstate,
         if (currstate.get_passes() >= 2) {
             auto score = currstate.final_score();
             result = SearchResult::from_score(score);
+            node->update(result.eval());
         } else if (m_nodes < MAX_TREE_SIZE) {
             float eval;
             const auto had_children = node->has_children();
@@ -189,12 +190,13 @@ SearchResult UCTSearch::play_simulation(GameState & currstate,
             next->invalidate();
         } else {
             result = play_simulation(currstate, next);
+            if (result.valid()) {
+                node->update(result.eval());
+            }
         }
     }
 
-    if (result.valid()) {
-        node->update(result.eval());
-    }
+
     node->virtual_loss_undo();
 
     return result;


Marcin1960:
uctnode.cpp(260): error C3861: 'get_pure_eval': identifier not found

Marcin1960:
@alreadydone 

Running your change on KGS as LeelaZeroT (up to 9 stones)

alreadydone:
@Marcin1960  Appreciated! Looking at the H4 [game](https://www.gokgs.com/gameArchives.jsp?user=LeelaZeroT) won against fater [2k], it no longer does 3-3 invasion for the first two moves, so I think this is in the right direction! How many playouts do you get? With the official branch the approach move is visible in Lizzie after about 1k visits, and on my weak machine I can't wait long enough for it to become the suggested move.

Unfortunately in the H9 game it still did 3-3 invasion. Lizzie shows the next suggested move is to attach. Maybe this is a policy suggestion and the winrate has been rounded down to 0 ... with ELF it will be even worse. I wonder whether higher accuracy in NN evaluation would help ... I'll also implement the other idea ASAP.

jkiliani:
I opened an issue in https://github.com/gcp/leela-zero/issues/1361 about the same problem, although I did not open a PR for it. About your PR: I think the function to scale the PUCT term with should use the inverse logistic function, i.e. logit log(x/(1-x)), to transform the winrates back to a continuous spectrum with a roughly constant distribution width around the parent node. To work both for low and high winrates equally, I'd recommend the derivative of the logit function as a scaling factor, normalised to give an output of 1 for balanced positions, which is 1/(4*Q*(1-Q)). E.g. for Q=0.01, the scaling factor would be ~25, same as for Q=0.99, while for Q=0.5 the scaling is 1. The current PR would likely work for very small winrates, but not all that well for large ones, i.e. won positions.

Hersmunch:
I think it would be a good idea to also have a test match to check that if facing a stronger opponent then it is also not worse than it is currently.  Maybe also with a handicap i.e. check that it doesn't lose more often from a what it thinks is a winning position. 
Am I remembering right that when facing a stronger opponent it is common that the winrate might actually be decent but then suddenly drop when a missed sequence is discovered?

jkiliani:
@alreadydone I agree that it is not important for high winrates to be treated symmetrically. But in that case I propose to use 4*Q*(1-Q) for Q<0.5, and 1 for Q>0.5, that way the search would behave exactly like it currently does while it's winning, and only very slightly different when it's at a slight disadvantage.

By the way, did you test that the code behaves correctly for Q = 0? I tried my version and it seemed that my code crashed in cases where puct was calculated to 0 resulting from Q = 0.

alreadydone:
Sadly I haven't been able to compile yet. If you have a sample sgf @Marcin1960 may be able to confirm. However notice that in the H9 games on KGS the winrate was already >99.995% (since Lizzie showed 100%) and maybe exactly 100% in some cases?.
Changing factor to 1 for Q>0.5 may or may not be better but I tend to leave it as it is since it's more natural to me; anyway the PR needs tests and it's just an additional test to find out which is better.
By the way I have a weak PC and it will be hard to test on Big Red II (needs to run two leelaz instances), so any tests are appreciated!

jkiliani:
I think many nodes have winrate rounded to 0 or 1 in highly unbalanced positions, since the neural net value head uses single precision floats. For this case, the search does not work efficiently whichever transform or function we use on puct and fpu reduction. A stronger player against handicap would likely be the Alphago Master (or classic Leela) architecture, where Monte Carlo rollouts are combined with the neural net evals. I'm very interested if your idea of only partially backing up the opponent evaluations can simulate a weaker opponent somehow similarly to rollouts.

alreadydone:
I can now compile without problem, but my focus temporarily shifted; see https://github.com/gcp/leela-zero/pull/1440.
I now plan to look at the effect of accurate evaluations after the precision fix before experimenting with idea 2.

jkiliani:
I don't think #1440 makes t=1 unnecessary in any way. It may be possible to make the search behave better with extreme winrates, but it doesn't solve the underlying problem that the neural net often makes bad decisions in uncommon situations, i.e. handicap play or after blunders, because it has no experience playing such situations. T=1 increases the variety of positions encountered, so it actually helps on that end, even if it may not necessarily suffice for optimal play against handicap.

gcp:
I think the underlying problem here is that you are assuming neural net evaluations have meaningful information when they're very high or very low. That assumption may or may not be true, but if it isn't, this entire thing is moot.

larspensjo:
@gcp So what you are saying is something like this?

A naive interpretation when comparing a1% win rate to a 2% win rate would be that the second one has twice as high chance to win. But given the average error on the win rate, it may be impossible to compare a 1% win rate to a 2% win rate in a meaningful way.

alreadydone:
> especially considering that numerical accuracy of network inference also limits even the potential quality of the evaluation.

When winrate is close to 0 or 1, the direct output of value head is the winrate in the log scale, so I don't think network inference will be particularly inaccurate in extreme situations. The relative error of the winrate due to rounding errors will not be higher in extreme situations.

The purpose of t=1 all game by my understanding is to allow more blunders that reverse the outcome of the game, so as to make the value head "less stratified" through training, but I argue that it may not be necessary with the precision fix. t=1 all game perhaps also makes encountered positions richer and helps with generalization.
High-handicap play is likely rare situations even with t=1 in the opening, but t=1 all game doesn't help to create similar situations. Situations with extreme winrates are not rare in no-resign games. Only situations after a blunder may be something new...

>the neural net often makes bad decisions in uncommon situations

The neural net correctly judges the situation to be hopeless with low winrate output; it's tree search's task to find out how to carry on. I argue that in such situations you should listen more to the winrates and less to the policy (e.g. you don't want to play 3-3 when given 4-stone handicap, giving out the whole center to the opponent), so I think accurate winrates are especially important here.

BTW my first commit in #1440 has absolutely no ill effect and is the correct way to calculate the quantity IMO.

jkiliani:
#1435 may well help to some extent, but I'm more dubious about #1440. From what I tried myself in handicap games against LZ, I think the problem is that LZ always assumes optimal play on the opponent's side, and in those circumstances, beating a handicap is as hopeless as the network assumes. Promising approaches in addition to t=1 which I think will alleviate the problem at least are training games specifically with handicap ( #1313) and your second suggestion regarding partial backpropagation of opponent moves.

alreadydone:
I agree that with 4- handicap stones the precision fix may have little effect, but for 5+ handicap stones precision is certainly a problem (yielding zero winrate). I plan to analyze the effect of c_puct (either with factor wr or wr(1-wr)) on these three [handicap games](https://www.reddit.com/r/cbaduk/comments/8ju8kz/youngsam_kim_8d_vs_ai_3_handicap_games_comments/) ([sgf](https://www.reddit.com/r/cbaduk/comments/8jozve/youngsam_kim_8d_egf_10d_aga_equivalent_tested_elf/)); these all showed ladder problems.

alreadydone:
I compiled Windows binaries at https://github.com/alreadydone/lz/releases/tag/c_puct but haven't been able to do thorough analysis. (My PC is weak anyway and cannot catch up with baduk1's playouts.) Looking at the 3-stone game with ELF, it seems that both linear and quadratic versions choose some moves differently, and the quadratic one chooses more differently (e.g. moves 78, 96, 100).

baduk1:
My lizzie 4 with ELF weights also crushed after I substituted leelaz file to a quadriple one. How can I solve this issue? Thanks

Hersmunch:
To follow up your report of a crash here https://github.com/gcp/leela-zero/issues/1470#issuecomment-391260648 I ran it up on the windows debugger. I get a null_ptr exception at the end of `uct_select_child()` when it tries to `best->inflate()`.
So this happens when the node has 0 visits and therefore `get_pure_eval()` returns -NaN meaning no best is found.
I think this is because there is a race in `play_simulation()` where two threads attempt to expand the same node. Only one manages to and the other gets to `if (node->has_children() && !result.valid())` which is true. The node still doesn't have any visits recorded yet (done further down in `node->update(result.eval())`). 
Normally this wouldn't matter as visits isn't used in `uct_select_child()`.

You reported that https://github.com/gcp/leela-zero/pull/1432 appears to resolve this but I can see it happen with it too, but less often.

Splee99:
What would be the difference if I use get_net_eval() instead of get_pure_eval()?

alreadydone:
Sorry for not being clear: I saw the crash with or without #1432. I have a patch for this crash to do update for newly visited node earlier within a lock; I tested it and it seems to work. @gcp Should I open a new PR or only add it here? The crash doesn't seem to appear without calling `get_pure_eval`, but it would be desirable for people to be able to use current Q-value more conveniently and not just the NNeval.

alreadydone:
@Splee99 get_net_eval is the NN eval of the node, and get_pure_eval is the average of the NN evals of the subtree under the node, dynamically updating. @jkiliani said above:

> By the way I think replacing initial parent FPU with dynamic parent is the right choice since tests with Leela Chess showed better performance for dynamic FPU. However it's likely that FPU reduction will need retuning with this, in Leela Chess the optimal FPU reduction with dynamic parent FPU was always smaller than with initial parent.

src/UCTNode.cpp
```diff
@@ -75,7 +75,6 @@ bool UCTNode::create_children(std::atomic<int>& nodecount,
     }
     // We'll be the one queueing this node for expansion, stop others
     m_is_expanding = true;
-    lock.unlock();
 
     const auto raw_netlist = Network::get_scored_moves(
```
Hersmunch: This takes a long time! I doubt we want to be locking over it.

src/UCTSearch.cpp
```diff
@@ -168,6 +168,7 @@ SearchResult UCTSearch::play_simulation(GameState & currstate,
         if (currstate.get_passes() >= 2) {
             auto score = currstate.final_score();
             result = SearchResult::from_score(score);
+            node->update(result.eval());
```
Hersmunch: Also needed on line 179

alreadydone: that one is done in UCTNode.cpp in create_children

src/UCTNode.cpp
alreadydone: I think once the tree grows large it's not so likely that two threads hit the same unexpanded node simultaneously? We won't be able to go below that node anyway if the NN result hasn't been returned yet? You won't even have the policy.

src/UCTSearch.cpp
Hersmunch: That updates `m_net_eval` but `node->update()` accumulates into `m_blackevals`.

src/UCTNode.cpp
Hersmunch: Agreed. Perhaps a way to solve it without extra locking time is to explicitly gate the search from going deeper if `m_net_eval` of the node is some initial value indicating it hasn't been populated yet?

src/UCTSearch.cpp
alreadydone: I think that is correct? The NN output is for side-to-move, and m_net_eval is replaced by 1-m_net_eval when it's white to move to get blackeval.

src/UCTNode.cpp
alreadydone: I am not sure what is the disadvantage of a lock that lasts long; I am not familiar with how it works. It should definitely halt a thread going into the node (that thread has to wait anyway), but does it affect other threads?

Hersmunch: I may have jumped to a conclusion too soon here. I was thinking this would block other threads relatively often but having a closer look I only can see that if another thread was exploring the expanding node's parent that it would have to wait e.g. in `uct_select_child()`

alreadydone: OK I now think extending the lock isn't necessary after all. The mutex unlock when link_nodelist returns (after all the psa_ratio stuff is done) signals that create_children has finished and the new node is ready for uct_select_child (in play_simulation, has_children will be true); however, the new node may not have been updated with the NN eval. So it suffices to do the update before going into link_nodelist. I moved the unlock back, compiled and tested, and the crash no longer occurs.

Splee99:
@alreadydone I have run several handi games between LZ and (original) leela and no crash was observed. However, I also don't see big improvement over the released version LZ. This version can boost up the winrate when it would be less than 0.1% with the released version, but it may also play 1 line silly moves when winrate is low. See the last few moves in here:
http://eidogo.com/#AZKEtQZc
Both this version and the released version lose 4 handi games against Leela 0.11 and won 3 handi games against Leela 0.11. I'll do more tests.

alreadydone:
@Splee99 Thank you for your testing, feedback and the game record! Can you share the setting you use to test LZ against Leela? I may want to do some tests with different parameters, and it would be good if we can compare the results.

So after some development, testing and bug fix, I am now ready for a release of the [source code](https://github.com/alreadydone/lz/tree/tree-search-lizzie) (for the changes see https://github.com/alreadydone/lz/pull/17) and a [Windows binary](https://github.com/alreadydone/lz/releases/tag/v1.0). This version of leelaz takes four additional command line options (defaulted at `--puct-factor 1 --backup-type 3 --backup-pct 50 --pseudo-backup on`). Since many features remain to be tested and optimized, I didn't add these changes to this PR.

I allow these many parameters to facilitate experimenting with them; it just need to modify a few characters in mylizzie.json to switch between parameters and see the effect in the MyLizzie GUI (does Lizzie support any command yet?). I haven't run any test matches, only some sanity checks about not invading 3-3 in 4-5H games and some quick self-play games (with asymmetric tree search). You are welcome to tweak the parameters or even modify the code to approach optimal strength in handicap games.

The `--puct-factor` option 0, 1, and 2 correspond to a constant factor before c_puct, a factor linear in winrate (my original proposal), and a quadratic factor (as @jkiliani suggested, which makes both sides more sensitive to the winrate), respectively. The default is set to 1.

For the other three options, some explanations are in order: MCTS as described in the AGZ paper works by doing many simulations (playouts) starting from the root node, i.e. the position where you want to choose a move to play. Each simulation, when reaching a leaf node in the search tree, will expand it, getting the winrate and policy from the neural net, and choose a move there to play, thereby adding one new node to the search tree; the winrate is backed up along the path this simulation traversed: for each node traversed, the new node's winrate will be added to the accumulated winrate of this node, and divided by the visit count (i.e. the number of times this node is traversed) to get the new evaluation (averaged winrates) of this node. I call this process the updating of this node.

My idea 2 is to partially update/backup winrates to simulate a weaker opponent by effectively allow the opponent (black) a lower number of playouts, and I proposed to do it by partially updating the black nodes for each simulation. Since during each simulation each move is chosen according the the fixed policy prior probabilities and the dynamically updated winrates, reducing winrate updates presumably has the effect of dumbing down black's choices. I implemented six types of backing up, specified through the `--backup-type` parameter; in all cases, the new node itself is always updated. Except for Type 1 and 2, backup depends on a probability p to be explained later.

0. Back up the winrate one level up (i.e. to the previous black's move) with probability p (originally proposed).
1. Always backup.
2. Never backup.
3. For each simulation, update all nodes with probability p, and update none of them with probability 1-p (default).
4. Independently for each node, update with probability p.
5. For a simulation of depth n, update the first p * n nodes in the path (some weird recent idea).

With Type 3 and 4, the effective visit count (number of updates) of black nodes will be close to p times the visit count of white nodes.
With Type 0, when p is small the effective visit counts of black nodes near the root node can be very low, and this hinders exploration (the PUCT formula encourages going into less visited nodes as the total visit count grows). Therefore I introduce "pseudo-update" to increment the visit count without actually changing the evaluation (averaged winrates). This is by default enabled, but you can turn it off with `--pseudo-backup off`.

The probability p is determined by the winrate at the root node and the option `--backup-pct` (at most 100, but negative values are accepted as well). Currently I am using a very crude formula, a quadratic function `calc_backup_pct` which approaches 100% quadratically at 50% winrate and --backup-pct at 0% winrate. It might be good to take the number of handicap stones into account.

Endnotes:
+ The precision fix isn't complete yet, so with 5+ stones some winrates may be literally indistinguishable from 0 or 1. 
+ Tree reuse may have a notable effect on these changes but there isn't a switch to allow convenient testing.)
+ Since it may be desirable to tune c_puct and fpu reduction after these changes, I enabled those command line options. 

Enjoy playing with this new gadget; any code contribution, suggestions of improvements, and testing is highly welcomed!

Splee99:
@alreadydone I used time_settings command to give both LZ and Leela 0.11 15 sec per move. Also for both the number of  threads is set to 2 and no ponder is used. Leela 0.11 is given nobook option.

By the way , it is amazing that LZ didn't prefer 3x3 invasion with this version.

alreadydone:
@Splee99 Thanks. Regarding 3-3 invasion, are you sure it's not due to using a different network? It has been observed that recent networks prefer 3-3 invasion less in the opening (link [1](https://www.reddit.com/r/cbaduk/comments/8l8mo0/threethree_invasion_is_now_a_rare_event_in_leela/) [2](https://www.reddit.com/r/cbaduk/comments/8lxrgp/leela_zero_opening_study_things_got_weird_lately/)). 

It seems the new version is not optimal in avoiding 3-3. In my tests using network 7c56 with 4H, the new version (default setting) needs 1796 playouts to avoid 3-3 on first move, the old linear version needs 1219 playouts, and the old quadratic version needs only 941 playouts, on average.
The old linear version should behave in the same way as `--puct-factor 1 --backup-type 1`, and the old quadratic version is `--puct-factor 2 --backup-type 1`. The official leelaz is `--puct-factor 0 --backup-type 1` and it does take much longer to avoid 3-3 because of insensitivity to winrates.

Splee99:
@alreadydone I haven't test many game runs, but for the same network I observed that the LZ release version sometimes played the first move at 3-3. With this version it is never the case.

alreadydone:
OK so you are comparing the official version with my versions. The **c_puct change** was done one week ago and I immediately [observed](https://github.com/gcp/leela-zero/pull/1435#issuecomment-389589675) the change from 3-3 invasion to corner approach; it's not clear how many playouts were achieved there, but my tests show 1k+ playouts usually suffice to avoid 3-3. What you used in https://github.com/gcp/leela-zero/pull/1435#issuecomment-391898894 should be no different from what @Marcin1960 used on KGS except for a multithreading bug fix. My latest release https://github.com/gcp/leela-zero/pull/1435#issuecomment-391984240 has the **partial backup** idea implemented; though I didn't see it makes 3-3 even less favorable, I expect it to improve in other aspect of handicap games.

Splee99:
@alreadydone I have downloaded your latest release code. There are so many parameters to tweak. The default setting seems to be not so optimal. I would explain from the following log:
```text
 P16 ->   17926 (V:  3.09%) (N: 62.24%) PV: P16 Q15 O14 R8 Q9 Q8 P8 Q6 O12 R9 Q10 P7 O8 Q12 O10 E18 E17 F18
 S17 ->      66 (V:  2.62%) (N:  5.46%) PV: S17 S18 P16 Q15 O14 N16 O16 N15 P14 O12 N13 N12 M13
 S18 ->       7 (V:  1.74%) (N:  1.59%) PV: S18 S17 R19 T16
 R19 ->       6 (V:  1.81%) (N:  1.19%) PV: R19 S17 P16 T16 N16 R8
  L4 ->       5 (V:  2.45%) (N:  0.54%) PV: L4 L5 M5 L6
  G6 ->       3 (V:  1.81%) (N:  0.92%) PV: G6 H5 P16
12.0 average depth, 25 max depth
13368 non leaf nodes, 1.35 average children
18021 visits, 5125036 nodes, 5715 playouts, 197 n/s
```
It seems like to think very deep in a single direction and the rest of possible moves are almost ignored. At least I would expect the number of visits of the second best move to be in the same order as the first choice. Could you please tell me how to tweak the parameters to overcome this?

alreadydone:
@Splee99 There has been a bug making partial backup not behaving as intended, which is fixed by the three commits I just added to `tree_search`. I haven't compiled off it so you may encounter compilation errors... I remember that this change makes the search more spread out, but when I tested 2-stone games as @OmnipotentEntity did in #1491 I didn't observe an improvement (I only ran 5 games and it was 2-3).

Splee99:
@alreadydone So how can I temporarily disable the backup?

alreadydone:
You mean disabling pseudo-backup? You can add the option --pseudo-backup off.

--pseudo-backup off means neither incremeningt visit count nor updating winrate with certain probability.
--pesudo-backup on means incrementing the visit count but not updating winrate with certain probability.
--backup-type 1 means always incrementing visit count and updating winrate (official leelaz behavior).

Splee99:
I disabled the pseudo backup option and also added my modification in #1514. The result is quite good. The 4 handicap game with hiratuka only took 100 moves to get a winrate near 50%. I think it totally makes sense to use get_pure_eval to select moves,  because the weak opponent probably can't generate too much virtualloss.

alreadydone:
Glad to hear such a combined approach works, and thank you for testing!
Virtual loss is for multi-threading, so I am not sure what you meant here.
If the winrate is expected to change a lot as search deepens then it definitely makes sense to use get_pure_eval (dynamical eval) instead of net_eval.

Splee99:
// Return the true score, without taking into account virtual losses.
float UCTNode::get_pure_eval(int tomove) const {

compared to:
float UCTNode::get_eval(int tomove) const {
    // Due to the use of atomic updates and virtual losses, it is
    // possible for the visit count to change underneath us. Make sure
    // to return a consistent result to the caller by caching the values.

alreadydone:
I only replaced net_eval by get_pure_eval; I didn't replace any get_eval.
Virtual loss is applied when a thread's simulation enters a node or the subtree below it; it decreases the winrate of the node, and thereby discourages other threads to go into this node. 
https://github.com/gcp/leela-zero/blob/9e903eda6e93fb65375ea03127e4f62268913175/src/UCTSearch.cpp#L165
Virtual loss is undone after the simulation is over and the newly expanded leaf node's winrate has been backed up to this node:
https://github.com/gcp/leela-zero/blob/9e903eda6e93fb65375ea03127e4f62268913175/src/UCTSearch.cpp#L198

Splee99:
Thanks. That's good to know.

Splee99:
I don't understand what the status of this PR is. It certainly helps increasing the strength of LZ. The logic is like use the winrate for choosing the best move instead of calculating winrate after a move is selected (which results in sometimes the second best move has the highest winrate). Maybe we can simplify it first like just including the basic c_puct modification and get this PR passed.

alreadydone:
The only change in this pull request so far is making tree search more sensitive to winrate (and using dynamical eval for fpu, but that's minor), which may not be a good thing in handicap games (pruning moves with low policy seems to work, which tells us we should focus on policy and not winrate, and we shouldn't trust low winrate signals too much since they are noisy (increasingly so with recent networks? maybe an effect of randomness all game)). So it seems the appropriate decision is to close this.

I've now implemented "dynamical komi" at https://github.com/gcp/leela-zero/issues/1599 to keep winrate within a certain range, so in a sense we no longer need to deal with low winrate situations. Hopefully that approach works better for handicap games.

Splee99:
While dynamic komi may have success on human games, I'm not able to let it win a 4 handi game with Hiratuka with the same setting I used before with LZ with this PR. Again I think this PR ensures LZ not playing 3-3 in the beginning and also is able to create high risk high reward situations such as trading big groups. Although your original intention is to use winrate, but the pure_eval value gives priority to some high risk high reward moves.

alreadydone:
pure_eval is not the main point of this PR. The point was making the winrate term more significant than the probability/visit count term in the PUCT formula, so the tree search is more sensitive to winrate. I will see if merging this PR into komi-v0.3 improves performance; the winrate is kept low at 10%, so this PR should make tree search 5 times as sensitive to winrate as with 50% winrate.

gcp:
I'll close this as the dynamic komi approach seems much more promising.

Issue #1405, opened 2018-05-11
Phoenix Go open sources


bjiyxo:
https://github.com/Tencent/PhoenixGo

And their weights here.

https://github.com/Tencent/PhoenixGo/releases

AncalagonX:
They released their 20-block weights only. According to the .step file, there were 639,200 steps performed on this individual set of weights.

We're entering the Golden Age of Go right before the computers fully take over. I'm glad the Chess world lived through this before, so we can know confidently that the Go community will survive this and even become better for it.

AncalagonX:
We've already set the precedent by including the much stronger "Elf weights" as a teacher. Let's now do the same with these weights if they were also truly AGZ'ed from the start.

fell111:
Can we convert the weight file to fit for Leela Zero even if there is no training source code? We could have another anchor for our target. That's great!

sbbdms:
"cronus" on CGOS = PhoenixGo

Matrok:
Are we confident that they are based on the AGZ approach? I mean, the program probably is (they claim it is), but how has it been trained?

jillybob:
How does this program compare to ELF?

22nsuk:
Is it compatible?

Ttl:
Looks like the network is slightly different from LZ. Input planes are the same as in AGZ but the network uses pre-activation architecture where batch norm and relu are before the convolution. Modifications are needed in the OpenCL code if we want to use the weights in LZ.

l1t1:
nice job

jkiliani:
Is PhoenixGo any easier to compile and run than ELF OpenGo? If so, it would be nice to know how good those weights actually are, i.e. if making LZ compatible for these would be worth the effort...

l1t1:
This time we open source PhoenixGo source code and a 20 block model. Training source code and other source code are not open source because they are deeply coupled with the cloud computing platform within WeChat.

----
http://sports.sina.com.cn/go/2018-05-11/doc-ihamfahw3402974.shtml

这次我们开源PhoenixGo对弈源码和一个20 block的模型。训练源码和其它源码由于跟微信内部的云计算平台耦合较深、不易分离，暂不开源。

TableCarpet:
LZ needs more than 20 blocks to be able to become stronger than best 20 blocks programs.

hydrogenpi:
AGZ next week

Ishinoshita:
[...]

PR #1440, opened 2018-05-17
WIP: Fix imprecision when winrate is extreme

alreadydone:
https://github.com/gcp/leela-zero/blob/c822e5e9623e796fb6dd87e7d8b587a31401ad40/src/Network.cpp#L974

When winrate is close to 0, adding 1 to it destroys precision and potentially hinder the effectiveness of MCTS. The first commit replaces the above by an equivalent but more direct and accurate formula.

With 3 passes in the opening, the imprecision is already very severe:
<details><summary>Commands</summary>
```
leelaz -t 1 -w 90560d926216c9842add40913a023fd59fde761a64014c48d8246648ddd272a3.gz -s 3836721493350202251
playmove b pass
playmove w q16
playmove b pass
playmove w d4
playmove b pass
playmove w q4
genmove b
```
</details>

before the commit: NN eval=0.000000178813934326171875 = 3 x 2^-24
after the commit: NN eval=0.0000001838196084236187743954360485076904296875 = 12935155 x 2^-46

The relative error (2.7%) isn't large, but if the winrate can only take a limited number of values it makes tree search unable to discern better moves.
With 4 passes the eval before commit is identically 0, and the tree search probably can only select children based on policy.

----
This is only a partial fix; winrate close to 1 will always be problematic by design of floating numbers, unless represented internally by 1-x. It also does not directly helps handicapped play: since the handicap stones are normally black, white has low winrate, but we currently only keeps m_blackeval internally and base all Q-value calculations on it, and we do 1-x to get white winrate which again destroys precision.

To remedy black/white disparity, my suggestion is to introduce m_whiteeval and backup both blackeval and whiteeval; when selecting child, we use the color that makes parent get_pure_b/w_eval <= 0.5. @gcp What do you think? Maybe we turn on whiteeval only in handicap games...

----
Note that 3 passes lead to ~2^-23 winrate and single precision number (which I assume we are using) can get as low as [2^-149](https://en.wikipedia.org/wiki/IEEE_754-1985#Single_precision) so we are probably safe with (arbitrarily placed) 9-stone handicap (8 passes). I'll do some tests using LZ and ELF networks to find out. For winrate out of range we may want to use winrate_out[0] directly and double precision etc.

By the way, does anyone has an explanation why ELF 4-stone value prior is higher than LZ? (See https://github.com/gcp/leela-zero/issues/1405#issuecomment-389029467)

### DIFF
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -970,8 +970,8 @@ Network::Netresult Network::get_scored_moves_internal(
     const auto winrate_out =
         innerproduct<256, 1, false>(winrate_data, ip2_val_w, ip2_val_b);
 
-    // Sigmoid
-    const auto winrate_sig = (1.0f + std::tanh(winrate_out[0])) / 2.0f;
+    // Sigmoid: tanh normalized to take value in (0,1)
+    const auto winrate_sig = 1.0f / (1.0f + std::exp(-2.0f * winrate_out[0]));
 
     Netresult result;
 


gcp:
I don't think our neural network computation is this accurate, especially with things like USE_HALF, so this seems beyond pointless to me.

alreadydone:
I know that 2.7% is less than the OpenCL self-check threshold (5%), but 100% relative error with 5+ handicap stones is definitely problematic. Anyway I believe that the new formula is the correct way to compute the quantity and always gives more accurate result so I see no reason not to adopt it. However as I realized this alone helps little with either black or white in handicapped games, so it'd better to take this change together with the upcoming ones.

Notice that when winrate is close to 0 or 1 the direct output (before sigmoid) of value head is the winrate in log scale, so the NN has no trouble discern between two tiny winrates; the only problem is that the values are extrapolation/generalization to unfamiliar situations, so the prediction accuracy is questionable, but I think we should still use the information prudently to improve the quality of the tree search, and I'll go on to add whiteeval.

I plan to move sigmoid out of `get_scored_moves_internal` to `create_children`. As far I can see this only breaks @roy7's #1089 and Training.cpp will have to do sigmoid. Any better suggestions? Modifying the type Netresult to include two winrates seems worse to me...

alreadydone:
@gcp 
Quick questions: 
https://github.com/gcp/leela-zero/blob/03005313520ada8dafdefda9afc4d8ca424f0fb6/src/UCTNode.cpp#L255
Since we are no longer using transposition table, counting parentvisits again seems unnecessary? How about introducing a m_total_visited_policy for UCTNode and updating it after L290 in case child.get_visits() =0? Why don't you like to use `else` with `if` for example in the following? https://github.com/gcp/leela-zero/blob/03005313520ada8dafdefda9afc4d8ca424f0fb6/src/UCTNode.cpp#L213

PR #1445, opened 2018-05-17
Make Debian leelaz package by "make package" with CMake

ChinChangYang:
This patch can create a Debian leelaz package. I have tested the patch in Ubuntu 18.04. The following step-by-step introduces how to create the leelaz package.

`git submodule update --init --recursive`
`mkdir build`
`cd build`
`cmake ..`
`make`
`make package`

The package `leelaz-0.15.deb` will be generated. We can install the package by:

`sudo dpkg -i leelaz-0.15.deb`

Check the installation by:

`which leelaz; which autogtp; which validation`
`/usr/bin/leelaz`
`/usr/bin/autogtp`
`/usr/bin/validation`

Uninstall the package by:

`sudo dpkg -r leelaz`

Check if the package has been removed by:

`which leelaz; which autogtp; which validation`

It should return an empty line.


marcocalignano:
Has anyone tried if autogtp from /usr/bin works? leelaz has to be in the same execution directory? Do you need special privileges for the execution directory?

ChinChangYang:
autogtp works from /usr/bin if you have this patch, which finds leelaz if ./leelaz does not exist.

No special privileges are needed to execute /usr/bin/autogtp and /usr/bin/leelaz in my test.

CMakeLists.txt
```diff
@@ -135,3 +135,17 @@ target_link_libraries(tests ${BLAS_LIBRARIES})
 target_link_libraries(tests ${OpenCL_LIBRARIES})
 target_link_libraries(tests ${ZLIB_LIBRARIES})
 target_link_libraries(tests gtest_main ${CMAKE_THREAD_LIBS_INIT})
+
+SET(MAJOR_VERSION 0)
+SET(MINOR_VERSION 15)
```
gcp: It is not acceptable to hardcode this here.

CMakeLists.txt
```diff
@@ -135,3 +135,17 @@ target_link_libraries(tests ${BLAS_LIBRARIES})
...
+SET(CPACK_GENERATOR "DEB")
+SET(CPACK_DEBIAN_PACKAGE_NAME "leelaz")
+SET(CPACK_DEBIAN_PACKAGE_VERSION "0.15")
```
gcp: Same.

CMakeLists.txt
marcocalignano: Have a look at this:
https://github.com/rpavlik/cmake-modules/blob/master/GetGitRevisionDescription.cmake

ChinChangYang: Here is also a good read.

http://ipenguin.ws/2012/11/cmake-automatically-use-git-tags-as.html

Including GetGitRevisionDescription is a big change to do because you have to add the module under this project. Making debian package should be independent to the work of automatically versioning.

I am not sure where to put the GetGitRevisionDescription module in this project, so I will stop here unless we have a good way to get leela-zero version in cmake.

gcp: I'm not sure how hard it is to get it from the config.h header?

ChinChangYang: We need a single source of version number. Maybe the version number can be defined in a configure file though, there are more than one way to “make” in this project. Thinking what the best way is...

ChinChangYang: This project does not have ./configure to define a global version number, so I prefer to use GetGitRevisionDescription.cmake to get git tags as version strings. To be done...

gcp: I referred to the config.h header as that is where the program itself gets its version number from, so that would be a stable source. If you are willing to work with GetGitRevisionDescription, then that's even better (it's also a stable source).

ChinChangYang:
In this change, two cmake files

`GetGitRevisionDescription.cmake` and
`GetGitRevisionDescription.cmake.in`

are copied from [https://github.com/rpavlik/cmake-modules](https://github.com/rpavlik/cmake-modules). These cmake files are distributed under the Boost Software License, which is compatible with the GNU GPL (see [here](https://www.gnu.org/licenses/license-list.html#GPLCompatibleLicenses)).

Thanks for the newly added cmake module, cmake can use git tags as version strings now. The steps given in the top post are tested in Ubuntu 18.04 again. Everything looks the same.

gcp:
The automatically generated dependencies are unfortunately a bit flakey, i.e. if I build it on my system, it will list curart as a dependency, instead of the generic OpenCL loader.

ChinChangYang:
If some of the generated dependencies are inaccurate, we can set the dependencies in the cmakefile manually. How to make true dependencies needs more testing...

Issue #1425, opened 2018-05-14
Missing separator for lz-analyze


bittsitt:
Commit 9fd7542d puts each move on different lines, but that makes it much harder to parse and maintain results since the client controller/gui doesn't know which results belongs together. Previously it could just throw away the old data, and replace it with the new data (from one line).  Now it has to maintain a list, and search it every time it receives new data to see if it is a new move, or if he just has to update the data of an existing move. This is much more cumbersome and complicated. 

The original Lizzie used a special token (begin/end) to group analysis results, but as you need to maintain some extra state, I put everything on one line to simplify. 

I therefore suggest to revert 9fd7542d.

PS.I assume LZ guarantees that if a move has been searched at least once, it will be maintained so that the doesn't have to check if a move has disappeared (which would be a real pain as you would need some concept of time or something..).

gcp:
This was intentional and according to the proposals in #1388.

> since the client controller/gui doesn't know which results belongs together

Why should "results belong together"? I think this is already betraying some assumptions that might perhaps hold in current LZ, but not in a generic Go engine. 

It's certainly possible to have a design that searches one move at a time, or only transmits updates for one move, i.e. the one it just searched and thus knows to be the only one whose values are changed.

>Now it has to maintain a list, and search it every time it receives new data to see if it is a new move, or if he just has to update the data of an existing move.
>This is much more cumbersome and complicated.

I think that's a reasonable thing to do. Can't you just store the move info in a hashtable?

>PS.I assume LZ guarantees that if a move has been searched at least once, it will be maintained so that the doesn't have to check if a move has disappeared

I think that's going to be true for LZ for now, but to say it's a guarantee? If a move was selected early, got some playouts, and then didn't get any more attention for a few million visits (because it's a self-atari that loses instantly), do you still want to display it?

I think it would be entirely reasonable for the engine to not post search updates for moves it hasn't searched in the last interval, BTW.

bittsitt:
I just think a stateless protocol where neither the client or leela zero needs to remember what has happened before is a lot easier for both parties. "These are the moves I currently consider the best". Finito.

The alternative has some implications:
- Only report those that has been updated => LZ will have to keep track of what has already been reported to the client, or report every time something changes (the latter I suspect could be a lot of data which seems unnecessary(?)). What about initial state? If the engine is busy investigating the second best move, how long will it take before the gui is informed what the best move is?

[...]

PR #1478, opened 2018-05-24
Add order to analysis data

bittsitt:
Been a bit busy, but finally got around to this one. Address #1425.

I realized that we have to make a copy of the data since it can change at any time, making sorting difficult if we don't have a copy.  I added a class to store the data. Open to other suggestions.

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -25,6 +25,7 @@
 #include <limits>
 #include <memory>
 #include <type_traits>
+#include <algorithm>
 
 #include "FastBoard.h"
 #include "FastState.h"
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -40,6 +40,37 @@ using namespace Utils;
 
 constexpr int UCTSearch::UNLIMITED_PLAYOUTS;
 
+class OutputAnalysisData {
+public:
+    OutputAnalysisData(std::string move, int visits, int winrate, std::string pv) :
+        m_move(move), m_visits(visits), m_winrate(winrate), m_pv(pv) {};
+
+    std::string infoString(int order) const {
+        std::string tmp = "info move " + m_move + " visits " + std::to_string(m_visits) +
+                          " winrate " + std::to_string(m_winrate);
+        if (order >= 0) {
+            tmp += " order " + std::to_string(order);
+        }
+        tmp += " pv " + m_pv;
+        return tmp;
+    }
+
+    friend bool operator<(const OutputAnalysisData& a, const OutputAnalysisData& b)
+    {
+        if (a.m_visits == b.m_visits) {
+            return a.m_winrate < b.m_winrate;
+        }
+        return a.m_visits < b.m_visits;
+    }
+
+private:
+    std::string m_move;
+    int m_visits;
+    int m_winrate;
+    std::string m_pv;
+};
+
+
 UCTSearch::UCTSearch(GameState& g)
     : m_rootstate(g) {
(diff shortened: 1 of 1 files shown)


src/UCTSearch.cpp
```diff
@@ -40,6 +41,37 @@ using namespace Utils;
 
 constexpr int UCTSearch::UNLIMITED_PLAYOUTS;
 
+class OutputAnalysisData {
+public:
+    OutputAnalysisData(std::string move, int visits, int winrate, std::string pv) :
```
gcp: const std::string&

src/UCTSearch.cpp
```diff
@@ -40,6 +41,37 @@ using namespace Utils;
...
+                          " winrate " + std::to_string(m_winrate);
+        if (order >= 0) {
+            tmp += " order " + std::to_string(order);
+        }
+        tmp += " pv " + m_pv;
+        return tmp;
+    }
+
+    friend bool operator<(const OutputAnalysisData& a, const OutputAnalysisData& b)
+    {
```
gcp: backet at the end of the previous line

src/UCTSearch.cpp
```diff
@@ -40,6 +41,37 @@ using namespace Utils;
 
 constexpr int UCTSearch::UNLIMITED_PLAYOUTS;
 
+class OutputAnalysisData {
+public:
+    OutputAnalysisData(std::string move, int visits, int winrate, std::string pv) :
+        m_move(move), m_visits(visits), m_winrate(winrate), m_pv(pv) {};
+
+    std::string infoString(int order) const {
+        std::string tmp = "info move " + m_move + " visits " + std::to_string(m_visits) +
```
gcp: auto tmp

Also I think the general style of the code would make this get_info_string().

src/UCTSearch.cpp
```diff
@@ -236,6 +268,9 @@ void UCTSearch::dump_stats(FastState & state, UCTNode & parent) {
 }
 
 void UCTSearch::output_analysis(FastState & state, UCTNode & parent) {
+    // We need to make a copy of the data before sorting
+    std::vector<OutputAnalysisData> sortable_data;
```
gcp: auto sortable_data =

src/UCTSearch.cpp
```diff
@@ -251,13 +286,24 @@ void UCTSearch::output_analysis(FastState & state, UCTNode & parent) {
...
         auto move_eval = node->get_visits() ?
-            static_cast<int>(node->get_pure_eval(color) * 10000) : 0;
-        gtp_printf_raw("info %s %s %s %d %s %d %s %s\n",
-                       "move", move.c_str(),
-                       "visits", node->get_visits(),
-                       "winrate", move_eval,
-                       "pv", pv.c_str());
+                         static_cast<int>(node->get_pure_eval(color) * 10000) : 0;
+        // Store data in array
+        sortable_data.push_back(OutputAnalysisData(move, node->get_visits(), move_eval, pv));
```
gcp: emplace_back and you don't need to mention the constructor then

src/UCTSearch.cpp
```diff
@@ -251,13 +286,24 @@ void UCTSearch::output_analysis(FastState & state, UCTNode & parent) {
...
+
+    }
+    // Sort array to decide order
+    std::stable_sort(rbegin(sortable_data), rend(sortable_data));
+
+    int i = 0;
```
gcp: auto

bittsitt:
Should've remembered the auto... emplace_back was really neat!

PR #1492, opened 2018-05-30
Continue searching deeper on cache hit

Hersmunch:
The basic idea here was to keep on searching (calling `UCTSearch::play_simulation()`) whenever there is a NNCache hit. My aim was to allow the search to best make use of looking up previous network evaluations of transpositions so it can better decide how to spend the remaining playouts. This mainly came about because I realised that by using T= 1 for self-play we will now be probably getting more cache hits whenever a node with lower visits is played, but I digress..

My first attempt continued until there was a new network evaluation but that obviously removes the time saving of using the NNCache in the first place. So I modified it to stop the cache hit chain playout on a cache miss but without performing the network evaluation and instead, return the result up from the last cache hit. Here's my initial testing results:

Fixed 800 visits with self-play settings:
`validation.exe -b leelaz_next -o "-g -v 800 --noponder -t 1 -q -d -r 5 -m 999 -n -w" -n a39b64cc -b leelaz_no_stop -o "-g -v 800 --noponder -t 1 -q -d -r 5 -m 999 -n -w" -n a39b64cc`
```
              wins        black       white
a39b64cc  108 46.15%   48 45.71%   60 46.51%
a39b64cc  126 53.85%   57 54.29%   69 53.49%
                      105 44.87%  129 55.13%
```
Fixed 800 visits with match settings:
`validation.exe -b leelaz_next -o "-g -v 800 --noponder -t 1 -q -d -r 5 -w" -n a39b64cc  -b leelaz_no_stop2 -o "-g -v 800 --noponder -t 1 -q -d -r 5 -w" -n a39b64cc  `
```
              wins        black       white
a39b64cc   87 44.85%   34 43.59%   53 45.69%
a39b64cc  107 55.15%   44 56.41%   63 54.31%
                       78 40.21%  116 59.79%
```
Fixed time with match settings (using #1459 and note that with only 10 mins my machine can't do loads of visits but I would estimate ~500-800 from how long the moves took):
`validation_ts.exe -b leelaz_next -o "-g --noponder -t 1 -q -d -r 5 -w" -n a39b64cc  --ts "600 0 0" -b leelaz_no_stop2 -o "-g --noponder -t 1 -q -d -r 5 -w" -n a39b64cc  --ts "600 0 0"`
```
              wins        black       white
a39b64cc  114 46.53%   46 45.54%   68 47.22%
a39b64cc  131 53.47%   55 54.46%   76 52.78%
                      101 41.22%  144 58.78%
```

So it looks a little better? I can't say exactly how much extra cpu time is used but from what I saw whilst they were running, in some games the changed binary was approaching +10% compared to next and in others it didn't appear to use any extra time.

After I did this I found the discussion in https://github.com/gcp/leela-zero/pull/662 and even that @roy7 [had mentioned](https://github.com/gcp/leela-zero/pull/662#issuecomment-359480316) something that sounds a bit like this. 
Either way, it looks like what I've done is probably going to be controversial. I mean it is naive in that the visit count of nodes along a cache hit chain will all be the same and their parents' visit counts will only think they had 1 visit even though many nodes were (re)discovered. I think the effect of this will be that transpositions will be slightly less likely to be played as they will have to make up for the lack of counted visits (though this will only be the length of the remaining cache hit chain).

So yeah, I'd be interested in hearing some discussion and won't be offended if this isn't merged. I had fun trying out a little idea :)

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -30,6 +30,7 @@
 #include <memory>
 #include <random>
 #include <string>
+#include <tuple>
 #include <vector>
 
 #include "FastBoard.h"
@@ -577,21 +578,22 @@ bool GTP::execute(GameState & game, std::string xinput) {
         cmdstream >> symmetry;
 
         Network::Netresult vec;
+        bool cache_hit;
         if (cmdstream.fail()) {
             // Default = DIRECT with no rotation
-            vec = Network::get_scored_moves(
+            std::tie(vec, cache_hit) = Network::get_scored_moves(
                 &game, Network::Ensemble::DIRECT, 0, true);
         } else if (symmetry == "all") {
-            for (auto r = 0; r < 8; r++) {
-                vec = Network::get_scored_moves(
-                    &game, Network::Ensemble::DIRECT, r, true);
+            for (auto sym = 0; sym < 8; sym++) {
+                std::tie(vec, cache_hit) = Network::get_scored_moves(
+                    &game, Network::Ensemble::DIRECT, sym, true);
                 Network::show_heatmap(&game, vec, false);
             }
         } else if (symmetry == "average" || symmetry == "avg") {
-            vec = Network::get_scored_moves(
+            std::tie(vec, cache_hit) = Network::get_scored_moves(
                 &game, Network::Ensemble::AVERAGE, 8, true);
         } else {
-            vec = Network::get_scored_moves(
+            std::tie(vec, cache_hit) = Network::get_scored_moves(
                 &game, Network::Ensemble::DIRECT, std::stoi(symmetry), true);
         }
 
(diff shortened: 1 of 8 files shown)


gcp:
Time to start testing a transposition table again. What you did here is a more limited/incorrect version of that, right?

gcp:
>Also will counting visits correctly mean that transpositions are more likely to get more visits and thus more likely to be played?

Yes, at least at the root.

>I mean it recovers the visits and winrate but why not (some of) the child tree (could be done with a method like this PR)?

This was proposed but simply never implemented IIRC.

Hersmunch:
I'll close this for now but I might have a go at putting this (fixed somehow) together with adding TTable back in, so the tree is "recovered" from the NNCache. I imagine testing will take weeks though..

Issue #1416, opened 2018-05-13
Optimal amount of visits per move


Ttl:
I made some analysis earlier that maybe lowering visits would be useful but it wasn't very rigorous (https://github.com/gcp/leela-zero/issues/1030#issuecomment-374328038, https://github.com/gcp/leela-zero/issues/1030#issuecomment-374246946). This is an attempt to make the earlier analysis more rigorous by defining a metric on how much information each visit gives.

A reasonable choice for the optimum number of visits in self play games would be the number of visits that gives the maximum amount of new information for the least amount of computation. Said otherwise we want to chose the number of visits used in each position such that with fixed visits/s we maximize new information/s. Training of the policy head is done using cross entropy and so it's very natural to use cross entropy as a measure of the gain in information. I use KL divergence in the analysis. It's basically same thing as cross entropy but gives 0 when the distributions are identical (cross entropy is KL divergence + constant). To calculate the KL divergence we need to know the true visit distribution for each move to which we compare to. The distribution of visits at very high visits (10000 in this case) is used as an approximation of the true distribution.

I made some changes to LZ that adds normalized visits count to `dump_stats` and prints the stats every 100 visits. I played 5 self-pay games using command: ```./leelaz -n -m 999 -d -t 1 --noponder -v 10000 -r 5 -w ecab83bb611574a6a3139ae0f9734c172167997a843fbb034dac88246cfd7199 2> playout_analysis_stderr.txt < po_analysis_cmd.txt```. In 5 games there are total of 813 moves so the dataset is not that small when we are analyzing moves and not games.

po_analysis_cmd.txt file has contents of:

```
time_settings 0 1 0
auto
```

Calculating KL divergence between the visit distribution at each visit count to the approximated true distribution (distribution at 10000 playouts) and averaging every move in the game gives following policy loss vs visits plot:

![policy_loss](https://user-images.githubusercontent.com/544240/39964851-c31d9872-568d-11e8-9f1a-eb6384d68037.png)

[...]

PR #1497, opened 2018-05-31
Non-pruning timemanagement for self-play

Ttl:
Adds support for timemanagement code in leelaz and autogtp that doesn't prune moves and only cuts off the search if the second best move can't overtake the best one with the remaining visits.

The visit distribution with timemanagement is very close to without it and it generates very similar training data. On average less visits are needed per move speeding up the game generation.

More discussion in #1416

### DIFF
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -72,8 +72,12 @@ static void parse_commandline(int argc, char *argv[]) {
         ("logfile,l", po::value<std::string>(), "File to log input/output to.")
         ("quiet,q", "Disable all diagnostic output.")
         ("timemanage", po::value<std::string>()->default_value("auto"),
-                       "[auto|on|off|fast] Enable time management features.\n"
-                       "auto = off when using -m, otherwise on")
+                       "[auto|on|off|fast|no_pruning] Enable time management features.\n"
+                       "auto = no_pruning when using -n, otherwise on.\n"
+                       "on = Cut off search when the best move can't change"
+                       ", but use full time if moving faster doesn't save time.\n"
+                       "fast = Same as on but always plays faster.\n"
+                       "no_pruning = For self play training use.\n")
         ("noponder", "Disable thinking on opponent's time.")
         ("benchmark", "Test network and exit. Default args:\n-v3200 --noponder "
                       "-m0 -t1 -s1.")
@@ -277,6 +281,8 @@ static void parse_commandline(int argc, char *argv[]) {
             cfg_timemanage = TimeManagement::OFF;
         } else if (tm == "fast") {
             cfg_timemanage = TimeManagement::FAST;
+        } else if (tm == "no_pruning") {
+            cfg_timemanage = TimeManagement::NO_PRUNING;
         } else {
             printf("Invalid timemanage value.\n");
             exit(EXIT_FAILURE);
@@ -284,7 +290,7 @@ static void parse_commandline(int argc, char *argv[]) {
     }
     if (cfg_timemanage == TimeManagement::AUTO) {
         cfg_timemanage =
-            cfg_random_cnt ? TimeManagement::OFF : TimeManagement::ON;
+            cfg_noise ? TimeManagement::NO_PRUNING : TimeManagement::ON;
     }
 
     if (vm.count("lagbuffer")) {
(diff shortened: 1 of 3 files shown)


Eddh:
For what it's worth I don't like these kind of modifications because it opens up new possibilities for the net to make its accuracy higher without actually learning anything useful. In this case the net could "cheat" by learning to recognize situations in which the search will be cut short in order to get a better policy loss. It would not help it in real games, and yet filters, layers might be wasted on that.

Ttl:
The visit distribution is very similar to one played at fixed visits. I don't see how knowing that the position got less visits than maximum would give advantage in prediction the policy distribution. Can you give an example?

gcp:
>In this case the net could "cheat" by learning to recognize situations in which the search will be cut short in order to get a better policy loss.

How would this happen?

src/Leela.cpp
```diff
@@ -277,6 +277,8 @@ static void parse_commandline(int argc, char *argv[]) {
             cfg_timemanage = TimeManagement::OFF;
         } else if (tm == "fast") {
             cfg_timemanage = TimeManagement::FAST;
+        } else if (tm == "simple") {
+            cfg_timemanage = TimeManagement::SIMPLE;
```
gcp: We already make quite a few things conditional on noise/random selection so this could as well.

I understand that above you've added a possibility for the server to add arbitrary options to enable this? That has the disadvantage that it requires all clients to be switched. (Unless AutoGTP behaves correctly with unknown options).

Ttl: I thought that server can give different commands to different versions. Would you rather have it default to simple?

I think that autogtp ignores unknown options. It parses json to ```map<QString, QString>``` and only checks keys for commands it expects.

gcp: >I thought that server can give different commands to different versions. 

Ha, good point.

>Would you rather have it default to simple?

We could do that if we detect noise is enabled. BTW. What about DONT_PRUNE or NO_PRUNING instead of SIMPLE?

Maybe FAST should have been named VERY_EARLY_EXIT or something. The naming here kinda sucks right now.

Eddh:
OK maybe I misunderstood something I will read in more detail.

autogtp/Management.cpp
```diff
@@ -269,6 +269,8 @@ QString Management::getOptionsString(const QJsonObject &opt, const QString &rnd)
     if (rnd != "") {
         options.append(" -s " + rnd + " ");
     }
+    // Additional options specified by server
+    options.append(getOption(opt, "other_options", "", ""));
```
marcocalignano: What is this for?

Ttl: Can be used to give arbitrary command line options to leelaz. For example timemanage, randomvisits, randomtemp. However I just removed it from this pull request since now non_pruning timemanagement is enabled by default.

Ttl:
Now enabled by default when noise is enabled. There is legitimate uses for using "-m" option in match games to get more random opening so I changed auto to look for noise option instead. Renamed simple to non_pruning. I didn't change FAST because the command line parameter for it is fast.

marcocalignano:
@Ttl  If you change things with the option you have to be aware of the save restore procedure, you have to store all the options  that you gave Leelaz that is way I would like not to have a general option flag.

Ishinoshita:
Hi @Ttl. Since I did some analysis of LZ177 training data that turned out to revolve around time_mgt_pruning without knowing it, let me share it:
[LZ177 training data - Effect of time management pruning on policy.pptx](https://github.com/gcp/leela-zero/files/2485012/LZ177.training.data.-.Effect.of.time.management.pruning.on.policy.pptx)

I can definitively see statistically significant differences in the distribution of policy targets, with  timemanagment pruning On or Off, but I cannot tell whether this is good or bad for training and if there is an issue with that option for self-play. Thus, I share the comparison, without any comment.

Just to illustrate directly in this thread:

image

image

Let me clarify that I did this comparison after removing *all games* affected by any of the new weirdnesses I found and I gonna mention in a new post on LZ177 traing data.

For the methodology, see [my git](https://github.com/Ishinoshita/Leela-Zero-Training-Data-Viewer).

PR #1522, opened 2018-06-05
Symmetry cleanup.

TFiFiE:
This salvages #1275 mainly to recover the `NUM_SYMMETRIES` constant and get rid of the duplicate symmetry and hash code.

### DIFF
--- a/src/FullBoard.cpp
+++ b/src/FullBoard.cpp
@@ -22,6 +22,7 @@
 #include <cassert>
 
 #include "FullBoard.h"
+#include "Network.h"
 #include "Utils.h"
 #include "Zobrist.h"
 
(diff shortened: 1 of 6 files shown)


src/FullBoard.cpp
```diff
@@ -68,12 +69,13 @@ std::uint64_t FullBoard::calc_ko_hash(void) {
     return res;
 }
 
-std::uint64_t FullBoard::calc_hash(int komove) {
+template<class Function>
```
Hersmunch: `Function` is used here but `Transform` in the declaration.

src/Network.h
```diff
@@ -73,6 +74,7 @@ class Network {
 
     static std::vector<net_t> gather_features(const GameState* const state,
                                               const int symmetry);
+    static std::pair<int, int> get_symmetry(const std::pair<int, int>& vertex, const int symmetry, const int board_size = BOARD_SIZE);
```
Hersmunch: I think line length has been kept to under 80 so far. There's a few other new places this has been exceeded.

gcp:
Very nice!

PR #1538, opened 2018-06-09
Script for converting minigo weights

Ttl:
Converts minigo weights to work with LZ. Some minigo weights are available here: https://console.cloud.google.com/storage/browser/minigo-pub

Some tests:

LZ weights 147 vs. LZ with minigo weights 303. 1000 visits, minigo weights has 3 handicap stones:

```
lz_minigo_303 v lz_147 (15/100 games)
board size: 19   handicap: 3 (free)   komi: 7.5
                wins                  avg cpu
lz_minigo_303      4 26.67%   (black)  945.66
lz_147            11 73.33%   (white)  615.45
```

Minigo vs LZ using the same 303-olympus weights with 100 playouts. LZ arguments: `-g -d -t 1 --noponder -p 100 --timemanage off -r 5`. Minigo used the default settings.
```
lz_303 v minigo_303 (100/100 games)
board size: 19   komi: 7.5
lz_303         86 86.00%       42 84.00%     44 88.00%    107.65
minigo_303     14 14.00%       6  12.00%     8  16.00%   4449.02
                               48 48.00%     52 52.00%
```

### DIFF
--- a/training/minigo/convert_minigo.py
+++ b/training/minigo/convert_minigo.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python3
+import tensorflow as tf
+import numpy as np
+import sys
+
+if len(sys.argv) < 2:
+    print('Model filename without extension needed as an argument.')
+    exit()
+
+sess = tf.Session()
+saver = tf.train.import_meta_graph(sys.argv[1]+'.meta')
+saver.restore(sess, sys.argv[1])
+
+if 0:
+    # Exports graph to tensorboard
+    with tf.Session() as sess:
+        writer = tf.summary.FileWriter('logs', sess.graph)
+        writer.close()
+
+trainable_names = []
+for v in tf.trainable_variables():
+    trainable_names.append(v.name)
+
+weights = []
+for v in tf.global_variables():
+    if v.name in trainable_names:
+        weights.append(v)
+    elif 'batch_normalization' in v.name:
+        # Moving mean and variance are not trainable, but are needed for the model
+        if 'moving_mean' in v.name or 'moving_variance' in v.name:
+            weights.append(v)
+
+if 0:
+    for w in weights:
+        print(w.name)
+
+def merge_gammas(weights):
+    out_weights = []
+    skip = 0
+    for e, w in enumerate(weights):
+        if skip > 0:
+            skip -= 1
+            continue
+        if 'kernel' in w.name and 'conv2d' in w.name and 'gamma' in weights[e+2].name:
+            kernel = w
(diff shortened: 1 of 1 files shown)


amj:
was this vs minigo's python code or the c++ code?

It obviously seems a little off that the same weights should be so bad in a different engine and it makes me wonder if we have a bug.  Can you show me the command used to invoke minigo?

brilee:
@amj was saying that LZ might be working with parallel reads = 1 whereas we're using parallel reads=8 by default. If that's correct, would you be open to trying the matchup again with `--parallel_readouts=1`?

Ttl:
I just wanted to verify that the weights were converted correctly. The match was run on a laptop and even with just 100 playouts it took over a day to play 100 games. I have other uses for it right now and don't want to repeat the match.

amj:
Ok, i can definitely understand that -- if you can compile our cc version it should be much faster :)

brilee:
Okay - thank you very much for reporting the results here.

Ttl:
The laptop this was played on had only Intel integrated GPU. As far as I know tensorflow can't use it, but LZs OpenCL code supports it. That's why there is such a large difference in the CPU use.

Ttl:
Shouldn't the equivalent Puct be twice as large in minigo because minigo uses winrate on scale [-1, 1] while LZ uses [0, 1]?

alreadydone:
If you compile leelaz with USE_TUNER defined
https://github.com/gcp/leela-zero/blob/d362ee86cf7ead85831c8a947c7fed580f63c8eb/src/config.h#L81
you should be able to specify options --puct ([0.69](https://www.reddit.com/r/cbaduk/comments/8j5x3w/first_play_urgency_fpu_parameter_in_alpha_zero/dz1ipi7/) for Minigo) and --fpu_reduction (0 for Minigo). Also LZ uses net_eval for FPU (before reduction). To use get_pure_eval (dynamic parent Q) instead, use my branch [patch-4](https://github.com/alreadydone/lz/tree/patch-4/), but revert the changes to fpu_reduction (L275) and puct (L294) in the [first commit](https://github.com/gcp/leela-zero/pull/1435/commits/4d3f6444cffb8354501f2806e98caf430cd73f9f).

amj:
@Ttl yes i would expect the equivalent puct to need to be twice as large as well -- very strange that it worked that way.

sethtroisi:
With @alreadydone's help I enabled `USE_TUNER` and set `--puct 0.69 --fpu_reduction 0`
I see very similar move selection now.

I'm going to disable rotation and see if I get identical results.

PR #1558, opened 2018-06-15
Code restructuring part 1 : no globals

ihavnoid:
Restructured code to remove:
- No global / thread_local OpenCL structures
- Network is now all non-static methods with instance variables
- NNCache is a member of Network, OpenCLScheduler is also a member of Network
- ThreadData is renamed to OpenCLContext (it's no longer thread-specific!)
- Other accompanying fixes

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -76,10 +76,11 @@ std::string cfg_options_str;
 bool cfg_benchmark;
 int cfg_analyze_interval_centis;
 
-std::unique_ptr<UCTSearch> GTP::search;
+Network GTP::network;
 
-void GTP::initialize(GameState & gamestate) {
-    search = std::make_unique<UCTSearch>(gamestate, cfg_weightsfile);
+void GTP::initialize() {
+    auto playouts = std::min(cfg_max_playouts, cfg_max_visits);
+    network.initialize(playouts, cfg_weightsfile);
 }
 
 void GTP::setup_default_parameters() {
@@ -192,6 +193,7 @@ std::string GTP::get_life_list(const GameState & game, bool live) {
 
 bool GTP::execute(GameState & game, std::string xinput) {
     std::string input;
+    static auto search = std::make_unique<UCTSearch>(game, network);
 
     bool transform_lowercase = true;
 
@@ -305,7 +307,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
     } else if (command.find("clear_board") == 0) {
         Training::clear_training();
         game.reset_game();
-        search->reset();
+        search = std::make_unique<UCTSearch>(game, network);
         gtp_printf(id, "");
         return true;
     } else if (command.find("komi") == 0) {
@@ -584,19 +586,19 @@ bool GTP::execute(GameState & game, std::string xinput) {
         Network::Netresult vec;
         if (cmdstream.fail()) {
             // Default = DIRECT with no symmetric change
-            vec = search->get_scored_moves(
+            vec = network.get_scored_moves(
                 &game, Network::Ensemble::DIRECT, Network::IDENTITY_SYMMETRY, true);
         } else if (symmetry == "all") {
             for (auto s = 0; s < Network::NUM_SYMMETRIES; ++s) {
-                vec = search->get_scored_moves(
+                vec = network.get_scored_moves(
                     &game, Network::Ensemble::DIRECT, s, true);
                 Network::show_heatmap(&game, vec, false);
             }
         } else if (symmetry == "average" || symmetry == "avg") {
-            vec = search->get_scored_moves(
+            vec = network.get_scored_moves(
                 &game, Network::Ensemble::AVERAGE, Network::NUM_SYMMETRIES, true);
         } else {
-            vec = search->get_scored_moves(
+            vec = network.get_scored_moves(
                 &game, Network::Ensemble::DIRECT, std::stoi(symmetry), true);
         }
 
@@ -630,7 +632,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
         cmdstream >> stones;
 
         if (!cmdstream.fail()) {
-            game.place_free_handicap(stones, search.get());
+            game.place_free_handicap(stones, network);
             auto stonestring = game.board.get_stone_list();
             gtp_printf(id, "%s", stonestring.c_str());
         } else {
@@ -752,9 +754,9 @@ bool GTP::execute(GameState & game, std::string xinput) {
         cmdstream >> iterations;
 
         if (!cmdstream.fail()) {
-            search->benchmark(&game, iterations);
+            network.benchmark(&game, iterations);
         } else {
-            search->benchmark(&game);
+            network.benchmark(&game);
         }
         gtp_printf(id, "");
         return true;
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -186,7 +186,9 @@ std::string GTP::get_life_list(const GameState & game, bool live) {
 
 bool GTP::execute(GameState & game, std::string xinput) {
     std::string input;
-    static auto search = std::make_unique<UCTSearch>(game);
+
+    // When visits are limited ensure cache size is still limited.
+    static auto search = std::make_unique<UCTSearch>(game, cfg_weightsfile);
 
     bool transform_lowercase = true;
 
@@ -300,7 +302,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
     } else if (command.find("clear_board") == 0) {
         Training::clear_training();
         game.reset_game();
-        search = std::make_unique<UCTSearch>(game);
+        search->reset();
         gtp_printf(id, "");
         return true;
     } else if (command.find("komi") == 0) {
@@ -579,19 +581,19 @@ bool GTP::execute(GameState & game, std::string xinput) {
         Network::Netresult vec;
         if (cmdstream.fail()) {
             // Default = DIRECT with no symmetric change
-            vec = g_network.get_scored_moves(
+            vec = search->get_scored_moves(
                 &game, Network::Ensemble::DIRECT, 0, true);
         } else if (symmetry == "all") {
             for (auto s = 0; s < Network::NUM_SYMMETRIES; ++s) {
-                vec = g_network.get_scored_moves(
+                vec = search->get_scored_moves(
                     &game, Network::Ensemble::DIRECT, s, true);
                 Network::show_heatmap(&game, vec, false);
             }
         } else if (symmetry == "average" || symmetry == "avg") {
-            vec = g_network.get_scored_moves(
+            vec = search->get_scored_moves(
                 &game, Network::Ensemble::AVERAGE, Network::NUM_SYMMETRIES, true);
         } else {
-            vec = g_network.get_scored_moves(
+            vec = search->get_scored_moves(
                 &game, Network::Ensemble::DIRECT, std::stoi(symmetry), true);
         }
 
@@ -625,7 +627,7 @@ bool GTP::execute(GameState & game, std::string xinput) {
         cmdstream >> stones;
 
         if (!cmdstream.fail()) {
-            game.place_free_handicap(stones);
+            game.place_free_handicap(stones, search.get());
             auto stonestring = game.board.get_stone_list();
             gtp_printf(id, "%s", stonestring.c_str());
         } else {
@@ -747,9 +749,9 @@ bool GTP::execute(GameState & game, std::string xinput) {
         cmdstream >> iterations;
 
         if (!cmdstream.fail()) {
-            g_network.benchmark(&game, iterations);
+            search->benchmark(&game, iterations);
         } else {
-            g_network.benchmark(&game);
+            search->benchmark(&game);
         }
         gtp_printf(id, "");
         return true;
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -76,11 +76,10 @@ std::string cfg_options_str;
 bool cfg_benchmark;
 int cfg_analyze_interval_centis;
 
-Network GTP::network;
+Network * GTP::network = nullptr;
 
-void GTP::initialize() {
-    auto playouts = std::min(cfg_max_playouts, cfg_max_visits);
-    network.initialize(playouts, cfg_weightsfile);
+void GTP::initialize(Network * net) {
+    network = net;
 }
 
 void GTP::setup_default_parameters() {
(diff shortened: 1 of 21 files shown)


ihavnoid:
I can't figure out why the test is failing - any help?

gcp:
Thanks! Give me a few moments to review this.

src/OpenCLScheduler.h
```diff
@@ -36,22 +37,13 @@ class OpenCLScheduler {
...
-                    std::vector<net_t> * out)
-            : input(in), output(out) {}
-    };
-
     std::vector<std::unique_ptr<OpenCL_Network>> m_networks;
     std::vector<std::unique_ptr<OpenCL>> m_opencl;
-    Utils::ThreadPool m_threadpool;
+    // XXX : I failed to figure out how to make this compile with unique_ptr
+    // especially on visual studio
+    std::vector<std::list<std::shared_ptr<OpenCLContext>>> m_context;
```
gcp: I might make sense to make a helper type for the innermost list. Would also make it clearer why 2 types of collections are being nested here.

src/UCTSearch.cpp
```diff
@@ -71,11 +71,23 @@ class OutputAnalysisData {
 };
 
 
-UCTSearch::UCTSearch(GameState& g)
+UCTSearch::UCTSearch(GameState& g, const std::string & weightsfile)
```
gcp: Meh. I don't think UCTSearch should know about a string indicating a file for the network. Wouldn't it make more sense if it took a Network& instead? GTP:: should then inject that ref into search.

This avoids some ugliness like GameState::place_free_handicap taking an UCTSearch*, even though it doesn't want an existing UCTSearch at all, it just really needs the info to set up the network for it.

You can probably also avoid that "global" search object you have in GTP now. The network that is specified on the commandline can be loaded an injected from the main function and cmdline parsing then.

src/GTP.cpp
```diff
@@ -76,6 +76,12 @@ std::string cfg_options_str;
 bool cfg_benchmark;
 int cfg_analyze_interval_centis;
 
+std::unique_ptr<UCTSearch> GTP::search;
```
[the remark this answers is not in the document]
ihavnoid: You probably figured out, but this was due to an attempt trying to remove the global Network and keep it as a member of UCTSearch.  Without this, OpenCL gets initialized on first GTP command hence introducing some wierdness.  (Comment will continue below)

src/UCTSearch.cpp
ihavnoid: Yeah I agree.  Trying to remove the global Network reference is causing more trouble than necessary, making other places looking odd for no apparent benefit.  I'll move the network variable to GTP and just pass a reference.

src/OpenCLScheduler.cpp
```diff
@@ -22,61 +22,72 @@
...
+                break;
+            }
+            queue_num++;
+        }
     }
 
-    auto f = m_threadpool.add_task([this, &input, &output_pol, &output_val]{
-        m_networks[current_thread_gpu_num]->forward(input, output_pol, output_val);
-    });
+    m_networks[ctx->m_gpu_num]->forward(input, output_pol, output_val, *ctx);
```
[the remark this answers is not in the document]
ihavnoid: Yes it is guaranteed, otherwise the condition var should not exit.  Will add assertion here.

ihavnoid:
Addressed 1st round of comments.
- Network is no longer child of UCTSearch, it's a reference to GTP::network
- Added assert on ctx
- Changed 'new OpenCLContext' to std::make_shared
- Added subtype for vector-of-list-of-pointer-of-OpenCLContext for readibility

src/OpenCLScheduler.h
```diff
@@ -36,22 +37,16 @@ class OpenCLScheduler {
...
-            : input(in), output(out) {}
-    };
-
     std::vector<std::unique_ptr<OpenCL_Network>> m_networks;
     std::vector<std::unique_ptr<OpenCL>> m_opencl;
-    Utils::ThreadPool m_threadpool;
-};
 
-extern OpenCLScheduler opencl;
+    // XXX : I failed to figure out how to make this compile with unique_ptr
```
gcp: Comment can go, can always try to fix it later.

src/OpenCLScheduler.cpp
```diff
@@ -22,61 +22,75 @@
...
-            // work for task preparation.
-            constexpr auto num_threads = 2;
-            for (auto i = 0; i < num_threads; i++) {
-                m_threadpool.add_thread([gnum] {
-                    current_thread_gpu_num = gnum;
-                });
-            }
-        }
-    } else {
+    for (auto gpu : gpus) {
```
gcp: const auto

src/OpenCLScheduler.cpp
```diff
@@ -22,61 +22,75 @@
...
                               std::vector<net_t>& output_val) {
-    if (m_networks.size() == 1) {
-        m_networks[0]->forward(input, output_pol, output_val);
-        return;
+    std::shared_ptr<OpenCLContext> ctx;
+    auto queue_num = size_t{0};
+    {
+        std::unique_lock<std::mutex> lk(m_context_lock);
+        m_context_condvar.wait(lk, [this]{
+            for (auto & ctxlist : m_context) {
```
gcp: const auto&

src/OpenCLScheduler.cpp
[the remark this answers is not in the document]
gcp: Ok I guess it's ensured by the predicate inside the condvar wait.

src/OpenCLScheduler.cpp
```diff
@@ -22,61 +22,75 @@
...
+                if (!ctxlist.empty()) {
+                    return true;
+                }
+            }
+            return false;
+        });
+        while (queue_num < m_context.size()) {
```
gcp: Can't this be a for loop indexing over m_context? Ok, I see that you need queue_num later.

src/OpenCLScheduler.h
```diff
@@ -36,22 +37,16 @@ class OpenCLScheduler {
...
+    // especially on visual studio
+    using OpenCLContextQueue = std::list<std::shared_ptr<OpenCLContext>>;
+    std::vector<OpenCLContextQueue> m_context;
```
gcp: Really a list of contexts per worker thread. m_work_threads? m_thread_contexts? Something else? Not sure.

src/OpenCLScheduler.cpp
```diff
@@ -22,61 +22,75 @@
...
-        }
-    } else {
+    for (auto gpu : gpus) {
         auto opencl = std::make_unique<OpenCL>();
         auto net = std::make_unique<OpenCL_Network>(*opencl);
-        opencl->initialize(channels, {});
-
+        opencl->initialize(channels, {gpu}, silent);
         m_opencl.push_back(std::move(opencl));
         m_networks.push_back(std::move(net));
```
gcp: Does this work correctly with --gpu 1 --gpu 3?

m_networks will be a size 2 array, yet ctx->m_gpu_num which is indexing into that can be 1 and 3.

src/Network.h
```diff
@@ -79,43 +72,78 @@ class Network {
...
-    static void forward_cpu(const std::vector<float>& input,
+    void forward_cpu(const std::vector<float>& input,
                             std::vector<float>& output_pol,
                             std::vector<float>& output_val);
 
 #endif
-};
 
+#ifdef USE_OPENCL
+    OpenCLScheduler m_opencl;
```
gcp: It feels a bit weird to have this owned by Network. I guess it's really here because it contains pointers to contexts which themselves identify where on the GPUs the network data is loaded.

src/OpenCL.cpp
```diff
@@ -412,26 +412,24 @@ const std::string sourceCode_sgemm =
 ;
 #endif
 
-thread_local ThreadData opencl_thread_data;
-
-void OpenCL::ensure_thread_initialized() {
-    if (!opencl_thread_data.m_is_initialized) {
+void OpenCL::ensure_thread_initialized(OpenCLContext &opencl_context) {
```
gcp: If this takes a context argument it is no longer thread specific, so the function name is wrong.

src/OpenCL.cpp
```diff
@@ -1093,11 +1097,14 @@ void OpenCL::initialize(const int channels, const std::vector<int> & gpus,
         throw std::runtime_error("Error building OpenCL kernels.");
     }
 
-    ensure_thread_initialized();
+    // dummy gpu number, putting bogus number so that we don't misuse it
+    OpenCLContext tdata(0xffff);
```
gcp: I don't think you can misuse anything here given that OpenCLContext is no longer a thread local. Just make it 0 or whatever is the first in cfg_gpu. There's also nothing thread related any more here.

ihavnoid:
No objection on any of your 2nd batch of comments.  Give me a couple of days to clean them up.

src/GameState.cpp
```diff
@@ -32,6 +32,7 @@
 #include "FullBoard.h"
 #include "KoState.h"
 #include "UCTSearch.h"
+#include "GTP.h"
```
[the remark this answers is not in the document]
ihavnoid: Correct.  Will be gone.

src/Network.h
```diff
@@ -28,8 +28,13 @@
...
 #include <fstream>
 
+#include "NNCache.h"
 #include "FastState.h"
-#include "GameState.h"
+#ifdef USE_OPENCL
+#include "OpenCLScheduler.h"
+#endif
+
+class GameState;
```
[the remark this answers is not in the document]
ihavnoid: ...and due to this, we need to remove the \#include "Network.h" from GameState.h and replace it with a forward declaration.  Well, the change you suggested seems to be more logical..

src/OpenCLScheduler.h
ihavnoid: Well, it's the 'pool' for OpenCL contexts.  Hence I just put it m_context_pool.  Not perfect but I don't think we should call this 'threads' anymore to avoid confusion (although yes it still has a lot of sense of the concept 'thread' here)

src/OpenCLScheduler.cpp
ihavnoid: Ouch x100.  Will rewrite code.

src/OpenCL.cpp
ihavnoid: I will still leave the comment that the thread number here is bogus, just to prevent anybody from being confused...

src/OpenCL.cpp
ihavnoid: renaming to ensure_thread_initialized

src/Network.h
ihavnoid: I agree with you, but I couldn't find a better place for this.  I have another batch of fixes (abstract interface for network evaluation backend) and this is replaced with a pointer to the abstract type.  This piece needs a bit more clever thought but no good idea yet...

src/OpenCLScheduler.cpp
ihavnoid: While trying to rewrite code, figured out that this isn't a bug - it is correct because m_gpu_num will be 0 or 1 (that is, this isn't the GPU number given by the command line argument.  This is the GPU index number on OpenCLScheduler.  Looking back, I don't like defining variables that are only used on OpenCLScheduler on a OpenCL data structure.  Will do a small rewrite to make it prettier...

ihavnoid:
Pushed 2nd batch of fixes.  Please have a look.

TFiFiE:
The following member functions of `Network` can stay **static**:
- `process_bn_var`
- `winograd_transform_f`
- `zeropad_U`
- `winograd_transform_in`
- `winograd_transform_out`
- `winograd_convolve3`
- `winograd_sgemm`

Also, `forward_cpu` can be declared **const** and the indentations of the function arguments in `Network.h` are in need of fixing.

ihavnoid:
Changed the functions' prototype to static, and fixed indentation.

src/Network.h
```diff
@@ -39,21 +43,9 @@ class Network {
...
-        float winrate;
-
-        Netresult() : policy(BOARD_SQUARES), policy_pass(0.0f), winrate(0.0f) {}
-    };
-
-    static Netresult get_scored_moves(const GameState* const state,
+    Netresult get_scored_moves(const GameState* const state,
                                       const Ensemble ensemble,
                                       const int symmetry = -1,
                                       const bool skip_cache = false);
```
TFiFiE: Indentation.

src/Network.h
```diff
@@ -67,8 +59,8 @@ class Network {
     static constexpr auto WINOGRAD_ALPHA = 4;
     static constexpr auto WINOGRAD_TILE = WINOGRAD_ALPHA * WINOGRAD_ALPHA;
 
-    static void initialize();
-    static void benchmark(const GameState * const state,
+    void initialize(int playouts, const std::string & weightsfile);
+    void benchmark(const GameState * const state,
                           const int iterations = 1600);
```
TFiFiE: Indentation.

src/Network.h
```diff
@@ -104,18 +96,53 @@ class Network {
     static void winograd_sgemm(const std::vector<float>& U,
                                const std::vector<float>& V,
                                std::vector<float>& M, const int C, const int K);
+    Netresult get_scored_moves_internal(const GameState* const state,
+                                               const int symmetry);
```
TFiFiE: Indentation.

src/Network.h
```diff
@@ -104,18 +96,53 @@ class Network {
...
                                       const int symmetry);
-    static Netresult get_scored_moves_internal(const GameState* const state,
-                                               const int symmetry);
+
+    bool probe_cache(const GameState* const state, Network::Netresult& result);
 #if defined(USE_BLAS)
-    static void forward_cpu(const std::vector<float>& input,
+    void forward_cpu(const std::vector<float>& input,
                             std::vector<float>& output_pol,
                             std::vector<float>& output_val);
```
TFiFiE: Indentation and, again, you can make this member function **const**.

TFiFiE:
Would it perhaps be better for `Network::initialize` to be the constructor instead? Any instance of `Network` that can't be immediately initialized could be replaced by a `unique_ptr` to it.

ihavnoid:
@TFiFiE I agree with you but my opinion is that this PR is getting too huge, and hence let's go with another code restructuring PR to fix that one.  The problem (if we call it a 'problem') is actually everywhere so it better be addressed all at once.  There are constructor implications too which may add some complexity here and there.

ihavnoid:
@gcp  - any next steps?

gcp:
There were some small cleanups I wanted to try (use unique_ptr's instead of the raw ones for the global objects, and see if some of the globals can be made function scoped) but might as well merge this in already and put the cleanups on top of it.

PR #1567, opened 2018-06-18
Initialize node eval immediately after NN eval

alreadydone:
This allows get_pure_eval (for the parent) be used in e.g. uct_select_child without causing crashes with multiple threads. Previously, when another thread visited the node, it could already have children but have no visits, so get_pure_eval wouldn't work. Discussions starting at https://github.com/gcp/leela-zero/pull/1435#issuecomment-391485532.

(Please ignore the commit descriptions and only look at the diffs.)


gcp:
This makes the code more complicated and error prone without gain.

Whatever is calling get_pure_eval should guard against m_visits == 0. Note that get_pure_eval actually contains debug assertions to make it very clear that is a precondition.

PR #1572, opened 2018-06-19
Tidy dependencies.

TFiFiE:
Can someone test this with Visual C++, because it's possible I may have gone overboard with the use of forward declarations?


gcp:
Our CI includes compiling with MSVC so given that it passes you should be good there.

gcp:
1) I prefer if config.h is always included and included first. It's useful for working around portability issues. compiler and platform differences etc. But to be useful it has to be "active" everywhere.

2) Forward declaring classes just to avoid including their defining header seems pretty dirty to me. I checked around a bit and I'm not the only one who believes so: https://google.github.io/styleguide/cppguide.html#Forward_Declarations

autogtp/Job.cpp
```diff
@@ -17,13 +17,12 @@
...
 #include "Job.h"
-#include "Game.h"
-#include "Management.h"
-#include <QTextStream>
-#include <chrono>
-
 #include <QFile>
+#include <QTextStream>
 #include <QThread>
+#include "Game.h"
```
marcocalignano: the includes of your files should go before the standard includes.

autogtp/Management.cpp
```diff
@@ -16,20 +16,19 @@
...
+#include <QProcess>
 #include <QRegularExpression>
-#include "Management.h"
-#include "Game.h"
-
+#include <QTextStream>
+#include <QThread>
+#include <QUuid>
+#include <cmath>
+#include "Worker.h"
```
marcocalignano: the includes of your files should go before the standard includes.

autogtp/Management.h
```diff
@@ -19,16 +19,19 @@
...
-#include <QThread>
-#include <QFileInfo>
-#include <QLockFile>
+#include <QStringList>
 #include <QVector>
 #include <chrono>
 #include <stdexcept>
-#include "Worker.h"
+#include "Order.h"
+#include "Result.h"
```
marcocalignano: the includes of your files should go before the standard includes.

marcocalignano: I do not like to include a `.h` that I do not need in a `.h` only because all the `.cpp` using the second `.h`  use also the first. I like the include to be explicit. This invalidate a lot of other changes in the `autogtp` code.

autogtp/Worker.h
```diff
@@ -19,11 +19,12 @@
...
 
+#include <QAtomicInt>
+#include <QString>
+#include <QThread>
 #include "Job.h"
 #include "Order.h"
-
-#include <QThread>
-#include <QMutex>
+#include "Result.h"
```
marcocalignano: I do not like to include a .h that I do not need in a .h only because all the .cpp using the second .h use also the first. I like the include to be explicit. This invalidate a lot of other changes in the autogtp code.

autogtp/Job.cpp
TFiFiE: That would be inconsistent with how it's done in `src`.

autogtp/Management.h
TFiFiE: For some reason, Qt's meta-object compiler complains if `Result` is forward declared instead.

autogtp/Worker.h
TFiFiE: Same as above.

TFiFiE:
I have (re)added the inclusion of `config.h` everywhere in `src`.

Building Leela takes a non-trivial amount of time, so if the use of forward declarations in lieu of `#include` statements results in less source files being re-compiled whenever something is changed, I don't see why those working with the code should be denied this benefit.

autogtp/Job.cpp
TFiFiE: That's because it's the complementary header file.

TFiFiE: `Game.h` gets included first because it's the header file belonging to `Game.cpp`.

gcp:
Please read the "cons" section in the link above. You may care about compile time, I care about maintenance issues. Forward class declarations won't be merged.

Issue #1535, opened 2018-06-08
AutoGTP should save its work in case a rebot is required mid-game


Nazgand:
I have killed AutoGTP at times, and have not noticed it starting from a mid-game position when running AutoGTP again.

Friday9i:
To close it and saving games, type "q + Enter", it saves file(s) and makes a clean exit, and reload everything when you launch it again : -)

gcp:
This would require dumping out the info every move, rather than every game. Not sure we want to do this.

Nazgand:
@Friday9i Thanks. Where did you learn that? I did not notice it in the README or the FAQ. Perhaps the README or FAQ should be updated? Does a manual exist which I am not aware of?

LRGB:
[**@gcp**](https://github.com/gcp) commented on [Jun 8, 2018, 2:04 PM GMT-3](https://github.com/gcp/leela-zero/issues/1535#issuecomment-395824441 "2018-06-08T17:04:35Z - Replied by Github Reply Comments"):
> This would require dumping out the info every move, rather than every game. Not sure we want to do this.

What about an autodump every 100 moves? I dont know how often people experience data loss to determine if this is worth the effort or not...

Nazgand:
I am satisfied knowing \`q + Enter'. If games are autosaved, it may be better to do in per unit of time (e.g. 10 minutes) rather than every n moves to account for differences in computing speed.

Nazgand:
Does some other arcane command exist which will finish the current game, then exit? The README should include this info.

d7urban:
Just do the q+Enter, then restart autogtp with the -s flag (for single game)

Nazgand:
A problem with q+Enter:
I needed to reboot, so I pressed q+Enter, then ran 

> autogtp -s

. A while later I decided I did not want to wait for 500 more moves before rebooting. I attempted q+Enter again and got no response. I pressed Ctrl+C, and data of 156 moves were lost.

alreadydone:
https://github.com/gcp/leela-zero/blob/91031bfbcb09b6dc379a62e4c5e84006f0560179/autogtp/main.cpp#L155

I think removing this line would solve this problem. Actually all three quitting conditions (timeout, #games limit, q+Enter) could be made mutually compatible, not mutually exclusive, as in https://github.com/alreadydone/lz/commit/d0f054a9ea14eab66b7ea58efbff2770fc05d146

marcocalignano:
[...]

PR #1574, opened 2018-06-20
Fix q+Enter with -s option

Nazgand:
As suggested https://github.com/gcp/leela-zero/issues/1535#issuecomment-397482619


marcocalignano:
No this is wrong you want q+Enter even without the other options.

Nazgand:
Yeah, I will close this as it is not a fix.

PR #1577, opened 2018-06-21
More draw handling.

TFiFiE:
If the use of different komi is going to be supported, then that should include integral ones, and that means the possibility of draws. I restrained myself however from introducing a contempt factor. :smile:

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -413,7 +413,7 @@ int UCTSearch::get_best_move(passflag_t passflag) {
     assert(first_child != nullptr);
 
     auto bestmove = first_child->get_move();
-    auto bestscore = first_child->get_eval(color);
+    auto bestscore = first_child->first_visit() ? 0.5f : first_child->get_pure_eval(color);
 
     // do we want to fiddle with the best move because of the rule set?
     if (passflag & UCTSearch::NOPASS) {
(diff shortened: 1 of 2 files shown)


src/UCTSearch.cpp
```diff
@@ -413,7 +413,7 @@ int UCTSearch::get_best_move(passflag_t passflag) {
     assert(first_child != nullptr);
 
     auto bestmove = first_child->get_move();
-    auto bestscore = first_child->get_eval(color);
+    auto bestscore = first_child->get_pure_eval(color);
```
gcp: No guarantee yet the first child has visits?

TFiFiE: Does that guarantee exist with `get_eval` though? Maybe make it `0.5f` or `1.0f` with no visits?

gcp: Your point is right, but get_eval works by coincidence due to virtual loss and so won't ever crash here, which the new code would.

TFiFiE: I've changed it, but I don't see how `first_child` is guaranteed to have a virtual loss.

gcp: Hmm, it's not (it's actually guaranteed not to in get_best_move). But if it isn't, then changing the calls from get_eval to get_pure_eval is a no-op obviously.

Maybe it's good to check what exactly happens with p = 1 and see if we get the edge cases right? And is specifying p = 0 possible?

alreadydone:
With maximizing winrate and not the score as the goal, drawn games may be the only chances for us to see AIs play perfect endgames ...

PR #1593, opened 2018-06-28
Delay deletion of game data of continued games

Nazgand:
This reduces data loss in some rare cases(killed process after a continued game), and allows other processes such as LeelaWatcher a chance to load the SGF file.

Not tested as I recently lost my ability to compile autogtp after failing to get my Intel GPU to be default while still allowing my NVIDIA to be used.
Someone, please test this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ the line that says "#define USE_OPENCL".
     # Clone github repo
     git clone https://github.com/gcp/leela-zero
     cd leela-zero/src
-    sudo apt install libboost-dev libboost-program-options-dev libopenblas-dev opencl-headers ocl-icd-libopencl1 ocl-icd-opencl-dev zlib1g-dev
+    sudo apt-get install libboost-dev libboost-program-options-dev libopenblas-dev opencl-headers ocl-icd-libopencl1 ocl-icd-opencl-dev zlib1g-dev g++
     make
     cd ..
     wget http://zero.sjeng.org/best-network
--- a/autogtp/Job.cpp
+++ b/autogtp/Job.cpp
@@ -75,8 +75,6 @@ Result ProductionJob::execute(){
         game.loadSgf(m_sgf);
         game.loadTraining(m_sgf);
         game.setMovesCount(m_moves);
-        QFile::remove(m_sgf + ".sgf");
-        QFile::remove(m_sgf + ".train");
     }
     do {
         game.move();
@@ -112,6 +110,10 @@ Result ProductionJob::execute(){
     default:
         break;
     }
+    if (!m_sgf.isEmpty()) {
+        QFile::remove(m_sgf + ".sgf");
+        QFile::remove(m_sgf + ".train");
+    }
     game.gameQuit();
     return res;
 }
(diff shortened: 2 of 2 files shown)


marcocalignano:
@Nazgand It should work like that but to be sure can you test the following:
- start a game 
- quit the game
- start it again 
- quit it again 
- check that old file are delted
- start again and check that the game starts from where it ended
If this test pass we can merge.

Nazgand:
Yeah, I asked for someone else to test it. At the moment, Ubuntu is screwed up such that I am unable to compile and have yet to figure out how to remedy this.

```
g++ -Wl,-O1 -o autogtp main.o Game.o Worker.o Order.o Job.o Management.o moc_Worker.o moc_Job.o moc_Management.o moc_Console.o   -lQt5Gui -lQt5Core -lGL -lpthread 
/usr/bin/x86_64-linux-gnu-ld: cannot find -lGL
collect2: error: ld returned 1 exit status
Makefile:165: recipe for target 'autogtp' failed
make: *** [autogtp] Error 1
```

I will probably backup things and switch to a fresh install of Linux Mint soon unless I can fix the problem.
Interestingly, leelaz compiles fines even though autogtp does not compile.

marcocalignano:
@Nazgand Tri to add this line in the autogtp project:
`QT -= gui`
just after the `CONFIG` lines. This should free from the QTGui and GL dependency. If you can compile and the test work we can also add that with another PR.

Nazgand:
I already destroyed my Ubuntu install before testing @marcocalignano 's suggestion.

Nazgand:
I tested it and autogtp correctly delays deletion of the files to when exiting with q+Enter or when the game ends.
However, if Ctrl+C is sent, the files do not get reused on subsequent runs which, in retrospect, is good because one would expect multiple autogtp processes running at the same time to not be uncommon.
I thus plan to rename the storefile*.bin to e.g. lockfile*.bin instead of removing it.
The lockfiles may be renamed to storefiles to recover training data. The lockfiles will also be removed at the end of the game.
I plan to add an `--unlock` flag to rename the lockfiles to storefiles.

BTW, Timeshift and Linux Mint are awesome.

Nazgand:
After a bit of digging through the code, I found `storefile" + unique + ".bin.lock` in Worker.cpp. This route seems more complicated than required, so I will look into making the *.sgf, *.train, and storefile*.bin all have the same identifier so delaying the deletion of the storefile will be easier.

marcocalignano:
@Nazgand The Lockfiles are used to sync different threads on the storefile so do not use  them.
You also need a unique identifier for each thread when you save the files.

gcp:
>This should free from the QTGui and GL dependency. If you can compile and the test work we can also add that with another PR.

validation.pro already has this, so it should be safe?

Nazgand:
Okay. Why not use the same UUID for the SGF name?
The pull request works as intended except when autogtp unexpectedly dies, leaving .sgf & .train files without a corresponding storefile. If the storefiles are not backed up under different names and deleted instead, as is currently the case, how easy would it be to regenerate them, I wonder.

marcocalignano:
@Nazgand you can also delete the storefile, in Management, after the game is finished and the .sgf and .train are also deleted.

gcp:
I'm still unclear whether to pull this or not?

Nazgand:
Well pulling this as is would be an improvement because then programs like LeelaWatcher could have access to the sgf for those using LeelaWatcher with the branch `next'.

The training data is not used if autogtp dies because of the absence of a storefile, yet the pull is an improvement regardless.

gcp:
I didn't notice this was against master instead of next until after I'd pulled it :( Had to back out.

marcocalignano:
And please do not merge it in next yet! All that lockfile work was to please our big costumer BIG-Red. 
Since my illness i was a bit out of the loop and miss this PR. Are there other that I should look at?

PR #1618, opened 2018-07-13
Packaging improvements

infinity0:
In preparation for uploading this to Debian, I had to tweak a few things relating to installation.

> Can you describe more about how to set ${CMAKE_INSTALL_BINDIR} when building by cmake or make?

`cmake -D` is how you set cmake variables in general.

> Validation can handle other engines than just leelaz, though. And I don't see much risk of autogtp clashing.

Can you suggest a better name for `validation`? For `autogtp` it's still specific to `leelaz` but that's not reflected in the name.

> @infinity0 There is already some works on making Debian package in the “next” branch. See #1445 for reference.

That creates a Debian binary package, which isn't how one gets packages into Debian official repositories (`ftp.debian.org`). I'm talking about the necessary changes to adhere to software Debian policy standards, to upload it to the Debian official repositories. In general you create a `debian/` directory and build a source package `.dsc` out of it, and use `dpkg-buildpackage` to build a binary package `.deb`. This carries many extra benefits like automatic packages with split debugging symbols.

To avoid latency I've gone ahead and uploaded leela-zero with my patches to Debian already and it's just [awaiting copyright review](https://ftp-master.debian.org/new/leela-zero_0.15-1.html). Will be happy to change the names of the binaries though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ add_subdirectory(gtest EXCLUDE_FROM_ALL) # We don't want to install gtest, exclu
 # Required Packages
 set(Boost_MIN_VERSION "1.58.0")
 set(Boost_USE_MULTITHREADED ON)
-find_package(Boost 1.58.0 REQUIRED program_options)
+find_package(Boost 1.58.0 REQUIRED program_options filesystem)
 find_package(Threads REQUIRED)
 find_package(ZLIB REQUIRED)
 find_package(OpenCL REQUIRED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,7 +108,7 @@ target_link_libraries(leelaz ${BLAS_LIBRARIES})
 target_link_libraries(leelaz ${OpenCL_LIBRARIES})
 target_link_libraries(leelaz ${ZLIB_LIBRARIES})
 target_link_libraries(leelaz ${CMAKE_THREAD_LIBS_INIT})
-install(TARGETS leelaz DESTINATION bin)
+install(TARGETS leelaz DESTINATION ${CMAKE_INSTALL_BINDIR})
 
 if(Qt5Core_FOUND)
     if(NOT Qt5Core_VERSION VERSION_LESS "5.3.0")
--- a/autogtp/CMakeLists.txt
+++ b/autogtp/CMakeLists.txt
@@ -7,4 +7,4 @@ add_executable(autogtp
 set_target_properties(autogtp PROPERTIES AUTOMOC 1)
 target_link_libraries(autogtp Qt5::Core)
 
-install(TARGETS autogtp DESTINATION bin)
+install(TARGETS autogtp DESTINATION ${CMAKE_INSTALL_BINDIR})
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -102,6 +102,7 @@ void GTP::setup_default_parameters() {
     cfg_max_visits = UCTSearch::UNLIMITED_PLAYOUTS;
     cfg_timemanage = TimeManagement::AUTO;
     cfg_lagbuffer_cs = 100;
+    cfg_weightsfile = leelaz_file("best-network");
 #ifdef USE_OPENCL
     cfg_gpus = { };
     cfg_sgemm_exhaustive = false;
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -168,32 +168,34 @@ size_t Utils::ceilMultiple(size_t a, size_t b) {
 #ifdef _WIN32
 // https://stackoverflow.com/a/3999597
 // Convert a wide Unicode string to an UTF8 string
-// Thanks Windows, saving developers time since 1995
 std::string utf8_encode(const std::wstring &wstr)
 {
-    if ( wstr.empty() ) return std::string();
-    int size_needed = WideCharToMultiByte(CP_UTF8, 0, &wstr[0], (int)wstr.size(), NULL, 0, NULL, NULL);
+    if (wstr.empty()) {
+        return std::string();
+    }
+    int size_needed = WideCharToMultiByte(CP_UTF8, 0, &wstr[0], (int)wstr.size(), nullptr, 0, nullptr, nullptr);
     std::string strTo(size_needed, 0);
-    WideCharToMultiByte(CP_UTF8, 0, &wstr[0], (int)wstr.size(), &strTo[0], size_needed, NULL, NULL);
+    WideCharToMultiByte(CP_UTF8, 0, &wstr[0], (int)wstr.size(), &strTo[0], size_needed, nullptr, nullptr);
     return strTo;
 }
 #endif
 
-std::string Utils::leelaz_file(std::string file) {
+const std::string Utils::leelaz_file(std::string file) {
 #ifdef _WIN32
     // https://stackoverflow.com/a/11032792 but avoiding the UNICODE mess and just using W variants
     wchar_t szPath[MAX_PATH];
-    if (!SUCCEEDED(SHGetFolderPathW(NULL, CSIDL_COMMON_APPDATA, NULL, 0, szPath)))
-        return NULL;
+    if (!SUCCEEDED(SHGetFolderPathW(nullptr, CSIDL_COMMON_APPDATA, nullptr, 0, szPath))) {
+        return std::string();
+    }
     PathAppendW(szPath, L"\\Leela Zero\\");
     boost::filesystem::path dir(utf8_encode(szPath));
 #else
     // https://stackoverflow.com/a/26696759
     const char *homedir;
-    if ((homedir = getenv("HOME")) == NULL) {
+    if ((homedir = getenv("HOME")) == nullptr) {
         struct passwd *pwd;
-        if ((pwd = getpwuid(getuid())) == NULL) { // NOLINT(runtime/threadsafe_fn)
-            return NULL;
+        if ((pwd = getpwuid(getuid())) == nullptr) { // NOLINT(runtime/threadsafe_fn)
+            return std::string();
         }
         homedir = pwd->pw_dir;
     }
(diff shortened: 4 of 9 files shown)


infinity0:
BTW I will be very happy to add any of you as co-maintainer of the Debian package. The source code is here: https://salsa.debian.org/infinity0/leela-zero/tree/master/debian

infinity0:
This PR is already from next branch.

ChinChangYang:
@infinity0 Ah, I did not notice, because I enter here by clicking links from email.

It looks good now.

Ttl:
One possible issue with installing to a system directory is that the OpenCL tuning file `leelaz_opencl_tuning` is only read and written in the working directory. If the tuning file doesn't exist in the working directory the tuner will be ran again which can take a while.

infinity0:
Typically that sort of thing can get stored in `~/.local/share/$pkgname`. For a system-wide install we could run it on post-install, store it in `/var/lib/$pkgname` then have `leelaz` copy it into the homedir so it's writeable, but so the first-run is much quicker.

Ttl:
I don't think that solution works on Windows since [HOME is not set](https://stackoverflow.com/questions/2552416/how-can-i-find-the-users-home-dir-in-a-cross-platform-manner-using-c).

infinity0:
We can use `%APPDATA%` instead but I'm not sure about the windows API function to get it, anyone got a windows machine to figure it out / test it on?

infinity0:
My package passed the Debian copyright review process and is now in the [Debian repositories](https://packages.debian.org/sid/leela-zero). I've just uploaded a tweak with some [improved documentation](https://salsa.debian.org/infinity0/leela-zero/blob/master/debian/README.Debian#L1) on the minor differences.

Please let me know if I should change anything (e.g. renaming `leelaz-autogp`, or changing the paths from `~/.local/share/leela-zero` to `~/.local/share/leelaz` or whatever). In the meantime I would still like to get this PR merged so that the Debian package has to carry less patches. I'll try to find some time to research the `%APPDATA%` situation on windows.

infinity0:
Yet another alternative to renaming the binaries is to instead install them in a non-PATH path like `/usr/lib/leelaz/libexec` and then symlink only `leelaz` into `/usr/games` which is on the PATH for human user accounts.

Lemme know what you guys prefer and I'll make the necessary changes to this PR.

ChinChangYang:
Make the naming consistent. The full name of this project is leela-zero, but the executable file name is its abbreviation, leelaz. In many other projects, such as gedit, gnugo, they only have one name.

Since this project is named as leela-zero, I believe leelaz should be renamed to leela-zero in the future work. Otherwise, this project should be renamed to leelaz.

gcp:
> I believe leelaz should be renamed to leela-zero in the future work.

I prefer this option, i.e. to make the binary just named leela-zero(.exe).

gcp:
>In the meantime I would still like to get this PR merged so that the Debian package has to carry less patches.

It should pass CI for that.

infinity0:
Sorry about the CI, I haven't had time to test on windows yet.

Shall I also rename the binary to `leela-zero` as part of this PR then?

ChinChangYang:
It is necessary to pass CI, but unnecessary to rename the binary to leela-zero. Renaming the binary may cause high level interface backward compatible problems, so we can rename the binary in the future work.

gcp:
>Shall I also rename the binary to leela-zero as part of this PR then?

Better to make the CI small.

I guess CI is partly failing because things like ```"%s/.cache/leela-zero/best-network"`` are probably not as appropriate to Windows.

CMakeLists.txt
```diff
@@ -108,7 +108,7 @@ target_link_libraries(leelaz ${BLAS_LIBRARIES})
 target_link_libraries(leelaz ${OpenCL_LIBRARIES})
 target_link_libraries(leelaz ${ZLIB_LIBRARIES})
 target_link_libraries(leelaz ${CMAKE_THREAD_LIBS_INIT})
-install(TARGETS leelaz DESTINATION bin)
+install(TARGETS leelaz DESTINATION ${CMAKE_INSTALL_BINDIR})
```
gcp: Can/will this default to bin if not set?

autogtp/CMakeLists.txt
```diff
@@ -1,10 +1,10 @@
 
 cmake_minimum_required(VERSION 3.1)
 
-add_executable(autogtp
+add_executable(leelaz-autogtp
```
gcp: Let's keep the autogtp name. Low risk of clashes and it might be what people expect by now.

validation/CMakeLists.txt
```diff
@@ -1,10 +1,10 @@
 
 cmake_minimum_required(VERSION 3.1)
 
-add_executable(validation
+add_executable(leelaz-validation
```
gcp: gtp-validation? Or leelaz-validation.  Or leela-zero-validation. All are fine.

src/GTP.cpp
```diff
@@ -96,6 +97,7 @@ void GTP::setup_default_parameters() {
     cfg_max_visits = UCTSearch::UNLIMITED_PLAYOUTS;
     cfg_timemanage = TimeManagement::AUTO;
     cfg_lagbuffer_cs = 100;
+    cfg_weightsfile = (boost::format("%s/.cache/leela-zero/best-network") % getenv("HOME")).str();
```
gcp: Probably not the appropriate location on Windows. (And HOME might not be set?)

infinity0:
What is the actual purpose of the autogtp and validation binaries? I couldn't find a short summary documented anywhere.

CMakeLists.txt
infinity0: Yes that's cmake's normal behaviour.

infinity0:
OK windows code fixed, tests passing. I've removed the binary renames for now until you guys make a final decision, in the meantime in Debian I'll just install them to a non-PATH location and symlink only `leelaz` into `/usr/games`.

src/Utils.cpp
```diff
@@ -156,3 +164,43 @@ size_t Utils::ceilMultiple(size_t a, size_t b) {
...
+    WideCharToMultiByte(CP_UTF8, 0, &wstr[0], (int)wstr.size(), &strTo[0], size_needed, NULL, NULL);
+    return strTo;
+}
+#endif
+
+std::string Utils::leelaz_file(std::string file) {
+#ifdef _WIN32
+    // https://stackoverflow.com/a/11032792 but avoiding the UNICODE mess and just using W variants
+    wchar_t szPath[MAX_PATH];
+    if (!SUCCEEDED(SHGetFolderPathW(NULL, CSIDL_COMMON_APPDATA, NULL, 0, szPath)))
```
gcp: Given that we have no installer who plunks things into AppData etc, there's no reason to assume the network will be there. For Windows it's fine to just keep the current behavior and avoid all this complication, I made remarks before because the Linux support was changing (and breaking) Windows. It doesn't necessarily need to do the same.

src/GTP.cpp
```diff
@@ -102,6 +102,7 @@ void GTP::setup_default_parameters() {
     cfg_max_visits = UCTSearch::UNLIMITED_PLAYOUTS;
     cfg_timemanage = TimeManagement::AUTO;
     cfg_lagbuffer_cs = 100;
+    cfg_weightsfile = leelaz_file("best-network");
```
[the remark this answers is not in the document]
infinity0: We get an error message about a nonexistent file "Could not open weights file: %"

src/Utils.cpp
infinity0: Are you sure?

It does not change the current behaviour regarding `--weights`, it only avoids an error if no `--weights` is given and that file exists. However it does change the behaviour of where the `leelaz_opencl_tuning` file is saved, but arguably this is better even on Windows, since now all invocations of leelaz will write to one standard file rather than dropping these files all over the filesystem depending on what the user's CWD was when they launched leelaz.

src/GTP.cpp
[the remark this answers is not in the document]
gcp: I think that's not so ideal?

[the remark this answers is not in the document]
infinity0: I've updated the current error message about it now.

infinity0:
Ping? I did everything except remove the Windows code as I thought you might reconsider and keep it in, but if you really want to remove it I can do that as well.

Once this is merged and 0.16 release I can then package lizzie for Debian so users have a nice GUI for it. The packaging is already done for that too, I am just waiting for a release here since it depends on the `next` branch and does not work with 0.15.

gcp:
>Ping? I did everything except remove the Windows code as I thought you might reconsider and keep it in, but if you really want to remove it I can do that as well

Yes please, looks fine otherwise.

infinity0:
Ok, done

gcp:
@infinity0 Thanks.

I somehow forgot that not just the weights file but also the tuning file were being put in those directories. In that context, the Windows changes would in fact have been desirable. Sorry about that.

infinity0:
@gcp Well it's not too late, if you want it back you can just `git revert 5517421` while on the `next` branch and edit the commit message before saving - it is like a reverse cherry-pick, the commit does not need to come from a previous point of the current branch. You probably still have commit 5517421 somewhere in your git object cache if you fetched or pulled from my fork recently.

AncalagonX:
I was no longer able to build /next/ as of yesterday on Windows 10 with Microsoft Visual Studio 2017, and guessed it was this commit. I reverted only this commit, and was then able to build without error as usual.

The error in VS2017 is:

`LINK : fatal error LNK1104	cannot open file 'libboost_filesystem-vc141-mt-1_65_1.lib' leela-zero C:\+++Main\++git_repos\leela-zero\msvc\VS2017\LINK`

I have confirmed there are no permissions issues with the directories involved.

There are a *lot* of Google hits for this error involving that exact .lib file, but it's way above my level to troubleshoot further.

infinity0:
@gcp BTW when can we expect version 0.16 to be released? If it's too long I might just package the next branch in Debian unstable already, so that I can also package lizzie which needs that branch.

gcp:
@infinity0 I'm still seeing various issues related to the fp16 handling and the interaction with drivers, so it looks like it might be a bit before there can be a stable release.

PR #1620, opened 2018-07-14
Restructuring part2 : Create abstract class ForwardPipe, which represents a class that has a forward() call

ihavnoid:
Created an abstract class ForwardPipe
- class OpenCLScheduler is a child of ForwardPipe
- created a CPUPipe which is a child of ForwardPipe, forward_cpu() and related methods moved to ForwardPipe
-- Hence instantiation of a different ForwardPipe() type will result in different backends
- Other accompanying fixes


src/Leela.cpp
```diff
@@ -81,6 +81,12 @@ static void parse_commandline(int argc, char *argv[]) {
         ("noponder", "Disable thinking on opponent's time.")
         ("benchmark", "Test network and exit. Default args:\n-v3200 --noponder "
                       "-m0 -t1 -s1.")
+#ifdef USE_CPU_ONLY
+        ("cpu-only", "Use CPU-only implementation.\n"
+                     "This build only has CPU implementation hence this option exists oly for compatability.")
```
Hersmunch: Typo "oly"

src/Network.cpp
```diff
@@ -375,55 +349,58 @@ void Network::initialize(int playouts, const std::string & weightsfile) {
...
+        m_forward_cpu.reset(new CPUPipe());
+        to_init.push_back(m_forward_cpu.get());
+    }
+#endif
 
-        const auto Upad = zeropad_U(conv_weights[weight_index],
-                                    channels, INPUT_CHANNELS,
-                                    m_ceil, k_ceil);
 
+    for(auto p : to_init) {
```
Hersmunch: Missing space between for and (

src/Network.cpp
```diff
@@ -375,55 +349,58 @@ void Network::initialize(int playouts, const std::string & weightsfile) {
...
+    
         // Winograd filter transformation changes filter size to 4x4
-        opencl_net->push_input_convolution(WINOGRAD_ALPHA, INPUT_CHANNELS,
-            channels, Upad,
+        p->push_input_convolution(WINOGRAD_ALPHA, INPUT_CHANNELS,
+            channels, conv_weights[weight_index],
             batchnorm_means[weight_index], batchnorm_stddivs[weight_index]);
         weight_index++;
-
+
```
Hersmunch: whitespace

src/Network.cpp
```diff
@@ -375,55 +349,58 @@ void Network::initialize(int playouts, const std::string & weightsfile) {
...
                                       batchnorm_means[weight_index],
                                       batchnorm_stddivs[weight_index],
-                                      Upad2,
+                                      conv_weights[weight_index + 1],
                                       batchnorm_means[weight_index + 1],
                                       batchnorm_stddivs[weight_index + 1]);
             weight_index += 2;
         }
-
+
```
Hersmunch: whitespace

src/Network.cpp
```diff
@@ -375,55 +349,58 @@ void Network::initialize(int playouts, const std::string & weightsfile) {
...
-            const auto Upad1 = zeropad_U(conv_weights[weight_index],
-                                         channels, channels,
-                                         m_ceil, m_ceil);
-            const auto Upad2 = zeropad_U(conv_weights[weight_index + 1],
-                                         channels, channels,
-                                         m_ceil, m_ceil);
-            opencl_net->push_residual(WINOGRAD_ALPHA, channels, channels,
-                                      Upad1,
+            p->push_residual(WINOGRAD_ALPHA, channels, channels,
+                                      conv_weights[weight_index],
```
Hersmunch: indentation

TFiFiE:
Why are occurrences of `net_t` being replaced by `float`?

src/CPUPipe.h
```diff
@@ -0,0 +1,91 @@
...
+                            const std::vector<float>& U,
+                            std::vector<float>& V,
+                            std::vector<float>& M,
+                            std::vector<float>& output);
+
+
+    int m_channels;
+
+    // Input + residual block tower
+    std::vector<std::vector<float>> conv_weights;
```
Hersmunch: should these have an `m_` prefix?

src/OpenCLScheduler.cpp
```diff
@@ -20,9 +20,39 @@
...
 #include "GTP.h"
 #include "Random.h"
+#include "Network.h"
+#include "Utils.h"
 #include "OpenCLScheduler.h"
 
+using Utils::ceilMultiple;
 
+static std::vector<float> zeropad_U(const std::vector<float>& U,
+                                      const int outputs, const int channels,
```
Hersmunch: indentation? Though I'm not sure exactly what rule we are following.

ihavnoid:
@TFiFiE net_t is replaced by float in some places because I wanted to make the forward() interface more general and allow different internal implementations with different types.  e.g., if we want to make fp16 a runtime option we can't use net_t here because the datatype should not be compile-time dependent.

A more general approach might be something like a bool, but that just seemed to be a bit too much at this point and hence didn't change the rest of the code.  If it seems to make sense to change the prototype of forward() to have boolean inputs then I can work on it too.

ihavnoid:
Addressed the first round of comments and pushed change.  Please have a look

src/Leela.cpp
```diff
@@ -81,6 +81,12 @@ static void parse_commandline(int argc, char *argv[]) {
         ("noponder", "Disable thinking on opponent's time.")
         ("benchmark", "Test network and exit. Default args:\n-v3200 --noponder "
                       "-m0 -t1 -s1.")
+#ifdef USE_CPU_ONLY
+        ("cpu-only", "Use CPU-only implementation.\n"
+                     "This build only has CPU implementation hence this option exists only for compatability.")
```
gcp: Just remove the option completely in this case and force the cfg var on?

src/OpenCL.cpp
```diff
@@ -614,8 +617,10 @@ void OpenCL_Network::forward(const std::vector<net_t>& input,
         queue.finish();
     }
 
-    std::memcpy(output_pol.data(), pinnedOutBufferHost_pol, finalSize_pol);
-    std::memcpy(output_val.data(), pinnedOutBufferHost_val, finalSize_val);
+    auto polptr = static_cast<net_t*>(pinnedOutBufferHost_pol);
+    auto valptr = static_cast<net_t*>(pinnedOutBufferHost_val);
+    std::copy(polptr, polptr + output_pol.size(), output_pol.begin());
```
gcp: begin(output_pol)

src/OpenCL.cpp
```diff
@@ -614,8 +617,10 @@ void OpenCL_Network::forward(const std::vector<net_t>& input,
...
+    std::copy(valptr, valptr + output_val.size(), output_val.begin());
```
gcp: Same

src/OpenCL.cpp
```diff
@@ -518,8 +518,11 @@ void OpenCL_Network::forward(const std::vector<net_t>& input,
     cl::Buffer & MBuffer = opencl_context.m_MBuffer;
     cl::CommandQueue & queue = opencl_context.m_commandqueue;
 
+    std::vector<net_t> net_t_input(input.size());
+    std::copy(input.begin(), input.end(), net_t_input.begin());
```
gcp: begin(input) etc

ihavnoid:
Addressed 2nd batch of comments - please have a look

ihavnoid:
Something went wrong while merging with latest next branch.  Seemingly playing random moves Will have a look

Issue #1514, opened 2018-06-04
Suggestion for suppressing network unfavorate moves


Splee99:
As we discussed in #1417, LZ tends to consider many weird moves when winrate is low and the number of visits is high with powerful GPUs. These moves have one thing in common, a low N value which I suppose is not favored by the neural network. I have tried the following modified Nodecomp function in UCTNode.cpp:
```text
        if (a.get_visits() != b.get_visits()) {
        	if (a.get_score() < 0.005) return true;
        	if (b.get_score() < 0.005) return false;
            return a.get_visits() < b.get_visits();
        }
```
where two lines are added to push the nodes with N value less than 0.5% to the end of all children. I have run several handicap games and the result is promising and no weird moves were played.

Whether this modification is useful in even games I'm not sure, but it indeed helps to avoid the weird moves.

gcp:
You need more concrete data than "result is promising".

petgo3:
I had a try on KGS (petgo3) with your suggestion. Weird moves appeared as well.

Splee99:
My test is based on move by move analysis. Here is the complete log of a 4 handicap with Hiratuka which is set with 20000 playouts. 
[4h1.log](https://github.com/gcp/leela-zero/files/2070267/4h1.log)

It is easy to see that all move candidates have N value bigger than 0.5% and LZ also got a game win. I have also run several similar games and LZ won all of them. Before the modification it was very hard to win for LZ with similar settings. I know I have to do more testing. 

Here is the sgf file.
[lz-vc-nohalf-hira-4h-30s14.sgf.txt](https://github.com/gcp/leela-zero/files/2070277/lz-vc-nohalf-hira-4h-30s14.sgf.txt)

@petgo3 Would you please post the sgf file with those weird moves, so I can take a closer look?

Splee99:
@petgo3 If you were referring to the 4 handi game just played on KGS, I couldn't repeat the move O1 on my computer. It is not even on the list.

sethtroisi:
closing older issues, no discussion for >6 months

PR #1623, opened 2018-07-14
An improved implementation of #1514 (screening out low net score children)

Splee99:
To recap the issue #1514, LZ sometimes plays weird first line or second line move when win rate is low. All those moves have one thing in common, low net scores. In this implementation those low score children will not be selected and other better moves will get more visits instead.
The ELF weight with only one playout is the best tester for this. The original LZ 0.15 often plays many weird moves in 5 handicap games and is unable to win ELF 1 playout with 3000 playouts. With the modified version LZ won many of the games. One example is here:
[elf1-5h-e1d4-3000po-cap-only1.sgf.txt](https://github.com/gcp/leela-zero/files/2195329/elf1-5h-e1d4-3000po-cap-only1.sgf.txt)


gcp:
Not going to pull in a change that ass a random magic parameter, nor that breaks the search scaling by making it unable to find good moves with low priors.

This will never be pulled.

alreadydone:
@Splee99 I haven't had time to try your modification yet, but I ran 9 games with my komi-v0.3 (3000 playouts, some with different parameters) against ELF 1 playout, and komi-v0.3 won only one game due to ladder, so your claim indeed sounds impressive. Will you keep testing and collect data and also test against humans? By the way, you can do such pruning at the time of creating children, but with a different metric: not the net score itself, but the ratio between the net score and the highest net score.
https://github.com/gcp/leela-zero/blob/8e9ca8eb073632141b3c4995f389fc9c481f3d67/src/UCTSearch.cpp#L207-L208
https://github.com/gcp/leela-zero/blob/8e9ca8eb073632141b3c4995f389fc9c481f3d67/src/UCTNodeRoot.cpp#L186
So you can change get_min_psa_ratio() to a fixed value like you did in this PR (or add it to the end in the second case). This should save some memory and computation but the gain is probably minor.

Splee99:
@alreadydone With this modification only LZ still plays weird first line or second line move. Then I changed cfg_puct from 0,8 to 0,2, those weird moves are completely gone and LZ can win ELF 1 playout with 2700 playouts.
[elf1-5h-e1d4-2700po-cap-0.9-puct0.2.sgf.txt](https://github.com/gcp/leela-zero/files/2199770/elf1-5h-e1d4-2700po-cap-0.9-puct0.2.sgf.txt)
I don't care this change will be ignored. I'll keep it for myself.

alreadydone:
Despite your tests, I find it highly suspicious that changing 0.8 to 0.2 would help avoid 1st/2nd line moves. This PR is filtering out low net score moves, while shrinking cfg_puct would make search focus more on winrate (which is not reliable when it's very low) and not policy prior. So the two changes seem to work against each other.

PR #1635, opened 2018-07-19
Restrict the use of "score".

TFiFiE:
I think that using "score" as a nonspecific term (and not when it, for example, refers to the count at the end of game) makes it unnecessarily hard to understand the code and see how it matches with the literature.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -585,19 +585,19 @@ bool GTP::execute(GameState & game, std::string xinput) {
         Network::Netresult vec;
         if (cmdstream.fail()) {
             // Default = DIRECT with no symmetric change
-            vec = network->get_scored_moves(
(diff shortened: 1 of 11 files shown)


lightvector:
Hehe, that was exactly one of the things that confused me when I was trying to understand some parts of Leela Zero - the fact that several distinct things were all called "score" in different places. Glad to see at least one other person had the same thought. :)

TFiFiE:
I also had trouble understanding what "psa" is supposed to mean, which I now figure should probably be read as `P(s,a)`, that is "probability given state and action".

gcp:
>I also had trouble understanding what "psa" is supposed to mean, which I now figure should probably be read as P(s,a), that is "probability given state and action".

Correct.

PR #1638, opened 2018-07-21
Replace if-else chain with switch statement.

TFiFiE:
Not sure if this runs afoul of the ruling style guide here, but it clearly looks prettier to me.


src/Network.cpp
```diff
@@ -233,36 +233,27 @@ std::pair<int, int> Network::load_v1_network(std::istream& wtfile) {
...
-        } else if (linecount == plain_conv_wts + 11) {
-            std::copy(cbegin(weights), cend(weights), begin(ip1_val_b));
-        } else if (linecount == plain_conv_wts + 12) {
-            std::copy(cbegin(weights), cend(weights), begin(ip2_val_w));
-        } else if (linecount == plain_conv_wts + 13) {
-            std::copy(cbegin(weights), cend(weights), begin(ip2_val_b));
+        } else {
+            const auto offset = linecount - plain_conv_wts;
+            if (offset == 3 || offset == 9) {
+                process_bn_var(weights);
```
gcp: If process_bn_var is in-place this can just be moved out of the loop and applied to bn_pol_2 and bn_val_v2.

src/Network.cpp
```diff
@@ -88,7 +88,9 @@ void Network::benchmark(const GameState* const state, const int iterations) {
              runcount.load(), elapsed, int(runcount.load() / elapsed));
 }
 
-void Network::process_bn_var(std::vector<float>& weights, const float epsilon) {
+template<class container>
+void process_bn_var(container& weights) {
+    constexpr float epsilon = 1e-5f;
```
TFiFiE: I guess you would want that to be `auto`.

Issue #1640, opened 2018-07-22
Can `lz-analyze` set an arbitrary color of next move?


zhanzhenzhen:
Sometimes you only know a position and which color to move, but you don't know the game record, so the next color couldn't be determined automatically.

I found `lz-genmove_analyze` can be used to set an arbitrary color to move. But it will actually do a move after analyzing. Could `lz-analyze` set it also?

zhanzhenzhen:
Just know that we can set the PL attribute in SGF, therefore we can specify which color to move for lz-analyze. I'll close it.

PR #1642, opened 2018-07-22
Add `input_pending` in `UCTSearch::think`

zhanzhenzhen:
Sometimes you only know a position and which color to move, but you don't know the game record, so the next color couldn't be determined automatically.

While "lz-analyze" command can do analyzing, we cannot choose which color to move arbitrarily.

"lz-genmove_analyze" can choose the color, but we can't press ENTER to stop.

To resolve it, I just did a small change: Adding `input_pending` in `UCTSearch::think`.

See: #1640

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -721,7 +721,7 @@ int UCTSearch::think(int color, passflag_t passflag) {
         keeprunning  = is_running();
         keeprunning &= !stop_thinking(elapsed_centis, time_for_move);
         keeprunning &= have_alternate_moves(elapsed_centis, time_for_move);
-    } while (keeprunning);
+    } while (!Utils::input_pending() && keeprunning);
 
     // stop the search
     m_run = false;


gcp:
I think it's better to add the color as an optional argument to lz-analyze. @bittsitt thoughts?

zhanzhenzhen:
@gcp If lz-analyze can be more like "think" than "ponder", that would be the best!

Sometimes (when dealing with passing), lz-analyze (using ponder) now couldn't suggest the move that genmove would play. So I also hope lz-analyze could give us more information.

gcp:
I suspect we can make lz-analyze take the new form and just check when parsing the arguments whether we have an interface that expects to send the old form.

i.e. if the first argument is not w/white/b/black assume the old format.

zhanzhenzhen:
Just know the PL property in SGF. So I'm no longer eager for lz-analyze to extend the syntax, though I'm still using a modified version of lz-genmove_analyze in my own program :)

gcp:
Implemented in #1872.

PR #1643, opened 2018-07-23
Winograd F(4x4, 3x3)

Ttl:
Replace F(2x2, 3x3) Winograd transformations with F(4x4, 3x3) transformations that have bigger tile size. F(2x2, 3x3) is theoretically 2.25 times faster than direct convolution and F(4x4, 3x3) is 4 times faster. Transformations are bigger though so in practice the gain is probably not as big.

Benchmarks on GTX 1050 Ti:
```
F(2x2, 3x3) (next):
 1600 evaluations in  7.93 seconds -> 201 n/s

F(4x4, 3x3) (This PR):
 1600 evaluations in  5.57 seconds -> 287 n/s
```

Batching is much more useful with the larger tile size. There are only 25 tiles on 19x19 board with the larger tile size when there was previously 100. For SGEMM it needs to be padded to 32 tiles so about 20 % of computation is wasted on the padding. With batch size of 4 I get 367 n/s, which would be 80 % faster than the current code.

The tree search part of batching is still missing so it can't be used even though the OpenCL part supports it. The good news is that because of the required padding for batch size of one, even a simple batching implementation with fixed size should be faster on pretty much all devices. I'm not familiar with that part of the code and I hope someone else can implement the tree search part of the batching support (@ihavnoid ?).

Intel iGPU users will also be pleasantly surprised that I spent some time in vectorizing the transformations so they will run much faster on Intel iGPUs. Performance increased on my laptop from 24 n/s to 49 n/s with this PR and batch size of 4 (34 n/s with batch size of 1).

### DIFF
--- a/src/CPUPipe.cpp
+++ b/src/CPUPipe.cpp
@@ -35,9 +35,6 @@
 #include "Network.h"
 #include "Im2Col.h"
 
-// Square root of 2
-#define SQ2 (1.4142135623730951f)
-
 void CPUPipe::initialize(int channels) {
     m_input_channels = channels;
 }
@@ -47,8 +44,8 @@ void CPUPipe::winograd_transform_in(const std::vector<float>& in,
                                     const int C) {
     constexpr auto W = BOARD_SIZE;
     constexpr auto H = BOARD_SIZE;
-    constexpr auto WTILES = W / WINOGRAD_M + (W % WINOGRAD_M != 0);
-    constexpr auto P = WTILES * WTILES;
+    constexpr auto WTILES = WINOGRAD_WTILES;
+    constexpr auto P = WINOGRAD_P;
 
     constexpr auto Wpad = 2 + WINOGRAD_M * WTILES;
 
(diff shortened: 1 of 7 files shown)


gcp:
Nice work.

Does combining this with USE_HALF start to showing rounding issues?

Eddh:
Does this mean the architecture will be different ?

gcp:
>Does this mean the architecture will be different ?

No, it's just a more efficient way of doing the computation. The outputs are still 3x3, i.e. equivalent to 3x3 convolutions.

Eddh:
OK. I guess this will be particularly useful on android :)

ihavnoid:
Nice work!

Regarding batching what I did was simply increased the number of threads to a huge number.  I didnt change anything on the tree search code itself. If there are enough threads needing a NN evaluation, then batching is used, if not, then it simply used the non batch mode implementation.  Something to gain on heavily threaded situations, nothing to lose on low thread situations.

I do wonder... did you test USE_HALF?  I do want to put a PR that supports runtime selection of precisions and that probably needs to be pushed after this gets merged...

Ttl:
`USE_HALF` still works and seems to be accurate enough. On empty board heatmap is identical and only the fourth decimal of the winrate is different.` USE_HALF` gives 10 % more n/s with this commit on GTX 1050 Ti.

gcp:
> I didnt change anything on the tree search code itself. If there are enough threads needing a NN evaluation, then batching is used, if not, then it simply used the non batch mode implementation

IIRC this was done with some slight delay? We should discuss batching strategies in an issue. It came to my mind that we might be able to use a global "number of threads in a lock" counter to detect when we lack tree parallelism for a batch.

kuba97531:
FWIW: I did some tests and I get consitently **worse** performance with this PR 7cfbb72b vs one commit before 6333b66.
Windows, Release build from VS2017. executable ` leelaz.exe -w 153 --gpu 0 -t 7` GPU is 1080Ti.
I have performed the benchmark on an asymetric position with commands:
`play B R16`, `play W D3`, `genmove B`

| Commit ID | n/s |
| --- | --- |
|this PR 7cfbb72b| [733,731,738]|
|previous 6333b66| [776, 770, 772]|

gcp:
You should file an issue. Also try the default number of threads.

I got 15-25% speedup on all NVIDIA cards I tested.

gcp:
Another thing to note is that this requires re-tuning. If you forced the full tuner, you might want to do so again.

fell111:
Ubuntu18.04 + 1070ti 
Run full tuner before each test. Tested with 50k visit.

Situation | n/s
------------ | -------------
Master with USE_HALF | 661
Next without USE_HALF | 589
Next with USE_HALF | 723

PR #1647, opened 2018-07-24
Increase error budget in tuner

### LINKED ISSUE #1645: Winograd (4x4,3x3), USE_HALF and 256x20
The combination doesn't seem to work:

```
Started OpenCL SGEMM tuner.
Will try 290 valid configurations.
Failed to find a working configuration.
Check your OpenCL drivers.
```

@ihavnoid @Ttl

Ttl:
Fixes #1645. Tuner has 0.5 % error with the previous SGEMM dimensions with 256 channel network but it increases to 1.3 % with the new dimensions required for the new Winograd convolution. The network output seems to be fine even with the larger error so just increase the error limits in the tuner.


gcp:
Looks like the self-check is set to accept 5% error in any case.

PR #1649, opened 2018-07-24
Runtime selection of fp16/fp32

ihavnoid:
This adds --use-half as a command line option, and users can select between fp16/fp32 based on command line.

### DIFF
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -90,6 +90,9 @@ static void parse_commandline(int argc, char *argv[]) {
                 "ID of the OpenCL device(s) to use (disables autodetection).")
         ("full-tuner", "Try harder to find an optimal OpenCL tuning.")
         ("tune-only", "Tune OpenCL only and then exit.")
+#ifdef USE_HALF
+        ("use-half", "Use half-precision OpenCL code.  Traades off some accuracy for higher performance")
+#endif
         ;
 #endif
     po::options_description selfplay_desc("Self-play options");
@@ -318,6 +321,12 @@ static void parse_commandline(int argc, char *argv[]) {
     if (vm.count("tune-only")) {
         cfg_tune_only = true;
     }
+
+#ifdef USE_HALF
+    if (vm.count("use-half")) {
+        cfg_use_half = true;
+    }
+#endif
 #endif
 
     if (vm.count("benchmark")) {
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -91,7 +91,7 @@ static void parse_commandline(int argc, char *argv[]) {
         ("full-tuner", "Try harder to find an optimal OpenCL tuning.")
         ("tune-only", "Tune OpenCL only and then exit.")
 #ifdef USE_HALF
-        ("use-half", "Use half-precision OpenCL code.  Traades off some accuracy for higher performance")
+        ("use-half", "Use half-precision OpenCL code.  Trades off some accuracy for higher performance")
 #endif
         ;
 #endif
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -345,26 +345,42 @@ void Network::initialize(int playouts, const std::string & weightsfile) {
 
     std::vector<ForwardPipe*> to_init;
 
+    bool use_selfcheck = true;
 #ifdef USE_OPENCL
     if (cfg_cpu_only) {
         myprintf("Initializing CPU-only evaluation.\n");
         m_forward = std::make_unique<CPUPipe>();
+        use_selfcheck = false;
     } else {
-        myprintf("Initializing OpenCL.\n");
-        m_forward = std::make_unique<OpenCLScheduler>();
-    }
+#ifdef USE_HALF
+        if (cfg_use_half) {
+            myprintf("Initializing OpenCL (half precision).\n");
+            m_forward = std::make_unique<OpenCLScheduler<half_float::half>>();
+            use_selfcheck = false;
+        } else {
+            myprintf("Initializing OpenCL (single precision).\n");
+            m_forward = std::make_unique<OpenCLScheduler<float>>();
+        }
 #else
+        myprintf("Initializing OpenCL (single precision).\n");
+        m_forward = std::make_unique<OpenCLScheduler<float>>();
+#endif
+    }
+
+#else //!USE_OPENCL
     myprintf("Initializing CPU-only evaluation.\n");
     m_forward = std::make_unique<CPUPipe>();
+    use_selfcheck = false;
 #endif
 
     to_init.emplace_back(m_forward.get());
-
 #ifdef USE_OPENCL_SELFCHECK
-    if (!cfg_cpu_only) {
+    if (use_selfcheck) {
         m_forward_cpu = std::make_unique<CPUPipe>();
         to_init.emplace_back(m_forward_cpu.get());
     }
+#else
+    (void)use_selfcheck;
 #endif
 
     for (const auto& p : to_init) {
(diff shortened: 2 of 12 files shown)


src/Leela.cpp
```diff
@@ -90,6 +90,9 @@ static void parse_commandline(int argc, char *argv[]) {
                 "ID of the OpenCL device(s) to use (disables autodetection).")
         ("full-tuner", "Try harder to find an optimal OpenCL tuning.")
         ("tune-only", "Tune OpenCL only and then exit.")
+#ifdef USE_HALF
+        ("use-half", "Use half-precision OpenCL code.  Traades off some accuracy for higher performance")
```
gcp: Typo.

src/Network.cpp
```diff
@@ -345,26 +345,42 @@ void Network::initialize(int playouts, const std::string & weightsfile) {
...
+        use_selfcheck = false;
     } else {
-        myprintf("Initializing OpenCL.\n");
-        m_forward = std::make_unique<OpenCLScheduler>();
-    }
+#ifdef USE_HALF
+        if (cfg_use_half) {
+            myprintf("Initializing OpenCL (half precision).\n");
+            m_forward = std::make_unique<OpenCLScheduler<half_float::half>>();
+            use_selfcheck = false;
```
gcp: How badly would the self-check need to be relaxed for HALF to pass?

gcp: The reason I'm asking is that the comment says "please test before enabling it".

And the user is going to wonder: test what? :-)

ihavnoid: I did a bit experimenting and ran the whole thing overnight (with self-checking everything) and I recall that eventually failed even with 100% margins.  Most of the cases the error was less than 1% so it was okay.

I even moved the self-check to the final results (from the output of the forward() call) and it still did fail with 100% margins.  It didn't seem to yield something too problematic (e.g., policy net moving a probability from 0.05 to 0.1) which isn't something that the tree search cannot fix, though.

I guess if we still want the USE_HALF self-check we have to do something like 'N cases with average error of XXX%' but I am not sure what is the right value/way to do this.

gcp: The self-check has a catch that "rounds" all small values to zero (because the relative error gets very big on a very small value). Maybe it's just a matter of slightly pulling that up.

gcp: I think Leela Zero Chess has a probabilistic self-check, in that it tolerates the occasional failure. Not sure how that combines with us already doing the self check once every xxx nodes though. At some point you're not going to catch buggy drivers either any more.

ihavnoid: Technically what we should test... is the strength of the engine (which one yields better win rate - speed vs. NN accuracy) and that can differ quite a bit depending on which GPU the user uses.  The last time I tested, I figured out that high-bandwidth GPUs (e.g., Tesla P100 from Google Cloud) doesn't yield much additional performance hence we were sacrificing accuracy for nothing.  In those cases --use-half would be worthless.

Maybe all I can say for now is to delete the comment? :)

Ttl: One option would be to use KL divergence or some other measure to calculate the error in the self check. KL divergence doesn't care if the low probabilities are little off unlike the current method.

ihavnoid: Okay, what I will do is:
- If it fails 2 out of 5 most recent checks it will assert-fail.
- Regarding the error margin... I will test a bit to see what is acceptable.

gcp:
The travis setup has a mode where it compiles leela-zero:gpu-half, so that likely needs some changes. Or maybe it can just be removed outright.

ihavnoid:
Added update on self-check.  Barfs when:
- 3 out of last 10 checks failed
- failure criteria is > 20% error on anything higher tha 1/361.0
- Checks are performed on final net output instead of output of forward() call

src/Network.cpp
```diff
@@ -495,20 +510,47 @@ T relative_difference(const T a, const T b) {
...
+        selfcheck_fail = true;
+    }
+    if (err_winrate > relative_error) {
+        selfcheck_fail = true;
+    }
+
+    LOCK(m_selfcheck_mutex, selfcheck_lock);
+    if (selfcheck_fail) {
+        m_selfcheck_fails.push_back(true);
+        if (std::count(m_selfcheck_fails.begin(), m_selfcheck_fails.end(), true) >= max_failures) {
```
gcp: begin(m_selfcheck_fails)

src/Network.cpp
```diff
@@ -495,20 +510,47 @@ T relative_difference(const T a, const T b) {
...
+    const auto err_pass = relative_difference(data.policy_pass, ref.policy_pass);
+    const auto err_winrate = relative_difference(data.winrate, ref.winrate);
+    if (err_pass > relative_error) {
+        selfcheck_fail = true;
+    }
+    if (err_winrate > relative_error) {
+        selfcheck_fail = true;
+    }
+
+    LOCK(m_selfcheck_mutex, selfcheck_lock);
```
gcp: I think you might be able to get rid of this lock by making m_selfcheck_fails a bitfield and CAS-ing it. (Optional)

src/config.h
```diff
@@ -73,6 +73,17 @@
  */
 #ifndef USE_CPU_ONLY
 #define USE_OPENCL
+
+/*
+ * USE_HALF: Include the half-precision OpenCL implementation when building.
+ * Unlike previous versions, this does not enable half-precision by default,
+ * it just compiles half-precision support.  You have to use the command line
+ * argument --use-half explicitly to enable half-precision.
+ * half-precision OpenCL gains performance on some GPUs while losing some
```
gcp: Obvious question: should we check it during tuning and automatically enable if beneficial?

gcp: #ifdef USE_HALF
const auto TUNER_KERNEL = std::string("XgemmBatchedHalf");
#endif

This stuff no longer works correctly. (Github didn't let me comment on the actual line)

ihavnoid: Thinking about autodetecting between fp16 and fp32, I think the best I can possibly try is:
- Do a netbench on both fp16 and fp32, and use fp16 only if fp16 is at least 5% faster.  5% is a bit arbitrary, but last time we tried engine strength was roughly equivalent with same playouts, so 5% gives a bit of margin.
- If fp16 selfcheck fails revert to fp32 - the revert process may be tricky due to possibility of violating time constraint if it were to happen in the middle of a game, though.  Hopefully this doesn't happen too frequently that it loses by time.
- autodetect is default but also allow forcing fp16 or fp32
Any better ideas?

src/Network.cpp
ihavnoid: Thought about quickly, but it seems to be more trouble than necessary - we need to implement an array of bitfields and then implement a circular buffer to track last N pass/fails.  This lock is unlikely to be performance-critical anyway (only happens one out of 2000 evals).

Any better idea is welcome but I can't figure out how to make it simple :)

ihavnoid:
Made some commits.  The autodetect might take a bit more time, eventually we have to get there, but I am not sure if we want it as a separate PR or part of this PR.

src/Network.cpp
gcp: >only happens one out of 2000 evals

That's a good point and makes my idea moot.

For reference:
You don't need an array or circular buffer. You just add the latest result like this ```bitfield = (bitfield << 1 | result) & 0xA;``` i.e. just relying on the shift and mask. Checking failure then is popcount(bitfield) > 2. Because the bitfield can be an int32, it can be atomically CASed.

src/config.h
gcp: >If fp16 selfcheck fails revert to fp32 

Not sure we need to handle this case with the relaxed settings.

fell111:
If we do not have auto detect, we need to add "USE_HALF" option in autogtp too. Otherwise there will be no parameter passes into leelaz.

PR #1652, opened 2018-07-24
Add Validation support for specifying GTP commands for each binary (2)

Hersmunch:
Second attempt at #1459 but this time not dependent on Qt 5.6. Instead, each binary argument has to be preceded by "--" (otherwise all GTP commands are sent to all binaries).
Note though that I haven't actually tried building with Qt 5.3.
```
validation.exe -h
Usage: validation.exe [options] [-- binary [--gtp-command...] [-- binary [--gtp-
command...]]]

Options:
  -?, -h, --help                    Displays this help.
  -v, --version                     Displays version information.
  -g, --gamesNum <num>              Play 'gamesNum' games on one GPU at the
                                    same time.
  -u, --gpus <num>                  Index of the GPU to use for multiple GPUs
                                    support.
  -s, --sprt <lower:upper>          Set the SPRT hypothesis (default
                                    '0.0:35.0').
  -k, --keepSgf <output directory>  Save SGF files after each self-play game.
  -n, --network <filename>          Networks to use as players in competition
                                    mode (two are needed).
  -o, --options <opt_string>        Options for the binary given by -b (default
                                    "-g -v 3200 --noponder -t 1 -q -d -r 0 -w").

  -c, --gtp-command <command>       GTP command to send to the binary on
                                    startup (default "time_settings 0 1 0").
                                    Multiple commands are sent in the order they

                                    are specified.
                                    Commands apply to the preceeding binary or
                                    both if specified before all binaries.

Arguments:
  [-- binary [--gtp-command...] [-- binary [--gtp-command...]]] Binary to
                                                                execute for the
                                                                game (default
                                                                ./leelaz).
                                                                Only
                                                                --gtp-command
                                                                options are
                                                                parsed after a
                                                                binary is
                                                                specified
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ training/tf/venv
 leelaz-model*
 *.orig
 leelaz_opencl_tuning
+/validation/validation.pro.user
+/build-validation-Desktop_Qt_5_10_0_MSVC2017_64bit-Debug
+/build-validation-Desktop_Qt_5_10_0_MSVC2017_64bit-Release
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,4 @@ training/tf/venv
 leelaz-model*
 *.orig
 leelaz_opencl_tuning
-/validation/validation.pro.user
-/build-validation-Desktop_Qt_5_10_0_MSVC2017_64bit-Debug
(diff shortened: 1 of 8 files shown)


gcp:
@marcocalignano Can you have a look?

gcp:
Compilation is fine with Qt 5.5 at least.

PR #1653, opened 2018-07-24
Fix a bug that an SGF file/string cannot contain 2 consecutive moves of the same color

### LINKED ISSUE #1469: SGF parsing: "setup" stones can't be followed by a black move
Currently, SGF files with "setup" stones are parsed correctly only when they are treated as handicap games (meaning it's followed by a white move), for example:

```
(;GM[1]FF[4]CA[UTF-8]SZ[19]AB[dd];W[pp];B[pd])
```

Parsed as:

```
   a b c d e f g h j k l m n o p q r s t 
19 . . . . . . . . . . . . . . . . . . . 19
18 . . . . . . . . . . . . . . . . . . . 18
17 . . . . . . . . . . . . . . . . . . . 17
16 . . . X . . . . . + . . . . .(X). . . 16
15 . . . . . . . . . . . . . . . . . . . 15
14 . . . . . . . . . . . . . . . . . . . 14
13 . . . . . . . . . . . . . . . . . . . 13
12 . . . . . . . . . . . . . . . . . . . 12
11 . . . . . . . . . . . . . . . . . . . 11
10 . . . + . . . . . + . . . . . + . . . 10
 9 . . . . . . . . . . . . . . . . . . .  9
 8 . . . . . . . . . . . . . . . . . . .  8
 7 . . . . . . . . . . . . . . . . . . .  7
 6 . . . . . . . . . . . . . . . . . . .  6
 5 . . . . . . . . . . . . . . . . . . .  5
 4 . . . + . . . . . + . . . . . O . . .  4
 3 . . . . . . . . . . . . . . . . . . .  3
 2 . . . . . . . . . . . . . . . . . . .  2
 1 . . . . . . . . . . . . . . . . . . .  1
   a b c d e f g h j k l m n o p q r s t 
```

But if followed by a black move (not a handicap game, but a problem, etc), for example, let's swap the order of the two moves:

```
(;GM[1]FF[4]CA[UTF-8]SZ[19]AB[dd];B[pd];W[pp])
```

It's parsed as:

```
   a b c d e f g h j k l m n o p q r s t 
19 . . . . . . . . . . . . . . . . . . . 19
18 . . . . . . . . . . . . . . . . . . . 18
17 . . . . . . . . . . . . . . . . . . . 17
16 . . . X . . . . . + . . . . . + . . . 16
15 . . . . . . . . . . . . . . . . . . . 15
14 . . . . . . . . . . . . . . . . . . . 14
13 . . . . . . . . . . . . . . . . . . . 13
12 . . . . . . . . . . . . . . . . . . . 12
11 . . . . . . . . . . . . . . . . . . . 11
10 . . . + . . . . . + . . . . . + . . . 10
 9 . . . . . . . . . . . . . . . . . . .  9
 8 . . . . . . . . . . . . . . . . . . .  8
 7 . . . . . . . . . . . . . . . . . . .  7
 6 . . . . . . . . . . . . . . . . . . .  6
 5 . . . . . . . . . . . . . . . . . . .  5
 4 . . . + . . . . . + . . . . .(O). . .  4
 3 . . . . . . . . . . . . . . . . . . .  3
 2 . . . . . . . . . . . . . . . . . . .  2
 1 . . . . . . . . . . . . . . . . . . .  1
   a b c d e f g h j k l m n o p q r s t 
```

zhanzhenzhen:
Fixes #1469.

As I'm not a C++ developer, please check my modifications carefully to see if they are all correct. I've tested, but maybe there're coding style or other problems.

This bug will be triggered if:

- There're 2 or more consecutive moves of the same color; or
- There's `AW` and the first move is WHITE; or
- There's no `AW` but there's `AB`, and the first move is BLACK.

The reason is that SGFTree passes `get_to_move()` as the argument when calling `get_move` method, so if all moves are not "black, white, black, white...", then it won't be loaded correctly. (Here all handicap/setup stones can be considered as one move and the color can be considered as black if there's no `AW`, or white if there's `AW`.)

I've replaced 2 places of `get_move`. There's still one place left, but that's related to training so it's impossible to trigger this bug (but maybe better also change it?).

The `follow_mainline_state` modification is required. But during my test, I found `populate_states` modification takes no effect, that is, it seems also OK if unchanged.

Another thing I don't know why is that `std::make_pair(FastBoard::INVAL, 0)` works, but if modified to `std::make_pair(FastBoard::INVAL, SGFTree::EOT)` then it will report error when compiling.

### DIFF
--- a/src/SGFTree.cpp
+++ b/src/SGFTree.cpp
@@ -71,14 +71,14 @@ GameState SGFTree::follow_mainline_state(unsigned int movenum) {
     for (unsigned int i = 0; i <= movenum && link != nullptr; i++) {
         // root position has no associated move
         if (i != 0) {
-            int move = link->get_move(result.get_to_move());
-            if (move != SGFTree::EOT) {
-                if (move != FastBoard::PASS && move != FastBoard::EMPTY
-                    && result.board.get_square(move) != FastBoard::EMPTY) {
+            auto colored_move = link->get_colored_move();
+            if (colored_move.first != FastBoard::INVAL) {
+                if (colored_move.second != FastBoard::PASS && colored_move.second != FastBoard::EMPTY
+                    && result.board.get_square(colored_move.second) != FastBoard::EMPTY) {
                     // Fail loading
                     return result;
                 }
-                result.play_move(move);
+                result.play_move(colored_move.first, colored_move.second);
             }
         }
         link = link->get_child(0);
@@ -241,9 +241,9 @@ void SGFTree::populate_states(void) {
 
         // XXX: maybe move this to the recursive call
         // get move for side to move
-        int move = child_state.get_move(m_state.get_to_move());
-        if (move != EOT) {
-            child_state.apply_move(move);
+        auto colored_move = child_state.get_colored_move();
+        if (colored_move.first != FastBoard::INVAL) {
+            child_state.apply_move(colored_move.first, colored_move.second);
         }
 
         child_state.populate_states();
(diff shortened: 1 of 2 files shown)


gcp:
>Another thing I don't know why is that std::make_pair(FastBoard::INVAL, 0) works, but if modified to std::make_pair(FastBoard::INVAL, SGFTree::EOT) then it will report error when compiling.

Seems to work for me. I did some minor cleanups here: #1654.

gcp:
Actually, the compiler error was likely caused because this was missing:
https://github.com/gcp/leela-zero/commit/7524a2f86defcd459486348a869d3a03ba12419f#diff-44ba5da097788068a1c8def978c3293cR40

PR #1657, opened 2018-07-26
Implemented NN eval fp16/fp32 autodetect

ihavnoid:
Runs both precisions for 1 seconds, and if fp16 is faster than fp32 by more than 5%, fp16 is used.
removes --use-half, replaces it with --precision [auto|single|half] option, default auto

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -65,7 +65,7 @@ std::vector<int> cfg_gpus;
 bool cfg_sgemm_exhaustive;
 bool cfg_tune_only;
 #ifdef USE_HALF
-bool cfg_use_half;
+std::string cfg_precision;
 #endif
 #endif
 float cfg_puct;
@@ -105,7 +105,7 @@ void GTP::setup_default_parameters() {
     cfg_sgemm_exhaustive = false;
     cfg_tune_only = false;
 #ifdef USE_HALF
-    cfg_use_half = false;
+    cfg_precision = "auto";
 #endif
 #endif
     cfg_puct = 0.8f;
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -65,7 +65,7 @@ std::vector<int> cfg_gpus;
 bool cfg_sgemm_exhaustive;
 bool cfg_tune_only;
 #ifdef USE_HALF
-std::string cfg_precision;
+precision_t cfg_precision;
 #endif
 #endif
 float cfg_puct;
@@ -105,7 +105,7 @@ void GTP::setup_default_parameters() {
     cfg_sgemm_exhaustive = false;
     cfg_tune_only = false;
 #ifdef USE_HALF
-    cfg_precision = "auto";
+    cfg_precision = precision_t::AUTO;
 #endif
 #endif
     cfg_puct = 0.8f;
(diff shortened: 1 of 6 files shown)


src/GTP.cpp
```diff
@@ -65,7 +65,7 @@ std::vector<int> cfg_gpus;
 bool cfg_sgemm_exhaustive;
 bool cfg_tune_only;
 #ifdef USE_HALF
-bool cfg_use_half;
+std::string cfg_precision;
```
gcp: Use an enum like time management.

src/Network.cpp
```diff
@@ -424,6 +462,20 @@ void Network::initialize(int playouts, const std::string & weightsfile) {
...
 #endif
+
+#ifdef USE_HALF
+    if (fp16net != nullptr) {
+        auto score_fp32 = benchmark_time(100);
+        std::swap(fp16net, m_forward);
+        auto score_fp16 = benchmark_time(100);
+        if (score_fp32 * 1.05f > score_fp16) {
+            std::swap(fp16net, m_forward);
+            myprintf("Using OpenCL single precision - %.2f n/s vs. %.2f n/s half).\n", score_fp32, score_fp16);
```
gcp: Probably going to lead users to wonder why the faster option wasn't chosen.

Ttl:
Can't choosing of the precision done when tuning to avoid benchmarking it on every startup? The startup already takes quite long with a large network. If both options need to be tuned anyway for the benchmark couldn't USE_HALF just be added as a tunable variable in the tuner and the used option chosen based on the tuner results?

Is there any need to use USE_HALF define in non-OpenCL code anymore? I don't see a reason why this couldn't be compiled all the time.

USE_HALF comment in config.h needs to be updated.

ihavnoid:
@Ttl  I like your suggestion to move the whole thing in the tuner.  Unfortunately that's going to need some code restructuring because we need to move the tuner outside of the OpenCL interface and have a unified tuner for everything.  Plus, we probably need to measure the non-GEMM part of the OpenCL code to make a final call so I think the whole tuning code needs to be changed.


Either we can drop this patch, or keep this patch until we have a better alternative.  Opinions?

ihavnoid:
Addressed code review round 1

ihavnoid:
@Ttl  I see one more problem - when we are doing multi-GPUs.  Normally it isn't a problem at all, but if we use different combinations of GPUs some may end up better with fp16 while others may be better off fp32.  As you see, the testing here happens for the whole system (all GPUs) instead of a single GPU.  Are we going to allow mixed precision (some GPUs using fp16 while others using fp32)?  How are we going to record these systems on the tuner?  Not sure how to address these questions.  Ideas ideas ideas...

gcp:
If you do this in the tuner, you'll have to make it depend on the network blocks (it already depends on filters), because changing those can vastly change the speed difference that fp16 makes (I see this with the current network on a lower end machine here - without fp16 speed for 256x20 totally plummets, I assume due to memory issues).

Based on that, and that fp16 coeff storage is something that people can't really predict from their hardware specs (unlike fp16 compute support), I'm ok with benchmarking this in startup (for now).

PR #1660, opened 2018-07-26
Compress duplicate evaluation code.

TFiFiE:


src/UCTNode.h
```diff
@@ -67,7 +67,7 @@ class UCTNode {
     float get_policy() const;
     void set_policy(float policy);
     float get_eval(int tomove) const;
-    float get_pure_eval(int tomove) const;
+    float get_pure_eval(int tomove, int virtual_loss = 0) const;
```
gcp: That naming is so awkward. What about making get_internal_eval take virtual_loss, and then having get_eval and get_pure_eval just call that? (Removing the default argument too)

TFiFiE: I at first actually thought of renaming `get_pure_eval` to `get_raw_eval` with this change. Would that be acceptable?

gcp: I think that's also better yes. It avoids the conflict that the get_eval call asks for the "pure" eval but supplies the virtual loss which is obviously so as not to actually get the pure eval. "raw" isn't so bad because it's more obvious it needs to get modified/processed.

It's cherry picking of wording nuances but given that people have been confused by bad naming in the source, small tweaks might help.

PR #1661, opened 2018-07-27
Remove separate half precision kernel

Ttl:
Use preprocessor defines to make single kernel support both single precision
and half precision storage.


ihavnoid:
I like this commit - perhaps we can always use a completely different codebase if we want to have half precision specific BLAS optimizations.

Issue #1662, opened 2018-07-27
fp16 vs. fp32


ihavnoid:
This is to record experiment results on the strength of fp16 and fp32 - using net d351f06e (192x15)

1) When equal number of visits

 visits | fp16 wins | fp32 wins
------- | ------------ | --------------
| 500 | 18 | 22 |
| 2000 | 19 | 21 |
| 4000 | 20 | 20 |
| 8000 | 18 | 22 |

2) when 5% more visits are given to fp16

visits | fp16 wins | fp32 wins
------- | ------------ | --------------
| 525/500 | 19 | 21 |
| 2100/2000 | 22 | 18 |
| 4200/4000 | 24 | 16 |
| 8400/8000 | 26 | 14 |

This suggests that the fp16 inaccuracy can be overcome by ~5% more visits when visit counts are high.  On low visits, NN accuracy is still crucial.

Friday9i:
In line with @jkiliani: we would need more games to get a significant result, it seems to be within noise and the difference due to 5% additionnal visits is too big to be reasonable (hence it is noise).

How many games are needed? Not easy, "it depends" ; -).
If there is a difference (eg 40% vs 60%) in winrate between fp16 and fp32, the number of games needed depends on the amplitude of the difference: with 60/40, around 200 games should be enough but if it is only 55/45, we need at least around 500 games. If it is 52/48, we need at least 2000 games, a pain...

Should the number of visits impact the result of fp16 vs fp32 tests? I assume it should not. Hence, to get significant results, I'd be tempted to get a lot of games (eg 1000 games) and to play them with a relatively low playouts/visits (maybe 200, to limit the time to get these games).
And if you are motivated, you can test the same at 50 visits: if results are the same, it would confirm the result does not depend on visits.

Friday9i:
Cool (but maybe you could begin with 400 games with the same number of visits: if it is is equivalent, no need to test +5% ; -).

ihavnoid:
Ran it overnight, message didn't change a whole lot though.

playoff | fp16 wins | fp32 wins | winrate(fp16)
------- | ----------- | ----------- | -----
525/500 | 208 | 192 | 52
2100/2000 | 134 | 112 | 54
4200/4000 | 80 | 63 |  56
8400/8000 | 74 | 53 | 58

Weather is getting too hot, shutting down machines.

tterava:
[...]

PR #1669, opened 2018-07-29
Thread scalability improvements

ihavnoid:
- On OpenCLScheduler, don't use condvars which tends to be slow because of thread sleep/wake
- Instead, use spinlocks and just have enough contexts to avoid sleeping
- Allow more threads than the CPU physically has.  This is required in many multi-GPU setups
  with low core counts (e.g., quad-core non-hyperthread with 2 GPUs)

Addresses #1662


src/Leela.cpp
```diff
@@ -207,11 +207,7 @@ static void parse_commandline(int argc, char *argv[]) {
 
     if (!vm["threads"].defaulted()) {
         auto num_threads = vm["threads"].as<int>();
-        if (num_threads > cfg_max_threads) {
-            myprintf("Clamping threads to maximum = %d\n", cfg_max_threads);
-        } else if (num_threads != cfg_num_threads) {
-            cfg_num_threads = num_threads;
-        }
+        cfg_num_threads = num_threads;
```
gcp: This needs to respect cfg_max_threads somehow or you risk crashing in OpenBLAS. Yanking up cfg_max_threads to ```std::max(1, std::min(SMP::get_num_cpus() * gpus, MAX_CPUS)); maybe?```

ihavnoid: No objection.  I don't think I am qualified enough to make a call here since I am not aware of how OpenBLAS crashes.  Will push an update soon.

ihavnoid: FYI will need a bit of changes (not too complicated though) simply because cfg_max_threads were originally calculated before we parsed the GPU options.

ihavnoid:
Pushed a change and resolved merge conflicts.

src/Leela.cpp
```diff
@@ -205,13 +205,45 @@ static void parse_commandline(int argc, char *argv[]) {
         cfg_gtp_mode = true;
     }
 
+#ifdef USE_OPENCL
+    if (vm.count("gpu")) {
+        cfg_gpus = vm["gpu"].as<std::vector<int> >();
+        // if we use OpenCL, we probably need more threads for the max so that we can saturate the GPU.
+        cfg_max_threads *= cfg_gpus.size();
```
gcp: This needs to respect MAX_CPUS.

The issue is that the OpenBLAS code has a rather weird threading design, and needs the maximum amount of thread to be set at compile time. It's 64 or 128 for common Linux distributions.

I am a bit annoying that forcing it to not thread (which we do in Network) does not avoid this problem, but from experience it does seem to crash if you exceed the built-in limit.

ihavnoid: Okay.  Pushed a fix that does it.

PR #1672, opened 2018-07-30
Half precision compute support

Ttl:
Use half precision computation on cards that support it when half precision is selected.

On Intel iGPU netbench goes from 34 n/s to 44 n/s when using fp16 computation. It's more inaccurate than just fp16 storage and requires relaxing the tuner error budget. However during the play network output is still very close to the fp32 output so I don't think the precision is a problem. Self-check wasn't an issue during self-play.

I would be interested if someone with high-end GPU with fp16 support could test this (AMD Vega, Nvidia V100 or P100, others?).


ihavnoid:
Awesome.  Unfortunately last time I checked, P100 didn't seem to support fp16 compute on OpenCL, at least the GPU drivers on Google cloud didnt. I tried to do fp16 compute support on Google cloud and gave up.

gcp:
Looks good, but unfortunately AMD cards like the RX560-RX580 will advertise fp16 support even though they only can do fp16 register storage and not fp16 compute.

The speedup might not be worth the precision loss on those. I have such a machine and I'll do some basic testing. If this makes them worse we'll need a command line flag (again) or extend the automatic tuning even more.

gcp:
Given that real fp16 support might be rarer than "fake" fp16 support, I'm OK with only allowing it to be enabled after a command line flag.

(If fake fp16 is no worse in practice, this is moot)

gcp:
Confirmed that on AMD Polaris cards the fp16 option will be selected even though the speedup is entirely negligible. I'm running a match to see if the extra error makes things overall worse, but that will take a while.

gcp:
88 wins, 87 losses
175 games played.

Safe to say, any precision loss from fp16 compute is small enough we don't need to care much about it.

gcp:
@Ttl if I see correctly this just always enables if supported, so I changed the comment.

One thing I wondered about: the card I tested stores intermediate results at fp16 but likely does the actual math in fp32. This might be the same for yours. I wonder if something with "real" fp16 support will see different precision losses.

AncalagonX:
A user with an AMD laptop GPU (*not* cpu/onboard GPU) could not run the /next/ build compiled immediately after this PR #1672 "Half precision compute support". 

When this user forced `--precision single`, the same leelaz.exe ran properly as expected.

The cause is presumably a failure in the full/half precision auto-detect routine. It is likely related to issue #1690. I'll also ping issue #1689 for this.

Here are his full logs:

```./leelaz.exe -w network
Using 2 thread(s).
RNG seed: 7924010773806070189
Leela Zero 0.15  Copyright (C) 2017-2018  Gian-Carlo Pascutto and contributors
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; see the COPYING file for details.
 
Detecting residual layers...v1...192 channels...15 blocks.
Initializing OpenCL (autodetect precision).
Detected 2 OpenCL platforms.
Platform version: OpenCL 1.2
Platform profile: FULL_PROFILE
Platform name:    Intel(R) OpenCL
Platform vendor:  Intel(R) Corporation
Device ID:     0
Device name:   Intel(R) Core(TM) i5-3230M CPU @ 2.60GHz
Device type:   CPU
Device vendor: Intel(R) Corporation
Device driver: 3.0.1.10891
Device speed:  2600 MHz
Device cores:  4 CU
Device score:  512
Device ID:     1
Device name:   Intel(R) HD Graphics 4000
Device type:   GPU
Device vendor: Intel(R) Corporation
Device driver: 10.18.10.4358
Device speed:  1100 MHz
Device cores:  16 CU
Device score:  612
Platform version: OpenCL 2.0 AMD-APP (1800.11)
Platform profile: FULL_PROFILE
Platform name:    AMD Accelerated Parallel Processing
Platform vendor:  Advanced Micro Devices, Inc.
Device ID:     2
Device name:   Turks
Device type:   GPU
Device vendor: Advanced Micro Devices, Inc.
Device driver: 1800.11 (VM)
Device speed:  500 MHz
Device cores:  6 CU
Device score:  1120
Device ID:     3
Device name:   Intel(R) Core(TM) i5-3230M CPU @ 2.60GHz
Device type:   CPU
Device vendor: GenuineIntel
Device driver: 1800.11 (sse2,avx)
Device speed:  2594 MHz
Device cores:  4 CU
Device score:  520
Selected platform: AMD Accelerated Parallel Processing
Selected device: Turks
with OpenCL 2.0 capability.
Half precision compute support: NO
 
Started OpenCL SGEMM tuner.
Will try 290 valid configurations.
Failed to find a working configuration.
Check your OpenCL drivers.
```

Ttl:
This PR shouldn't have any effect if fp16 compute is not supported.

I guess the issue is with GPU drivers mishandling the half precision storage somehow? Anyhow it would be nice if automatic choosing of fp16/fp32 could handle this kind of case where fp16 doesn't work at all.

PR #1692, opened 2018-08-03
Include Eigen as BLAS replacement.

gcp:
This vendors in Eigen 3.3.5 as a default matrix/vector library.
This has a load of advantages:

* It can be used as a replacement for a cBLAS library when it is
  not available, cannot be found, or is outdated compared to the
  compiler or CPU.
* Because Eigen is header only, it significantly eases the build
  prequisite requirements.
* The Eigen code paths are much more readable from a mathematical
  perspective.
* Eigen can optimize more heavily for known matrix sizes. The
  current code doesn't yet take advantage of this, though.

The downsides:

* Eigen might be a bit slower than other BLAS libraries. (Nevertheless,
  on my system it is faster than OpenBLAS)
* Binaries built with Eigen are optimized for the CPU it was compiled
  on and don't port as well to other CPUs. So you need seperate binaries
  for wider client support.


gcp:
This still need some changes in CMake (I could use help!). If no BLAS library is found, USE_BLAS should be unset. If a BLAS library was found, USE_BLAS should be set.

I also am not 100% sure whether to vendor in the relevant part of the Eigen library, or to use a submodule here.

gcp:
Another option is to trim down the vendored library a bit further. I'm sure entire subdirs can be removed here.

ihavnoid:
Awesome.  If this works and we eventually get rid of openblas, does it mean that we can remove the MAX_CPU constraint?

ihavnoid:
Plus, I also tested it on my machine (i8-8700, running CPU-only) and it seems that it's at least 20% faster compared to the Haswell-optimized OpenBLAS library that I have on my machine.  I wonder how it behaves on other machines, especially Ryzens?

roy7:
I believe @gcp has at least one Ryzen system so perhaps he can let us know. :)

MartinDevelopment:
I did 2 tests on a Ryzen 1600X and when from 27 n/s to 31 n/s and from 25 n/s to 29 n/s.
So, using this PR I get around a 15-16% speedup.

tterava:
On a Ryzen 7 i get about 10% speedup using this PR.

gcp:
>Awesome. If this works and we eventually get rid of openblas, does it mean that we can remove the MAX_CPU constraint?

You could set it to 9999 if USE_BLAS is disabled, because yes, Eigen does avoid all the OpenBLAS threading problems.

Getting rid of OpenBLAS is trickier when providing binaries, which is the situation releases and things like distro packages belong to. You may need separate binaries for x87, SSE, AVX, AVX512, etc...

I guess in theory AutoGTP could detect which binary to run, but that doesn't solve the problem for the distros...

gcp:
I involuntarily ended up testing this on a Penryn, and there it drops the speed from 3 nodes per second to 2 nodes per second. So it's not faster everywhere. Though, I'd say, probably faster on most machines people still care about.

Ttl:
Seems to be slightly slower on Intel i5-2500k:

```
next:
  100 evaluations in 12.79 seconds -> 7 n/s

include_eigen:
  100 evaluations in 15.69 seconds -> 6 n/s
```

kaorahi:
This PR is impressively useful for me.
"include_eigen" is 1.5 times faster than "next" in my environment (note PC without GPU, Debian testing (buster), openblas).
Its speed is comparable to MKL.

BTW, I need the option "-I." to compile include_eigen as follows.

    $ git checkout origin/next
    HEAD is now at bd36100 Don't refer to stone locations as "squares".
    $ make CXXFLAGS='-DUSE_CPU_ONLY' > /dev/null
    $ ./leelaz -g --benchmark -t 4 -w ~/157_192x15/d351f06e446ba10697bfd2977b4be52c3de148032865eaaf9efc9796aea95a0c.gz |& grep 'n/s'
    3203 visits, 1098074 nodes, 3202 playouts, 34 n/s
    
    $ git checkout origin/include_eigen
    Previous HEAD position was bd36100 Don't refer to stone locations as "squares".
    HEAD is now at 82ea37d Trim vendored Eigen library.
    $ make CXXFLAGS='-I. -DUSE_CPU_ONLY' > /dev/null
    $ ./leelaz -g --benchmark -t 4 -w ~/157_192x15/d351f06e446ba10697bfd2977b4be52c3de148032865eaaf9efc9796aea95a0c.gz |& grep 'n/s'
    3203 visits, 1098644 nodes, 3202 playouts, 50 n/s
    
    $ uname -a
    Linux xxx 4.17.0-1-amd64 #1 SMP Debian 4.17.8-1 (2018-07-20) x86_64 GNU/Linux
    $ cat /proc/cpuinfo | egrep 'model name|cache size|cpu cores' | sort -ru
    model name      : Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
    cpu cores       : 4
    cache size      : 8192 KB
    $ update-alternatives --display libblas.so.3-x86_64-linux-gnu | grep current
      link currently points to /usr/lib/x86_64-linux-gnu/openblas/libblas.so.3

gcp:
What's the error you get when -I. isn't there? Or is this with the Makefile? That hasn't been updated yet.

I need some time to work with CMake and make it default to using this, but allow a flag to force BLAS use. (It seems faster on most new systems, so I'd default this on. Would also work better for highly threaded use)

kaorahi:
"make CXXFLAGS='-DUSE_CPU_ONLY'" says "Eigen/Core: No such file or directory".
"make CXXFLAGS='-I. -DUSE_CPU_ONLY'" is ok.

    $ git checkout origin/include_eigen
    HEAD is now at 82ea37d Trim vendored Eigen library.
    
    $ make CXXFLAGS='-DUSE_CPU_ONLY'
    Detected OS: Linux
    make CC=gcc CXX=g++ \
            CXXFLAGS='-DUSE_CPU_ONLY -Wall -Wextra -Wno-ignored-attributes -pipe -O3 -g -ffast-math -flto -march=native -std=c++14 -DNDEBUG'  \
            LDFLAGS=' -flto -g' \
            leelaz
    make[1]: Entering directory '/home/xxx/leelaz_src/src'
    g++ -DUSE_CPU_ONLY -Wall -Wextra -Wno-ignored-attributes -pipe -O3 -g -ffast-math -flto -march=native -std=c++14 -DNDEBUG -MD -MP -c -o Network.o Network.cpp
    Network.cpp:34:10: fatal error: Eigen/Core: No such file or directory
     #include <Eigen/Core>
              ^~~~~~~~~~~~
    compilation terminated.
    make[1]: *** [Makefile:61: Network.o] Error 1
    make[1]: Leaving directory '/home/xxx/leelaz_src/src'
    make: *** [Makefile:5: default] Error 2
    
    $ make CXXFLAGS='-I. -DUSE_CPU_ONLY'
    ...
    g++ -flto -g -o leelaz Network.o FullBoard.o KoState.o Training.o TimeControl.o UCTSearch.o GameState.o Leela.o SGFParser.o Timing.o Utils.o FastBoard.o SGFTree.o Zobrist.o FastState.o GTP.o Random.o SMP.o UCTNode.o UCTNodePointer.o UCTNodeRoot.o OpenCL.o OpenCLScheduler.o NNCache.o Tuner.o CPUPipe.o  -lboost_program_options -lpthread -lz -lopenblas -lOpenCL
    make[1]: Leaving directory '/home/xxx/leelaz_src/src'
    $ ls leelaz
    leelaz*

gcp:
I guess the code should say ```#include "Eigen/Core"``` not ```#include <Eigen/Core>```

kaorahi:
Yes. "-I." is not needed for ```#include "Eigen/Core"```.

    $ git diff -U0
    diff --git a/src/CPUPipe.cpp b/src/CPUPipe.cpp
    index a52e30c..047e5b4 100644
    --- a/src/CPUPipe.cpp
    +++ b/src/CPUPipe.cpp
    @@ -31 +31 @@
    -#include <Eigen/Core>
    +#include "Eigen/Core"
    diff --git a/src/Network.cpp b/src/Network.cpp
    index eb17ab5..8885d30 100644
    --- a/src/Network.cpp
    +++ b/src/Network.cpp
    @@ -34 +34 @@
    -#include <Eigen/Core>
    +#include "Eigen/Core"
    diff --git a/src/Tuner.cpp b/src/Tuner.cpp
    index 2c96ef7..0355022 100644
    --- a/src/Tuner.cpp
    +++ b/src/Tuner.cpp
    @@ -48 +48 @@
    -#include <Eigen/Core>
    +#include "Eigen/Core"
    
    $ make CXXFLAGS='-DUSE_CPU_ONLY'
    ...

infinity0:
Why is it necessary to vendor in eigen? It's already available in Ubuntu/Debian as `libeigen3-dev`, version 3.3.5 was just added to Debian unstable a few days ago.

gcp:
Because we support Windows too.

Maybe a git submodule is better?

infinity0:
Submodule sounds good yes, like what's done with gtest at the moment. Those get excluded from the `git archive` tarball from the releases page and is easier to deal with than manually excluding vendored stuff.

gcp:
Switched to a submodule, rebased.

Now just adjusting CMake stuff, and probably the existing MSVC projects.

src/Tuner.cpp
```diff
@@ -78,16 +91,10 @@ static void sgemmBatched_ref(const std::vector<net_t>& a,
         auto offset_v = batch * n * k;
         auto offset_m = batch * m * n;
 
-        // Calculates transpose(tranpose(A) * B)
```
gcp: @Ttl I believe this comment (from the existing sources) is actually incorrect?

I get correct working if I put C = A * B_T, or C = (B * A_T)^T, but that's not the same as C = (A_T * B)^T.

gcp: @Ttl BTW, the CPU version of the winograd transforms might benefit from using Eigen in the matrix math. I find that it certainly becomes more readable!

Ttl: It's row major vs column major issue. The column major C = A * B^T probably does makes more sense.

gcp:
I'm a bit confused why Travis only ran the last test-blas now, and is not showing the other build steps.

gcp:
```
$ {:script=>nil}
{:script=: command not found
The command "{:script=>nil}" exited with 127
```

Clearly my understanding of Travis syntax is 0.

gcp:
This will probably break the MSVC project files, who'll need an update to their include dirs.

ChinChangYang:
It is possible to test MSVC build by appveyor. I have made one in my local branch. See below for example:

version: '{build}'
configuration: Release
platform: x64
environment:
  matrix:
  - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
skip_commits:
  files:
    - '**/*.md'
    - '**/.gitignore'
    - scripts/
    - training/
    - AUTHORS
    - COPYING
install:
- cmd: nuget restore msvc\VS2017 -PackagesDirectory msvc\packages
cache: msvc\packages -> appveyor.yml
build_script:
- cmd: >-
 
    msbuild /t:build msvc/VS2017/leela-zero.vcxproj

TFiFiE:
I really don't like the unnecessary submodule dependency in Linux (see #1881 and #1883). Can't we just add the `libeigen3-dev` package to the Ubuntu instructions and change `src/Makefile` to include `/usr/include/eigen3`?

gcp:
>It is possible to test MSVC build by appveyor. 

It's already included in this pull.

gcp:
>Can't we just add the libeigen3-dev package to the Ubuntu instructions

I don't see how a submodule that just works on every platform isn't strictly massively better than that.

ChinChangYang:
@gcp In this pull, appveyor helps us building leelaz by cmake. That is totally different to what users build leelaz by double-clicking msvc/*.sln. The closest way to test msvc build is running the msbuild command. I have been working on this. It will be able to test if *.vcxproj is correctly updated.

gcp:
Ah, I see!

That would be useful. 

It would've been nice to retire the .sln files and rely on the CMake generated ones generally, but this would complicate things for users too much because CMake probably can't do the automated package installation (without a lot of extra work?).

ChinChangYang:
@gcp I will make a new pull request to auto-build leelaz by cmake and msbuild in the appveyor.yml once when it is ready.

infinity0:
It's easy enough to ignore the submodule, I assume cmake should find the system version if it exists and the user didn't `init` the submodule.

Or just `apt-get install leela-zero`, that's why I packaged it in the first place.

PR #1703, opened 2018-08-07
Allow unlimited thinking during opponent time

Cabu:
Add a new command line option --deepponder to allow unlimited thinking during opponent time even when --playouts or --visits are set.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -47,6 +47,7 @@ using namespace Utils;
 // Configuration flags
 bool cfg_gtp_mode;
 bool cfg_allow_pondering;
+bool cfg_deep_ponder;
 int cfg_num_threads;
 int cfg_max_threads;
 int cfg_max_playouts;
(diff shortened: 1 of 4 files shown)


AncalagonX:
Is `--deepponder` the best name for this option? It sounds like it has a functional purpose beyond what it actually does—which is removing time/playout/visit limits for pondering. While reading another thread about this before I saw this PR, I myself was genuinely unclear on what "deep ponder" was referring to.

Since we're creating a new command-line option, we should ensure it is both clear and concise as to its purpose—for example `--unlimitedponder`.

Marcin1960:
@AncalagonX "One example would be calling it --unlimitedponder"

Or do the reverse, make unlimited into default, and provide option --limit-pondering.

Cabu:
I have taken deep ponder not having a better name. Unlimited ponder is good for me too, I really don't care about the name of the option. I don't really like changing the default, because most probably some scripts/programs are working based on these defaults. Is there is a consensus about the naming of the option I will be pleased to make the change.

gcp:
Limiting playouts or visits and then doing an unlimited amount of them while pondering has no sensible use case.

Issue #1702, opened 2018-08-07
Request option to limit playouts without restricting/disabling pondering


evdwerf:
On GridMaster many people like to play without clock, but this does not work well with Leela zero. Either the engine thinks very long or, if I limit the number of visits, it becomes weaker than necessary. A good solution would be to limit the number of playouts/visits when leelaz is to move (so that it plays fast), but allow unlimited pondering when the user is thinking (so that it plays as strong as possible).

Unfortunately leelaz does not support this and even considers such a setting nonsensical! 
```
$ ./leelaz --weights 64x5.gz -g --playouts 100
Using 8 thread(s).
RNG seed: 8625826165390380976
Nonsensical options: Playouts are restricted but thinking on the opponent's time is still allowed. Add --noponder if you want a weakened engine.
```

Would it be possible to remove this restriction in the next release?

roy7:
I think for a while --visits was allowed with pondering but I think maybe it was removed. I've wanted a feature like that myself in the past, for the same reason. Use opponent time for more strength but move fast otherwise. :)

Cabu:
--visit is allowed and is not restricted like --playouts, but it limit pondering too. There is no possibility to do what evdwerf ask and my patch to implement what I called at the time "deep pondering" has been rejected.

evdwerf:
The deepponder proposal probably works, but I don't think we really need an
additional command line argument. The way I do it in my program is to
simply configure max playouts independent of pondering and simply not check
max playouts when pondering.

Cabu:
I prefer the command line option because some people whant it and some other doesn't want it and it leave the choice to them to have infinite pondering or not.
I am in favor of command line because I am a script junky and tools that doesn't accept commands via the command line are some kind of pain in the *ss to automate. But if it is really a problem, perhaps we could do it as a GTP command?

evdwerf:
I'm not arguing against the command line option. I'm saying we don't really
need an *extra* command line option. If people don't want pondering they
can use --noponder (which they are forced to do now anyway if they specify
--playouts). For the behavior you call 'deepponder' we could just omit
--noponder (if leelaz would allow it), but it's fine with me either way.
Even a gtp init command would be fine; GridMaster can handle any of that.
The only problem is getting leelaz to accept the desired settings.

Cabu:
[...]

PR #1708, opened 2018-08-08
Allow pondering to be unrestricted by playout & visit caps.

TFiFiE:
An alternative to fulfilling #1702 that differs from #1703 by creating a generalized ponder option.


AncalagonX:
There are prominent use cases where people would want to limit playouts or visits while leaving pondering enabled, and it has long been a requested feature with little or no technical side effects.

Ponder should be enabled for both limited playouts and limits visits.

The only wrinkle is whether a) pondering should respect the playout or visit cap, or b) be allowed to run unchecked until it becomes LZ's turn. These two options should be a configurable toggle, but that can be dealt with in a future PR.

gcp:
See the discussion. Most of all this needs a sensible use case, not pull requests.

Issue #1709, opened 2018-08-08
"Half-precision compute support" auto-detection fails on AMD laptop GPU (dedicated, not CPU-integrated)


AncalagonX:
A user with an AMD laptop GPU (*not* cpu/onboard GPU) could not run the /next/ build compiled immediately after this PR #1672 "Half precision compute support". 

**When this user forced `--precision single`, the very same leelaz.exe ran properly as expected.** Below, `Device ID 2` is the user's GPU which is properly selected but still leads to a crash during LZ's startup.

I presume the cause is a failure in the full/half precision auto-detect routine, but can't be certain. It is possibly related to issues #1701 and #1690 (I'll also ping issue #1689 for this). I already posted a duplicate of this report on PR #1672, but now realize a separate issue thread should be started for it.

Here are this user's full logs, and he is happy to provide further information and testing:

```./leelaz.exe -w network
Using 2 thread(s).
RNG seed: 7924010773806070189
Leela Zero 0.15  Copyright (C) 2017-2018  Gian-Carlo Pascutto and contributors
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; see the COPYING file for details.
 
Detecting residual layers...v1...192 channels...15 blocks.
Initializing OpenCL (autodetect precision).
Detected 2 OpenCL platforms.
Platform version: OpenCL 1.2
Platform profile: FULL_PROFILE
Platform name:    Intel(R) OpenCL
Platform vendor:  Intel(R) Corporation
Device ID:     0
Device name:   Intel(R) Core(TM) i5-3230M CPU @ 2.60GHz
Device type:   CPU
Device vendor: Intel(R) Corporation
Device driver: 3.0.1.10891
Device speed:  2600 MHz
Device cores:  4 CU
Device score:  512
Device ID:     1
Device name:   Intel(R) HD Graphics 4000
Device type:   GPU
Device vendor: Intel(R) Corporation
Device driver: 10.18.10.4358
Device speed:  1100 MHz
Device cores:  16 CU
Device score:  612
Platform version: OpenCL 2.0 AMD-APP (1800.11)
Platform profile: FULL_PROFILE
Platform name:    AMD Accelerated Parallel Processing
Platform vendor:  Advanced Micro Devices, Inc.
Device ID:     2
Device name:   Turks
Device type:   GPU
Device vendor: Advanced Micro Devices, Inc.
Device driver: 1800.11 (VM)
[...]

Issue #1689, opened 2018-08-03
Need testers on real fp16 cards


gcp:
I just merged autodetection + use of fp16 support:
https://github.com/gcp/leela-zero/pull/1672

On a card that has *real* fp16 support, this greatly increases performance. The question is about precision->strength at fixed playouts. I have cards with partial fp16 support, but no real ones, like P100/V100 or AMD Vega.

No-one responded in the original pull request, so I'll ask here again: if someone with such a card could verify the strength loss (if any!) at fixed visits or playouts, that would be useful for us.

bood:
Does Titan V has real fp16 support too? if yes, I can try this on a friend's machine.

SebastienFaurite:
It seems that it does support fp16 and the performance should double. 
https://devtalk.nvidia.com/default/topic/1027346/titan-v-fp16-performance/

fishcu:
Does the Titan XP have real fp16 support? if yes, I can try this. Would need some instructions on what to test exactly (scripts to run, etc.)

akdtg:
The performance of my Intel HD Graphics 620 increases from 33 n/s to 42 n/s.
But how to verify the strength loss?

gcp:
>The performance of my Intel HD Graphics 620 increases from 33 n/s to 42 n/s.

@Ttl tested this and it was fine, but I am not sure if Intel HD does the compute in 16-bit or if it is just freeing up registers or something.

Ttl:
I'm not 100% sure either what the Intel GPUs do, but unlike AMD GPU it is faster with fp16 compute than with fp16 storage and some sources list that half precision should have twice as high GFLOPs (https://www.techpowerup.com/gpudb/2911/hd-graphics-620-mobile). CLBlast SGEMM tuner was also about 50% faster with half precision.

I can run some fp16 vs fp32 games on my laptop this weekend but it would be still good to also have tests with high-end GPUs.

akdtg:
Well, this fp16 build works very unstable on my PC.
Before this merge to "next" branch, the typical rate of the self-play game was about 35 seconds per move.

With f1p16 build, the 1st game was played with 52971 ms/move, the 2nd one was aborted due to 'Error in OpenCL calculation: Update your GPU drivers or reduce the amount of games played simultaneously.' error, and the 3rd game was played with 28448 ms/move.

roy7:
Half precision compute support: NO

[...]

PR #1710, opened 2018-08-08
Tuner fixes

Ttl:
Fix error calculation (Missing batch_size divider).

Better error reporting when no working configuration could be found.

Change reference data to have less rounding errors with half precision.

Replace BLAS reference SGEMM with custom code that gives transposed output like the OpenCL SGEMM.

Fixes issue with too large error when using fp16 compute and 256 channel network: https://github.com/gcp/leela-zero/issues/1689#issuecomment-411060714.

### DIFF
--- a/src/Tuner.cpp
+++ b/src/Tuner.cpp
@@ -35,16 +35,6 @@
 #include "Utils.h"
 #include "Random.h"
 
-#ifdef __APPLE__
-#include <Accelerate/Accelerate.h>
-#endif
-#ifdef USE_MKL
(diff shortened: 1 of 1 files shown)


roy7:
If we don't have one already should there be a fallback to fp32 if no working config can be found for fp16?

Ttl:
It's not implemented yet, but should probably be. It would fix #1709.

AncalagonX:
The user from #1709 is available for further testing if needed. Let me know (our Discord would be fastest).

Issue #1707, opened 2018-08-08
UCTNode::uct_select_child bottlenecking on heavily threaded cases?


ihavnoid:
I was working on a distributed version of leela-zero (nothing fancy, just offloading NN evaluations to remote servers).  The problem I started to see when scaling up to large number of threads was that UCTNode::uct_select_child starts to become a bottleneck, and CPU usage soars when leela-zero seems to be evaluating through a long chain of obvious moves (all threads searching through a narrow passage).  Profiling shows that it is wasting CPU time on the...… spinlocks of uct_select_child.

Any idea how to improve this?  I tried to create a slightly different version with read-write-locks, seemed to be better, but no idea if this is the right thing to do.  Any comments?

Eddh:
Hmm. Maybe a dumb idea, but what if each node remembers the child that was selected last time it was selected, and also the action value of the second highest node. With some math that I'm too lazy to do right now, you could compute if the second node has any chance of surpassing the first one with the next visit, and if not, just go straight to the last selected child. It would not change the mutex thing, but the accelerated speed might help (except if the mutex lock is actually the longest thing in the function, idk about that).

gcp:
Are there any other cases besides re-expansion in which m_children can change underneath it? Because that's the thing the lock protects, IIRC...

ihavnoid:
> Are there any other cases besides re-expansion in which m_children can change underneath it? Because that's the thing the lock protects, IIRC...

I see there is an inflate() call that needs protection... but maybe a bit of experimenting (with assertions) might work.  I thought there was a branch that tried to make the tree smaller on unprofitable situations but it seems that  the branch didn't make it to the next branch...?

ihavnoid:
I put a PR which changes UCTNode's mutex into a read-write lock.  Seems to help the problem, at least in a general fashion.  I was a bit hesitant to lift the lock as @gcp suggested, as subtle changes in the future might break the code if it makes the lock necessary again.

ihavnoid:
#1744 has a 2nd attempt for this (don't lock UCTNode when node is fully expanded)

PR #1725, opened 2018-08-12
UCTNode : change lock to read-write-lock

ihavnoid:
uct_select_child is now protected with read-write-lock so that it does not bottleneck when
there are a lot of threads working on a narrow search path.  Improves GPU utilization on
massively parallel cases with lots of threads.

Address #1707 
May address #1554 
>He also reported that current multi-GPU support can only max out up to 5x1080Ti (~98% usage); with 10x1080Ti the GPU usage plunges below 50%.


bubblesld:
I did some test when the 40b training was writing the model (not using gpu). I could not test the case of new version on 40b (probably lacks of gpu memory. the new version of leelaz consumes more memory, and the training already occupied most of gpu memory)

It is not a precise test. From my observation, there is no clear improvement in gpu usage for the 20b weight. And the gpu usage for the 40b weight with the current leelaz is already high.

20b weight using current leelaz 

20b_ori

20b weight using new leelaz

20b_new

40b weight using current leelaz

40b_ori

ihavnoid:
@bubblesld this addresses some atypical cases (not necessarily corner case) so probably overall utilization should be fine even before the patch.  In my experience the utilization drop happens around the mid game.  Have you seen the utilization change over time?

bubblesld:
@ihavnoid I have only observed a few minutes. From my past experience, the gpu usage drops only between moves. I will not test it longer now since I do not want to affect the main training.

gcp:
I've been thinking about this, but I am inclined not to take it. There's 2 reasons:

a) This is non-trivial and it's going to take quite some effort to ensure there's no subtle bugs in a hand rolled rwlock implementation (we already had some).

b) AFAIK the only writer that the lock is protecting against is the node expansion. So what this patch really achieves is that node selection becomes non blocking (after expansion, but that obviously needs to happen first). So I think we don't need read locks at all, except for the edge case of low memory and partially expanded nodes. Instead, we can just block on the node not being expanded and after that the lock isn't needed at all?

This brings us to c) being blocked on the expansion is very bad, so perhaps we can check for this in the node selection one level up: any node that is being expanded should get a huge malus in the node winrate so we don't end up blocked by selecting it with a 2nd thread.

Am I missing anything here?

The combination of both tricks should remove almost all blocking in the MCTS loop.

Issue #1733, opened 2018-08-15
Quantize network files


gcp:
Won't fundamentally solve #1731 but a simple and fully backwards compatible optimization:

We already know that using fp16 doesn't really affect the network strength. This means we should be able to write out the network files with much less significant digits than we are now, and maybe cut the size of everything almost in half.

Given the experience with things like Bfloat16, it may even be that less digits than in "half" format may be needed, as long as the comma is in the right spot (which it is given that we're text based).

This could be nicely tested by quantizing some existing networks and scheduling matches between them. It would be interesting to see how far we can go before the precision loss is really noticeable.

l1t1:
write out the network files into ieee 754 float can save half space than double, but write as plain text cannot.

gcp:
Writing the files in binary (be it half, double or Bfloat16) isn't supported by the clients nor the training code nor the server and needs a network format update. 

We already write out files with float precision, I'm saying we can use even less than that.

ihavnoid:
Another hillarious but effective trick is to replace all instances of 0.xxx to .xxx, e.g., print .372 instead of 0.372 - most numbers are less than 1.000 so this should save quite a bit of size.

l1t1:
compress to gz .372 is same as 0.372

[nozero.txt.gz](https://github.com/gcp/leela-zero/files/2289827/nozero.txt.gz)
[zero.txt.gz](https://github.com/gcp/leela-zero/files/2289828/zero.txt.gz)
[nozero.txt](https://github.com/gcp/leela-zero/files/2289829/nozero.txt)
[zero.txt](https://github.com/gcp/leela-zero/files/2289830/zero.txt)

date| time|size|file
-|-|-|-
2018-08-15 | 17:35            |   174 |nozero.txt
2018-08-15 | 17:36              |  60 |nozero.txt.gz
2018-08-15 | 17:35             |  203 |zero.txt
2018-08-15 | 17:36              |  59 |zero.txt.gz

gjm11:
Leela Nozero. (Sorry.)

gcp:
>So even under very aggressive compression settings, simply removing the 0 is a net gain.

It's still less things that have to be encoded as being present, makes sense.

gcp:
I did a test with:

```
#!/usr/bin/env python3
import sys


def write_block(f, b):
    for w in b:
        f.write(' '.join(map(str, w.flatten())) + '\n')

with open(sys.argv[1], 'r') as w:
[...]

PR #1740, opened 2018-08-17
Binary file format

OmnipotentEntity:
This is a first attempt at defining a binary file format.

I fully expect something to be wrong, and I believe that in fact there is something already wrong with the value net output (though the policy net seems fine at least), though I cannot locate what I've done wrong.  So if someone who is more comfortable with the format of the network can chime in that would be helpful.

As mentioned in #1733, this is not enough still.  Because neither the training pipeline nor the server yet has support for this file format.  But this should be enough to get started on those.

The latest 20b network is only 48042715 bytes (46MB) when uncompressed.  It does compress a bit further with gzip, down to 43MB.  This is under half of the gzip compressed file on the server, so this is a big gain in that area at the very least.

### DIFF
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -30,6 +30,7 @@
 #include <boost/utility.hpp>
 #include <boost/format.hpp>
 #include <boost/spirit/home/x3.hpp>
+// #include <ctime>
 
 #ifdef __APPLE__
 #include <Accelerate/Accelerate.h>
@@ -366,13 +367,13 @@ std::pair<int, int> Network::load_v3_network(std::istream& wtfile) {
         auto sign = bytes >> 15;
 
         if (exponent == 0) { // Subnormal number
-            return mantissa * pow(2.0, -24.0) * (sign ? -1 : 1);
+            return mantissa / (double)(1 << 24) * (sign ? -1 : 1);
         } else if (exponent == 32) { // Infinite number
             // Don't bother distinguishing, this is a failure case
             return std::numeric_limits<float>::infinity();
         } else {
-            float significand = 1 + mantissa * pow(2.0, -10.0);
-            return significand * pow(2.0, exponent - 15) * (sign ? -1 : 1);
+            float significand = 1 + mantissa / (double)(1 << 10);
+            return significand * (double)(1 << exponent) / (double)(1 << 15) * (sign ? -1 : 1);
         }
     };
 
@@ -499,6 +500,11 @@ std::pair<int, int> Network::load_v3_network(std::istream& wtfile) {
       process_bn_var(m_bn_pol_w2);
       process_bn_var(m_bn_val_w2);
 
+      // We got all the data, but we overread on the last few bytes?
+      if (!good()) {
+          myprintf("\nWarning, it seems the file was slightly too short?\n");
+      }
+
       // Finally, the file should be exhausted.  Double check.
       size_t offset = wtfile.tellg();
       wtfile.seekg(0, std::ios_base::end);
@@ -562,11 +568,19 @@ std::pair<int, int> Network::load_network_file(const std::string& filename) {
                 m_value_head_not_stm = false;
             }
 
+            std::pair<int,int> res;
+
+            // std::clock_t begin = clock();
             if (format_version == 3) {
-                return load_v3_network(buffer);
+                res = load_v3_network(buffer);
             } else {
-                return load_v1_network(buffer);
+                res = load_v1_network(buffer);
             }
+            // std::clock_t end = clock();
+
+            // myprintf("Loaded network in %g seconds.\n", double(end - begin) / CLOCKS_PER_SEC);
+
+            return res;
         }
     }
     return {0, 0};


src/Network.cpp
```diff
@@ -267,6 +267,334 @@ std::pair<int, int> Network::load_v1_network(std::istream& wtfile) {
...
+    };
+
+    for (int block=0; block < 1 + (2 * blocks); ++block) {
+        // Convolution Weights
+        {
+            int count = filters * filters * 9;
+            if (!block) { // Very first has a different shape because it's the input layer
+                count = filters * 162;
+            }
+            std::vector<float> weights = process(count);
```
ihavnoid: Suggest putting
```
auto weights = process(count);
```
and something similar for rest of the process() calls.

src/Network.cpp
```diff
@@ -267,6 +267,334 @@ std::pair<int, int> Network::load_v1_network(std::istream& wtfile) {
...
+    }
+
+    // Size 1
+    {
+        std::vector<float> weights = process(m_ip2_val_b.size());
+        if (!weights.size()) {
+            return {0, 0};
+        }
+        std::copy(cbegin(weights), cend(weights), begin(m_ip2_val_b));
+    }
```
ihavnoid: This looks like a lot of repetitive code.  How about moving the code body into process() calls and handle the ```return {0,0};``` case as an exception throw?

ihavnoid:
I like the implementation - but suggest using ```auto``` whenever possible, for consistency with rest of the leela-zero code.

Ttl:
The output with #1736 is 52 MB compressed. If the space saving from binary is just 9 MB I'm not sure if it's worth the trouble.

ihavnoid:
@Ttl I guess we have to see the size of the compressed binary format.  Plus, this is without losing any precision, and this code actually includes half-precision binary formats too.  My hope is that half precision weight will further save area while not losing much strength, but that will need some experimenting.

One question that I had was whether it is okay to lose precision on the final fully connected layers (when half precision is used to store weights).  The previous half precision implementations used half precision only within the residual tower so it isn't totally equivalent.  I would like to make sure that half-precision binary weights are okay, or at least keep the final few layers to be single precision even when half precision format is used.

ihavnoid:
Okay perhaps I misunderstood - the 48MB numbers are for the fp16 type.  So yes probably we aren't saving a lot of disk.  But anyway will probably need to see the strength difference.

OmnipotentEntity:
@Ttl I could also attempt some form of value length encoding to save more space.  The reason why the space saving seems less than it ought to is because we're already not storing a lot of entropy.  Though I'd argue that 9 MB on 52MB is still significant, as it's around 18%.

OmnipotentEntity:
This should be everything outside of the training code, which I will be the first to admit I do not fully understand and do not really have a good method of testing.  May I get a review?  Is the file format sane and fully featured?  Does the code make sense and is it readable?  Is this something that the project is interested in?

ihavnoid:
My opinion is that we should eventually migrate to some form of binary format.  I believe the binary format should be at least a little bit faster than the text format (since we don't need any parsing at all) if done properly.  The current text files aren't that useful for debugging anyway (large files with very long lines - too hard to do anything on a text editor) but this may be just me.

The reason I say "eventually" is because changing file formats can cause quite a bit of trouble - we have to force everybody to move beyond a certain version (which we sort of do when forcing a minimum autogtp/leelaz version for training) but I am not sure what else we are breaking...

OmnipotentEntity:
Did a little bit of performance tuning:

Type (20x256) | Seconds
------------ | -------------
v3 | 0.731481
v3.gz | 0.652128
v1 | 1.35178
v1.gz | 1.13941

ihavnoid:
@OmnipotentEntity - maybe not 'breaking', but at least this will make things be a bit more complicated.  I think we should still keep the 32-bit formats for training, but only the distribution versions of the net will be binary-converted to 16-bit.  This adds a bit more complexity.

Other than that, my only other concern is that we also reduced precision for the final 'dual head' part of the net.  Our current half precision net implementation applies half precision on the residual towers, and the 'dual head' is still single precision.  I see that the final dual-head parameters have quite big numbers (~70) and on some older nets I recall seeing values in the range of thousands.  My suggestion is that we should be a bit more conservative (after all, the 'dual head' is only a very small part of the weights) unless we have proof that it is okay to use fp16 even on the final heads.

src/Network.cpp
```diff
@@ -30,6 +30,7 @@
 #include <boost/utility.hpp>
 #include <boost/format.hpp>
 #include <boost/spirit/home/x3.hpp>
+// #include <ctime>
```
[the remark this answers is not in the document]
OmnipotentEntity: I commented out the timing code, so that other people could replicate my findings.  I'll remove it entirely if need be.

OmnipotentEntity:
A kind person allowed me to use his computer to set up a match.  So far the result after 120 games is exactly 60-60.

```
leelaz-next-3 v leelaz-next-1 (120/400 games)
board size: 19   komi: 7.5
leelaz-next-3     60 50.00%       24 40.00%     36 60.00%    337.32
leelaz-next-1     60 50.00%       24 40.00%     36 60.00%    327.70
                                  48 40.00%     72 60.00%
```

gcp:
Quick thoughts:

a) We're not sure what the best float format is, see "half" vs "Bfloat16" discussion. Ironically the text format can deal with both. Would be nice to have some flexibility there is we have a binary one.

b) Without working training code we can't deploy such nets.

c) If we are revising the network format, we should add beta and gamma to the batchnorm layers. We currently encode the beta into the convolution bias of the layer before, but we can't encode gamma.

gcp:
> May I get a review? Is the file format sane and fully featured? Does the code make sense and is it readable? Is this something that the project is interested in?

I hope I sufficiently addressed this. We can continue in an issue if you want.

Improving the network format is something we should do, but it has a large impact as it needs changes all over, so I'd prefer to do it right or not at all.

gcp:
Whethe process_bn_var should be done at writing time is another thing to consider.

PR #1741, opened 2018-08-17
add possibility to specify maximum tree size as a command line argument

kuba97531:
I see two good reasons to include this change

1) Enable longer analysis on stronger computers with more RAM.

I have been hitting the 100M nodes limit after few minutes of analysis. I think it's unrealistically low given that it's not very uncommon for people to have systems with 16+ Gigabytes of ram. Also it seems that the constant has been chosen when it was believed that memory consumption is roughly node_count * 56 (equal to `sizeof(UCTNode)`) which is no longer the case with the smart `UCTNodePointer`. I have observed as little memory consumption as 18 bytes per node instead of the 56 at 50 millions nodes.

2) Enable earlier min_psa_ratio optimisation on slower computers with less RAM.

Some people might want to reduce the number as it affects the search via `get_min_psa_ratio`.  https://github.com/kuba97531/leela-zero/blob/9e903eda6e93fb65375ea03127e4f62268913175/src/UCTSearch.cpp#L148-L149


gcp:
This kind of change should take a size in megabytes or gigabytes, not a completely meaningless (to the user) nodes number.

(I have a strong preference for allowing this to be set through a GTP "setoption" style command. Adding command line settings to the engine is bad design and we already have a few hundred too many. But maybe I can live with adding another one for now.)

src/UCTSearch.h
```diff
@@ -126,6 +127,7 @@ class UCTSearch {
     std::atomic<bool> m_run{false};
     int m_maxplayouts;
     int m_maxvisits;
+    int m_maxnodes;
```
gcp: This should probably be a size_t.

TFiFiE:
See also #1244.

kuba97531:
I initially tried to set the memory limit, but it turned out that simple node_count * size of node is off by a factor of 3-4, but I will try to find out how can I calculate the proper memory consumption or at least something that's not that far off.

Regarding the interface I will add it as an interface command instead of command line argument. 
After some thinking however I would like to additionally implement setting this via an environment variable (so priority being interface setting, env var, default). The reason being that this is really a global setting for all calculations I will ever run on a single machine and being forced to manually specify it for each client will be painful. Would that be OK?

gcp:
>I will try to find out how can I calculate the proper memory consumption or at least something that's not that far off.

That would be good, doesn't need to be perfect. Just an estimate, if possibly on the "high" side so we don't exceed it.

>After some thinking however I would like to add option to set it globally for a machine

That's harder because the clients don't (shouldn't) know about each other. Whatever controls the client will have to make the division.

kuba97531:
I have removed the part about letting user configure the max_tree_size setting for now, and replaced the node count limit for a real memory size limit.
This will allow to build 3-4 times bigger trees already. 
If this is accepted I will work on the interface part.

MartinDevelopment:
I don't think we need an extra command option for this, since most people won't even hit 100M nodes. I only reach between 40-65M nodes when I play a match at 250k visits.

kuba97531:
>  most people won't even hit 100M nodes
I don't see how is that relevant. Most people won't but some people will. And on my computer few minutes is enough to hit this artificial limit (less than 2GB currently, 5GB after my change and I would be happy to be able to run analysis until the tree is 50G+ if I want to see the limit of LZ in a certain position). Ability to increase the max tree size doesn't hurt those who don't care nor it increases the code complexity.

kuba97531:
@MartinDevelopment Also please notice that you're affected by the limit already when you reach 50M nodes. 
After half of the capacity is consumed the `min_psa_ratio` changes and the expansion of the tree slows down dramatically and it must have some impact on game strength.
In current next branch I reach 50M nodes already after 150k visits and 100M limit only after couple of milions.

MartinDevelopment:
Interesting, never noticed that. Is it possible to test if this affects strength by using a low amount of visits and a low node limit?

kuba97531:
@MartinDevelopment 
I did make some test with 10k visits and it seems that 50% difference in number of nodes doesn't change much.

Version A would be current next - reporting on average 2.3M nodes
Version B would have artificial small limit for the tree size - reporting on average 1.6M nodes

I stopped the match after it got to a 11-11 tie. Just enough to reject hypothesis that one of the sides is 2:1 favorite.

gcp:
>it must have some impact on game strength.

The impact is very, very low. You can probably find some test results if you look at the commit that added it.

src/GTP.cpp
```diff
@@ -172,6 +176,13 @@ const std::string GTP::s_commands[] = {
     "heatmap",
     "lz-analyze",
     "lz-genmove_analyze",
+    "set_option",
+    ""
+};
+
+const std::string GTP::s_options[] = {
+    "max_tree_size <size in MB> - set the maximum size of a game tree in megabytes",
+    "max_cache_size <size in MB> - set the maximum size of a Neural Network result cache in megabytes",
```
gcp: No, please. The user specifies a single memory usage maximum and the engine figures out a suitable allocation distribution between both. For example, by keeping the same proportion as whatever default we have now, with some reasonable limits for NNCache (that may already be enforced in that code anyway!).

How is the user supposed to guess what suitable values are here? They shouldn't.

src/GTP.cpp
```diff
@@ -882,10 +893,64 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
+                    gtp_fail_printf(id, "incorrect value for cache size");
+                    return true;
+                }
+            }
+        }
+        std::string options_out_tmp("");
+        for (int i = 0; s_options[i].size() > 0; i++) {
+            options_out_tmp = options_out_tmp + "\n" + s_options[i];
+        }
+        gtp_printf(id, options_out_tmp.c_str());
```
gcp: Is this for debugging?

src/UCTNodePointer.cpp
```diff
@@ -60,6 +65,8 @@ UCTNodePointer& UCTNodePointer::operator=(UCTNodePointer&& n) {
 
 void UCTNodePointer::inflate() const {
     if (is_inflated()) return;
+    m_tree_size += sizeof(UCTNode);
```
gcp: Would this correctly account for the memory consumed by the contained vectors?

gcp:
@ihavnoid Can you have a look at the UCTNodePointer change to track real memory usage?

ihavnoid:
I will give it a try, but I expect there need to be some fudge factor because of the malloc pointer management overhead.  Will try and see what that overhead might be...

gcp:
If we can't track it accurately anyway, I'd prefer to have us estimate a 95th percentile value or something from generating some statistics rather than trying to count it.

What concerns me a bit is that those estimates wouldn't scale with board size. A count, even if flawed, indeed does not have that problem.

src/GTP.cpp
kuba97531: Ok. I am ok with making one setting "max_memory" but I would be happy to keep the ratio modifiable if you'll agree. 
For example for people who do a lot of undo/redo while analyzing with Lizzie only the cache size matters and for them even 1:1 would be a good ratio, whereas for basic playing it seems tat 1:10 is roughly fine.
So even if simple end-user would only care about total memory then the tools like Lizzie could request more NNCache heavy ratio.

src/GTP.cpp
kuba97531: set_option without arguments is printing available options

kuba97531:
I have tested the estimates and at least on windows what system reports as leelaz memroy consumption is very close to what I get when calculating network size + nn cache size + m_tree_size.
Sometimes I get smaller and sometimes bigger numbers but always +/- 10%. I know that I don't fully understand memory allocation, but at least empirically it seems to be doing all right.

ihavnoid:
Used a 128x10 net to speed up test, trying to see where the memory consumption saturates:
Tree size 100MB : 591464kB
Tree size 200MB : 626688kB
Tree size 300MB : 688852kB
Tree size 500MB : 800916kB
Tree size 1000MB : 1.032GB
Tree size 2000MB : 1.430GB

ihavnoid:
Based on my (pretty unscientific) test I think we are off by 2x.  Don't know why but I will look into the code.

ihavnoid:
In general, I recommend adding assertions on tree size underflow / overflow.  That is, make sure tree size is zero when the tree is wiped out, and make sure the tree size is positive before decrementing.

src/UCTNodePointer.cpp
```diff
@@ -25,18 +25,21 @@
...
         delete read_ptr();
+        sz += sizeof(UCTNode);
     }
+    m_tree_size -= sz;
 }
 
 UCTNodePointer::UCTNodePointer(UCTNodePointer&& n) {
-    if (is_inflated()) {
-        delete read_ptr();
-    }
```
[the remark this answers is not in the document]
kuba97531: I have thought that this is dead code. Here is my logic (also confirmed by tests):
This is a constructor so I this is a fresh instance and is guaranteed that `m_data = 1` and is not inflated. We only have to care here about the possibly inflated argument.
Is this logic flawed?

src/UCTNodePointer.cpp
```diff
@@ -25,18 +25,21 @@
...
     m_data = n.m_data;
     n.m_data = 1; // non-inflated garbage
+    m_tree_size += sizeof(UCTNodePointer);
```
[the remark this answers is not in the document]
kuba97531: I thought that the destructor of other instance will still be called and trigger decrementing the counter.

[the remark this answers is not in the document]
ihavnoid: We clobbered the original pointer on the line above.

src/UCTNodePointer.cpp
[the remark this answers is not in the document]
ihavnoid: Good point... I think you are correct that it is dead.

src/UCTNodePointer.cpp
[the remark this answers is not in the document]
ihavnoid: Hmm... thinking a bit more I think you are right.      This is incrementing pointer size not the node itself.

ihavnoid:
Okay, I added some more myprint() on some locations and figured out that the stat itself isn't off by 2x - it's the get_min_psa_ratio() itself preventing the tree from growing beyond a certain range.  Possibly expected, but I wonder if the PSA ratio is going to mislead people.  For example, we specified 2GB but will never hit anywhere close to 2GB (probably closer to 1GB).

Not sure what to do here...

src/GTP.cpp
kuba97531: > The user specifies a single memory usage maximum and the engine figures out a suitable allocation distribution between both

@gcp 
Is it acceptable if in the program I count the memory consumption as size of tree + nn cache + network + some empirically achieved magical number?
I don't know the program enough to count what is the whole memory consumption. E.g. when the program starts there seems to much higher memory consumption than the pure network size. 

> How is the user supposed to guess what suitable values are here? They shouldn't.

I am willing to accept your judgment, but I disagree. As I see it there are two classes of users
1) End-user who want to play. 90% of them are using some clients and indeed don't need to know or care about gtp protocol at all.
2) Client authors who are able and potentially willing to understand which is more relevant to the use of their program.

kuba97531:
@ihavnoid 
> the stat itself isn't off by 2x - it's the get_min_psa_ratio() itself preventing the tree from growing beyond a certain range

I think that's perfectly OK. With the current 100M nodes limit half of it is consumed after 150k playouts and after few million playouts the search stops completely. With bigger limit 50% will be reached later and complete stop even later.

> there seems to be a counter 'leak' on UCTNodePointer::release.

You're right. I will fix that

> In general, I recommend adding assertions on tree size underflow / overflow

I will do that.

gcp:
>Not sure what to do here...

Behavior seems to be "working as intended" to me.

i.e. user specifies the max they can tolerate and we're trying to make good use of it, which due our current design means only maxing it out in extreme circumstances.

kuba97531:
I would also like to add a gtp command to dump what the mechanism used in this PR thinks is the current memory usage (mostly for verification purposes). What would be the good name / output format for this command? How about:
`lz-estimatememory` 
```
Current estimated memory consumption: 1691MB
Network + overhead: 500MB 
Search tree: 1094MB
NNCache: 97MB
```

src/Network.cpp
```diff
@@ -898,3 +898,50 @@ std::pair<int, int> Network::get_symmetry(const std::pair<int, int>& vertex,
...
+    //polcy head
+    result += OUTPUTS_POLICY * sizeof(float); //m_bn_pol_w1
+    result += OUTPUTS_POLICY * sizeof(float); //m_bn_pol_w2
+    result += OUTPUTS_POLICY * NUM_INTERSECTIONS * POTENTIAL_MOVES * sizeof(float); //m_ip_pol_w
+    result += POTENTIAL_MOVES * sizeof(float); //m_ip_pol_b
+
+                                               //value ehead
+    result += m_conv_val_w.size() * sizeof(float);
+    result += m_conv_val_b.size() * sizeof(float);
+    result += OUTPUTS_VALUE * sizeof(float); //m_bn_val_w1;
```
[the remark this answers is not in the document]
kuba97531: Comments are on the lines where the variable is not used to calculate the size (the line adds a constant)

src/Network.cpp
```diff
@@ -898,3 +898,50 @@ std::pair<int, int> Network::get_symmetry(const std::pair<int, int>& vertex,
...
+    result += m_conv_pol_w.size() * sizeof(float);
+    result += m_conv_pol_b.size() * sizeof(float);
+
+    //polcy head
+    result += OUTPUTS_POLICY * sizeof(float); //m_bn_pol_w1
+    result += OUTPUTS_POLICY * sizeof(float); //m_bn_pol_w2
+    result += OUTPUTS_POLICY * NUM_INTERSECTIONS * POTENTIAL_MOVES * sizeof(float); //m_ip_pol_w
+    result += POTENTIAL_MOVES * sizeof(float); //m_ip_pol_b
+
+                                               //value ehead
```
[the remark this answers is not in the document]
kuba97531: thanks for reading and spotting that!

src/Network.cpp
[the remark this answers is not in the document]
kuba97531: Oh. There isn't. I fixed that.

TFiFiE:
I'd use "[MiB](https://en.wikipedia.org/wiki/Mebibyte)" instead of "MB" in order to be completely unambiguous. Also, I'd replace every occurrence of `1048576LL` with a constant initialized as `1024LL * 1024LL`.

src/GTP.cpp
```diff
@@ -172,6 +176,13 @@ const std::string GTP::s_commands[] = {
     "heatmap",
     "lz-analyze",
     "lz-genmove_analyze",
+    "lz-estimatememory",
+    "lz-setoption",
+    ""
+};
+
+const std::string GTP::s_options[] = {
+    "max_memory <total memory in MB> <optional: fraction for Neural Network Cache>- set the maximum memory (in MB) leelaz should utilize. Optionally set fraction of memory remaining for the tree search which should be used by Neural Network cache (by default 0.1)",
```
gcp: For the working of lz-setoption, can you take a look at the UCI spec (for chess engines) and see how setoption works there? I'd prefer it if lz-setoption were to follow the same format, but perhaps adding float support.

It doesn't provide any way to provide a description to the GUI, though (but the options can be named more descriptively!), but I think that's still better than this format which is OK for someone working in the terminal, or someone who adapts their GUI specifically to LZ's idiosyncrasies, but not really where we want to go for Go engines in general.

So a session would look like:
```
lz-setoption
option name Maximum Memory Use (MiB) type spin default 2048 min 128 max 131072
option name Percentage of memory for cache type spin default 10 min 1 max 99

lz-setoption name Maximum Memory Use (MiB) value 1024
lz-setoption name Percentage of memory for cache value 20
```

src/GTP.cpp
```diff
@@ -882,10 +894,86 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
         }
+        return true;
+    } else if (command.find("lz-estimatememory") == 0) {
+        auto base_memory = (size_t)(s_network->get_estimated_size() * Network::CONSERVATIVE_MEMORY_OVERHEAD);
+        auto tree_size = UCTNodePointer::get_tree_size();
+        auto cache_size = s_network->get_estimated_cache_size();
 
+        auto total = base_memory + tree_size + cache_size;
+        printf("Estimated total memory consumption: %d MB.\n", total / 1048576LL);
+        printf("Network with program overhead: %d MB / Search tree: %d MB / %d\n", base_memory / 1048576LL, tree_size / 1048576LL, cache_size / 1048576LL);
```
gcp: Watch the line length, <80 if possible.

src/GTP.cpp
```diff
@@ -882,10 +894,86 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
         return true;
-    }
+    } else if (command.find("lz-setoption") == 0) {
+        std::istringstream cmdstream(command);
+        std::string tmp, option;
 
+        cmdstream >> tmp >> option;
+
+        if (!cmdstream.fail()) {
```
gcp: Maybe option handling could be factored out to a separate function, especially as we'll want to add more.

src/NNCache.h
```diff
@@ -29,6 +29,17 @@
 
 class NNCache {
 public:
+
+    /*
+    Maximum size of the cache in number of items.
+    */
+    static constexpr int MAX_CACHE_COUNT = 150'000;
```
gcp: You need to define these in the .cpp file too or you'll get obscure compilation errors at some point. (What you have here is only a declaration).

Check the other constants in FastBoard.cpp/h if you don't immediately get what I mean.

src/UCTNodePointer.cpp
```diff
@@ -25,18 +25,34 @@
 
 #include "UCTNode.h"
 
+std::atomic<size_t> UCTNodePointer::m_tree_size = { 0 };
```
gcp: Feels a bit weird to have this here instead of UCTSearch but it's probably a bit more convenient I guess.

Though, code like here https://github.com/gcp/leela-zero/pull/1741/files#diff-70aab7683fe64f9428b8bc47d1d183c0R179
would look more natural if it was part of UCTSearch.

kuba97531: I don't see a perfect solution and see some good reasons to keep it on either class, so I decided to keep it where the implementation is easier. I'm open to suggestions on how to improve it however.

kuba97531:
rebased on top of next

src/GTP.cpp
```diff
@@ -875,10 +899,158 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
         }
+        return true;
+    } else if (command.find("lz-memory_report") == 0) {
+        auto base_memory = (size_t)(s_network->get_estimated_size() *
+            Network::CONSERVATIVE_MEMORY_OVERHEAD);
+        auto tree_size = UCTNodePointer::get_tree_size();
+        auto cache_size = s_network->get_estimated_cache_size();
+
+        auto total = base_memory + tree_size + cache_size;
+        printf("Estimated total memory consumption: %d MiB.\n", total / MiB);
```
gcp: This probably needs gtp_printf, make sure the output is legal according to the GTP spec?

src/GTP.cpp
```diff
@@ -875,10 +899,158 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
+                gtp_fail_printf(id, reason.c_str());
+            }
+            return true;
+        }
+        else {
+            gtp_fail_printf(id, "incorrect value");
+            return true;
+        }
+    }
+    if (name == "percentage of memory for cache") {
```
gcp: else if

src/GTP.cpp
```diff
@@ -875,10 +899,158 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
+        valuestream >> max_memory_in_mib;
+        if (!valuestream.fail()) {
+            bool result;
+            std::string reason;
+            std::tie(result, reason) = set_max_memory(max_memory_in_mib * MiB,
+                cfg_max_cache_ratio_percent);
+            if (result) {
+                gtp_printf(id, reason.c_str());
+            }
+            else {
```
gcp: } else {

(consistent with rest of codebase)

src/GTP.cpp
```diff
@@ -875,10 +899,158 @@ bool GTP::execute(GameState & game, std::string xinput) {
...
+    // resize cache
+    s_network->nncache_resize(max_cache_count);
+
+    return std::make_pair(true, "Setting max tree size to " +
+        std::to_string(cfg_max_tree_size / MiB) + " MiB and cache size to " +
+        std::to_string(max_cache_size / MiB) +
+        " MiB");
+}
+
+bool GTP::execute_setoption(int id, std::string command) {
```
gcp: Unless you need to modify command, pass by const ref.

kuba97531:
I have rebased the code and addressed the review points. I also added some more magical parameters to the memory estimation and at least on my computer it's very good at the moment. I would welcome confirmation from people on other systems / cpu version.

Thanks for encouragement and taking the time to review my changes. I welcome any additional comments / nits given I am not a C++ programmer.

kuba97531:
Fixed CPU version.

src/GTP.cpp
kuba97531: Converted to gtp_printf.
I think I now follow the GTP spec regarding multiline responses.

kuba97531:
Thank you for accepting my code and nice comments. I'm sorry for abundance of minor bugs, styling issues and typos. I'll try to do a better job next time!

PR #1179, opened 2018-04-10
Thread-safe UCTNodePointer

ihavnoid:
This makes almost all UCTNodePointer operations thread-safe.
The only exceptions are destructors and when it is 'moved out'
Should even handle concurrent inflate() calls properly.

PR #1744, opened 2018-08-19
Thread scalability attempt 2

ihavnoid:
Attempt 2 of #1707 (see comments from #1725)

May address #1554 

> He also reported that current multi-GPU support can only max out up to 5x1080Ti (~98% usage); with 10x1080Ti the GPU usage plunges below 50%.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -70,12 +70,13 @@ bool UCTNode::create_children(Network & network,
     if (!expandable(min_psa_ratio)) {
         return false;
     }
-    // Someone else is running the expansion
-    if (m_is_expanding) {
+
+    // We'll be the one queueing this node for expansion, stop others
+    // If it was already true, someone else is running the expansion
+    bool expanding = m_is_expanding.exchange(true);
+    if (expanding) {
         return false;
     }
-    // We'll be the one queueing this node for expansion, stop others
-    m_is_expanding = true;
     lock.unlock();
 
     const auto raw_netlist = network.get_output(
@@ -242,7 +243,11 @@ void UCTNode::accumulate_eval(float eval) {
 }
 
 UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
-    LOCK(get_mutex(), lock);
+    // make sure expansion is done
+    while (m_is_expanding.load()) {}
+
+    // Once we made sure that the expansion is done, m_children is
+    // not going to be further modified, hence we don't need a lock here.
 
     // Count parentvisits manually to avoid issues with transpositions.
     auto total_visited_policy = 0.0f;
@@ -291,6 +296,8 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
         }
     }
 
+    // the inflate() method has to be protected by the parent's lock
+    LOCK(m_nodemutex, lock);
     assert(best != nullptr);
     best->inflate();
     return best->get();
--- a/src/UCTNode.h
+++ b/src/UCTNode.h
@@ -114,7 +114,7 @@ private:
     std::atomic<double> m_blackevals{0.0};
     std::atomic<Status> m_status{ACTIVE};
     // Is someone adding policy priors to this node?
-    bool m_is_expanding{false};
+    std::atomic<bool> m_is_expanding{false};
     SMP::Mutex m_nodemutex;
 
     // Tree data


src/UCTNode.cpp
```diff
@@ -291,6 +296,8 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
         }
     }
 
+    // the inflate() method has to be protected by the parent's lock
```
gcp: Oh, right, this....

What was the thread restriction on inflate? If it's just to avoid concurrent inflates, we're fine. But if there's restrictions in reading out the node info while it's being inflated, this might not work. But I think we replace the data item in an atomic transaction, so it should work?

src/UCTNode.cpp
```diff
@@ -242,7 +243,11 @@ void UCTNode::accumulate_eval(float eval) {
 }
 
 UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
-    LOCK(get_mutex(), lock);
+    // make sure expansion is done
```
gcp: I wonder if this needs a check that the node can never be expanded again (i.e. against min_psa_ratio).

I guess during each search, every node only gets expanded or re-expanded once. So either we are expanding the node, or it's already expanded. But it can't get re-expanded in the middle.

src/UCTNode.cpp
ihavnoid: Oh right... this won't work.  Thread-safe UCTNodePointer was #1179 and it was dropped.  See methods like UCTNodePointer::get_policy() and you can see nothing is atomic here.

So, it's either we revive #1179 or revive #1725 or find something else or just give up.  I think #1725 is simpler than #1179 as long as the RWLock has no bugs.  read-write locks are well understood concepts...

One thing is that I was running with this patch enabled for at least 40 hours and nothing crashed.  Maybe there is something that prevents any obvious crash?

src/UCTNode.cpp
ihavnoid: If we were to get this working then I think the right way to do this is instead of having a 'is expanding' flag, change it to an enum which is basically not expanded / expanding / expanded three-state thingy.

src/UCTNode.cpp
gcp: >Oh right... this won't work. Thread-safe UCTNodePointer was #1179 and it was dropped. See methods like UCTNodePointer::get_policy() and you can see nothing is atomic here.
...
>One thing is that I was running with this patch enabled for at least 40 hours and nothing crashed. Maybe there is something that prevents any obvious crash?

m_data is a 64-bit integer. On x86/x64 it will get atomically updated (the entire cache line, really), just the ordering isn't guaranteed (but by default x86 does very sane things). But nothing is going to see a half-updated value...

We can revive #1179. I closed it because there was no gain at the time and we found the bug that prompted its creation.

We can also do the RWLock. I'm not sure yet what the best approach is. But I think with very high concurrency (batching?), the atomic approach avoids taking any locks at all, and was little write traffic, whereas the RWLocks still need to write something every time there is a reader.

gcp: On the other hand, the RWLock approach splits up the complexity into the RWLock implementation. Reasoning about the tree is probably easier with RWLocks.

Could we test how both approaches behave?

ihavnoid:
Okay, what I will do is count how many times a spinlock 'spins'.  Gather stats on how frequently it happens out of a -v 20000 -t 24 run with a 20x256 net.  Three binaries:
- next branch
- this branch
- #1725 - back to the read-write lock

ihavnoid:
Regarding the 'extra memory writes' - I am not concerned at all about those extra accesses because it is 100% likely that we will hit cache anyway.  We are reading adjacent cache lines anyway hence there is no extra read, maybe one extra write per uct_select_child, this will be something like 50k more cache line (64B) writes per second (even assuming worst case) which is something like 3.2MB/sec more memory bandwidth... out of 30+GB/sec of DRAM bandwidth available from recent Intel cores and AMD Ryzens.

ihavnoid:
Have been thinking for a while, and I think this is an okay compromise:
- Start from this patch
- Revive #1179
- Once node is expanded, the lock is 'destroyed' and further locks will assert-fail - this makes the whole thing explicit that the node is modifiable only if it is single thread context.
- Fix any other cases the lock is legitimately needed - remaining references of the lock is only used in single thread contexts anyway hence the lock is not needed at all.

I wish we had some way to assert-fail if we are not in a single thread context but still modifying m_children.  Will think if there is an easy, clean way.

ihavnoid:
Upon trying a bit more, I figured out that there is no 'safe' timing to disable a lock.  The lockless implementation is getting out of control - ugh.  Need to think a bit more.

ihavnoid:
Okay, I pushed #1764 which is another attempt.  I think #1764 is much better and cleaner... but others may disagree and hence would like comments.

Closing this one.

Issue #1748, opened 2018-08-19
GPU memory use


Ttl:
Next branch seems to require quite lot of GPU memory because of bigger Winograd transformations and precision autodetection.

Winograd F(4x4, 3x3) expands filters from 3x3 to 6x6 for total of four times increase in memory. At startup with precision autodetection both single and half precision are allocated at the same time. As a result we need about six times more memory than is required to store the weights. Using 40b network I measured about 1.5 GB peak GPU memory use at the startup and 960 MB in normal play with single precision.

How much memory use is excessive? I think there are few older cards that don't have this much memory. Colab is also limited to 512 MB.

I have a branch that does the filter transformation at runtime allowing the filters to be stored untransformed in GPU memory and dropping memory usage to about one quarter. It does have about 25% of performance hit with batch size of 1 so I wouldn't want to turn it on by default. Finding how much free GPU memory there is doesn't seem to be very easy with OpenCL either.

ihavnoid:
> That should be fixable and a fast win wrt peak usage?
The downside is that if we create one at a time, and end up picking the 'wrong' one, we have to recreate the net.  I will test how long it takes to create an opencl net.

ihavnoid:
A quick test shows that we add roughly a second initializing the 20b nets and pushing the weights to the newly created net on my i7-8700 + GTX1080.  Probably much slower on older machines, but maybe it's better than simply crashing.  The time consumed seems to be mostly on pushing the weights from CPU to GPU, but maybe there is some chance of optimization there, too.

gcp:
Going to close this. We did a round of optimizations. Let's file new issues for further work.

PR #1750, opened 2018-08-20
Network initialization restructuring

ihavnoid:
- Create one net at a time when doing fp16/fp32 autodetect.  Saves some GPU memory
- Create an internal lambda which initializes the nets
- Use std::copy to copy vectors to reduce runtime

Related to #1748 - possibly addresses concern, plus makes the code cleaner

### DIFF
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -318,6 +318,22 @@ std::pair<int, int> Network::load_network_file(const std::string& filename) {
 }
 
 void Network::initialize(int playouts, const std::string & weightsfile) {
+#ifdef USE_BLAS
+#ifndef __APPLE__
+#ifdef USE_OPENBLAS
+    openblas_set_num_threads(1);
+    myprintf("BLAS Core: %s\n", openblas_get_corename());
+#endif
+#ifdef USE_MKL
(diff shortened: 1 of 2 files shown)


ihavnoid:
Added one more (obvious) performance optimization on zeropad_U : shaves out another 1 second to startup

src/Network.cpp
```diff
@@ -372,132 +388,111 @@ void Network::initialize(int playouts, const std::string & weightsfile) {
...
-                    fp16net = std::make_unique<OpenCLScheduler<half_float::half>>();
-                    fp16net->initialize(channels);
-                    to_init.emplace_back(fp16net.get());
-                } catch (std::runtime_error) {
-                    myprintf("Failed to initialize half precision net.  Resorting to single precision.\n");
-                    fp16net.reset();
+                    m_forward = init_net(std::make_unique<OpenCLScheduler<float>>());
+                    score_fp32 = benchmark_time(100);
+                } catch (...) {
+                    // empty - if exception thrown just throw away fp16 net
```
mmartin: Incorrect comment—should probably be “throw away **fp32** net”?

PR #1757, opened 2018-08-22
Replace "size" with the unambiguous "length" where applicable.

TFiFiE:
This is a non-functional change and does not defy the SGF and GTP specifications.


MartinDevelopment:
I think the current way is less confusing and I see no reason to change it.

gcp:
Nobody talks about board length though.

>already manifests itself in the current code

```constexpr int board_squares = num_intersections``` (I'm just kidding, I think the code can just use num_intersections?)

TFiFiE:
It could very well be the case that the use of "size" was a deliberate choice in the SGF and/or GTP specifications to facilitate a possible future extension supporting oblong boards (by accepting two numbers in their respective context).

PR #1764, opened 2018-08-23
"Lockless" UCTNode (2nd attempt)

ihavnoid:
- Get rid of lock, replace it with a std::atomic<ExpandState>
- The FSM acts as the lock when it is needed, and a simple state flag normally, (mostly) eliminating the lock.  The only situation we spin on a lock is when the node is under expansion and other threads are waiting for the node expansion to be done.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -52,23 +52,17 @@ bool UCTNode::create_children(Network & network,
                               GameState& state,
                               float& eval,
                               float min_psa_ratio) {
-    // check whether somebody beat us to it (atomic)
-    if (!expandable(min_psa_ratio)) {
-        return false;
-    }
-    // acquire the lock
-    bool success = acquire_expanding();
-    if (!success) {
-        return false;
-    }
-
     // no successors in final state
     if (state.get_passes() >= 2) {
-        expand_cancel();
         return false;
     }
 
-    // check whether somebody beat us to it (after taking the lock)
+    // acquire the lock
+    if (!acquire_expanding()) {
+        return false;
+    }
+
+    // can we actually expand?
     if (!expandable(min_psa_ratio)) {
         expand_cancel();
         return false;
@@ -236,7 +230,7 @@ void UCTNode::accumulate_eval(float eval) {
 }
 
 UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
-    check_expanded();
+    wait_expanded();
 
     // Count parentvisits manually to avoid issues with transpositions.
     auto total_visited_policy = 0.0f;
@@ -314,13 +308,13 @@ private:
 };
 
 void UCTNode::sort_children(int color) {
-    decl_single_thread_use();
+    enforce_single_thread_use();
     std::stable_sort(rbegin(m_children), rend(m_children), NodeComp(color));
     finish_single_thread_use();
 }
 
 UCTNode& UCTNode::get_best_root_child(int color) {
-    check_expanded();
+    wait_expanded();
 
     assert(!m_children.empty());
 
@@ -331,12 +325,13 @@ UCTNode& UCTNode::get_best_root_child(int color) {
     return *(ret->get());
 }
 
-size_t UCTNode::count_nodes() const {
+size_t UCTNode::count_nodes_and_clear_expand_state() {
     auto nodecount = size_t{0};
     nodecount += m_children.size();
+    m_expand_state = ExpandState::INITIAL;
     for (auto& child : m_children) {
-        if (child.get_visits() > 0) {
-            nodecount += child->count_nodes();
+        if (child.is_inflated()) {
+            nodecount += child->count_nodes_and_clear_expand_state();
         }
     }
     return nodecount;
@@ -380,7 +375,7 @@ void UCTNode::expand_cancel() {
 #endif
     assert(v == ExpandState::EXPANDING);
 }
-void UCTNode::check_expanded() {
+void UCTNode::wait_expanded() {
     while (m_expand_state.load() == ExpandState::EXPANDING) {}
     auto v = m_expand_state.load();
 #ifdef NDEBUG
@@ -388,7 +383,7 @@ void UCTNode::check_expanded() {
 #endif
     assert(v == ExpandState::EXPANDED);
 }
-void UCTNode::decl_single_thread_use() {
+void UCTNode::enforce_single_thread_use() {
     auto v = m_expand_state.exchange(ExpandState::SINGLE_THREAD_USE);
 #ifdef NDEBUG
     (void)v;
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -64,7 +64,7 @@ bool UCTNode::create_children(Network & network,
 
     // can we actually expand?
     if (!expandable(min_psa_ratio)) {
-        expand_cancel();
+        expand_done();
         return false;
     }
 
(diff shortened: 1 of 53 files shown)


src/UCTNode.cpp
```diff
@@ -61,22 +57,22 @@ bool UCTNode::create_children(Network & network,
...
+        return false;
+    }
+
     // no successors in final state
     if (state.get_passes() >= 2) {
+        expand_cancel();
         return false;
     }
+
     // check whether somebody beat us to it (after taking the lock)
```
gcp: Is this case still possible now? If we acquired_expanding then we are the only ones here, right?

Can we still re-expand nodes that were created under limited memory?

src/UCTNode.cpp
```diff
@@ -362,3 +359,47 @@ bool UCTNode::valid() const {
...
+    assert(v == ExpandState::EXPANDING);
+}
+void UCTNode::expand_cancel() {
+    auto v = m_expand_state.exchange(ExpandState::INITIAL);
+#ifdef NDEBUG
+    (void)v;
+#endif
+    assert(v == ExpandState::EXPANDING);
+}
+void UCTNode::check_expanded() {
```
gcp: This should indicate it's really a wait, so wait_expanded?

src/UCTNode.cpp
```diff
@@ -362,3 +359,47 @@ bool UCTNode::valid() const {
...
+    while (m_expand_state.load() == ExpandState::EXPANDING) {}
+    auto v = m_expand_state.load();
+#ifdef NDEBUG
+    (void)v;
+#endif
+    assert(v == ExpandState::EXPANDED);
+}
+void UCTNode::decl_single_thread_use() {
```
gcp: request? enforce?

lock? :-)

src/UCTNodePointer.cpp
```diff
@@ -26,17 +26,18 @@
 #include "UCTNode.h"
 
 UCTNodePointer::~UCTNodePointer() {
-    if (is_inflated()) {
-        delete read_ptr();
+    auto v = std::atomic_exchange(&m_data, INVALID);
```
gcp: Do you need the exchange here? We can't be running this with multiple threads accessing it anyway.

src/UCTNodePointer.cpp
```diff
@@ -26,17 +26,18 @@
...
+    if (is_inflated(v)) {
+        delete read_ptr(v);
     }
 }
 
 UCTNodePointer::UCTNodePointer(UCTNodePointer&& n) {
-    if (is_inflated()) {
-        delete read_ptr();
+    auto nv = std::atomic_exchange(&n.m_data, INVALID);
```
gcp: What's the use case where setting n.m_data to invalid would be needed?

src/UCTNodePointer.cpp
```diff
@@ -45,52 +46,71 @@ UCTNodePointer::UCTNodePointer(std::int16_t vertex, float policy) {
...
+        if (is_inflated(v)) return;
+
+        auto v2 = reinterpret_cast<std::uint64_t>(
+            new UCTNode(read_vertex(v), read_policy(v))
+        ) | POINTER;
+        bool success = m_data.compare_exchange_strong(v, v2);
+        if (success) {
+            return;
+        } else {
+            // this means that somebody else also modified this instance.
```
gcp: Could they have done anything else beside inflate it?

src/UCTNode.cpp
gcp: This can probably be fixed by not bailing out if acquire_expanding was already set to EXPANDED?

src/UCTNodePointer.cpp
ihavnoid: Not really, but no harm either.  We can change it to non-exchange if you insist.

src/UCTNodePointer.cpp
ihavnoid: This is the move constructor so we have to make sure the original instance no longer owns the pointer hence have to overwrite it with something.  Otherwise the destructor of the original instance will delete it.  Did I get your question correctly?

src/UCTNodePointer.cpp
ihavnoid: UCTNodePointer should be considered read only if it is not inflated.  That was the assumption - this is supposed to be mimicking  lazy initialization.

ihavnoid:
In general, the whole code assumes that an expanded node is not going to get modified on any multi thread context, hence we can't support re-expansion or any other type of node manipulation.  If we wanted to support such behavior, we should rather move to a read-write-lock based solution to make sure there are no uct_select_child calls running when we want to do the re expansion.

I will work on other change requests when I have a PC.

ihavnoid:
...and after reading the code a bit more it seems that this patch is indeed breaking the re-expansion.  It won't crash or corrupt itself, but it simply will refuse the re-expansion and move on.

I guess if we want to support re-expansion we should rather live with a read-write lock as fixing this to support re-expansion is not going to make anything simpler.  @gcp - what do you think?

src/UCTNode.cpp
ihavnoid: Not really a locking thing, so will say 'enforce'

src/UCTNodePointer.cpp
gcp: Simple assignment?

src/UCTNodePointer.cpp
gcp: OK!

gcp:
>This can probably be fixed by not bailing out if acquire_expanding was already set to EXPANDED

This obviously wouldn't have worked because nodes rely on EXPANDED being set as meaning they can traverse m_children.

src/UCTNode.cpp
```diff
@@ -47,36 +47,26 @@ bool UCTNode::first_visit() const {
...
+
+    // acquire the lock
+    if (!acquire_expanding()) {
         return false;
     }
-    // Someone else is running the expansion
-    if (m_is_expanding) {
+
+    // can we actually expand?
+    if (!expandable(min_psa_ratio)) {
```
gcp: You introduced a bug now: if you choose not to expand a node because it is already expanded up to the appropriate min_psa_ratio, you will leave the node in INITIAL state, but it should really be left to EXPANDED.

gcp:
>But maybe it is going to add too much possible complexity on an already complicated piece of code.

Yes, this is complicated, but:

1) It makes a whole lot of the hidden threading assumptions in the code very explicit, and also makes the evolution of the node very explicit.

2) It removes almost all blocking in the tree search aside from the case where we're waiting for an expansion, and the other pull request also avoids that.

So I like this a lot.

ihavnoid:
Okay, the auto merge failed.  Will look into what's wrong later today.

src/UCTNode.h
```diff
@@ -113,13 +113,50 @@ class UCTNode {
...
+        // EXPANDING is responsible of finishing the expansion and then
+        // move to EXPANDED, or revert to INITIAL if impossible
+        EXPANDING,
+
+        // expansion done.  m_children cannot be modified on a multi-thread
+        // context, until node is destroyed.
+        EXPANDED,
+
+        // m_children is being modified on a single-thread context.
+        SINGLE_THREAD_USE
```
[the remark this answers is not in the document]
ihavnoid: Okay, removing this.

ihavnoid:
I think I addressed all remaining issues.  Please have a look.

PR #1765, opened 2018-08-23
Don't use "void" as function parameter.

TFiFiE:


ihavnoid:
It is quite explicit that we are using c++14 here - C++14 explicitly states that f(void) and f() are identical. See 8.3.5 and C1.7

gcp:
I think it's very common to leave it out in modern C++ and in any case the code's current usage isn't even consistent with it.

PR #1782, opened 2018-08-27
ELF training

alreadydone:
This is not intended to be merged for now but just to facilitate the training of ELF weights with LZ-produced data. Ideally the weights file should be read to determine whether it's v1 or v2 and apply the change in this commit if it's v2, but reading the network is in another py file IIUC. net2net should just work, so it may be a good idea to net2net to 256x40 (or 256x30 as some has suggested) and train with this modified code.
@Ttl Does this look working to you? I am worried that if we don't keep sampling 50% positions from black-won games and 50% from white-won games we may mess up the value head. https://github.com/pytorch/ELF/issues/74#issuecomment-409087307 How do you think we can achieve this?

### DIFF
--- a/training/tf/chunkparser.py
+++ b/training/tf/chunkparser.py
@@ -212,6 +212,8 @@ class ChunkParser:
         winner = float(text_item[18])
         if not(winner == 1.0 or winner == -1.0):
             return False, None
+        if stm == 1
+            winner = -winner
         winner = int((winner + 1) / 2)
 
         version = struct.pack('i', 1)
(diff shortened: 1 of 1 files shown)


Ttl:
It's not enough to flip the value head. You need to add the batchnorm gammas too.

I wouldn't worry about sampling.

alreadydone:
The only difference between ELF (v2) networks and LZ (v1) networks at inference time is inverting the value, so why would there be additional difference at training time? After SWA there are no longer identically zero lines in the weights files; is that relevant to the discussion here?

Ttl:
ELF was trained with batch norm gammas enabled while LZ has them fixed to 1. For inference gammas were multiplied into convolution weights, but this doesn't work during training when batch statistics are used instead of moving averages. During training activations will be incorrectly normalized to unit variance.

PR #1793, opened 2018-08-30
Isolate and clean up text-to-vertex conversion.

TFiFiE:
Case-sensitive coordinates are a thing [in SGF](https://www.red-bean.com/sgf/go.html#types), [not GTP](https://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html#sec:simple-entities).

### DIFF
--- a/src/FastBoard.cpp
+++ b/src/FastBoard.cpp
@@ -19,6 +19,8 @@
 #include "FastBoard.h"
 
 #include <cassert>
+#include <cctype>
+#include <algorithm>
 #include <array>
 #include <iostream>
 #include <queue>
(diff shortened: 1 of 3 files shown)


src/FastBoard.cpp
```diff
@@ -430,6 +432,34 @@ std::string FastBoard::move_to_text(int move) const {
     return result.str();
 }
 
+int FastBoard::text_to_move(std::string move) const {
```
gcp: const ref?

src/FastBoard.cpp
```diff
@@ -430,6 +432,34 @@ std::string FastBoard::move_to_text(int move) const {
...
+
+    if (move == "pass") {
+        return PASS;
+    } else if (move == "resign") {
+        return RESIGN;
+    } else if (move.size() < 2 || !std::isalpha(move[0]) || !std::isdigit(move[1]) || move[0] == 'i') {
+        return NO_VERTEX;
+    }
+
+    int column = move[0] - 'a';
```
gcp: auto

PR #1794, opened 2018-08-31
Improve MTCS a bit

bood:
Search thread should explore other nodes in this case, this would save
the search from some useless searches.

It has benefit for batching support too. Before this change, all
threads could be busy waiting for the first node being expanded.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -273,6 +273,11 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     for (auto& child : m_children) {
         if (!child.active()) {
             continue;
+        } else {
+            child.inflate();
+            if (child->m_is_expanding) {
+                continue;
+            }
         }
 
         auto winrate = fpu_eval;
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -271,12 +271,14 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     auto best_value = std::numeric_limits<double>::lowest();
 
     for (auto& child : m_children) {
-        if (!child.active() || (child.is_inflated() && child->m_is_expanding)) {
+        if (!child.active()) {
             continue;
         }
 
         auto winrate = fpu_eval;
-        if (child.get_visits() > 0) {
+        if (child.is_inflated() && child->m_is_expanding) {
+            winrate = - fpu_reduction - 1.0;
+        } else if (child.get_visits() > 0) {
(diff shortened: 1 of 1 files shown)


src/UCTNode.cpp
```diff
@@ -273,6 +273,11 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     for (auto& child : m_children) {
         if (!child.active()) {
             continue;
+        } else {
+            child.inflate();
```
ihavnoid: This is going to increase the memory consumption by a huge amount and makes the whole UCTNodePointer optimization pointless.  If it isn't inflated you can skip the m_is_expanding check and assume it isn't being expanded.

ihavnoid: One more comment - this is going to conflict with the 'lockless UCTNode' implementation down there since it doesn't have the m_is_expanding anymore (it uses a separate enum)

bood: Thanks. I've updated the code to save memory.
For the conflicting, I think we have to live with it until one is merged, then we can do the rebase.
The change needs to be proved to be correct first anyway.

bood:
By the way, I didn't see any significant performance increase on my MBP though, maybe because the saving is small. Anyone with other cards can double check.

dtracers:
isn't the main saving for multiple cards?

src/UCTNode.cpp
```diff
@@ -271,7 +271,7 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     auto best_value = std::numeric_limits<double>::lowest();
 
     for (auto& child : m_children) {
-        if (!child.active()) {
+        if (!child.active() || (child.is_inflated() && child->m_is_expanding)) {
```
gcp: You will now crash if there is only one valid successor. (or assert at best != nullptr).

I'd set winrate to 0.0 or something, like a huge virtual loss.

bood: Does `- fpu_reduction - 1.0` make sense? I feel 0.0 is not enough since the eval may give 0 win rate in extreme conditions and we're using `parent - fpu_reduction` for unvisited nodes.

gcp: As long as it's a very low value it will work. But I'm not sure we care about MCTS scaling if the critical path has all 0.0 winrates.

bood: We usually don't. But current batching implementation requires that the expanding node not being selected again, or the full batch waiting will be stuck.

PR #1795, opened 2018-09-01
Network : remove unused residual tower weight values

ihavnoid:
We restructured the code ad moved everything to ForwardPipe.  No need to leave it in the memory anymore.  This saves about half a gigabyte of memory footprint on 20x256 nets

### DIFF
--- a/src/CPUPipe.cpp
+++ b/src/CPUPipe.cpp
@@ -265,25 +265,25 @@ template <size_t spatial_size>
 void batchnorm(const size_t channels,
                std::vector<float>& data,
                const float* const means,
-               const float* const stddivs,
+               const float* const stddevs,
                const float* const eltwise = nullptr) {
     const auto lambda_ReLU = [](const auto val) { return (val > 0.0f) ?
                                                           val : 0.0f; };
     for (auto c = size_t{0}; c < channels; ++c) {
         const auto mean = means[c];
-        const auto scale_stddiv = stddivs[c];
+        const auto scale_stddev = stddevs[c];
         const auto arr = &data[c * spatial_size];
 
         if (eltwise == nullptr) {
             // Classical BN
             for (auto b = size_t{0}; b < spatial_size; b++) {
-                arr[b] = lambda_ReLU(scale_stddiv * (arr[b] - mean));
+                arr[b] = lambda_ReLU(scale_stddev * (arr[b] - mean));
             }
         } else {
             // BN + residual add
             const auto res = &eltwise[c * spatial_size];
             for (auto b = size_t{0}; b < spatial_size; b++) {
-                arr[b] = lambda_ReLU((scale_stddiv * (arr[b] - mean)) + res[b]);
+                arr[b] = lambda_ReLU((scale_stddev * (arr[b] - mean)) + res[b]);
             }
         }
     }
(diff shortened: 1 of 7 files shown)


ihavnoid:
Just to be clear, there is no guarantee that the memory is actually saved because it has a lot to do with how  libc manages memory.  If it refuses to return memory back to OS, then it will still eat up memory footprint.   But at least this is a workable, simple change. :)

src/CPUPipe.cpp
```diff
@@ -348,6 +348,24 @@ void CPUPipe::push_input_convolution(unsigned int /*filter_size*/,
...
+void CPUPipe::move_residual(unsigned int /*filter_size*/,
+                            unsigned int /*channels*/,
+                            unsigned int /*outputs*/,
+                            std::vector<float>&& weights_1,
+                            std::vector<float>&& means_1,
+                            std::vector<float>&& variances_1,
+                            std::vector<float>&& weights_2,
+                            std::vector<float>&& means_2,
+                            std::vector<float>&& variances_2) {
+    m_conv_weights.push_back(std::move(weights_1));
```
gcp: This is the case where you *really* want emplace_back, isn't it? So you get a move constructor, not a copy constructor...?

gcp: Also fix the input convolution?

src/ForwardPipe.h
```diff
@@ -49,6 +49,22 @@ class ForwardPipe {
                                const std::vector<float>& means_2,
                                const std::vector<float>& variances_2) = 0;
 
+    // have a better way to deal with situations when caller doesn't need the prior net?
+    // implement this.
+    virtual void move_residual(unsigned int filter_size,
```
gcp: Can't you support both cases by the combination of emplace_back and the universal references?

src/Network.cpp
```diff
@@ -499,19 +509,31 @@ void Network::initialize(int playouts, const std::string & weightsfile) {
...
     if (use_selfcheck) {
-        m_forward_cpu = init_net(std::make_unique<CPUPipe>());
+        m_forward_cpu = init_net(std::make_unique<CPUPipe>(), true);
     }
 #else
     (void)use_selfcheck;
 #endif
+
+    // ForwardPipe instances are initialized.  We don't need the residual tower nets anymore
+    // clear() doesn't guarantee memory freeing, so this is an alternative...
```
gcp: Use clear() + shrink_to_fit().

gcp:
Apparently another trick is to make the functions take the vectors by value(!) and then emplace_back(std::move(value_arg)) into the members.

This will copy in normal cases, and move when the vector argument is std::moved from the callers' site.

It makes no sense to me but then again so does most of C++ these days.

src/CPUPipe.cpp
ihavnoid: emplace_back(std::move(…)) works.
emplace_back(…) still consumes peak memory footprint.

Also fixed convolution input but the gains should be much smaller.

src/Network.cpp
ihavnoid: Got it.

ihavnoid:
Hmm... call by value + std::move might work.  I will just rewrite it that way as that seems to be the most c++14-ish way.

Makes no sense (haha) but it seems that will work. :)

ihavnoid:
Thinking a bit more, I think I was trying to solve the problem using a solution that is too complicated.  Pushed a change that simply stores the pointer of the weights on CPUPipe.

gcp:
Seems like we'd want a separate network-data class and pass shared ptrs to that?

ihavnoid:
3rd attempt pushed.  This time everything is not encapsulated into a ForwardPipeWeights subclass, and shared_ptr<const ForwardPipeWeights> is passed around.

Issue #1760, opened 2018-08-23
Tree search batch support


bood:
I spent some time trying to make batch work at tree search side, but runs into bugs I cannot solve for now. so I'm looking for some help here.

Basic idea is to queue all the forward requests into one list and let each GPU pull them when idle. (see my commit message too). I added tons of printfs for debugging purpose too.

But when I test, it's always the second batch forward got failed and crashed the whole program. Example output:
```
Boods-MBP:bin Bood$ ./leelaz -w w141.gz -p 40 -t 4 --benchmark --noponder --precision single
forward wait
1: found 1 entries (total 1)
1: left 0 entries
1: prepare inputs from 1 entries
1: copy 6498 6498
1: forwarding 0, 1
opencl init 1
opencl enqueue buffer 1
opencl queue finish 1 2888 1444
opencl output 1
1: gather outputs
1: notify 0
forward done
NN eval=1.000000
Playouts: 0, Win: 100.00%, PV:
2: found 1 entries (total 1)
forward wait
2: left 0 entries
2: prepare inputs from 1 entries
2: copy 6498 6498
2: forwarding 0, 1
opencl init 1
opencl enqueue buffer 1
opencl queue finish 1 2888 1444
Abort trap: 6
```

It seems to be related to OpenCLContext, since if I change the context every time, it can survive for more batches.

Code at (I added tons of printfs for debugging too):
https://github.com/bood/leela-zero/tree/batch-prototype

@ihavnoid @Ttl You probably know this area better, do you have any ideas?

gcp:
Remark unrelated to the OpenCL stuff: If you're implementing this, and waiting till you have a full batch (i.e. not time based) you may want to use the trick to avoid UCT-selecting any node where expansion is currently happening, to ensure the batch fills with different positions.

bood:
It does not wait for a full batch now (I tried but it seems to be too tricky, it's hard to know when to stop). Do you think it should?

I feel that if tree expansion is quick enough, there should always be enough requests to merge to a full batch. If not, you can barely gain performance even with full batch.

gcp:
>Do you think it should?

I don't know, it's something to test. But without the above trick or at least counting the number of threads that are in a lock, I think it won't work.

ihavnoid:
I picked up the branch, but it is terminating after running -p 4000.  On the destructor of OpenCLScheduler.  So probably a bit different.

```
[...]

PR #1803, opened 2018-09-04
Tree search batch support

bood:
See details in #1760 and let's discuss more with the code available now.

@gcp Please take a look if there is major flaws in this implementation. Currently batching in pondering is explicitly disabled.


src/UCTSearch.cpp
```diff
@@ -663,6 +669,22 @@ void UCTSearch::increment_playouts() {
...
+    Time elapsed;
+    int elapsed_centis = Time::timediff_centis(start, elapsed);
+    while(!predict_stop_thinking(elapsed_centis, time_for_move)) {
+        Time elapsed2;
+        elapsed_centis = Time::timediff_centis(start, elapsed2);
+        std::this_thread::sleep_for(200ms);
+    }
+    m_network.set_batching(false);
+}
+
```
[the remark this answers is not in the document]
bood: It's actually a brilliant idea! But it requires all threads knows when to stop. Current implementation only main thread knows when to stop thinking, while other threads just pull is_running. So if main thread waits in the queue, it'll stuck.

I still like this idea, we could improve the code a bit to see if we can do this.

src/UCTSearch.cpp
```diff
@@ -759,6 +788,8 @@ int UCTSearch::think(int color, passflag_t passflag) {
 }
 
 void UCTSearch::ponder() {
+    m_network.set_batching(false);
+
```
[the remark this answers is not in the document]
bood: I just didn't implement it yet. The predict stop thing is not obvious if we use the timer approach.

src/OpenCLScheduler.cpp
```diff
@@ -182,28 +195,92 @@ template <typename net_t>
...
+                    auto end = begin;
+                    std::advance(end, count);
+                    std::move(begin, end, std::back_inserter(inputs));
+                    m_forward_queue.erase(begin, end);
+                    break;
+                }
+                else {
+                    // wait with pred does not work anymore, not sure why yet
+                    m_cv.wait(lk);
+                    //m_cv.wait(lk, [this](){ return !m_running || !m_forward_queue.empty(); });
```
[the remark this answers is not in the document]
bood: It'll stuck after my full batch changes, not sure why. It was fine before my changes.
And I'm not sure why we need the predicts here, since when notified, the queue should never be empty.

src/OpenCLScheduler.cpp
```diff
@@ -86,28 +88,39 @@ void OpenCLScheduler<net_t>::initialize(const int channels) {
...
-        auto opencl = std::make_unique<OpenCL<net_t>>();
-        auto net = std::make_unique<OpenCL_Network<net_t>>(*opencl);
-        opencl->initialize(channels, gpu, silent);
-        m_opencl.push_back(std::move(opencl));
-        m_networks.push_back(std::move(net));
+        {
+            auto opencl = std::make_unique<OpenCL<net_t>>();
+            auto net = std::make_unique<OpenCL_Network<net_t>>(*opencl);
+            opencl->initialize(channels, gpu, silent, cfg_batch_size);
+            m_opencl.push_back(std::move(opencl));
```
gcp: nit: emplace_back (same for all other uses)

src/OpenCLScheduler.cpp
```diff
@@ -182,28 +195,92 @@ template <typename net_t>
...
+        {
+            std::unique_lock<std::mutex> lk(m_mutex);
+            while (true) {
+                if(!m_running) return;
+                count = std::min(m_forward_queue.size(), size_t(cfg_batch_size));
+                if (count < cfg_batch_size && m_is_batching) {
+                    count = 0;
+                }
+                if (count > 0) {
+                    auto begin = m_forward_queue.begin();
```
gcp: nit: begin(x) over x.begin() (same for all other uses, end(), etc)

src/UCTSearch.cpp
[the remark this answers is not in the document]
gcp: The main thread could signal a bunch of notification vars, and this sleep_for could be a timed_wait for those.

src/config.h
```diff
@@ -87,8 +87,7 @@ static constexpr auto POTENTIAL_MOVES = NUM_INTERSECTIONS + 1; // including pass
 
 /* Maximum supported batch size for OpenCL.
  */
-static constexpr auto MAX_BATCH = 1;
-static_assert(MAX_BATCH == 1, "MAX_BATCH != 1 not implemented");
+static constexpr auto MAX_BATCH = 8;
```
gcp: Do we have an actual hard limit? Shouln't we just check cfg_batch_size instead?

OpenCL allocation might need to recheck that the available memory is large enough if cfg_batch_size changes. 

Defaulting it to some larger value is probably going to break too many low RAM cards.

src/OpenCLScheduler.cpp
```diff
@@ -182,28 +195,92 @@ template <typename net_t>
...
-                m_context_pool[queue_num].pop_front();
-                break;
+        std::unique_lock<std::mutex> lk(m_mutex);
+        m_forward_queue.push_back(entry);
+    }
+    m_cv.notify_one();
+    entry->cv.wait(lk);
+}
+
+std::atomic<size_t> batch_stats[2];
```
ihavnoid: Do we want to keep this in global context?

src/config.h
Ttl: Extra memory use from batching is only about 3 MB * batch_size for single precision and half that for half precision. I think that MAX_BATCH can be safely removed and replaced with `cfg_batch_size` in `OpenCL.cpp`.

src/OpenCLScheduler.cpp
bood: No, I'll remove that. But it's convenient for reviewers so I leave it here for now.

src/config.h
bood: I believe MAX_BATCH is used no where already. I'll remove this.

Ttl:
This deadlocks for me at the fourth move using command: `./leelaz -w d351f06e446ba10697bfd2977b4be52c3de148032865eaaf9efc9796aea95a0c.gz -t 16 -d -v 800 --noponder --precision half --batchsize 8` and typing `auto`.

bood:
@Ttl It's related to tree re-use. If a new move already has enough visits, the predict timer turns off batching too soon, even before the main thread finished preparing the root node.

src/UCTSearch.cpp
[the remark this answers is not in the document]
bood: After some thoughts, it appears to me simply turning off batching after main thread exits the loop would be enough. Only thing needs to worry is when node children are not enough to form a batch, we can just turn off batching for this case since move choices are not that many anyway.

src/OpenCLScheduler.cpp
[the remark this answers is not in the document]
bood: I think the inner loop already covered this, so we don't need it anyway.

bood:
Pushed a bunch of fixes. Addressed most of the comments, simplify the waiting logic (timer removed). @Ttl I believe the problem has been solved. Please verify if you want.

I'm also going to help with your CuDNN porting after this is merged. I see a lot of redundant code in CuDNNScheduler and OpenCLScheduler, I think a new base class like BatchScheduler code help with the code sharing.

Ttl:
I can verify that it doesn't deadlock anymore with the latest changes.

bood:
One more thing to notice is LZ currently only allows 64 threads for max, it looks like a reason why we don't see a performance gain with larger batch size.

This is what I got on a Titan V with w173  

  | name             | net        | perf |
  |------------------|------------|------|
  | b=4 t=24         |   w173  | 1540 |
  | b=4 t=64         |   w173  | 1639 |
  | b=8 t=64         |   w173  | 1669 |
  | b=6 t=64         |   w173  | 1790 |

roy7:
My 32 batch full tune only got to 7571 GFLOPS so maybe I should save those earlier results, lol.

roy7:
Does --benchmark not work with the batching PR? 

> ./leelaz -w ~/code/leela-zero/src/33986b7f9456660c0877b1fc9b310fc2d4e9ba6aa9cee5e5d242bd7b2fb1b166.gz --precision half --benchmark
> Thinking at most 2678400.0 seconds...
> NN eval=0.533232
> 

Just hangs, no cpu or gpu being used.

roy7:
It doesn't hang if I specify a # of threads. I can't specify a number under the batch size. However default threads is 2. So default threads < batch size is maybe the hang cause?

Ttl:
Some tests using equal nodes with different batch size and threads against next with "-t 1":

```
Command line: " -g -d -t 8 -r 10 --noponder -w d351f06e446ba10697bfd2977b4be52c3de148032865eaaf9efc9796aea95a0c.gz --precision half --batchsize 8"

lz_batch8_t16 v lz_next (20/400 games)
board size: 19   komi: 7.5
               wins              black          white      avg cpu
lz_batch8_t16     1  5.00%       1   10.00%     0  0.00%     81.16
lz_next          19 95.00%       10 100.00%     9 90.00%     81.36
                              11  55.00%     9 45.00%

lz_batch4_t4 v lz_next (20/400 games)
board size: 19   komi: 7.5
lz_batch4_t4      5 25.00%       3  30.00%     2 20.00%     81.12
lz_next          15 75.00%       8  80.00%     7 70.00%     88.54
                              11 55.00%     9 45.00%

lz_batch8_t8 v lz_next (20/400 games)
board size: 19   komi: 7.5
lz_batch8_t8      8 40.00%       4  40.00%     4  40.00%     78.97
lz_next          12 60.00%       6  60.00%     6  60.00%     83.73
                                 10 50.00%     10 50.00%
```

Number of games is low so there is pretty high variance, but batching is clearly weaker especially at high batch sizes. I guess it makes sense to use as low amount of threads as necessary. With too many threads there are many positions waiting for the batch.

roy7:
genmove on empty board /next is ~3400 n/s but using 4 batch size with 4 threads is ~5200 n/s. So that's a very significant speed boost. Although @ttl's results are concerning. :/

This is using a 20b network.

ihavnoid:
I propose trying this:

https://github.com/ihavnoid/leela-zero/tree/batch-full

What I did was have OpenCLScheduler predict how long it wants to wait for the batch to fill up - if the prediction misses (and the batch doesn't fill up) then we do a scalar evaluation and wait longer next time.  If the prediction misses (and the batch over-fills) then we do a batch evaluation and wait shorter next time.

The benefit is that UCTSearch no longer needs to worry when it needs to turn on/off batching to avoid deadlock (I believe we can have pretty weird corner cases here, especially on endgame when we only have one or two legal moves) and doesn't need to be aware of any batching behavior.

Ttl:
With equal time batching seems to be stronger:

```
lz_batch v lz_next (27/400 games)
board size: 19   komi: 7.5
lz_batch     18 66.67%       8  57.14%     10 76.92%   1352.06
lz_next       9 33.33%       3  23.08%     6  42.86%   1165.80
                             11 40.74%     16 59.26%
```

This was with `--batchsize 5 -t 6`. 10s per move. Otherwise same command as before.

Batch size 5 results in minimal zero padding in sgemm and seemed to perform well.

roy7:
Interesting, with 5 batch 5 thread I used about 50% GPU for about 5200 n/s but when I use 5 batch 6 thread I used about 70% GPU with about 6800 n/s. 6800 n/s is the highest I think I can remember seeing before.

roy7:
I tried some high batch values and would get an OpenBLAS error message and core dump. Digging around I found issues like https://github.com/xianyi/OpenBLAS/issues/1735 which led me to try upgrading from 0.3.2 to 0.3.3 and that fixed the problem.

roy7:
@bood mentioned in Discord that the sweet spot seems to be 2x-3x as many threads as batches. I tried 64/64 at first since Leela's thread cap is 64, but didn't get more than about 5K nps. (I can get about 6K nps on /next with 2 threads as-is.)

I tried 8 batch with 16 threads, and hit an amazing 9100 nps. Perhaps the 64 failure was lack of threads. Full tune 5327.2 GFLOPS.

I also tried 32 batch 64 threads and hit 8900 nps. Full tune 9274.0 GFLOPS.

lp200:
2 core cpu can not run with default parameters.

log
>Threads number = 2 must be larger than batch size = 4

bood:
@lp200 You must make threads number >= batch number to make it work, default batch size is 4. e.g. add `-t 8`

lp200:
I know:smiley:
I think that it must work with default parameters.

ihavnoid:
My suggestion is to clamp the batch size to the number of threads.  @bood what do you think?

nerai:
Just for your information, I've worked on this patch for a while and changed the architecture a lot. My results so far, compared to it as well as to LZ-next with increased thread count, are promising. But it's at too early a stage to be confident. I'll post if/when I get more results.

bood:
@gcp Message received. I think we also need to solve some corner case before this can be merged. As @ihavnoid described, with the current implementation (without an extra timer thread), when the unexpanded node number is less than a batch size (in the very end of the game), it could be stuck.

However, I'm not a fun of @ihavnoid 's proposal either, an self-adjusting time wait in critical path just does not seem right to me. But do you have any performance data for your proposal?

@nerai Looking forward to see your implementation. If it turns out to be better, please do submit another PR.

ihavnoid:
@bood - It mostly filled the batches while hitting 100% GPU utilization, so I can say that it's at least not too bad.  I didn't measure the strength but I anticipate it won't be weaker than any other heuristic here.

Self-adjusting isn't so bad simply because the NN evaluation execution time is mostly consistent - the runtime is independent of input data so predictions will (IMHO) stabilize quite quickly.

bood:
> I didn't measure the strength but I anticipate it won't be weaker than any other heuristic here.

You don't have to measure the strength, just `nodes per second` should be enough.

ihavnoid:
Further tweaked my branch (https://github.com/ihavnoid/leela-zero/tree/batch-full) including putting in reasonable default batch sizes per thread counts.  Showing reasonable throughput for both single and multi-GPUs with roughly 20% more evals on 10 thread / 5 batch / 2 GPU configs.

However, it isn't showing that the engine is stronger when configured to do 10 second/move:

```
36/80 games played

                   A     B
A leelaz-baseline       18-18
B leelaz-batch    18-18
```

This is using 10 threads for both baseline and batched binaries.  Did confirm that the batched runs have more evals per second but still, strength is roughly equal.  This is using the latest 40x256 net.

nerai:
@ihavnoid Thank you for confirming this. So it seems this was no fluke.

The following is speculation.

At first I thought this is due to the (intentional) redirection of parallel search threads, i.e. the most interesting branches are not looked at because they are being processed already. As a consequence, more low-importance branches are processed and thus take space in the evaluation. While throughput increases, this also increases the latency until good branches are considered. Probably, this averages out to no difference in effective strength (of batched vs nonbatched).

However, this whole phenomenon should only be noticeable with quite fast times. At 10s, many (>1000) batches will be processed. So once the tree is a couple steps deep, almost all effort should go into useful nodes. Is that right? Are VL possibly an issue here?

I hope that made sense. I'm not very familiar with UCT search.

nerai:
In the fork of @bood, there is waiting for the batch to fill up. Why is that?

My understanding is that immediate processing is strictly superior. The GC would never be idle then, and will always process as many data as possible. If one batch does not get filled up, the next one will be more full. It seems like a standard case of a provably-best greedy algorithm.

nerai:
@ihavnoid Did you mean to refer to https://github.com/ihavnoid/leela-zero/tree/batch-full or https://github.com/ihavnoid/leela-zero/tree/full-batch ? The latter seems right.

ihavnoid:
Yeah, the latter one.

ihavnoid:
After running 80 games (10 threads, 2 GPU, 10 seconds/move) it seems that there is an improvement:

```
80/80 games played

                   A     B
A leelaz-baseline       31-49
B leelaz-batch    49-31
```
This is https://github.com/ihavnoid/leela-zero/tree/full-batch vs. the current next branch.

gcp:
>At first I thought this is due to the (intentional) redirection of parallel search threads, i.e. the most interesting branches are not looked at because they are being processed already. As a consequence, more low-importance branches are processed and thus take space in the evaluation. While throughput increases, this also increases the latency until good branches are considered. Probably, this averages out to no difference in effective strength (of batched vs nonbatched).

Yes, this is the main concern with very aggressive batching. Just looking at NPS is not enough because the search becomes less efficient.

>However, this whole phenomenon should only be noticeable with quite fast times. At 10s, many (>1000) batches will be processed. So once the tree is a couple steps deep, almost all effort should go into useful nodes. Is that right? Are VL possibly an issue here?

It's true that on a longer time control, you'll have a bigger tree and hence also more opportunity for parallelism, but the critical path in the tree might still be quite narrow, so I wouldn't be so sure about this.

gcp:
Closing this. I understand there's a new try now with splitting up the MCTS select->update phases to have arbitrary sized batches with a single thread.

PR #1814, opened 2018-09-07
Always expect 2 arguments after "play" command.

TFiFiE:
As [demanded by GTP](https://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html#SECTION00073300000000000000), improving the input handling of `GameState::play_textmove` in the process (which now would crash if given a pass or resignation).


gcp:
@TFiFiE Are you willing to add unit tests for this?

PR #1818, opened 2018-09-08
OpenCL : don't initialize weights using CL_MEM_COPY_HOST_PTR

ihavnoid:
On many OpenCL implementations, CL_MEM_COPY_HOST_PTR | CL_MEM_READ_ONLY keeps a backing store of the data.  Hence we need to keep two copies of the weight (one on host, one on GPU) spending more memory.

This was done so that the driver can swap in/swap out memory from GPU when GPU memory is in a pinch.  Based on how leela-zero works, I believe this behavior won't help much since we are re-scanning through all the weights every millisecond or so (it will add too much pressure to the PCI-e bus)

So, instead of using CL_MEM_COPY_HOST_PTR, just create a buffer and then copy the weights using enqueueWriteBuffer - saves about 300MB of host memory on a 256x40 weight.

### DIFF
--- a/src/OpenCL.cpp
+++ b/src/OpenCL.cpp
@@ -112,12 +112,16 @@ void OpenCL_Network<net_t>::add_weights(size_t layer,
     }
 
     auto weightSize = size * sizeof(net_t);
-    m_layers.back().weights.emplace_back(
+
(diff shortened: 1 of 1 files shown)


ihavnoid:
Just in case, I measured the performance and no measurable difference.

Issue #1808, opened 2018-09-05
Can't compile next commit 73f1f93 on VS2007


jun63cn:
FastBoard.cpp
  FastState.cpp
  FullBoard.cpp
  GameState.cpp
  GTP.cpp
  KoState.cpp
  Leela.cpp
  Unknown compiler version - please run the configure tests and report the results
d:\downloads\leela-zero-73f1f934169a62d7b054aefccecb41029de0fc5f\src\leela.cpp(213): warning C4267: “*=”: 从“size_t”转换到“int”，可能丢失数据
  Network.cpp
  Unknown compiler version - please run the configure tests and report the results
d:\downloads\leela-zero-73f1f934169a62d7b054aefccecb41029de0fc5f\msvc\packages\boost.1.65.1.0\lib\native\include\boost\spirit\home\x3\support\subcontext.hpp : warning C4819: 该文件包含不能在当前代码页(936)中表示的字符。请将该文件保存为 Unicode 格式以防止数据丢失
d:\downloads\leela-zero-73f1f934169a62d7b054aefccecb41029de0fc5f\msvc\packages\boost.1.65.1.0\lib\native\include\boost\spirit\home\support\char_encoding\iso8859_1.hpp : warning C4819: 该文件包含不能在当前代码页(936)中表示的字符。请将该文件保存为 Unicode 格式以防止数据丢失
d:\downloads\leela-zero-73f1f934169a62d7b054aefccecb41029de0fc5f\src\network.cpp(198): warning C4244: “=”: 从“__int64”转换到“int”，可能丢失数据
d:\downloads\leela-zero-73f1f934169a62d7b054aefccecb41029de0fc5f\src\network.cpp(373): warning C4267: “参数”: 从“size_t”转换到“const int”，可能丢失数据
d:\downloads\leela-zero-73f1f934169a62d7b054aefccecb41029de0fc5f\src\network.cpp(380): warning C4267: “参数”: 从“size_t”转换到“const int”，可能丢失数据
d:\downloads\leela-zero-73f1f934169a62d7b054aefccecb41029de0fc5f\src\network.cpp(406): warning C4267: “参数”: 从“size_t”转换到“const int”，可能丢失数据
[...]

Issue #771, opened 2018-01-28
Hindsight Experience Replay and komi


Hersmunch:
So I came across a paper on [Hindsight Experience Replay](https://arxiv.org/pdf/1707.01495.pdf) from watching this [Two Minute Papers video](https://www.youtube.com/watch?v=Dvd1jQe3pq0). It has got me wondering whether it could be applied to LZ. Specifically, whether the use of variable komi would allow for each self-play game to provide two games' worth of training information.

I admit I don't know what research is out there about training to cope with variable komi so please tell me if this has already been considered and point me towards it so I can find out more. 

In terms of using HER for LZ my thoughts are that as well as the existing game result being used for training, a second result could be produced by using a different komi when scoring, such that the loser is given just enough points to win the game. Maybe this algorithm description from the paper would help:

![algorithm](https://i.imgur.com/Kvibj4W.png?1)

It does sound like the situation they say it can be used for is different to ours in that we always have a winner in each game whereas their reward is sparse. I'm also sceptical about whether a single komi input could be used as a "goal state" and although I can conceptually see how this approach could help the value output understand scoring, I'm finding it harder to see how it could help the policy by rewarding bad moves. 

So yeah, any thoughts? I'm not specifically arguing for or against adding variable komi but could this be applied? Please tell me that I'm barking up the wrong tree if that is the case!

Edit: OK, nevermind, I should have read what "[off-policy vs on-policy](https://stats.stackexchange.com/questions/184657/what-is-the-difference-between-off-policy-and-on-policy-learning)" is first. Oh well, I hope the links are of interest to some people anyway.

PR #1825, opened 2018-09-09
Encode komi in the side to move NN input planes

Hersmunch:
To allow for games to be played with selectable but constant komi.
I have also updated the training data output to change the side to move indication such that it encodes the komi. I haven't tried touching the python code to make use of the komi value during training (I have no way to run it) but have confirmed that the dumped data has no changed when 7.5 komi is used (will also work for -7.5).
Hopefully this will be a small enough change to help start getting #1772 pulled in.

Note that I haven't put any code in to cope with komi being changed mid-game and that data from `save_training` without this change cannot be loaded with the change.
Also I backed out #1618 locally due to #1808, to be able to compile.

### DIFF
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -99,7 +99,7 @@ float Network::benchmark_time(int centiseconds) {
     std::atomic<int> runcount{0};
 
     GameState state;
-    state.init_game(BOARD_SIZE, 7.5);
+    state.init_game(BOARD_SIZE, TRAINED_UNIT_KOMI);
 
     // As a sanity run, try one run with self check.
     // Isn't enough to guarantee correctness but better than nothing,
@@ -915,6 +915,10 @@ void Network::fill_input_plane_pair(const FullBoard& board,
     }
 }
 
+float Network::get_normalised_komi(const GameState* const state) {
+    return 0.5f + (state->get_komi() / (2.0f * TRAINED_UNIT_KOMI));
+}
+
 std::vector<float> Network::gather_features(const GameState* const state,
                                             const int symmetry) {
     assert(symmetry >= 0 && symmetry < NUM_SYMMETRIES);
@@ -929,9 +933,6 @@ std::vector<float> Network::gather_features(const GameState* const state,
     const auto white_it = blacks_move ?
                           begin(input_data) + INPUT_MOVES * NUM_INTERSECTIONS :
                           begin(input_data);
-    const auto to_move_it = blacks_move ?
-        begin(input_data) + 2 * INPUT_MOVES * NUM_INTERSECTIONS :
-        begin(input_data) + (2 * INPUT_MOVES + 1) * NUM_INTERSECTIONS;
 
     const auto moves = std::min<size_t>(state->get_movenum() + 1, INPUT_MOVES);
     // Go back in time, fill history boards
@@ -943,7 +944,15 @@ std::vector<float> Network::gather_features(const GameState* const state,
                               symmetry);
     }
 
-    std::fill(to_move_it, to_move_it + NUM_INTERSECTIONS, float(true));
+    const auto black_to_move_it = begin(input_data) + 2 * INPUT_MOVES * NUM_INTERSECTIONS;
+    const auto white_to_move_it = black_to_move_it + NUM_INTERSECTIONS;
+    const auto pos_norm_komi = get_normalised_komi(state);
+    const auto neg_norm_komi = 1.0f - pos_norm_komi;
+    const auto black_norm_komi = blacks_move ? pos_norm_komi : neg_norm_komi;
+    const auto white_norm_komi = blacks_move ? neg_norm_komi : pos_norm_komi;
+
+    std::fill(black_to_move_it, black_to_move_it + NUM_INTERSECTIONS, black_norm_komi);
+    std::fill(white_to_move_it, white_to_move_it + NUM_INTERSECTIONS, white_norm_komi);
 
     return input_data;
 }
--- a/src/Training.cpp
+++ b/src/Training.cpp
@@ -68,7 +68,8 @@ std::ostream& operator <<(std::ostream& stream, const TimeStep& timestep) {
     stream << timestep.net_winrate << ' ';
     stream << timestep.root_uct_winrate << ' ';
     stream << timestep.child_uct_winrate << ' ';
-    stream << timestep.bestmove_visits << std::endl;
+    stream << timestep.bestmove_visits << ' ';
+    stream << timestep.stm_komi << std::endl;
     return stream;
 }
 
@@ -92,6 +93,7 @@ std::istream& operator>> (std::istream& stream, TimeStep& timestep) {
     stream >> timestep.root_uct_winrate;
     stream >> timestep.child_uct_winrate;
     stream >> timestep.bestmove_visits;
+    stream >> timestep.stm_komi;
     return stream;
 }
 
@@ -164,9 +166,17 @@ TimeStep::NNPlanes Training::get_planes(const GameState* const state) {
     return planes;
 }
 
-void Training::record(Network & network, GameState& state, UCTNode& root) {
+float Training::get_stm_komi(const GameState* const state) {
+    const auto komi = Network::get_normalised_komi(state);
+    const auto stm_komi =
+        (FastBoard::BLACK == state->board.get_to_move() ? 1.0f - komi : komi);
+    return stm_komi;
+}
+
+void Training::record(Network& network, GameState& state, UCTNode& root) {
     auto step = TimeStep{};
     step.to_move = state.board.get_to_move();
+    step.stm_komi = get_stm_komi(&state);
     step.planes = get_planes(&state);
 
     const auto result = network.get_output(
@@ -263,9 +273,8 @@ void Training::dump_training(int winner_color, OutputChunker& outchunk) {
             out << plane[plane.size() - 1];
             out << std::dec << std::endl;
         }
-        // The side to move planes can be compactly encoded into a single
-        // bit, 0 = black to move.
-        out << (step.to_move == FastBoard::BLACK ? "0" : "1") << std::endl;
+        // The side to move planes are used to encode the komi
+        out << step.stm_komi << std::endl;
         // Then a POTENTIAL_MOVES long array of float probabilities
         for (auto it = begin(step.probabilities);
             it != end(step.probabilities); ++it) {
@@ -340,6 +349,7 @@ void Training::process_game(GameState& state, size_t& train_pos, int who_won,
 
         auto step = TimeStep{};
         step.to_move = to_move;
+        step.stm_komi = get_stm_komi(&state);
         step.planes = get_planes(&state);
 
         step.probabilities.resize(POTENTIAL_MOVES);
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -577,7 +577,7 @@ void GTP::execute(GameState & game, const std::string& xinput) {
     } else if (command.find("komi") == 0) {
         std::istringstream cmdstream(command);
         std::string tmp;
-        float komi = 7.5f;
+        float komi;
         float old_komi = game.get_komi();
 
         cmdstream >> tmp;  // eat komi
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -532,8 +532,7 @@ int main(int argc, char *argv[]) {
     auto maingame = std::make_unique<GameState>();
 
     /* set board limits */
-    auto komi = 7.5f;
-    maingame->init_game(BOARD_SIZE, komi);
+    maingame->init_game(BOARD_SIZE, TRAINED_UNIT_KOMI);
 
     if (cfg_benchmark) {
         cfg_quiet = false;
--- a/src/Network.h
+++ b/src/Network.h
@@ -95,6 +95,7 @@ public:
     static void show_heatmap(const FastState * const state,
                              const Netresult & netres, const bool topmoves);
 
+    static float get_normalised_komi(const GameState* const state);
     static std::vector<float> gather_features(const GameState* const state,
                                               const int symmetry);
     static std::pair<int, int> get_symmetry(const std::pair<int, int>& vertex,
--- a/src/SGFTree.cpp
+++ b/src/SGFTree.cpp
@@ -55,7 +55,7 @@ void SGFTree::init_state() {
     // Initialize with defaults.
     // The SGF might be missing boardsize or komi
     // which means we'll never initialize properly.
-    m_state.init_game(std::min(BOARD_SIZE, 19), KOMI);
+    m_state.init_game(std::min(BOARD_SIZE, 19), TRAINED_UNIT_KOMI);
 }
 
 const KoState * SGFTree::get_state(void) const {
@@ -157,7 +157,7 @@ void SGFTree::populate_states() {
         strm >> bsize;
         if (bsize == BOARD_SIZE) {
             // Assume default komi in config.h if not specified
-            m_state.init_game(bsize, KOMI);
+            m_state.init_game(bsize, TRAINED_UNIT_KOMI);
             valid_size = true;
         } else {
             throw std::runtime_error("Board size not supported.");
--- a/src/Training.h
+++ b/src/Training.h
@@ -49,6 +49,7 @@ public:
     NNPlanes planes;
     std::vector<float> probabilities;
     int to_move;
+    float stm_komi;
     float net_winrate;
     float root_uct_winrate;
     float child_uct_winrate;
@@ -91,6 +92,7 @@ public:
 
 private:
     static TimeStep::NNPlanes get_planes(const GameState* const state);
+    static float get_stm_komi(const GameState* const state);
     static void process_game(GameState& state, size_t& train_pos, int who_won,
                              const std::vector<int>& tree_moves,
                              OutputChunker& outchunker);
--- a/src/config.h
+++ b/src/config.h
@@ -43,7 +43,7 @@
 
 /*
  * BOARD_SIZE: Define size of the board to compile Leela with, must be an odd
-   number due to winograd tiles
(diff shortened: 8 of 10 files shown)


Hersmunch:
Thanks for having a look.
> Nice! This allows training games and data be generated with different komi.

Yeah, allowing autogtp to send gtp commands is probably something I could look at. I might take a while though.
> For training support, it seems we just need to...

:) I'm pretty sure I saw somewhere in the tfprocess code that treats the stm planes as uint8s.. I doubt it is too much work for someone who knows what they are doing.

> I hope we first train the network with post-endgame positions (where all territories are determined) with different komi input to make the network's winrate assessments close to reality at least near the endgame

I like the idea. I'm not sure how we would ensure the end position sampling scheme is unbiased though. Sounds similar to #771.

alreadydone:
I guess we can just replace `uint8` in https://github.com/gcp/leela-zero/blob/b290f47434563b74d814f4fa910b429dea97e2de/training/tf/tfprocess.py#L152 by `float32` since things are converted to float at L156 anyway. @Ttl Does that seem right? I had trouble making tensorflow work on my machine, so I'm also unable to test for now.
After fractional stm support is ready, probably @pangafu will be willing to train networks with data with different komi input (he's now actively training LM weights again and concerned about handicap performance).
Possible sources of post-endgame positions are: LZ no-resign training games, and online/human games that ends with a score. For LZ games we can probably use a pass by either side as the sign that the game is over, and score all subsequent positions with Sabaki, Zen, or whatever; we will only be modifying the stm value of existing training data, but we may want to exclude the policy loss when stm is not 0 or 1 (since the visit counts weren't generated when the stm is not 0 or 1). For online/human games we may just modify dump_supervised to set stm to values corresponding to score+0.5 and score-0.5 (these are the most natural values; one may also sample from some distribution of stm values but that's complicated and probably unnecessary) and winner to B and W; however, probably only the final position can be used.
I'd like to see networks be trained with different komi input since it potentially allows the engine play/analyze under different komi (a desired feature) and allows my dynamic komi code be simplified significantly (no need of separate stm_komi and opp_komi, for example) to make it easier to get reviewed and merged.

gcp:
Code looks OK to me, question is what to do with the compatibility with the current training code.

I'm already slightly surprised this works because it looked like it would write 1.0 for side to move and I wasn't sure the python code could accept that.

Hersmunch:
I was expecting it to output 1.0 and 0.0 and was surprised when I saw just 1 and 0 respectively when I was testing it. Having a quick google it seems that this is the default behaviour when a precision isn't specified. Lucky for us.

Considering the SAI approach in #1835, would you prefer this be made optional/configurable somehow?

wpstmxhs:
This change may cause a problem with NNCache.
When a user wants to change komi during a game, the results stored in NNCache may be inconsistent with new komi.

pangafu:
@alreadydone @Hersmunch 
After replace

```
leela-zero/training/tf/tfprocess.py

Line 152 in b290f47

planes = tf.decode_raw(self.planes, tf.uint8) 
by float32 since things are converted to float at L156 anyway.

```
I get error like: 
```
Use the `axis` argument instead
  File "/home/ww/anaconda3/lib/python3.6/site-packages/tensorflow/python/client/session.py", line 1278, in _do_call
    return fn(*args)
  File "/home/ww/anaconda3/lib/python3.6/site-packages/tensorflow/python/client/session.py", line 1263, in _run_fn
    options, feed_dict, fetch_list, target_list, run_metadata)
  File "/home/ww/anaconda3/lib/python3.6/site-packages/tensorflow/python/client/session.py", line 1350, in _call_tf_sessionrun
    run_metadata)
tensorflow.python.framework.errors_impl.InvalidArgumentError: Input to reshape is a tensor with 519840 values, but the requested shape has 2079360
         [[Node: Reshape = Reshape[T=DT_FLOAT, Tshape=DT_INT32, _device="/job:localhost/replica:0/task:0/device:GPU:0"](DecodeRaw/_677, Reshape/shape)]]
         [[Node: Mean_1/_8207 = _Recv[client_terminated=false, recv_device="/job:localhost/replica:0/task:0/device:CPU:0", send_device="/job:localhost/replica:0/task:0/device:GPU:0", send_device_incarnation=1, tensor_name="edge_174187_Mean_1", tensor_type=DT_FLOAT, _device="/job:localhost/replica:0/task:0/device:CPU:0"]()]]

During handling of the above exception, another exception occurred:

  File "../../train/parse.py", line 178, in <module>
    main()
  File "../../train/parse.py", line 174, in main
    tfprocess.process(train_parser, test_parser)
  File "/home/ww/lztrain/train/tfprocess.py", line 418, in process
    losses = self.measure_loss(batch, training=True)
  File "/home/ww/lztrain/train/tfprocess.py", line 406, in measure_loss
    self.winner: batch[2]})
  File "/home/ww/anaconda3/lib/python3.6/site-packages/tensorflow/python/client/session.py", line 877, in run
    run_metadata_ptr)
  File "/home/ww/anaconda3/lib/python3.6/site-packages/tensorflow/python/client/session.py", line 1100, in _run
    feed_dict_tensor, options, run_metadata)
  File "/home/ww/anaconda3/lib/python3.6/site-packages/tensorflow/python/client/session.py", line 1272, in _do_run
    run_metadata)
  File "/home/ww/anaconda3/lib/python3.6/site-packages/tensorflow/python/client/session.py", line 1291, in _do_call
    raise type(e)(node_def, op, message)
tensorflow.python.framework.errors_impl.InvalidArgumentError: Input to reshape is a tensor with 519840 values, but the requested shape has 2079360
         [[Node: Reshape = Reshape[T=DT_FLOAT, Tshape=DT_INT32, _device="/job:localhost/replica:0/task:0/device:GPU:0"](DecodeRaw/_677, Reshape/shape)]]
         [[Node: Mean_1/_8207 = _Recv[client_terminated=false, recv_device="/job:localhost/replica:0/task:0/device:CPU:0", send_device="/job:localhost/replica:0/task:0/device:GPU:0", send_device_incarnation=1, tensor_name="edge_174187_Mean_1", tensor_type=DT_FLOAT, _device="/job:localhost/replica:0/task:0/device:CPU:0"]()]]

Caused by op 'Reshape', defined at:
  File "../../train/parse.py", line 178, in <module>
    main()
  File "../../train/parse.py", line 168, in main
    macrobatch=BATCH_SIZE // RAM_BATCH_SIZE)
  File "/home/ww/lztrain/train/tfprocess.py", line 172, in init
    planes = tf.reshape(planes, (batch_size, 18, 19*19))
  File "/home/ww/anaconda3/lib/python3.6/site-packages/tensorflow/python/ops/gen_array_ops.py", line 6199, in reshape
    "Reshape", tensor=tensor, shape=shape, name=name)
  File "/home/ww/anaconda3/lib/python3.6/site-packages/tensorflow/python/framework/op_def_library.py", line 787, in _apply_op_helper
    op_def=op_def)
  File "/home/ww/anaconda3/lib/python3.6/site-packages/tensorflow/python/util/deprecation.py", line 454, in new_func
    return func(*args, **kwargs)
  File "/home/ww/anaconda3/lib/python3.6/site-packages/tensorflow/python/framework/ops.py", line 3155, in create_op
    op_def=op_def)
  File "/home/ww/anaconda3/lib/python3.6/site-packages/tensorflow/python/framework/ops.py", line 1717, in __init__
    self._traceback = tf_stack.extract_stack()

InvalidArgumentError (see above for traceback): Input to reshape is a tensor with 519840 values, but the requested shape has 2079360
         [[Node: Reshape = Reshape[T=DT_FLOAT, Tshape=DT_INT32, _device="/job:localhost/replica:0/task:0/device:GPU:0"](DecodeRaw/_677, Reshape/shape)]]
         [[Node: Mean_1/_8207 = _Recv[client_terminated=false, recv_device="/job:localhost/replica:0/task:0/device:CPU:0", send_device="/job:localhost/replica:0/task:0/device:GPU:0", send_device_incarnation=1, tensor_name="edge_174187_Mean_1", tensor_type=DT_FLOAT, _device="/job:localhost/replica:0/task:0/device:CPU:0"]()]]

```

alreadydone:
@pangafu Try my branch https://github.com/alreadydone/lz/tree/patch-32 (diffs: https://github.com/gcp/leela-zero/compare/next...alreadydone:patch-32?expand=1#diff-a435af694cbd4a20a19214444357b40d). Haven't tested. If there's any problem, @Ttl can probably help.

pangafu:
@alreadydone 
It report:

```
  File "../../train/parse.py", line 178, in <module>
    main()
  File "../../train/parse.py", line 174, in main
    tfprocess.process(train_parser, test_parser)
  File "/home/ww/lztrain/train/tfprocess.py", line 414, in process
    batch = next(train_data)
  File "/home/ww/lztrain/train/chunkparser.py", line 382, in parse
    for b in gen:
  File "/home/ww/lztrain/train/chunkparser.py", line 367, in batch_gen
    s = list(itertools.islice(gen, self.batch_size))
  File "/home/ww/lztrain/train/chunkparser.py", line 358, in tuple_gen
    yield self.convert_v2_to_tuple(r)
  File "/home/ww/lztrain/train/chunkparser.py", line 266, in convert_v2_to_tuple
    planes = np.unpackbits(np.frombuffer(planes, dtype=np.float32))
ValueError: buffer size must be a multiple of element size
```

pangafu:
I use data with float stm in original train code, it seem can work, but it most likely round float to int,  so the train data I made maybe correct.

alreadydone:
https://github.com/gcp/leela-zero/compare/next...alreadydone:stm4komi?expand=1 seems to be working

Hersmunch:
Thanks for the info. If you don’t mind sharing, I’d like to know more about how you are generating training data. Like the number of games you have generated and the distribution for Komi over the games etc. Thanks in advanced.

pangafu:
@Hersmunch 
Yes, I will explain the data sometime later, I prepare it for about a week, the data is very large and complex..

pangafu:
@Hersmunch I make a brief readme of OZ-serial's data, please refer to 
https://github.com/pangafu/LeelaMasterWeight/blob/master/LeelaMasterOZ.md

src/Training.cpp
```diff
@@ -263,9 +273,8 @@ void Training::dump_training(int winner_color, OutputChunker& outchunk) {
             out << plane[plane.size() - 1];
             out << std::dec << std::endl;
         }
-        // The side to move planes can be compactly encoded into a single
-        // bit, 0 = black to move.
-        out << (step.to_move == FastBoard::BLACK ? "0" : "1") << std::endl;
+        // The side to move planes are used to encode the komi
+        out << step.stm_komi << std::endl;
```
Hersmunch, line 277: @gcp Since this changes the format of the data to a float should the file format version be change to 3 (on line 308)?

Hersmunch:
btw @alreadydone has made a PR on my branch for the training script changes (see https://github.com/Hersmunch/leela-zero/pull/1#pullrequestreview-199397185) :D

Unfortunately I'm not in a position to review them thoroughly so I am wondering if anyone can help with that OR should I just merge them and let the review happen here? I fear the latter might be more awkward if there are several review rounds.

src/Training.cpp
alreadydone: Good point. But is the binary format v2 actually in use other than as a temporary intermediate format? If not we may just change the definition/specification of v2... It's clear that the original v2 format cannot support stm-for-komi.

gcp: I don't think so? We never cycled the training format because that's such a mess during a run.

I've never used the v2 on-disk format, there hasn't been a point for me.

alreadydone:
Oh since the komi can be changed without restarting leelaz, I think we should include komi in the hash for NNCache. (proposed before but I forgot about it)

pangafu:
Maybe the dump code of different komi should changed.
I had made some code do the things bellow, but I don't know is it correct, still under test~

1. Filter all W+R, B+R games, the games left only W+N.n and B+N.n;

2. All games now is make in 7.5 komi, so if a game is W+2.5, when komi set to 7.5, 8.5, 9.5, the white side is still win, so I can make komi<10 data with this game. If a game is B+2.5, when komi set to 7.5, 6.5, 5.5, the black side can still win, so I can make komi>5 data with this game.

Hersmunch:
I've now merged in @alreadydone's changes to the training which I've been told have been tested. But I've then done a little tidying and have no way to test them. So review and testing would be appreciated.

Hersmunch:
> Oh since the komi can be changed without restarting leelaz, I think we should include komi in the hash for NNCache. (proposed before but I forgot about it)

Agreed. I propose that that can be added in a later PR as it isn't required for basic self-play and training. (Also I remember looking at doing this when I was originally looking at this and couldn't figure out how to change the hash function..)

Hersmunch:
@pangafu The approach you describe sounds quite like [Hindsight Experience Replay](https://arxiv.org/abs/1707.01495) - see #771. Here's my very rough understanding (not backed up by any evidence at all!) of why it is likely to not be "the best approach" for training a NN for Go:

1. The best move to make is likely to be dependent upon how much Komi there is. Therefore the search is probably not going to be able to produce the best training target without knowledge of the Komi. This is the "am I winning safely and can therefore play solid moves to secure points" vs "I might be behind so need to make the game more complicated" decision problem. All the training games we have so far "know" to aim for 7.5 Komi and so will guide the search to meet that criteria.
Based on this I'm guessing that by using the approach you are trying out it will effectively "add noise" to the training data because it won't be clear to the NN what it is aiming for. Of course this might cause the training process to tend toward assuming the "worst case" i.e. that the Komi is large or small enough to result in a loss and therefore pick "greedy" actions to try and get enough points to overcome the largest Komi you have trained it with.

2. By reusing the same board positions there will be more correlation in the training data overall, which in turn is more likely to tend toward overfitting.

It also isn't clear to me whether you are using resigned games at all in your training. I think that they can and should still be used (but as they are, without any komi changes) because they provide the wide range of positions which will help to avoid overfitting.

Anyway, I guess I should point out that this PR is aimed at actually playing self-play games with various Komi values so that both the above problems are avoided. The training can then be performed as it is now but with komi as an input. However, there is the issue of deciding what distribution of komi values should be used for self-play games though. SAI took an interesting approach of having a chance of changing komi during self-play games to bring unbalanced positions back to being balanced. It sounds like a "relatively simple" way to "let the training process explore by itself" without having to impose too much upon it.

src/Training.cpp
Hersmunch: Thanks for confirming.

alreadydone:
I think it's a good idea to augment the training data for the value head with komi inputs different from the komi actually used in the games, but also agree that such data probably shouldn't be used to train the policy head. (but note that SAI produces a komi-independent policy (or the policy under a komi that makes the game even))

However I find that there's something not right in pangafu's reasoning that I didn't realize before (when I was first told the methodology). If white wins a game under 7.5 komi, it should have won the game under 8.5, 9.5, ..., 361.5 komi, no matter whether the game is won by resignation or has a score, since increasing komi is always to white's advantage; similarly for black. If the result is W+2.5, you may be tempted to argue that white would still win when komi is set to 6.5 and 5.5, but note that
+ both sides may fail to remove some dead stones
+ black might have given up in the endgame  and play suboptimal moves since it's losing anyway (the winrate may be already low and policy flat if this is a no-resign game)
+ even if both sides fought hard and all dead stones were removed, black may had a move that lead to a sure W+0.5, but instead chose a complicated variation which could lead to a black win if white failed to respond correctly. In this case, white will not win under 6.5 komi.

ivysrono:
+1

src/Network.cpp
```diff
@@ -914,6 +914,11 @@ void Network::fill_input_plane_pair(const FullBoard& board,
     }
 }
 
+float Network::get_normalised_komi(const GameState* const state) {
+    const auto norm_komi = 0.5f + (state->get_komi() / (2.0f * TRAINED_UNIT_KOMI));
+    return norm_komi;
```
TFiFiE: Why not just return the expression outright? Even for the sake of clarity, the identifier is completely redundant.

src/Network.cpp
```diff
@@ -942,7 +944,15 @@ std::vector<float> Network::gather_features(const GameState* const state,
                               symmetry);
     }
 
-    std::fill(to_move_it, to_move_it + NUM_INTERSECTIONS, float(true));
+    const auto black_to_move_it = begin(input_data) + 2 * INPUT_MOVES * NUM_INTERSECTIONS;
+    const auto white_to_move_it = black_to_move_it + NUM_INTERSECTIONS;
+    const auto pos_komi = get_normalised_komi(state);
+    const auto neg_komi = 1.0f - pos_komi;
+    const auto black_komi = blacks_move ? pos_komi : neg_komi;
+    const auto white_komi = blacks_move ? neg_komi : pos_komi;
```
TFiFiE: Conversely, having these variable names end with something like `_norm_komi` would be proper.

pangafu:
@Hersmunch @alreadydone 
Untile now, the test of OZ-serial show it make a great success, OZ13(white) can hadicap zen7(black) with 4 stone at komi 0, with komi engine. 

The color plan komi is awesome~

TFiFiE:
While trying to salvage this pull request, I discovered there might be an issue because #2191 changes the cast of the `planes` variable in `TFProcess`.

If having the cast removed is still no problem now, then https://github.com/TFiFiE/leela-zero/commits/komi is an updated and cleaned up branch containing the changes here.

TFiFiE:
One option could be to bring back the cast with:
```
if self.model_dtype != tf.float32:
    planes = tf.cast(planes, self.model_dtype)
```
and another would be to just change:
`planes = tf.decode_raw(self.planes, tf.float32)`
to:
`planes = tf.decode_raw(self.planes, self.model_dtype)`

Also pinging @godmoves.

ivysrono:
merged to komi not next?

TFiFiE:
I did that out of prudence, because the format in which the training data is saved and loaded for the resume feature has changed.

ivysrono:
So the same as https://github.com/leela-zero/leela-zero/pull/2370 ?

PR #1836, opened 2018-09-14
lz-analyze policy prior output

alreadydone:
Per request at https://github.com/gcp/leela-zero/pull/1772#discussion_r215905542
It's not my original work; the diff was communicated to me by 榜爷 (who operated LZ in Tencent competition).

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -40,15 +40,14 @@
 using namespace Utils;
 
 constexpr int UCTSearch::UNLIMITED_PLAYOUTS;
-
 class OutputAnalysisData {
 public:
-    OutputAnalysisData(const std::string& move, int visits, int winrate, std::string pv) :
-        m_move(move), m_visits(visits), m_winrate(winrate), m_pv(pv) {};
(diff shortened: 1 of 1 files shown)


src/UCTSearch.cpp
```diff
@@ -67,6 +66,7 @@ class OutputAnalysisData {
     std::string m_move;
     int m_visits;
     int m_winrate;
+    std::string m_N_num; // policy prior
```
gcp: Rename this to m_policy_prior please, and store it as a float. (Do the conversion at display time)

gcp:
Note the point about storing it as a float and doing the conversion at display time, not when creating the storage struct.

(I prefer this because, for example, if you wanted to work with the struct and sort the output somehow, you'd need to redo the conversion twice)

alreadydone:
Just made the change. Sorry somehow I neglected that part of your comment.

gcp:
@alreadydone Thanks. Note that I changed the "N" in the output to "prior" as well, and changed the format to be consistent with "winrate".

Issue #1753, opened 2018-08-21
fp16 auto detection is lower performance than --precision half


gcp:
Radeon RX 560 Series (POLARIS11, DRM 3.26.0, 4.18.3-1-MANJARO, LLVM 6.0.1)

./leelaz -w networks/7141e6979284910b96f766cac59f05f80ac10ac840f9f1c21387174b6d5cb35d.gz -t 4 --precision auto (this last part can be omitted as it's the default)
Half precision compute support: YES
Using OpenCL half precision (at least 5% faster than single)
3204 visits, 1109102 nodes, 3203 playouts, 54 n/s

./leelaz -w networks/7141e6979284910b96f766cac59f05f80ac10ac840f9f1c21387174b6d5cb35d.gz  --benchmark -t 4 --precision half
3204 visits, 1108805 nodes, 3203 playouts, 106 n/s

It looks like we test half precision storage, but "forget" to activate half precision compute.

Note the gigantic speedup even though this card has no dual-speed fp16 math support.

gcp:
Oh right, if we detect "real" fp16 support then we always use fp16 math, and the case where we only use fp16 coeff storage doesn't exist.

I'm going to bet the AMD drivers are too inaccurate when benchmarking.

gcp:
This is very reproducible:

```
[morbo@ivux build]$ ./leelaz -w autogtp/networks/7141e6979284910b96f766cac59f05f80ac10ac840f9f1c21387174b6d5cb35d.gz  --benchmark -v 800 -t 2 --precision single
Thinking at most 2678400.0 seconds...
NN eval=0.530273
Playouts: 66, Win: 53.28%, PV: D3 C4 D5 C5 D6 B7 D16 Q4 P17 Q15 O3 K4 R3
Playouts: 140, Win: 53.25%, PV: D3 C4 D5 C5 D6 B7 D16 Q4 P17 Q15 O3 M3 K3 K4
Playouts: 216, Win: 53.26%, PV: D3 C4 D5 C5 D6 B7 D16 Q4 P17 Q15 O3 M3 K3 K4 J4
Playouts: 316, Win: 53.22%, PV: D3 C4 D5 C5 D6 B7 D16 Q4 P17 Q15 O3 K4 R3 Q3 R4 Q5
Playouts: 396, Win: 53.20%, PV: D3 C4 D5 C5 D6 B7 D16 Q4 P17 Q15 O3 K4 R3 Q3 R4 Q5
[...]

PR #1842, opened 2018-09-17
Reduce network memory usage when autodetecting

ihavnoid:
When doing auto precision detection, make sure prior implementation is destroyed before trying new implementation


ihavnoid:
Not sure, but may have to do something with #1753

gcp:
Looks pretty ugly but even a m_forward.reset(init_net()) wouldn't avoid the double memory allocation as far as I can tell.

Issue #1849, opened 2018-09-19
Assertion failure in UCTNode::wait_expanded()


dvdkhlng:
Compiling leela zero for debugging (make debug) results in assertion failures on GTP command "auto":

    ./leelaz -w networks/dec5b9d88877fe876a308b42370335c8f5465235d0104c889bdda3b42993aa28.gz 

(entering "auto" <enter>)

    [..]
      update_root, 1185058 -> 263105 nodes (22.2% reused)
    Thinking at most 36.3 seconds...
    NN eval=0.538457
    leelaz: UCTNode.cpp:386: void UCTNode::wait_expanded(): Assertion `v == ExpandState::EXPANDED' failed.
    leelaz: UCTNode.cpp:386: void UCTNode::wait_expanded(): Assertion `v == ExpandState::EXPANDED' failed.
    Aborted

ihavnoid:
I think I know why.  We are returning nodes to INITIAL state before a new move, which is assert-failing when we call get_pv because it assumes nodes are expanded if we want to print it out.

Will come with a fix, but the easiest workaround for now would be not printing anything, that is, don't call dump_analysis nor output_analysis in the middle of think().

ihavnoid:
Pushed a patch.  I think it addresses possible failures.  It seemed to be more than what I initially thought - please have a look.

ihavnoid:
FYI it probably won't help #1823

gcp:
Hmm, not too fond of that patch.

What the error case exactly? Some thread is in wait_expanded() and it gets surprised because the node gets reset to INITIAL underneath it?

But how can this happen as count_nodes_and_clear_expand_state() is supposed to be called when we're single threaded?

gcp:
Ah, I get it. It passes the .load() immediately because the node *isn't* in EXPANDING, but then expects the result of a note not being in EXPANDING to be EXPANDED, whereas it simply wasn't (re)EXPANDED.

gcp:
I suggest we take your patch to not reset fully expanded nodes, and change the logic so get_pv bails out if the node isn't fully expanded (which we only expect *really* deep in the PV).

gcp:
This should be fixed now.

PR #1851, opened 2018-09-19
Fixes assert failure on wait_expanded()

ihavnoid:
If a node is fully expanded but is reveretd to INITIAL state, there is no chance
it returns to EXPANDED state.  Don't revert nodes to INITIAL state if it is fully
expanded.

Addresses other smaller bugs too

May address #1849

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -178,6 +178,11 @@ bool UCTNode::has_children() const {
 }
 
 bool UCTNode::expandable(const float min_psa_ratio) const {
+    if (m_min_psa_ratio_children == 0.0) {
+        // if we figured out that we are fully expandable
+        // it is impossible that we stay in INITIAL state
+        assert(m_expand_state.load() != ExpandState::INITIAL);
+    }
     return min_psa_ratio < m_min_psa_ratio_children;
 }
 
@@ -330,7 +335,9 @@ UCTNode& UCTNode::get_best_root_child(int color) {
 size_t UCTNode::count_nodes_and_clear_expand_state() {
     auto nodecount = size_t{0};
     nodecount += m_children.size();
-    m_expand_state = ExpandState::INITIAL;
+    if (expandable()) {
+        m_expand_state = ExpandState::INITIAL;
+    }
     for (auto& child : m_children) {
         if (child.is_inflated()) {
             nodecount += child->count_nodes_and_clear_expand_state();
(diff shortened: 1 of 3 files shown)


src/UCTSearch.cpp
```diff
@@ -522,12 +521,18 @@ int UCTSearch::get_best_move(passflag_t passflag) {
...
-std::string UCTSearch::get_pv(FastState & state, UCTNode& parent) {
+std::string UCTSearch::get_pv(GameState & state, UCTNode& parent) {
     if (!parent.has_children()) {
         return std::string();
     }
 
+    float eval;
+    // we do this anyway even if we have children.  We don't want to crash the run
+    // just because the previonsly-expanded node is in the INITiAL state.
+    parent.create_children(m_network, m_nodes, state, eval, get_min_psa_ratio());
```
gcp: Bah!

What about changing wait_expanded to bail before the assert if the node is in INITIAL?

gcp: This is really tricky. get_pv wants to traverse m_children during the search so it really needs to wait until the node is fully expanded, or give up. Can we make it bail if any of the nodes it encounters is in that state? Together with your previous batch to not reset fully expanded nodes, this should work? PVs will be truncated shorter but that is OK.

src/UCTSearch.cpp
```diff
@@ -522,12 +521,18 @@ int UCTSearch::get_best_move(passflag_t passflag) {
     return bestmove;
 }
 
-std::string UCTSearch::get_pv(FastState & state, UCTNode& parent) {
+std::string UCTSearch::get_pv(GameState & state, UCTNode& parent) {
     if (!parent.has_children()) {
```
gcp: This should be a check for EXPANDED instead, because we can't safely traverse partially expanded nodes while the search is still going on.

src/UCTNode.cpp
```diff
@@ -330,7 +330,9 @@ UCTNode& UCTNode::get_best_root_child(int color) {
 size_t UCTNode::count_nodes_and_clear_expand_state() {
     auto nodecount = size_t{0};
     nodecount += m_children.size();
-    m_expand_state = ExpandState::INITIAL;
+    if (expandable()) {
+        m_expand_state = ExpandState::INITIAL;
```
gcp: This is a good idea.

ihavnoid:
Pushed the change which simply bails out when node is not fully expanded.  Added assertion that node should not stay in INITIAL state if m_min_psa_ratio_children is zero.

PR #1852, opened 2018-09-19
Update .gitignore to inclue ".vs/"

AncalagonX:
When cloning and opening an .sln solution (such as the `/next/` branch) in Visual Studio 2017, the `.vs` folder is created at the base of the project to store temporary user-specific environment settings. By default, Visual Studio 2017's Github plugin instantly detects these as changed files which must be synced. In my case, three files are created:

`ProjectSettings.json - Contains user-set project config settings`
`slnx.sqlite - Contains directory information and other details specific to the user's machine`
`VSWorkspaceState.json - Tracks open files and expanded directory views in the user's workspace`

This minor addition to the base `.gitignore` file resolves this, and as far as I can tell does not ignore any MSVC/VS files which *should* be managed under our git source control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ leelaz-model*
 leelaz_opencl_tuning
 /build-autogtp-*
 /build-validation-*
+.vs/


kuba97531:
I know that it doesn't hurt to add it to the `.gitignore` but why do you need it?
You could use `msvc` folder as a root of your VS projects, The relevant VS solution files are already inside the `msvc` directory and `msvc/.gitignore` already contains `.vs`.

AncalagonX:
It seems that the .vs folder in the base project directory is created when using "Folder View" in the Solution Explorer to traverse out of the main project directories specified in `leela-zero2017.sln`. One must do so in order to easily access non-sln files through Visual Studio 2017, such as the python files contained in `/training/tf/`. I've been dealing with this annoyance for a while, so I've finally submitted a tiny PR.

PR #1873, opened 2018-09-25
Autoselect half mode for fp16 compute.

gcp:
Rework the Network initialization to pull out the OpenCL benchmarking
for precision autodetection. Add support to ForwardPipe to report
whether it is necessary to run the benchmarks. If the answer is no, and
fp16 compute works, we assume that's what we want.

This avoids the benchmarking overhead on modern AMD cards and probably
on the latest ones from NVIDIA too.

### DIFF
--- a/src/ForwardPipe.h
+++ b/src/ForwardPipe.h
@@ -45,6 +45,7 @@ public:
     virtual ~ForwardPipe() = default;
 
     virtual void initialize(const int channels) = 0;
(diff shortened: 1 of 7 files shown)


gcp:
> So far we haven't found any card that advertises fp16 compute but misbehaves. Right?

I'm sure that if there are any they'll surface after I merge this :-)

PR #1882, opened 2018-09-26
Include Eigen package reference for VS2017 build.

kuba97531:


gcp:
Eigen is included as a git submodule, there is no need for packages.

Issue #1675, opened 2018-07-30
Manual pruning capability


TFiFiE:
As a way to make LZ more useful as an analysis tool, what about introducing a (real-time) option to disregard moves from the search? Nodes can already be deactivated by the time management, so it wouldn't even have to be an intrusive change.

gcp:
uci searchmoves?

TFiFiE:
I think it makes more sense with Go engines to indicate what to exclude rather than what to include.

dfannius:
Do you just want to exclude certain moves at the root position, or do you want all play to be in a local area for a while (e.g., solving a tsumego problem)?

One possible problem with the former could be that it might want to just make a sente exchange where it's allowed to and then come back to the area it really wants to play in. One possible problem with the latter is that the policy and value heads assume that moves can be played all over the board, so they might be silently assuming a followup outside of the constrained region and be very incorrect as a result.

These problems are all hypothetical for now, of course.

gcp:
>I think it makes more sense with Go engines to indicate what to exclude

That's a good point. We should add this, because unlike include moves this can't really be sensibly handled by playing out moves and checking the evaluation after that.

betterworld:
I implemented a feature to avoid moves and it seemed to work just fine during my first tests.

Regarding @dfannius' comment, in order to avoid a "quick sente exchange" before the forbidden move, I included a feature to forbid a certain move to be played for the next N moves.

E.g. the gtp command "avoid b q16 +8" makes Black ignore the possibility to play q16 for the next 8 moves.

roy7:
Hey this could be awesome for my bots who sometimes lose to people passing too early and trying to get a win from the auto score when the board is mostly empty. In theory then I could do like "avoid b pass 90" would prevent black from passing in the first 90 moves of the game? (Say my bot is black.) This way if an opponent starts passing at move 30 I'll just keep playing more stones.

wonderingabout:
@betterworld @roy7 
great idea

AncalagonX:
This is a simple yet excellent feature that can be used to greatly improve game review in tools such as Lizzie. There are many in-game situations and other reasons why a user would want to forbid a move from consideration and force LZ's efforts elsewhere, and I've quietly wished for a function like this a number of times in my games.

It also would allow for more creative players to do things like solve tsumegos as mentioned above, or forbid common opening moves like 3x3 invasions, or force LZ to always begin on Tengen as black, and so forth.

[...]

PR #1891, opened 2018-09-28
Separate FPU-reduction setting for root.

TFiFiE:
One way to make LZ more useful as (the back end of) an analysis tool, would be to make it possible to force it to start out visiting all the possible moves in a position and give a quick evaluation of even those that are normally never considered within any foreseeable time. Being able to override the first-play-urgency reduction at just the root would be one way to make this possible.

### DIFF
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -118,6 +118,7 @@ static void parse_commandline(int argc, char *argv[]) {
         ("puct", po::value<float>())
         ("softmax_temp", po::value<float>())
         ("fpu_reduction", po::value<float>())
+        ("fpu_root_reduction", po::value<float>())
         ;
 #endif
     // These won't be shown, we use them to catch incorrect usage of the
@@ -360,6 +361,14 @@ static void parse_commandline(int argc, char *argv[]) {
         }
     }
 
+    // Do not lower the expected eval for root moves that are likely not
+    // the best if we have introduced noise there exactly to explore more.
+    cfg_fpu_root_reduction = cfg_noise ? 0.0f : cfg_fpu_reduction;
+#ifdef USE_TUNER
+    if (vm.count("fpu_root_reduction")) {
+        cfg_fpu_root_reduction = vm["fpu_root_reduction"].as<float>();
+    }
+#endif
 
     auto out = std::stringstream{};
     for (auto i = 1; i < argc; i++) {
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -251,16 +251,10 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
         }
     }
 
-    auto numerator = std::sqrt(double(parentvisits));
-    auto fpu_reduction = 0.0f;
-    // Lower the expected eval for moves that are likely not the best.
-    // Do not do this if we have introduced noise at this node exactly
-    // to explore more.
-    if (!is_root || !cfg_noise) {
-        fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy);
-    }
+    const auto numerator = std::sqrt(double(parentvisits));
+    const auto fpu_reduction = (is_root ? cfg_fpu_root_reduction : cfg_fpu_reduction) * std::sqrt(total_visited_policy);
     // Estimated eval for unknown nodes = original parent NN eval - reduction
-    auto fpu_eval = get_net_eval(color) - fpu_reduction;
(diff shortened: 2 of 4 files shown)


src/UCTNode.cpp
```diff
@@ -251,16 +251,13 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
...
-    auto numerator = std::sqrt(double(parentvisits));
-    auto fpu_reduction = 0.0f;
+    const auto numerator = std::sqrt(double(parentvisits));
     // Lower the expected eval for moves that are likely not the best.
     // Do not do this if we have introduced noise at this node exactly
     // to explore more.
-    if (!is_root || !cfg_noise) {
-        fpu_reduction = cfg_fpu_reduction * std::sqrt(total_visited_policy);
-    }
+    const auto fpu_reduction = (is_root ? cfg_fpu_root_reduction : cfg_fpu_reduction) * std::sqrt(total_visited_policy);
```
Hersmunch: The comment about noise should probably move with where it is accounted for i.e. leela.cpp line 364.

src/Leela.cpp
```diff
@@ -118,6 +118,7 @@ static void parse_commandline(int argc, char *argv[]) {
         ("puct", po::value<float>())
         ("softmax_temp", po::value<float>())
         ("fpu_reduction", po::value<float>())
+        ("fpu_root_reduction", po::value<float>())
```
Hersmunch: If this option is to be useful for analysis, can you please explain why it is a "tuning option"?

gcp:
Wouldn't exposing PUCT be closer to what one would expect?

But I'm extremely unlikely to pull this nor any kind of tuning option that affects strength. If you want the evaluation of a move that the engine does not consider, then PLAY IT OUT and you will get the eval immediately.

gcp:
At the very least we should have a discussion on how to best achieve the stated aim, and what's most useful for people analyzing.

TFiFiE:
A "heatmap" of winrates for the whole board would, especially for weaker players, be much more helpful than manually playing out a few different moves, in terms of getting general ideas about a position, like how urgent a local response is (i.e. how harmful tenuki would be) or what the best move would be in any given area.

I still think think it's cleaner code to have a separate variable for the root reduction, even if it doesn't get its own tuning option. Would you pull this if I remove `fpu_root_reduction`?

gcp:
>A "heatmap" of winrates for the whole board would, especially for weaker players, be much more helpful than manually playing out a few different moves, in terms of getting general ideas about a position, like how urgent a local response is (i.e. how harmful tenuki would be) or what the best move would be in any given area.

Don't the prior values achieve this? The GUI could use ```heatmap```. 

Alternatively, if you want the actual winrate, you need to play and eval (1 visit) every move, but you have the same problem in this pull.

Note my comment in issue #1675.

>I still think think it's cleaner code to have a separate variable for the root reduction, even if it doesn't get its own tuning option. Would you pull this if I remove fpu_root_reduction?

Sure.

TFiFiE:
OK, I have removed `fpu_root_reduction`. Please reopen and merge.

> Don't the prior values achieve this? The GUI could use `heatmap`.

Not only are priors not winrates, the relationship between them doesn't even have to be monotonic; one could imagine that, in some position, a move that isn't necessarily that much of a loss, might nevertheless get a very low prior because the network has learnt there is almost always a better one in that sort of situation. Think [crude moves](https://senseis.xmp.net/?Zokusuji) or sente moves that only waste ko threats.

alreadydone:
Looking at your discussion and checking the code, I find that children of the root is sorted not by noised policy but by the original policy. This will cause some moves with higher noised policy be expanded *after* lower noised policy but higher original policy. In extreme cases (e.g. very flat noised policy & Q>>net_eval, or most of the tree reused), it seems possible that the move with highest noised policy would never be expanded. Shoule we fix this by calling stable_sort again?

gcp:
>This will cause some children (Edit: only the first one) with higher noised policy be expanded after children with lower noised policy but higher original policy.
>Should we fix this by calling stable_sort in dirichlet_noise?

Yes, this is a small bug that's neat to fix.

gcp:
>Please reopen and merge.

I don't see the changes and github won't let me re-open due to "branch was force pushed or recreated".

alreadydone:
The code in the link above actually doesn't work, since atomics can't be moved. I end up doing it [this way](https://github.com/alreadydone/lz/commit/00507428ce0196744e6faab8022e997f8159935f#diff-c56c48893bb7a5b7ffa8bb597772d378).

PR #1928, opened 2018-10-12
UCTSearch : fix assert-fail when memory is completely full

ihavnoid:
We need to call UCTNode::create_children() even if we aren't expanding due to low memory, because that moves our node's state from INITIAL to EXPANDED.

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -189,6 +189,10 @@ float UCTSearch::get_min_psa_ratio() const {
     if (mem_full > 0.5f) {
         // Memory is almost exhausted, trim more aggressively.
         if (mem_full > 0.95f) {
+            // if completely full just stop expansion by returning an impossible number
+            if (mem_full >= 1.0f) {
+                return 2.0f;
+            }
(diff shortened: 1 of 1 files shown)


ihavnoid:
```
Build started
git clone -q https://github.com/gcp/leela-zero.git C:\projects\leela-zero
fatal: early EOF
fatal: The remote end hung up unexpectedly
fatal: index-pack failed
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
Command exited with code 128
```
appveyor fails again?

gcp:
Doesn't this risk regressing us to the state again where more simulations cause the moves to be weaker?

This seems like a nasty workaround for the actual bug.

gcp:
Hmm, OK I guess it just ends up running the same codepath because success will be false?

ihavnoid:
> Hmm, OK I guess it just ends up running the same codepath because success will be false?

Yes correct.  Plus, IMHO it is logical that 'not expanding' being equivalent to 'expand only if policy is larger than 2.0'.

Issue #1779, opened 2018-08-26
Ladder & Life-n-death under belly in LZ


yeechutngeow:
AQ was released for PC and android recently, and various people have been pitching it against LZ. One observation is the frequency at which LZ failed miserably on ladder and life-n-death. Below is an example of me pitting LZ/Gridmaster against AQ on android phone. LZ is black, set to 1 move per 10 sec.
![img_20180825_091752](https://user-images.githubusercontent.com/2901215/44624454-cd689c00-a921-11e8-848d-5f7d12481b78.jpg)

Some points for discussion:
- the rate at which the current MCTS strategy will fully evaluate a ladder and life-n-death fully is near zero.
- the winrate in the middle of a ladder or a capturing race is undefined or meaningless until the full sequence is evaluated.
- which means the NN will not get to learn about ladder and many life-n-death patterns.
- AQ has menu like "show ladder capture", "show ladder escape", "show false eyes", which means AQ has built-in logic for these things.
- I like to argue that LZ will need similar built-in logic. A recognition system to guide the tree search to reach a local resolution before applying NN evaluation.

alreadydone:
Can't you use LZ in AQ Android?

jokkebk:
This seems hardly news, as Android phones will get very low number of visits, resulting in misread ladders and L&D in much higher frequency than on PC with high visit count. Or are these "various people" seeing Leela being beat by AQ due to ladders on PC with high visit counts as well?

The "we should implement ladder logic outside the zero framework" argument has been hashed a couple times, and so far the answer has consistently been, that Leela Zero is (for the time being at least) going forward with pure zero philosophy (I'm not GCP or a code contributor so just summarizing the previous discussions based on my memory of them), that means no additional inputs or logic for ladders or simple L&D. Yes it means more visits are needed to avoid blunders, and yes it might mean AQ will kick Leela's ass on phones for some time.

Maybe after the Zero approach has run its course and we have 256x40 super-AI for go, a fork of the project could be made with additional features (and a dozen other improvement ideas that have been pitched), and we could see if a stronger AI would emerge using a "non-zero" approach.

yeechutngeow:
[...]

Issue #1754, opened 2018-08-21
Ladder input feature?


bsteuber:
We know that the original AlphaGo had a ladder feature in the network inputs, which worked out really well. For the Zero version, they got rid of it - not because it works better, but because it reads more impressive to use absolutely no human knowledge.

After talking to some Deepmind people, I am quite certain that reintroducing this feature to LZ would get rid of all ladder problems once at for all, so as someone using LZ a lot for analysis (in fact, I just opened a commercial LZ cloud analysis service) I would really love a version with this feature.

I understand that this isn't part of the original project goal (after all, it would have to be renamed “Leela One“). But it seems several people here are willing to experiment with forks and have access to a lot of computing power. So I hope this post might inspire one of you to try this out.

gcp:
Out of scope for this project.

gcp:
There's several people who already made a merger of LZ with Ray's ladder detection AFAIK.

That said I haven't seen LZ fall for a ladder in ages, unless we're talking about a lost position, in which case a ladder is really the least of the problems you'll see.

bsteuber:
Just as an example of ladders still being a big problem for LZ, take https://zero.sjeng.org/viewmatch/28ccb9f11e068608c208f015b1ad18f720ecbdd972101dbfc0828b3f49b86c7f?viewer=wgo from reddit. Its upper right corner variation was played in a lot of other recent games, too (without the ladder working).
When analyzing the position with Lizzie on more or less current next branch and current 20b network, I get these screenshots:

**After >20k playouts:**
![ladder_1](https://user-images.githubusercontent.com/213112/44426195-48caf600-a58e-11e8-8d37-edf04468e546.png)

**After >50k playouts:**
![ladder_2](https://user-images.githubusercontent.com/213112/44426202-4d8faa00-a58e-11e8-900a-174c59bd8042.png)

So, escaping the ladder is a winning move, but even with it being just one move ahead, LZ needs >50k playouts to find it. Black played a losing variation but got unpunished. 

I believe things like that happen far more more open than noticed....

gcp:
Oh, so it's not playing bad ladder at all? It just needs search to overcome a low prior for escaping that way.

gcp:
This kind of thing won't be fixed by ladder detection at all.

bsteuber:
[...]

Issue #1447, opened 2018-05-18
 is the rn ladder check thought useful to improve leelaz?


l1t1:
https://github.com/kobanium/Ray/blob/b219b4d880aa2eba64609d0927b1000f14915c8f/src/Ladder.cpp#L22

https://github.com/zakki/Ray/issues/162

bood:
This is not how zero approach works.

PR #1941, opened 2018-10-17
Simple ladder detector

kaorahi:
This small PR (about 10 lines essentially) adds a ladder detector to leelaz. I know that it is out of the scope of this project (#1447, #1754, #1779). But it practically benefits cpu-only users without large change of existing code. It is not for making leelaz stronger but for comfortable analysis of human games by Lizzie, Go Review Partner, etc. The detector is turned off by default.

Example 1 (ref. https://github.com/gcp/leela-zero/issues/1754#issuecomment-414805195)

    (;B[dd];W[pp];B[pd];W[dp];B[cq];W[cp];B[dq];W[ep];B[fr];W[qf];B[qq];W[qp];B[pq];W[op];B[oq];W[np];B[nq];W[nc];B[nd];W[md];B[ne];W[pc];B[mc];W[oc];B[ld])

The network 182 (68d7c8fc) finds ladder escaping from Example 1 within 500 playouts with the following option.

    leelaz --ladderlen 3 ...

Here, "3" is the length of moves for ladder detection. It takes more than 1000 playouts without this option.

Actually, this PR does not check ladders seriously for simplicity. It only detects any repeated patterns of every 4 moves and calls play_simulation() recursively as long as the same pattern continues. This can cause false positive. For example, repeated pushes from the following Example 2 are regarded as "ladder". But it seems harmless because the pattern is soon broken naturally.

Example 2

    (;B[pp];W[pd];B[dp];W[cd];B[ec];W[dc];B[ed];W[cf];B[nc];W[qf];B[pc];W[qc];B[qb];W[oc];B[pb];W[od];B[qd];W[ob];B[qe];W[pf];B[jc];W[qn];B[rf];W[rg];B[re];W[oh];B[pn];W[pm];B[on];W[ro];B[qq];W[jq];B[hq];W[mq];B[nl];W[ok];B[nr];W[om];B[nm];W[nn];B[no])

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -58,6 +58,7 @@ int cfg_max_cache_ratio_percent;
 TimeManagement::enabled_t cfg_timemanage;
 int cfg_lagbuffer_cs;
 int cfg_resignpct;
+int cfg_ladder_len;
 int cfg_noise;
 int cfg_random_cnt;
 int cfg_random_min_visits;
@@ -138,6 +139,7 @@ void GTP::setup_default_parameters() {
     cfg_fpu_reduction = 0.25f;
     // see UCTSearch::should_resign
     cfg_resignpct = -1;
+    cfg_ladder_len = 0;
     cfg_noise = false;
     cfg_random_cnt = 0;
     cfg_random_min_visits = 1;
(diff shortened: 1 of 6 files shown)


LL145:
We could just forbidding ladder detection in training games. and use it only in real matches (not self play matches)

kaorahi:
Yes. I expect that this PR accelerates search of "middle part" of any ladders, though it does not help search of "beginning part" and "end part".

gcp:
Won't be added for obvious reasons. The CPU argument is silly, there are many optimizations that can be made on the input size by adding knowledge.

PR #1949, opened 2018-10-23
Add a feature to exclude moves from the search.

betterworld:
(These instructions have been updated to match the current version.)

The new GTP feature "avoid" specifies moves to avoid. It is implemented as a new argument for the `lz-analyze` and `lz-genmove_analyze` commands.
Syntax is `lz-analyze ARGS [avoid <color> <coords> <number_of_moves>] [avoid ...]` (same for `lz-genmove_analyze`).
The specified coords will not be considered for this color starting from the current move for the next `number_of_moves` of moves. Instead of coordinates, `pass` or `resign` may also be specified.

Example: `lz-analyze 100 avoid b q16,q4,d16,d4 1 avoid w k10 2` forbids Black to play a Hoshi and White to play Tengen for the next move.

An additional form is `allow`. "Allow"ing a list of moves is equivalent to "avoid"ing all other moves, except pass or resign:
`lz-analyze 100 allow b r16,r17,q16 1` would force Black to play in the upper-right corner.
`lz-analyze 100 allow b r16,r17,q16 1 avoid b pass,resign 1` would do the same and forbid pass and resign at the same time. Note that `pass` and `resign` are the only possible `avoid` moves that can be specified together with `allow`.

This addresses issue https://github.com/gcp/leela-zero/issues/1675

### DIFF
--- a/src/FastState.cpp
+++ b/src/FastState.cpp
@@ -77,7 +77,19 @@ bool FastState::is_move_legal(int color, int vertex) const {
            vertex == FastBoard::RESIGN ||
            (vertex != m_komove &&
                 board.get_state(vertex) == FastBoard::EMPTY &&
+                !is_to_avoid(color, vertex, m_movenum) &&
                 !board.is_suicide(vertex, color));
+}
+
+bool FastState::is_to_avoid(int color, int vertex, size_t movenum) const {
+    for (auto& move : m_moves_to_avoid) {
+        if (color == move.color &&
+                vertex == move.vertex &&
+                movenum >= move.from_move &&
+                movenum <= move.to_move)
+            return true;
+    }
+    return false;
 }
 
 void FastState::play_move(int vertex) {
@@ -130,6 +142,15 @@ void FastState::set_passes(int val) {
 void FastState::increment_passes() {
     m_passes++;
     if (m_passes > 4) m_passes = 4;
+}
+
+void FastState::add_move_to_avoid(int color, int vertex, size_t from_move, size_t to_move) {
+    MoveToAvoid item;
+    item.color = color;
+    item.vertex = vertex;
+    item.from_move = from_move;
+    item.to_move = to_move;
+    m_moves_to_avoid.push_back(item);
 }
 
 int FastState::get_to_move() const {
--- a/src/FastState.cpp
+++ b/src/FastState.cpp
@@ -73,12 +73,12 @@ void FastState::reset_board() {
 }
 
 bool FastState::is_move_legal(int color, int vertex) const {
-    return vertex == FastBoard::PASS ||
-           vertex == FastBoard::RESIGN ||
-           (vertex != m_komove &&
-                board.get_state(vertex) == FastBoard::EMPTY &&
-                !is_to_avoid(color, vertex, m_movenum) &&
-                !board.is_suicide(vertex, color));
+    return !is_to_avoid(color, vertex, m_movenum) && (
+              vertex == FastBoard::PASS ||
+                 vertex == FastBoard::RESIGN ||
+                 (vertex != m_komove &&
+                      board.get_state(vertex) == FastBoard::EMPTY &&
+                      !board.is_suicide(vertex, color)));
 }
 
 bool FastState::is_to_avoid(int color, int vertex, size_t movenum) const {
--- a/src/FastState.cpp
+++ b/src/FastState.cpp
@@ -84,19 +84,14 @@ bool FastState::is_move_legal(int color, int vertex) const {
 
 bool FastState::is_to_avoid(int color, int vertex, size_t movenum) const {
     for (auto& move : cfg_analyze_tags.m_moves_to_avoid) {
-        if (color == move.color &&
-                vertex == move.vertex &&
-                movenum >= move.from_move &&
-                movenum <= move.to_move) {
+        if (color == move.color && vertex == move.vertex && movenum <= move.until_move) {
             return true;
         }
     }
     if (vertex != FastBoard::PASS && vertex != FastBoard::RESIGN) {
         bool active_allow = false;
         for (auto& move : cfg_analyze_tags.m_moves_to_allow) {
-            if (color == move.color &&
-                    movenum >= move.from_move &&
-                    movenum <= move.to_move) {
+            if (color == move.color && movenum <= move.until_move) {
                 active_allow = true;
                 if (vertex == move.vertex) {
                     return false;
--- a/src/FastState.cpp
+++ b/src/FastState.cpp
@@ -91,6 +91,18 @@ bool FastState::is_to_avoid(int color, int vertex, size_t movenum) const {
             return true;
         }
     }
+    if (!cfg_analyze_tags.m_moves_to_allow.empty()
+            && vertex != FastBoard::PASS && vertex != FastBoard::RESIGN) {
+        for (auto& move : cfg_analyze_tags.m_moves_to_allow) {
+            if (color == move.color &&
+                    vertex == move.vertex &&
+                    movenum >= move.from_move &&
+                    movenum <= move.to_move) {
+                return false;
+            }
+        }
+        return true;
+    }
     return false;
 }
 
--- a/src/FastState.cpp
+++ b/src/FastState.cpp
@@ -74,35 +74,12 @@ void FastState::reset_board() {
 }
 
 bool FastState::is_move_legal(int color, int vertex) const {
-    return !is_to_avoid(color, vertex, m_movenum) && (
+    return !cfg_analyze_tags.is_to_avoid(color, vertex, m_movenum) && (
               vertex == FastBoard::PASS ||
                  vertex == FastBoard::RESIGN ||
                  (vertex != m_komove &&
                       board.get_state(vertex) == FastBoard::EMPTY &&
                       !board.is_suicide(vertex, color)));
-}
-
-bool FastState::is_to_avoid(int color, int vertex, size_t movenum) const {
-    for (auto& move : cfg_analyze_tags.m_moves_to_avoid) {
-        if (color == move.color && vertex == move.vertex && movenum <= move.until_move) {
-            return true;
-        }
-    }
-    if (vertex != FastBoard::PASS && vertex != FastBoard::RESIGN) {
-        bool active_allow = false;
-        for (auto& move : cfg_analyze_tags.m_moves_to_allow) {
-            if (color == move.color && movenum <= move.until_move) {
-                active_allow = true;
-                if (vertex == move.vertex) {
-                    return false;
-                }
-            }
-        }
-        if (active_allow) {
-            return true;
-        }
-    }
-    return false;
 }
 
 void FastState::play_move(int vertex) {
--- a/src/FastState.cpp
+++ b/src/FastState.cpp
@@ -37,6 +37,7 @@
 #include "FastBoard.h"
 #include "Utils.h"
 #include "Zobrist.h"
+#include "GTP.h"
 
 using namespace Utils;
 
@@ -82,7 +83,7 @@ bool FastState::is_move_legal(int color, int vertex) const {
 }
 
 bool FastState::is_to_avoid(int color, int vertex, size_t movenum) const {
-    for (auto& move : m_moves_to_avoid) {
+    for (auto& move : cfg_analyze_tags.m_moves_to_avoid) {
         if (color == move.color &&
                 vertex == move.vertex &&
                 movenum >= move.from_move &&
@@ -143,14 +144,6 @@ void FastState::set_passes(int val) {
 void FastState::increment_passes() {
     m_passes++;
     if (m_passes > 4) m_passes = 4;
-}
-
-void FastState::add_move_to_avoid(int color, int vertex, size_t from_move, size_t to_move) {
-    m_moves_to_avoid.emplace_back(color, from_move, to_move, vertex);
-}
-
-void FastState::clear_moves_to_avoid() {
-    m_moves_to_avoid.clear();
 }
 
 int FastState::get_to_move() const {
--- a/src/FastState.cpp
+++ b/src/FastState.cpp
@@ -86,8 +86,9 @@ bool FastState::is_to_avoid(int color, int vertex, size_t movenum) const {
         if (color == move.color &&
                 vertex == move.vertex &&
                 movenum >= move.from_move &&
-                movenum <= move.to_move)
+                movenum <= move.to_move) {
             return true;
+        }
     }
     return false;
 }
@@ -145,12 +146,11 @@ void FastState::increment_passes() {
 }
 
 void FastState::add_move_to_avoid(int color, int vertex, size_t from_move, size_t to_move) {
-    MoveToAvoid item;
-    item.color = color;
-    item.vertex = vertex;
-    item.from_move = from_move;
-    item.to_move = to_move;
-    m_moves_to_avoid.push_back(item);
+    m_moves_to_avoid.emplace_back(color, from_move, to_move, vertex);
+}
+
+void FastState::clear_moves_to_avoid() {
+    m_moves_to_avoid.clear();
 }
 
 int FastState::get_to_move() const {
--- a/src/FastState.cpp
+++ b/src/FastState.cpp
@@ -91,17 +91,21 @@ bool FastState::is_to_avoid(int color, int vertex, size_t movenum) const {
             return true;
         }
     }
-    if (!cfg_analyze_tags.m_moves_to_allow.empty()
-            && vertex != FastBoard::PASS && vertex != FastBoard::RESIGN) {
+    if (vertex != FastBoard::PASS && vertex != FastBoard::RESIGN) {
+        bool active_allow = false;
         for (auto& move : cfg_analyze_tags.m_moves_to_allow) {
             if (color == move.color &&
-                    vertex == move.vertex &&
                     movenum >= move.from_move &&
                     movenum <= move.to_move) {
-                return false;
+                active_allow = true;
+                if (vertex == move.vertex) {
+                    return false;
+                }
             }
         }
-        return true;
+        if (active_allow) {
+            return true;
+        }
     }
     return false;
 }
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -284,10 +284,10 @@ AnalyzeTags GTP::parse_analyze_tags(std::istringstream & cmdstream, const GameSt
 
         if (tag == "avoid" || tag == "allow") {
             std::string textcolor, textmoves;
-            size_t to_movenum;
+            size_t until_movenum;
             cmdstream >> textcolor;
             cmdstream >> textmoves;
-            cmdstream >> to_movenum;
+            cmdstream >> until_movenum;
             if (cmdstream.fail()) {
                 break;
             }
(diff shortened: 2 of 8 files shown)


src/GTP.cpp
```diff
@@ -197,6 +197,7 @@ const std::string GTP::s_commands[] = {
     "lz-genmove_analyze",
     "lz-memory_report",
     "lz-setoption",
+    "avoid",
```
TFiFiE: Prefix the command with `lz-` (like the ones above it), because it's not part of the standard protocol (see https://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html#sec:private-extensions).

src/FastState.cpp
```diff
@@ -121,6 +133,15 @@ void FastState::increment_passes() {
     if (m_passes > 4) m_passes = 4;
 }
 
+void FastState::add_move_to_avoid(int color, int vertex, size_t from_move, size_t to_move) {
+    MoveToAvoid item;
+    item.color = color;
+    item.vertex = vertex;
+    item.from_move = from_move;
+    item.to_move = to_move;
+    m_moves_to_avoid.push_back(item);
```
Hersmunch: I'm pretty sure gcp will ask you to add a constructor and use emplace_back()

src/GTP.cpp
```diff
@@ -952,6 +953,43 @@ void GTP::execute(GameState & game, const std::string& xinput) {
...
+        } else if (textcolor == "b" || textcolor == "black") {
+            color = FastBoard::BLACK;
+        } else {
+            gtp_fail_printf(id, "cannot parse color");
+            return;
+        }
+
+        std::stringstream movenumstream(text_to_movenum);
+        movenumstream >> to_movenum;
+        if (text_to_movenum[0] == '+')
```
Hersmunch: Missing braces

Ishinoshita:
Can specify a series of moves you want to avoid or just one? Suppose for instance that you want black to start a game avoiding any of the Hoshi. Or that you don't want to see any 3-3  invasion during the first 50 moves. Can you list the the 4 Hoshi or the 4 San-San?

Ishinoshita:
Great, thank you @betterworld!

MartinDevelopment:
An avoid clear command would be nice to empty the list.

gcp:
>The specified coords will not be considered for this color starting from
the current move until the specified move number.

I don't like the extra GUI vs engine state that this ends up carrying.

This should be implemented via a tag on lz-analyze or lz-genmove_analyze, with the clear being implicit after the command ends.

(Basically, when adding GTP extensions, model them after UCI, and do the exact opposite as what XBoard/WinBoard/CECP does. This from plenty of experience dealing with both protocols on both sides...)

betterworld:
@gcp 
Your idea sounds like a more elegant design. However I am wondering if it is so easy to implement.

Consider three scenarios.

1. You run lz-analyze without any tags, stop it, then run it again.
2. You run lz-analyze with an "avoid" tag, then run it again but without the tag. (Or the other way round.)
3. You run lz-analyze with an "avoid" tag, then run it again with the same tag.

Forgive me if my understanding of the move search algorithm and lz-analyze (or UCI) is not perfect. The way I understand it, in the first case, the search tree is re-used for lz-analyze. In the 2nd case, you can't really re-use the search tree because some of the visit counts are much too large or much too small and wouldn't make sense in the new context. In the 3rd case, you would like to re-use the search tree just like in the 1st case.

Would this be possible to implement without keeping the state in the engine?

Another point: In issue https://github.com/gcp/leela-zero/issues/1675, @roy7 said he would like to use `lz-avoid` in his bot to keep it from passing. In a bot that just uses `genmove`, the `lz-analyze` tag couldn't be used.

As for the relative move numbers, I think you're right, absolute move numbers aren't really required.

roy7:
For my personal use case, if all I had to do was issue the lz-avoid command and use lz-genmove_analyze instead of genmove, that's fine with me. I need to alter the bot interface code anyway to inject lz-avoid at game initialization. I'd just change it to use lz-genmove_analyze all of the time instead of ever using genmove (if lz-genmove_analyze is the same as genmove, just honoring lz-avoid instructions).

src/GTP.h
```diff
@@ -69,6 +99,7 @@ extern std::string cfg_options_str;
 extern bool cfg_benchmark;
 extern bool cfg_cpu_only;
 extern int cfg_analyze_interval_centis;
+extern AnalyzeTags cfg_analyze_tags;
```
betterworld: I was thinking where the best place is to store these settings. I ended up storing it here because it is right next to the existing variable `cfg_analyze_interval_centis`.

By adding the new class, I intended to make it straightforward to implement more of these tags in the future.

betterworld:
> You could even throw away tree reuse when there are tags as a first implementation (the cache will still work).

Thanks for this hint. The cache seems for work pretty well here. My patch clears the tree at the start and the end of the `ponder` function whenever the `avoid` tag has been used. I'd be happy to know if there is a more elegant way to do this.

src/GTP.cpp
```diff
@@ -494,6 +550,10 @@ void GTP::execute(GameState & game, const std::string& xinput) {
                 }
             }
         }
+        if (!process_analyze_tags(id, cmdstream, game)) {
```
gcp: I think you can attempt to parse tags in both the above "syntax not understood" failure branches. (again, to allow transitioning to passing the interval as a tag)

src/UCTSearch.cpp
```diff
@@ -780,6 +786,11 @@ int UCTSearch::think(int color, passflag_t passflag) {
 }
 
 void UCTSearch::ponder() {
+    bool disable_reuse = !cfg_analyze_tags.m_moves_to_avoid.empty();
+    if (disable_reuse) {
```
gcp: This will disable tree reuse, but the information in the NNCache will still be incorrect unless the tags are included in the hash, or the cache is cleared here.

gcp: Disregard this, I forgot that our cache doesn't store the tree search outcome, just the neural network output (it's not a transposition table).

gcp:
Looks very good. My only wish item remaining is a provision for:

>Also this makes me wish we'd made it ```lz-analyze w interval 100``` but oh well. We can still add support for stating it that way.

In other words, work even if the interval isn't specified but there are tags instead.

It might be a good idea to handle parsing these commands to the unit tests.

src/UCTSearch.cpp
```diff
@@ -508,8 +516,10 @@ int UCTSearch::get_best_move(passflag_t passflag) {
             // We didn't consider passing. Should we have and
             // end the game immediately?
 
+            if (!m_rootstate.is_move_legal(color, FastBoard::PASS)) {
```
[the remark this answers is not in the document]
betterworld: Do you mean when `bestmove` already has the value PASS? I don't think this can happen because the tree search will ignore moves that are not legal.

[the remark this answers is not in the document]
Hersmunch: I see what you are saying. I had a search through and can see that before searching from `m_root` `prepare_root_node()` calls `create_children()` and it looks like in there `PASS` is assumed to be legal and is always added to `m_children` (via `nodelist`). Maybe that should be changed instead of doing something in this function though?

Hersmunch:
What happens if the user tags all moves (including pass and resign) as to be avoided? :)

TFiFiE:
I see no code handling any situation in which there are no allowed moves. At the root, one could handle this case by returning an error message, whereas deeper inside the search, propagating a forced loss for the side to move is probably the way to go.

Also, I'd like to see the possibility of concisely phrasing what moves to allow rather than what to avoid, which could be accomplished by accepting `all` as a move argument and `allow` as a tag (with later tags overriding earlier ones). That way, for example, allowing only 3-3 or 3-5 as Black's first move in the search could be done with something like: `lz-analyze 100 avoid b all 1 allow b r17 1 allow b r15 1` (with no need to mention the other symmetric intersections)

gcp:
Instead of ```avoid all``` + ```allow```, the presence of the ```allow``` tag could imply ```avoid all``` as I think it doesn't make sense otherwise?

TFiFiE:
While making the simultaneous use of both `avoid` and `allow` tags erroneous? Sounds fine.

betterworld:
@gcp 
> In other words, work even if the interval isn't specified but there are tags instead.

I added a commit which tries to accomplish this.  In order to simplify all the possible code paths and failure handling, I figured it was best to add the parsing code for all the arguments into the same function.

I think I can look into adding the `allow` tag. Is the naming fine? Maybe if the one tag is called `allow`, the other should be called `forbid`? Or we call it `searchmoves` instead of `allow`, which is similar to the UCI version?

TFiFiE:
`include` and `exclude` are also possibilities.

To avoid repeating them, the tags could also be made to accept a variable number of `(color, coordinate, number)` triples. If that would be problematic on account of possible future ambiguous commands, one could always enforce that multiple triples must be enclosed with brackets.

betterworld:
@TFiFiE 
> To avoid repeating them, the tags could also be made to accept a variable number of (color, coordinate, number) triples. If that would be problematic on account of possible future ambiguous commands, one could always enforce that multiple triples must be enclosed with brackets.

Good idea. I believe in most cases the `color` and `number` would stay the same, so I added the comma-separated syntax for the coordinates only. This is easier to parse, too :)

When I tried to investigate @Hersmunch's question about what happens when all moves are forbidden, I noticed that I had problems issuing GTP commands that are longer than 4096 bytes. This should not be an issue any more with the new `,` syntax.

betterworld:
@Hersmunch 
> What happens if the user tags all moves (including pass and resign) as to be avoided? :)

As you noticed in your code comments, I forgot to cover all possible cases where a `PASS` can happen.  My latest commit https://github.com/gcp/leela-zero/pull/1949/commits/e0620ddfd9250a3a8f908068316b7c40d687db60 fixes this. I fixed it in a way that removes the `PASS` node only if there are other legal moves left. So if you `avoid` all legal moves including `pass`, LZ will still pass, otherwise there would be a segmentation fault.

src/UCTSearch.cpp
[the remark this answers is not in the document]
Hersmunch: Yes. I meant in addition to what you have already done but I prefer your method of only excluding pass if there is another move that can be played

gcp:
@betterworld I've been slacking here due to travel but also a bit due to interactions with #2032. I have not forgotten and really appreciate your help!

dbosst:
I added a pull request (betterworld#1) to his avoidmove fork to allow for compressed point lists (ie q16:t19). I hope it gets added to this pull request.

(Not sure how to manage this on someone else's pull request & fork ... when its out of date of origin)

wonderingabout:
good work and ideas, keep up everyone :+1:

src/FastState.cpp
```diff
@@ -62,11 +63,35 @@ void FastState::reset_board() {
...
+    for (auto& move : cfg_analyze_tags.m_moves_to_avoid) {
+        if (color == move.color && vertex == move.vertex && movenum <= move.until_move) {
+            return true;
+        }
+    }
+    if (vertex != FastBoard::PASS && vertex != FastBoard::RESIGN) {
+        bool active_allow = false;
+        for (auto& move : cfg_analyze_tags.m_moves_to_allow) {
+            if (color == move.color && movenum <= move.until_move) {
+                active_allow = true;
```
gcp: Wouldn't we want to set this as soon as cfg_analyze_tags.m_moves_to_allow is not empty? What's the reason it is inside this if?

src/GTP.cpp
```diff
@@ -152,7 +152,7 @@ void GTP::setup_default_parameters() {
     cfg_cpu_only = false;
 #endif
 
-    cfg_analyze_interval_centis = 0;
+    cfg_analyze_tags = AnalyzeTags();
```
gcp: Coding style prefers AnalzeTags{}.

src/GTP.cpp
```diff
@@ -243,6 +243,146 @@ std::string GTP::get_life_list(const GameState & game, bool live) {
     return result;
 }
 
+AnalyzeTags GTP::parse_analyze_tags(std::istringstream & cmdstream, const GameState & game) {
+    std::string tag;
```
gcp: Use auto for consistency with the rest of the code.

src/GTP.cpp
```diff
@@ -243,6 +243,146 @@ std::string GTP::get_life_list(const GameState & game, bool live) {
...
+    AnalyzeTags tags;
+
+    /* Default color is the current one */
+    tags.m_who = game.board.get_to_move();
+
+    bool avoid_not_pass_resign_b = false, avoid_not_pass_resign_w = false;
```
gcp: Same.

src/GTP.cpp
```diff
@@ -243,6 +243,146 @@ std::string GTP::get_life_list(const GameState & game, bool live) {
...
+            } else {
+                break;
+            }
+
+            if (until_movenum < 1) {
+                break;
+            }
+            until_movenum += game.get_movenum() - 1;
+
+            for (auto move : moves) {
```
gcp: This can take a (const) reference.

src/GTP.h
```diff
@@ -29,6 +29,49 @@
...
+
+class AnalyzeTags {
+public:
+    bool m_invalid;
+    std::vector<MoveToAvoid> m_moves_to_avoid, m_moves_to_allow;
+    int m_interval_centis;
+    int m_who;
+
+    AnalyzeTags() {
+        clear();
```
gcp: You could use uniform initialization here, i.e. make this = default, and just do

```
public:
    bool m_invalid{true};
    etc..
```

This is currently used by most of the other code. Do away with clear, and just assign from an empty instance when clearing is needed.

This is a matter of taste but I think it would end up slightly cleaner.

src/GTP.h
```diff
@@ -29,6 +29,49 @@
...
+    {}
+
+    bool operator==(const MoveToAvoid other) const {
+        return color == other.color &&
+            until_move == other.until_move && vertex == other.vertex;
+    }
+};
+
+class AnalyzeTags {
+public:
```
gcp: If it's not going to be POD, I'd prefer the members are private and we use accessors.

src/FastState.cpp
betterworld: Thanks for the review!

Each `allow` and `avoid` tag has its own move number range.

I.e. `allow b q16 1 avoid b r3 10` will set the `m_moves_to_allow` to a non-empty list. However from move 2 to 10 (and also for all moves beyond that) it as quasi-empty. The `active_allow` flag indicates that we are in a move number range where `allow` is in effect, which means that all non-allowed moves will be avoided.

src/GTP.cpp
betterworld: Is this because you prefer it from a coding-style point of view? I think for plain int vectors it would not make much of a difference.

src/GTP.cpp
betterworld: Just asking to make sure I understand you correctly as I am a bit puzzled here: You are referring to the variables in the function body below this line, and not to the `tag` variable or the return value of the function, right? The rest of the codebase does not seem to use `auto` for return types.

Anyway, I think I'll turn the function into a constructor or a method of `AnalyzeTags` when I make the members of the class private as you requested.

TFiFiE:
> I fixed it in a way that removes the `PASS` node only if there are other legal moves left. So if you `avoid` all legal moves including `pass`, LZ will still pass, otherwise there would be a segmentation fault.

This still strikes me as unnatural compared to the more "mathematically extensional" behavior of signalling an error at the root and propagating a forced loss deeper in the search.

gcp:
This appears to be failing in CI, but it looks like confusing due to the force-push? I'll retrigger.

gcp:
@betterworld given that it's still failing in CI, maybe you forgot to commit a change?

dbosst:
> > This appears to be failing in CI
> 
> Yeah this is really strange. Travis shows error messages like this:
> 
> ```
> /src/src/UCTSearch.cpp:765:9: error: 'cfg_analyze_interval_centis' was not declared in this scope
> ```
> 
> however I completely removed the `cfg_analyze_interval_centis` variable from the source. So I guess it is building the wrong revision?
> 
> I have double-checked that `cfg_analyze_interval_centis` does not appear in [7eaf7b2](https://github.com/gcp/leela-zero/commit/7eaf7b26a97139b5f6d8c218e58a418688767aaa), which is the latest commit in my branch.

Your branch is behind `gcp:next` ... actually the problem is that plus because from one of my pull requests (#2114), which gcp already merged, where I used `cfg_analyze_interval_centis` (where you removed it from UCTSearch.cpp)

betterworld:
@dbosst 

Oh right, thank you. I think I understand what happened.

I rebased my avoidmove branch so I think it should build in the CI now.

Ishinoshita:
1). Is there a syntax to exclude pass move?

2). Is there a way to exclude pass move not only at root but in the whole search tree?

3). A way to do 1) or 2) for both sides? 

Thanks!

betterworld:
@gcp Thanks for merging!

@Ishinoshita 

> Is there a syntax to exclude pass move?

Just use `pass` instead of coordinates

> Is there a way to exclude pass move not only at root but in the whole search tree?

`lz-analyze avoid b pass 700` should work (can't test right now)

> A way to do 1) or 2) for both sides?

`lz-analyze avoid b pass 700 avoid w pass 700`

roy7:
The move # provided is the number of moves of search depth and not the move # since start of the game right? So if I want to never pass in the first 90 moves of the game I need to start with 'avoid b pass 90 avoid w pass 90' and tick the 90 down as I make moves until after move 90 I'll stop using the lz-analyze command?

(Intent is to stop my bots from passing on a mostly empty board and losing a game to OGS trying to auto-score a big mostly empty board. If human is going to start passing empty, fine, my bot will keep playing anyway. :) )

Thanks!

Ishinoshita:
@roy7 Lol! I asked @betterworld with the very same intent. I noticed that with 4 or 5 handicap stones, both white then black prefered move is pass :-(. I wanted to do some test with handicap, but not with the variable komi  version of LZ; this has petered out...

gcp:
There was a request once to only allow the top x moves to be searched. 

I think variations of that can be built via the ```allow``` tag, but there's still the problem that you can't easily know the move ordering by prior. Just doing a 1 node search won't work because lz-analyze doesn't post moves with no visits, so you won't see their prior. I think I proposed adding a minmoves tag to lz-analyze at some point, which would enforce those to be posted. Then you'd do:

```
lz-setoption name Playouts value 1
lz-analyze minmoves 5
(store top 5 prior moves in the GUI, which in this case would be the 5 mentioned)
lz-analyze allow a1 b1 c1 d1 e1
```

TFiFiE:
To have a front end force the engine to distribute visits over all the possible root moves as uniformly as possible (to get a [winrate heatmap](https://github.com/featurecat/lizzie/pull/321)), I assume the mechanism would be to periodically set the search to visit only those children with less visits than some measure of central tendency, but I'm not sure which one would be best; I'm leaning towards the midrange.

betterworld:
@roy7 

> The move # provided is the number of moves of search depth and not the move # since start of the game right? So if I want to never pass in the first 90 moves of the game I need to start with 'avoid b pass 90 avoid w pass 90' and tick the 90 down as I make moves until after move 90 I'll stop using the lz-analyze command?

Yes, it is the move number relative to the current move number. In my first version, it was the absolute move number unless you prefixed it with `+`, but we decided it is simpler to always use relative move numbers.

PR #1950, opened 2018-10-23
Suppress GCC 8.2 warning in CL/cl2.hpp.

TFiFiE:
GCC is now also giving me the following, but I haven't looked deeply into how this should be properly addressed here:
```
/usr/include/CL/cl_version.h:34:104: note: #pragma message: cl_version.h: CL_TARGET_OPENCL_VERSION is not defined. Defaulting to 220 (OpenCL 2.2)
```

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -3,14 +3,14 @@ THE_OS := $(shell uname -s)
 default:
 	@echo "Detected OS: ${THE_OS}"
 	$(MAKE) CC=gcc CXX=g++ \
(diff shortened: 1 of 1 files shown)


gcp:
We define CL_HPP_TARGET_OPENCL_VERSION.

Is this a bug in the OpenCL C++ headers? If so we should try to fix it there.

gcp:
I reported this as https://github.com/KhronosGroup/OpenCL-CLHPP/issues/53

The warning here might be fixed by c72cb3adeb780e6f5a3640e7bfc234c54bc1242c, can you confirm?

TFiFiE:
Yeah, that removed it.

PR #1953, opened 2018-10-24
Option to start out with visits for each move.

TFiFiE:
This introduces an option to force a full-width search with a given number of visits (as requested in featurecat/lizzie#321 and obviously improving on what was originally in #1891). Note that such behavior cannot be replicated by changing c<sub>puct</sub>, even if only for the root.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -58,6 +58,7 @@ int cfg_max_cache_ratio_percent;
 TimeManagement::enabled_t cfg_timemanage;
 int cfg_lagbuffer_cs;
 int cfg_resignpct;
+int cfg_starting_root_child_visits;
 int cfg_noise;
 int cfg_random_cnt;
 int cfg_random_min_visits;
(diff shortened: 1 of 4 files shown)


gcp:
Why not implement this by executing each move and doing a normal N visit search after it?

It seems pointless to add this option as the GUI can already achieve exactly what it wants.

gcp:
I commented on the Lizzie feature request. Closing this.

kuba97531:
I think it's only possible to do it now if N is a known constant.
If you want it to keep pondering (say 100 playouts to each move at a time) all the moves you'd have to go back and forth between all hundreds of moves.
Additionally it will not work at all because leelaz is prunning a tree immediately after you play a move.

gcp:
No, but the cache would save you in this case? You still have to reconstruct the tree but it will be very fast as you don't actually need to evaluate the NN.

For the pondering case, I would double N and keep looping, i.e. do one visit for all moves, do two visits for all moves, etc.

gcp:
>Note that such behavior cannot be replicated by changing cpuct, even if only for the root.

What if we would expose the policy softmax and PUCT? I'm hardly a fan of that, but if make/eval/unmake is problematic because of tree destruction, it would be another option. I'd rather fix Leela, though!

Can we discuss in #1675?

PR #1955, opened 2018-10-25
(AUTOMATED) Google Cloud free trial instructions

wonderingabout:
Feel free to highlight anything that can be corrected

Thanks for reading.

edit : first thing is the section "System Package corruption should be moved to the bottom, as it is only important if you are using manual instance creation (and not managed instance group), this was part of the old instructions


gcp:
Should we perhaps keep these instructions elsewhere (a Google Doc) and just include a link to it in the repo?

wonderingabout:
i think they are fine in github, but i am not sure what we should do

please give me 10 more minutes, i'm submitting the last commit

wonderingabout:
just done
these new enhancements simplify instructions and have been tested to work successfully

this branch is finalized

wonderingabout:
just finished all minor changes related to this new optimized script

wonderingabout:
i understand, but the advantage of this script is that all packages will always be up to date (it doesnt download any file with wget, but only apt-get update and ppa)

the only thing that will need to be updated is nvidia-driver-410, which anyone can do with a commit

for master branch we just have to slightly modify the script, let me send it here in a few minutes

gcp:
Again, I do not think this belongs on Leela Zero's github. Random changes in Google Cloud could require modifying the instructions. Please, please put it elsewhere (Google Docs should be fine and appropriate) so we can just link to it.

wonderingabout:
ok, but then would people who want to modify it have to copy paste all the instructions in a new google docs link ?
how will they access the source article ?

i cannot guarantee that i will always be here to modify it if such a situation happens

gcp:
You can share access with a specific Google Doc with multiple people, so that works. Alternatively, yes, someone could just make a copy if you are no longer around.

If I pull this in, and it sits in master, now I suddenly have to make update commits to the master branch (and the next branch) to keep it up to date. I appreciate your optimism that this will keep working but I am careful here knowing Google :-)

If it can be independently updated by some trusted people who actually use it, it has much better odds of saying up to date.

wonderingabout:
ok i understand,

i thought the github format would be more convenient for collective contributed update, but i may have failed to think it bigger

one last thing i'd like to suggest, is to merge the updated README.md file in my branch

i separated contributing with/without my own hardware, because it is not clear to many people that cloud is actually independant from their own hardware, and that it works even if their hardware is powered off.

https://github.com/wonderingabout/leela-zero/blob/next/README.md

gcp:
Yes, that would be the idea, i.e. to put the link with some minimal explanation in an updated README.md here.

wonderingabout:
ok then,

should i create 2 branches (one to update MASTER and one to update NEXT),
with 1 google docs file (that says 1)if you want to use master branch use this script, and 2) if you want to use next branch, use that script ?)?

gcp:
Just update next. I am trying to get a v0.16 release and then next will be merged to master.

wonderingabout:
ok,
i will have a look at how google docs works (never used to create files before, for privacy, etc), and i'll add to the google docs file a mediafire folder containing all screenshots used in the tutorial (if they need to be reused)

i would like to ask while i have the opportunity which trusted users i can/should add to the doc
this pull request can be closed then, thank you for your time and advice

wonderingabout:
@gcp @alreadydone , did the google docs : 

it continues here : https://github.com/gcp/leela-zero/pull/1959

PR #1964, opened 2018-10-29
Correct initialize different size sgf

zliu1022:
The sgf size initialized should be same as leelazero program's size settings, so leelazero can open size 9 , size 7 sgf file corretly.

### DIFF
--- a/src/SGFTree.cpp
+++ b/src/SGFTree.cpp
@@ -44,7 +44,7 @@ void SGFTree::init_state() {
     // Initialize with defaults.
     // The SGF might be missing boardsize or komi
     // which means we'll never initialize properly.
-    m_state.init_game(19, 7.5f);
+    m_state.init_game(BOARD_SIZE, 7.5f);
 }
 
 KoState * SGFTree::get_state() {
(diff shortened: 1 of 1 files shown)


gcp:
https://www.red-bean.com/sgf/properties.html#SZ
		Default value: game specific
			       for Go: 19 (square board)
			       for Chess: 8 (square board)

If the SGF has an SZ property, then Leela Zero will handle it here:
https://github.com/gcp/leela-zero/blob/1953dc865bf054f5d49d36c4e6852320c28f35fb/src/SGFTree.cpp#L137

If it does not have one, then per the above it's 19x19.

zliu1022:
When I compile leelazero with BOARD_SIZE 7 or 9, we 'd better init game with BOARD_SIZE not the fixed size 19, otherwise it will crash according to below code:

`
void KoState::init_game(int size, float komi) {
    assert(size <= BOARD_SIZE);
`

gcp:
What about std::min(BOARD_SIZE, 19)?

zliu1022:
Ok. Sure. 
I update the code.

> What about std::min(BOARD_SIZE, 19)?

Issue #1973, opened 2018-11-01
--full-tuner can do 3 tunings instead of 1 or 2


liujn2018:
I have downloaded the latest source and compiled on google colab.
However, when I did the full tuner, the tuner process ran again and again.

My command is:
`!cd run_lz_in_gg && ./leelaz -w networks/2da87ea8da0f54e87b70159e6bb82811b61d1c31091b6e019fbe62aeaa803b9c.gz --full-tuner
`

The output is:
`Using 2 thread(s).
RNG seed: 11407022574244404640
Leela Zero 0.16  Copyright (C) 2017-2018  Gian-Carlo Pascutto and contributors
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; see the COPYING file for details.

BLAS Core: built-in Eigen 3.3.5 library.
Detecting residual layers...v1...256 channels...40 blocks.
Initializing OpenCL (autodetecting precision).
Detected 1 OpenCL platforms.
Platform version: OpenCL 1.2 CUDA 9.2.176
Platform profile: FULL_PROFILE
Platform name:    NVIDIA CUDA
Platform vendor:  NVIDIA Corporation
Device ID:     0
Device name:   Tesla K80
Device type:   GPU
Device vendor: NVIDIA Corporation
Device driver: 396.44
Device speed:  823 MHz
Device cores:  13 CU
Device score:  1112
Selected platform: NVIDIA CUDA
Selected device: Tesla K80
with OpenCL 1.2 capability.
Half precision compute support: No.
Detected 1 OpenCL platforms.
Platform version: OpenCL 1.2 CUDA 9.2.176
Platform profile: FULL_PROFILE
Platform name:    NVIDIA CUDA
Platform vendor:  NVIDIA Corporation
Device ID:     0
Device name:   Tesla K80
Device type:   GPU
Device vendor: NVIDIA Corporation
Device driver: 396.44
Device speed:  823 MHz
Device cores:  13 CU
Device score:  1112
Selected platform: NVIDIA CUDA
Selected device: Tesla K80
with OpenCL 1.2 capability.
Half precision compute support: No.

Started OpenCL SGEMM tuner.
Will try 5355 valid configurations.
[...]

PR #1986, opened 2018-11-03
Fix for #1973 - full tuner broken for heterogeneous GPUs and auto precision detection

ihavnoid:
See #1973 

--full-tuner implies --tune-only
--full-tuner requires an explicit precision

### DIFF
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -244,9 +244,8 @@ static void parse_commandline(int argc, char *argv[]) {
         }
     }
     if (cfg_precision == precision_t::AUTO) {
-        // Auto precision is not supported for tune only cases.
-        // Since full tuner implies tune only, check that too.  #1973
-        if (cfg_sgemm_exhaustive || cfg_tune_only) {
+        // Auto precision is not supported for tune only cases. #1973
+        if (cfg_tune_only) {
             printf("Automatic precision not supported when tuning only\n");
             printf("Please add '--precision single' or '--precision half'\n");
             exit(EXIT_FAILURE);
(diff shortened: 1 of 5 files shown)


Hersmunch:
Are the changes to tfprocess meant to be included?

ihavnoid:
Argh, my bad.  Will push a change that reverts my experiments.

ihavnoid:
@Hersmunch - okay reverted the tensorflow code changes.

src/Leela.cpp
```diff
@@ -238,6 +243,15 @@ static void parse_commandline(int argc, char *argv[]) {
             exit(EXIT_FAILURE);
         }
     }
+    if (cfg_precision == precision_t::AUTO) {
+        // Auto precision is not supported for tune only cases.
+        // Since full tuner implies tune only, check that too.  #1973
+        if (cfg_sgemm_exhaustive || cfg_tune_only) {
```
Hersmunch: Is there a good reason for repeating the full tuner implying tune only handling? I’d prefer if the decision is made only in one place.

ihavnoid: Yes I think it makes sense.  I'll push achange.

src/Leela.cpp
```diff
@@ -238,6 +243,15 @@ static void parse_commandline(int argc, char *argv[]) {
             exit(EXIT_FAILURE);
         }
     }
+    if (cfg_precision == precision_t::AUTO) {
+        // Auto precision is not supported for tune only cases.
+        // Since full tuner implies tune only, check that too.  #1973
```
Hersmunch: You missed the comment :)

gcp:
Did some minor edits.

Remove references to issue numbers - only makes sense if you can look it up on github, and you might as well use git blame to understand why code is there.

Replaced std::list by std::vector.

Pushed as 4a044d9.

But I will have to back this out: it's not compatible with AutoGTP, which needs to tune once but has no idea about what precision it should use.

PR #2014, opened 2018-11-13
Optimized out and out_in kernels

Ttl:
Very minor speedup of about 2% with batch size of 1. With batch size of 5 I get speedup of about 5% with half precision and 12% with single precision.

Out transformation memory accesses are almost completely coalesced with the new kernel. In transformation has still very bad memory access pattern when loading the input from memory but I didn't find any improvement to it yet.

With these changes OpenCL backend using batch size of 5 is faster than cudnn with batch size of 32 on my PC.


gcp:
Tested on AMD Polaris, it also helps about 4% there.

PR #2019, opened 2018-11-15
Option to use "full-tuner" without "tune-only".

TFiFiE:
This restores the possibility removed by #2004, if requested explicitly with `--full-tuner --tune-only=no`;
`--tune-only` without argument will still work though.


gcp:
What's the use case for this?

TFiFiE:
It would save a bit of time if initializing the engine becomes slower in the future with differently compressed networks.

PR #2029, opened 2018-11-18
Refactor to allow autogtp to use Engine

Hersmunch:
Engine was added in #1652 as part of allowing Validation to send GTP commands. This change expands the use of Engine to autogtp for consistency and hopefully making it easier for a later change to allow for the server to use GTP commands for self-play and/match games.

I wrote this a while ago and remember doing some quick testing but I haven't had time to go through this as thoroughly as I would like. Especially as I don't want to break autogtp! I decided to make a PR to either help me get back on it or allow someone else to reuse it. So yeah, I'd appreciate any review feedback and suggestions for testing.

### DIFF
--- a/autogtp/Job.cpp
+++ b/autogtp/Job.cpp
@@ -48,17 +48,21 @@ void Job::init(const Order &o) {
 }
 
 ProductionJob::ProductionJob(QString gpu, Management *parent) :
-Job(gpu, parent)
+    Job(gpu, parent),
+    m_engine(Engine(QString(), QString()))
 {
 }
 
 ValidationJob::ValidationJob(QString gpu, Management *parent) :
-Job(gpu, parent)
+    Job(gpu, parent)
 {
+    for (auto engine : m_engines) {
+        engine = Engine(QString(), QString());
+    }
 }
 
 WaitJob::WaitJob(QString gpu, Management *parent) :
-Job(gpu, parent)
+    Job(gpu, parent)
 {
 }
 
--- a/autogtp/Job.cpp
+++ b/autogtp/Job.cpp
@@ -27,14 +27,12 @@
 
 Job::Job(QString gpu, Management *parent) :
     m_state(RUNNING),
-    m_option(""),
-  m_gpu(gpu),
-  m_boss(parent)
+    m_gpu(gpu),
+    m_boss(parent)
 {
 }
 
 void Job::init(const Order &o) {
-    m_option = " " + o.parameters()["options"] + m_gpu + " -g -q -w ";
     QStringList version_list = o.parameters()["leelazVer"].split(".");
     if (version_list.size() < 2) {
         QTextStream(stdout)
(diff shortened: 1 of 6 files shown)


autogtp/Job.cpp
```diff
@@ -131,7 +129,7 @@ void ProductionJob::init(const Order &o) {
 
 Result ValidationJob::execute(){
     Result res(Result::Error);
-    Game first("networks/" + m_firstNet + ".gz",  m_option);
+    Game first(m_engines[0]);
```
marcocalignano: Change This to ```m_engineFirst``` like we do for the SFG strings

autogtp/Job.cpp
```diff
@@ -140,7 +138,7 @@ Result ValidationJob::execute(){
         first.setMovesCount(m_moves);
         QFile::remove(m_sgfFirst + ".sgf");
     }
-    Game second("networks/" + m_secondNet + ".gz", m_option);
+    Game second(m_engines[1]);
```
marcocalignano: Change this to ```m_engineSecond```

autogtp/Job.cpp
```diff
@@ -205,8 +204,10 @@ Result ValidationJob::execute(){
 
 void ValidationJob::init(const Order &o) {
     Job::init(o);
-    m_firstNet = o.parameters()["firstNet"];
-    m_secondNet = o.parameters()["secondNet"];
+    m_engines[0].m_network = "networks/" + o.parameters()["firstNet"] + ".gz";
```
marcocalignano: Same as above

autogtp/Job.h
```diff
@@ -79,8 +79,7 @@ class ValidationJob : public Job {
     void init(const Order &o);
     Result execute();
 private:
-    QString m_firstNet;
-    QString m_secondNet;
+    Engine m_engines[2];
```
marcocalignano: Same as above

marcocalignano:
The code looks good but I would like to test it for a bit. 
@Hersmunch How many Games did you do with this code?

Hersmunch:
My machine isn’t fast enough to get match games :( I didn’t run many self play games either. If I recall correctly, just a few to check the games start ok and that the sgf player names are ok.

I’ll probably have time tomorrow to go through the changes you’ve proposed. Thanks!

Hersmunch:
Got a few more games but no matches. I'll leave it running overnight.

gcp:
Is someone else with a fast GPU willing to test this branch?

marcocalignano:
@gcp  Do you want to test this in next?
@Hersmunch Can you change the little things i reviewed  and then make a PR in next?

Hersmunch:
@marcocalignano I think I have done but do give me specific instructions if you think otherwise as I can't see anything I've missed.

gcp:
>Do you want to test this in next?

Sounds good. Development branch is for development :-)

marcocalignano:
@Hersmunch  Perfect! I didn't notice the last commit!

PR #2031, opened 2018-11-19
Microsoft azure cloud free trial added (master branch)

wonderingabout:
arround 250 hours Tesla V100 computing for free

the doc is here : 
https://docs.google.com/document/d/1DMpi16Aq9yXXvGj0OOw7jbd7k2A9LHDUDxxWPNHIRPQ/edit

General display : 
https://github.com/wonderingabout/leela-zero/tree/azure-cloud-docs

i added computing time in hours for free as an incentive and indication
prices have been stable in the last years so the estimate should stay accurate

these 2 contributors have edit rights on the azure and google cloud docs,
they can also add other people to edit the doc (the people added will have the right to add people too, so better choose them wisely)

@alreadydone 
@roy7

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,10 @@ Many cloud companies offer free trials (or paid solutions, not discussed here)
 that are usable for helping the leela-zero project.
 
 There are community maintained instructions available here:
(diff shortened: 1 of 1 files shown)


gcp:
Please make pull requests to the next branch only.

wonderingabout:
yes indeed

in the past the first link only had google cloud free trial instructions, but mentionned the name of other cloud companies giving free trials with GPUs even though it didnt give any instruction for azure or oracle

this is why i removed azure and oracle from the first link now

however i dont understand why i should commit to next
forgive my lack of experience, but as it was only documentation (no code change), i thought it would be better to include the new azure instructions on master branch too

can you clarify for me please
thanks

Issue #1093, opened 2018-03-24
Add unit tests for FullBoard


barrybecker4:
Now that I have experience from adding gtests for FastBoard, I would like to add them for FullBoard.
I discovered, while adding tests for FastBoard, that some of the methods (count_pliberties, is_eye, is_suicide) only make sense to test in the context of a FulBoard.

If it makes sense to do so, I can open issues for the following and implement them as time permits:
 - Either remove set/get_square (because it is unused) or make m_square private in FastBoard and allow access only with set/get_square. My preference is for the latter.
 - Refactor out FastBoardSerializer class from FastBoard. I think it would simplify the code to extract all the logic having to do with board serialization to a separate class. It would have methods for serialize, text_to_move, text_to_move_sgf and probably starpoint.
 - I think there should be an update_board method in FastBoard that contains all the logic from FullBoard::update_board except the part that deals with the Zobrist hashing. Then the FullBoard::update_board method would just do the Zobrist update and call FastBoard::update_board. If we did this, then we could create sample FastBoards in the unit tests and tests for count_pliberties, is_eye, is_suicide, and area_score would make sense. As it is now, it really only makes sense to test those methods in the context of a FullBoard.

barrybecker4:
While writing the FullBoard tests to validate is_eye(v), I think I may have discovered a bug. Hopefully it is just my misunderstanding.
All the cases make sense to me except for the last one in the list below (the point at "b5").
```
/*
  Only single points surrounded by own color are counted as eyes.
      a b c d e
    5 X . X X .  5
    4 . . X . X  4
    3 X X . X .  3
    2 . O X O .  2
    1 . . . . .  1
      a b c d e
*/
TEST(FullBoardTest, IsEyeOnBlackField) {
    FullBoard b = create_5x5_all_black();
    EXPECT_EQ(true, b.is_eye(FastBoard::BLACK, b.get_vertex(4, 4))); // black eye
    EXPECT_EQ(true, b.is_eye(FastBoard::BLACK, b.get_vertex(3, 3))); // black eye
    EXPECT_EQ(false, b.is_eye(FastBoard::WHITE, b.get_vertex(3, 3))); // not white eye
    EXPECT_EQ(false, b.is_eye(FastBoard::BLACK, b.get_vertex(2, 2))); // potentially false eye, so not real eye
    EXPECT_EQ(false, b.is_eye(FastBoard::BLACK, b.get_vertex(1, 1))); 
[...]

PR #2043, opened 2018-11-24
Next 1093 board tests and refactor bb

### LINKED ISSUE #709: Add Unit tests for FastBoard
I think there should be unit tests added for many of the classes, but FastBoard seems like a good place to start.

I don't have much experience adding unit tests for C++ projects, so some guidance would be appreciated. Should we use Google Test, CppUnit, or MS native c++ unit test framework? Probably Google Test is best, unless someone has reasons for something else. I see that there is a gtest directory at the root, but it is empty.

Once tests are added, how can they be run automatically when building?

barrybecker4:
These changes add tests for FullBoard and FastBoard, and they also make some changes to these classes which I believe to be improvements.
The two main changes that I made were:

1. Moved update_board into FastState so that the unit tests that manipulate the state of the board will give valid games results when moves are added. Without this change the expected results in FastBoard seemed inaccurate. The part of update_move that updates the Zobrist remains in FullBoard and is called via "record_position" which does nothing in FastBoard, but upates the hash in FullBoard. It should be clear now that the sole purpose of FullBoard is to add management of the Zobrist hashes to FastBoard. 
1. The logic related to serializing FastBoard was extracted out into a separate FastBoardSerializer class. 

If these changes are accepted, I will close #709 and #1093, and will move on to adding tests for FastState, KoState, and GameState.


barrybecker4:
The travis-ci build says

> [1Berror pulling image configuration: unknown blob
> travis_time:end:06b8d5bc:start=1543083220405234099,finish=1543083221594615864,duration=1189381765
> [0K[31;1mThe command "docker pull ubuntu:16.04" failed and exited with 1 during .[0m

Not sure what that means.

gcp:
This needs a rebase.

Issue #2036, opened 2018-11-21
PTX inline assembly for using tensor cores


ihavnoid:
On #2007 I mentioned that I will attempt writing inline assembly to use the tensor cores.  Well, I think I got a functional version of hgemm that uses the wmmv instructions!

https://github.com/ihavnoid/hgemmtest

Currently it runs around 30ms on a Tesla v100 (tested on Google Compute Platform).  This is similar to what I got while running the leela-zero tuner, so it's hardly any improvement yet.  Not bad for the first implementation though. :)

```
(228/290) KWG=32 KWI=8 MDIMA=8 MDIMC=8 MWG=16 NDIMB=8 NDIMC=8 NWG=32 SA=1 SB=1 STRM=0 STRN=0 VWM=2 VWN=4 0.0317 ms (3716.1 GFLOPS)
```

Since this is only the first iteration, hopefully we can better tune the matrix multiplication code, and one day when the assembly version performs much better we can attempt to merge it into leela-zero.  I need some help from people who has better expertise on fine-tuning these kind of kernels.

ihavnoid:
Plus, I would like to have some people test the code on how it performs on an Turing GPU

ihavnoid:
Correction : 30us.  Not 30 ms.  Got a bit more tuning (m32n8k16 instead of m16n16k16) and now I am at 25us. :)

Ttl:
Very interesting. It might be hard to utilize the hardware fully at batch size of 1. Batching in Winograd increases the n-dimension in sgemm (n = batch_size * tiles and tiles = 25 for F(4x4, 3x3)) allowing use of bigger tiles.

Currently transformations take about 30 - 40% of runtime so doubling the sgemm speed results in about 50% nps increase. Cudnn actually performs better using im2col + sgemm instead of Winograd at low batch sizes even though it's theoretically more work, but it probably uses tensor cores more effectively.

NVidia doesn't provide profiler for opencl anymore and the only way to profile opencl kernels is to compile them with cuda. It isn't as hard it seems, kernels usually work just by making some defines (For example: https://github.com/CNugteren/CLBlast/blob/master/src/kernels/opencl_to_cuda.h). Host code also needs to be written with cuda and there are examples for that too in CLBlast.

sewim:
It is going to be happened? Interesting...
So basically, Leela zero will be released normal version and also Nvidia specific version?

alreadydone:
[...]

PR #2049, opened 2018-11-27
Tensor Core support with PTX inline assembly

ihavnoid:
See #2036 for details.  ~10% performance gain on a V100 cluster.  Validated that doesn't lose strength on equal playout (vs. current next branch)

### DIFF
--- a/src/OpenCL.cpp
+++ b/src/OpenCL.cpp
@@ -59,7 +59,7 @@ const std::string sourceCode_common =
     #include "kernels/common.opencl"
 ;
 
-static const std::string sourceCode_tensorcore_test = 
+static const std::string sourceCode_tensorcore_test =
     #include "kernels/tensorcore_test.opencl"
 ;
 
--- a/src/OpenCL.cpp
+++ b/src/OpenCL.cpp
@@ -406,9 +406,9 @@ void OpenCL_Network<net_t>::convolve3(OpenCLContext & opencl_context,
 
         // tensorcore implementation uses a different dimension
         if (tce) {
-            local_sgemm = {32 * mdimc/16, ndimc/16, 1};
-            size_sgemm = {32 * m_ceil / 16 * mdimc / mwg,
-                          n_ceil / 16 * ndimc / nwg,
+            local_sgemm = {32 * mdimc/32, ndimc/8, 1};
+            size_sgemm = {32 * m_ceil / 32 * mdimc / mwg,
+                          n_ceil / 8 * ndimc / nwg,
                           cl::size_type(WINOGRAD_TILE)};
         }
         queue.enqueueNDRangeKernel(sgemm_kernel, cl::NullRange,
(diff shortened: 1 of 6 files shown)


gcp:
@ihavnoid is this solid enough to pull?

Does @Ttl perhaps want to help reviewing?

Ttl:
I don't have GPU with tensor cores so I can't test this, but looking at the code I don't really see any issues.

ihavnoid:
I think the code itself is okay, but after second thought I think we can wait until we have the cuda implementation so that we can skip merging this if cuda performs much better than this.  I think there is no need to hurry merging, it's not like there are other outstanding branches that might conflict a lot (or... do we?)

Biggest concern I have is that I tested this only on a v100 in Google cloud.

gcp:
>Biggest concern I have is that I tested this only on a v100 in Google cloud.

From what I understand, the optimal use of Tensor Cores on RTX (compute capability 7.5) devices is different from that on 7.0 devices, due to the Tensor Core vs CUDA ratio. I'm not sure if this has any effect on your code.

But it could be seen in some benchmarks of the RTX cards, where SGEMM benchmark kernels did not get the expected speed on the RTX cards because they weren't compiled with 7.5 support.

jpturcotte:
If that could help, I would be ready to test this on a desktop with a RTX 2070. I could also test the CUDA backend, but I would need help to set all this up. I can setup Git and MSVC, but I am not sure what to do after that.

gcp:
First observation: we are only checking one specific instruction to see whether it's supported: 4fc611335a36f4648997806ce14cadd111185385

But the tuner selects:

(363/380) KWG=16 KWI=2 MDIMA=8 MDIMC=8 MWG=32 NDIMB=32 NDIMC=32 NWG=32 SA=0 SB=0 STRM=0 STRN=0 TC E=1 VWM=2 VWN=2 0.0236 ms (5008.7 GFLOPS)

Which seems inconsistent with https://github.com/gcp/leela-zero/pull/2049/files#diff-248059ab88d371db2d8ef2fe42733d7cR84

gcp:
Hmm, I guess that is working as intended and indicates that this card does support them. We report TC support as soon as any of the instructions work, and then filter in the tuner.

gcp:
Factory clocked RTX 2080, benchmark with ffe8ba44.

next:
4t = 408
8t = 446
16t = 560

tensorcores:
4t = 507
8t = 579
16t = 549

Tuner goes from 3 TFLOPS to 5 TFLOPS.

Seems good enough to pull, even if we're clearly having some bottleneck elsewhere.

lp200:
I hope ihavnoid/batch-full branch will be merged
It shows great performance with tensor core.

V100 256x20 net

-t 16  batch 5
 6400 evaluations in  2.11 seconds -> 3039 n/s

-t 32 batch 12
 6400 evaluations in  1.57 seconds -> 4085 n/s

ihavnoid:
> Hmm, I guess that is working as intended and indicates that this card does support them. We report TC support as soon as any of the instructions work, and then filter in the tuner.

Yes correct.  There seems to be some GPUs that only support m16n16k16 format while others support more formats.  Testing all three fails to pick up those GPUs.

ihavnoid:
> I hope ihavnoid/batch-full branch will be merged
> It shows great performance with tensor core.
> 
> V100 256x20 net
> 
> -t 16 batch 5
> 6400 evaluations in 2.11 seconds -> 3039 n/s
> 
> -t 32 batch 12
> 6400 evaluations in 1.57 seconds -> 4085 n/s

I'll work on a cleanup.  Currently it merge-conflicts on various places.

PR #2052, opened 2018-11-29
Add ability for the server to specify an sgf as the initial position for self-play and match games

Hersmunch:
I've cherry-picked some of @amato-gianluca's commits from SAI (I hope that's ok?). Specifically those to allow the server to start self-play games from an sgf (i.e. some of those listed here https://github.com/sai-dev/sai/commits/sai-0.15/autogtp/Management.cpp). I've then extended this to also cover match games.

See https://github.com/sai-dev/sai/issues/7 for a possible use for this in self-play. I think allowing similar functionality for match games could be useful and is a small change, so why not.

If these are accepted as sound enough reasons for including this PR then I'll raise a server issue asking @roy7 for help testing. All I can do and have done so far is get a couple of self-play games (based on existing server responses) and they look like they ran fine.
http://zero.sjeng.org/view/0808e80c508107b8cdb7c5b3ca5a11c5d3b1f9569a50fb2cb14de910d3af2ac8?viewer=wgo
http://zero.sjeng.org/view/c658b51e65892faf440c11ed85b76502641bc7de385fff473eafa45980f3bfa3?viewer=wgo [with a store and restore partway through]

The basic idea is to allow for optionally adding this to the JSON:
```
    hash_sgf_hash: "7dbccc5ad9eb38f0135ff7ec860f0e81157f47dfc0a8375cef6bf1119859e537",
    moves_count: "92"
```
Thoughts?

### DIFF
--- a/autogtp/Game.cpp
+++ b/autogtp/Game.cpp
@@ -349,11 +349,6 @@ bool Game::loadSgf(const QString &fileName, int moves) {
     return sendGtpCommand(qPrintable("loadsgf " + fileName + ".sgf " + QString::number(moves)));
 }
 
-bool Game::loadTraining(const QString &fileName, int moves) {
-    QTextStream(stdout) << "Loading " << fileName + ".train with " << moves << " moves" << endl;
-    return sendGtpCommand(qPrintable("load_training " + fileName + ".train " + QString::number(moves)));
-}
-
 bool Game::fixSgf(const QString& weightFile, const bool resignation) {
     QFile sgfFile(m_fileName + ".sgf");
     if (!sgfFile.open(QIODevice::Text | QIODevice::ReadOnly)) {
--- a/autogtp/Game.cpp
+++ b/autogtp/Game.cpp
@@ -344,9 +344,9 @@ bool Game::loadSgf(const QString &fileName) {
     return sendGtpCommand(qPrintable("loadsgf " + fileName + ".sgf"));
 }
 
-bool Game::loadSgf(const QString &fileName, int moves) {
+bool Game::loadSgf(const QString &fileName, const int moves) {
     QTextStream(stdout) << "Loading " << fileName + ".sgf with " << moves << " moves" << endl;
-    return sendGtpCommand(qPrintable("loadsgf " + fileName + ".sgf " + QString::number(moves)));
+    return sendGtpCommand(qPrintable("loadsgf " + fileName + ".sgf " + QString::number(moves + 1)));
 }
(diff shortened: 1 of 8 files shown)


amato-gianluca:
We have no problems if you use sai code in leelaz. Commit 40ff01c is quite ugly and uses the deprecated std::tmpnam function. A more elegant solution would probably require using std::tmpfile (or the higher-level QtTemporaryFile class) and passing around the file handler until it is time to delete the sgf.

marcocalignano:
But if you do not have the training data it is useless for the training. When autogtp save/restore from the HD it also save/restore the training data. If you want the server to distribute already started game it needs to provide also the training data.

Hersmunch:
@marcocalignano, whilst I agree that restoring training data would be a good thing to include (perhaps in this PR?), it isn't needed for the SAI use-case example I linked to (although their aim was later achieved using a different method).

@amato-gianluca, thanks for the source and your feedback. If the premise for this PR is accepted then I'll look at tidying up the implementation.

Hersmunch:
Rebased. Will leave it running for a while locally before seeing what can be done about the `std::tmpnam()` usage.

Hersmunch:
I went through and replaced `std::tmpnam()` with `QTemporaryFile` and then realised I should probably just use `QUuid` like the other code that makes temporary files. Oh well. I've also changed matches to only need one sgf file so now the optional syntax is the same as for self-play jobs.
I haven't yet looked at adding optional JSON for specifying a training file to start from for self-play but I do intend to do this as it won't be hard. Since the training files aren't hosted for easy download I guess I'll make it go look for gzipped files at "https://zero.sjeng.org/training" unless anyone has any other suggestions.

I have now tested this by locally modifying the order construction. So I've now started from mid-way through an sgf and checked that storing and restoring data for both self-play and match games all appears to work.

Hersmunch:
@gcp thanks for merging. Just wanted to point out that in your commit comment it incorrectly says that training files are downloaded:
> * Automatic downloading of sgf and training files.

Also, when this comes to be released it should be noted that as the elements of a match order have changed (just `sgf` rather than `sgfFirst` and `sgfSecond`) it will not be possible for users to store from the current version using `q+Enter` and then restore with the new version.

I saw in your commit comment that loading of training data is not needed by what is your opinion on whether it is a sensible thing to support at all? I'm currently thinking it is not worth doing because the server could duplicate any training data up to the position in an sgf starting point, thus making sending it back and forth pointless.

PR #2057, opened 2018-11-30
Add ability for LZ to maintain a chosen winrate against any opponent of any strength

AncalagonX:
This may or may not fit within the intended scope of `gcp/leela-zero`, so no harm if it ends up not being accepted, though many people have expressed interest in a function like this for practicing against LZ without being destroyed in every game. This is my first substantive PR for the project, so hopefully we can at least take the time to bring the code up to standard so I know what to expect when I submit additional PRs shortly down the road.

This PR adds the ability for custom winrate targeting using the new command-line option `--winratetarget x` where `x` is an integer ranging from 0 to 100 percent. The engine will attempt to play moves on its turns to force the board state into always maintaining a certain winrate throughout the game, regardless of how skillfully or poorly its opponent plays. During reading and pondering, LZ always assumes its opponent will play each of its moves optimally at full strength. If the opponent "surprises" LZ by playing a weaker move than expected, LZ's search automatically compensates by guiding it towards weaker PVs that bring the board state back to the user's desired winrate. In practice, this works *very* well.

The basic MCTS implementation that balances exploration vs exploitation with `value = winrate + puct` remains absolutely unchanged. Instead, `winrate` in the value function for LZ's turns is calculated based on *how close LZ's assessed winrate was to the user's desired winrate*. The magic happens on just two lines of UCTNode.cpp., right after `auto value = winrate + puct;`:

```
		if (!unmodified_search && !is_opponent_move) {
			value = (1 - abs(winrate_target_value - winrate)) + puct;
		}
```

(The `unmodified_search` bool guarantees no change in `uct_select_child` behavior unless the user has input a valid value for `--winratetarget`)

There is also some relevant support code in UCTSearch.cpp to take care of the following:

- New bool "is_opponent_move" passed to uct_select_child to indicate whether a visit being made at any arbitrary depth is an "engine move" or an "opponent move" (essentially whether `depth` is odd or even). This is what decides whether to use the "winrate-targeting value function" for the engine's own moves, or the "full strength, regular LZ search" for its opponent's moves.
- New bool "is_pondering_now" is true/false depending on whether engine is pondering during opponent's turn. Used to adjust the odd/even `depth` determination to return the proper bool value for "is_opponent_move" during pondering.

### DIFF
--- a/src/UCTSearch.cpp
+++ b/src/UCTSearch.cpp
@@ -40,6 +40,7 @@
 using namespace Utils;
 
 constexpr int UCTSearch::UNLIMITED_PLAYOUTS;
+bool is_pondering_now = false;
 
 class OutputAnalysisData {
 public:
@@ -224,8 +225,16 @@ SearchResult UCTSearch::play_simulation(GameState & currstate,
         }
     }
 
+	auto depth =
+		int(currstate.get_movenum() - m_rootstate.get_movenum());
+	auto apparent_depth = depth;
+	if (is_pondering_now == true) { // Adjust depth by 1 to account for thinking/pondering during opponent's turn
+		apparent_depth =
+			(1 + int(currstate.get_movenum() - m_rootstate.get_movenum()));
+	}
+
     if (node->has_children() && !result.valid()) {
-        auto next = node->uct_select_child(color, node == m_root.get());
+        auto next = node->uct_select_child(color, node == m_root.get(), ((apparent_depth % 2) != 0));
         auto move = next->get_move();
 
         currstate.play_move(move);
@@ -778,6 +787,7 @@ int UCTSearch::think(int color, passflag_t passflag) {
 }
 
 void UCTSearch::ponder() {
+	is_pondering_now = true;
     update_root();
 
     m_root->prepare_root_node(m_network, m_rootstate.board.get_to_move(),
@@ -818,6 +828,7 @@ void UCTSearch::ponder() {
     dump_stats(m_rootstate, *m_root);
 
     myprintf("\n%d visits, %d nodes\n\n", m_root->get_visits(), m_nodes.load());
+	is_pondering_now = false;
 
     // Copy the root state. Use to check for tree re-use in future calls.
     m_last_rootstate = std::make_unique<GameState>(m_rootstate);
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -62,6 +62,7 @@ int cfg_noise;
 int cfg_random_cnt;
 int cfg_random_min_visits;
 float cfg_random_temp;
+int cfg_winrate_target;
 std::uint64_t cfg_rng_seed;
 bool cfg_dumbpass;
 #ifdef USE_OPENCL
@@ -144,6 +145,7 @@ void GTP::setup_default_parameters() {
     cfg_random_cnt = 0;
     cfg_random_min_visits = 1;
     cfg_random_temp = 1.0f;
+	cfg_winrate_target = 100;
     cfg_dumbpass = false;
     cfg_logfile_handle = nullptr;
     cfg_quiet = false;
--- a/src/GTP.h
+++ b/src/GTP.h
@@ -45,6 +45,7 @@ extern int cfg_noise;
 extern int cfg_random_cnt;
 extern int cfg_random_min_visits;
 extern float cfg_random_temp;
+extern int cfg_winrate_target;
 extern std::uint64_t cfg_rng_seed;
 extern bool cfg_dumbpass;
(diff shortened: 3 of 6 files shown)


AncalagonX:
Since only the `winrate` is modified (and only for LZ's own moves in a PV), the search favors high-policy moves for initial searches as usual. This is highly desired as it ensures a strong, robust, and generally logical search while the `winrate` component gathers weight in the `value = winrate + puct` function.

Some added points about using targeted winrates:

LZ will play very poor moves at times in response to the opponent playing bad moves. These can manifest as first-line moves inside established LZ territory, as self-atari in opponent territory, and very rarely even passes to force the winrate back into its proper zone. LZ's search is surprisingly well-suited for accurately assessing how its own "bad moves" will affect the game's winrate.

Since the targeted winrate is active throughout the entire game, games set to target winrates between 20% and 80% often end up as very close games, frequently with exactly 0.5-point margins. This is because an 80% winrate very late in the game does not carry the same "score advantage" that an 80% winrate would in the early game.

Additionally, LZ slightly favors ko fights when winrates around 50% are chosen, as a kind of variant of the "horizon effect".

src/UCTSearch.cpp
```diff
@@ -224,8 +225,16 @@ SearchResult UCTSearch::play_simulation(GameState & currstate,
         }
     }
 
+    auto depth =
+        int(currstate.get_movenum() - m_rootstate.get_movenum());
+    auto apparent_depth = depth;
+    if (is_pondering_now == true) { // Adjust depth by 1 to account for thinking/pondering during opponent's turn
+        apparent_depth =
```
gcp: This seems like a fickle way to handle this. What about recording the "computer color" at the root and just comparing to that?

AncalagonX: Thanks for taking the time to look this over.

One issue with remembering computer color is that sometimes players "switch places" with the computer during a game for various reasons, so it might be best to set/update this whenever LZ enters pondering.

I will submit a more elegant handling of this tomorrow.

gcp: Are you still working on this?

AncalagonX: Yes, incoming very shortly thanks to this reminder.

AncalagonX: Rewrite nearly finished, along with a few questions about best practices when I post it

AncalagonX:
I've cleaned up the code, and implemented it hopefully in ways that are more elegant and more in line with the existing LZ codebase. One of the major new changes is the decision to pass `bool is_pondering` and `int depth` directly to `uct_select_child()` in UCTNode.cpp. This provides more tools to users who wish to modify LZ's search behavior in creative ways, such as this PR.

Even if the PR happens to fall too far outside the scope of the main LZ project, it's still super helpful to get code corrections, style suggestions, or other any feedback that will make my future PRs to the project smoother for all involved.

Ishinoshita:
@AncalagonX Very appealing! But what is the feel from a human perspective? Does this LZ alternate very strong move with very odd ones? Or does it pick suboptimal moves in a subtle enough way so that you don't feel like it's obviously curbing its strength by playing cat and mouse? 

Can you attached a sgf of one the here above game? Thanks.

AncalagonX:
@Ishinoshita You can see the SGFs from my bot `YourRank` on KGS for games using this function. If you are SDK-level (around 6-12 months of consistent practice), you will not notice LZ playing sub-par moves more than once or twice in a game.

This PR works exactly as intended, and its only (minor) snag is that the policy head is trained to locate the _most_ optimal moves quickly, rather than moves with sub-par, user-defined winrates. This is not a significant problem for this PR, because the policy head is wide enough for LZ's MCTS to quickly locate appropriate variations. A modified version of this branch has beaten two certified Korean professionals by 0.5 points on Tygem, and regularly ends SDK/DDK (low-level amateur) games within 10 points—though usually even closer.

However, I do not recommend this PR be accepted _solely_ on the basis of intending to provide compelling games to users of all skill levels. The problem is that players using this feature for a few games quickly come to realize that no matter how much they do _or don't_ think, they will never be punished by LZ for it because LZ will adjust the board situation to match the desired winrate. There is a lot more to write about this, and there are solutions, but they go too far beyond this PR and LZ's current official scope.

AncalagonX:
I'm interesting in working out any other problems or deficiencies in this code, _especially_ those issues that are likely to pop up again in later PRs I submit, even if this particular PR is ultimately closed and not accepted. My next major PR enables user-defined search width intended to be used with game review tools such as Lizzie, so I'd ideally like to learn as much as possible from this PR before I submit another one.Thank you for all the feedback so far.

PR #2061, opened 2018-11-30
Next 1093 board tests and refactor bb

barrybecker4:
I did an interactive rebase so that the history is now clean an linear.
There are now just 2 commits for the 2 main changes.


barrybecker4:
Canceling this request so I can fix the style check and try again. It appears that a newline is required at the end of a file.

Issue #1914, opened 2018-10-05
Scalability: 40b is now the best LZ net for game analysis and long games


Friday9i:
Here is an update of the scalability of nets, with LZ181 vs LZ157 included (the pink curve):
![image](https://user-images.githubusercontent.com/37338822/46563194-68309f00-c900-11e8-8f2b-b63ad3d095a0.png)
Legend:
- x-axis: number of visits for the first net (eg LZ181 for the LZ181/LZ157 curve)
- y-axis: ratio of visits needed by second net to get 50% winrate (eg for LZ181/LZ157 and with 100 visits for LZ181, we see the curve at 3 because LZ157 needs ~3 times more visits to be as strong as LZ181, ie to get ~50% winrate).

Interpretation:
- With 1 visit, larger nets are much better than smaller nets. Eg LZ157 needs quite a lot of visits to compete with LZ181@1 visit: about 13 visits are needed for LZ157!
- With more visits, from about 30 visits to about 300 visits, the ratio needed to compete goes down to a minimum, and the minimum is quite flat (from ~30 to ~300 visits). In the case of LZ181 vs LZ157, LZ157 needs only ~3 times more visit to compete with the larger net (but as it is about 4x faster, it confirms LZ157 is still a bit better on time-parity with relatively low visits)
- With more visits, larger nets scale systematically better than smaller nets. For 40b vs 15b, this curve goes up too, but it is still relatively slow: the ratio goes up to ~4 with 1000 visits and ~5 with 3000 visits (ie ~16K visits needed for LZ157 when LZ181 has 3000 visits). Hence, LZ181 is better than LZ157 on time parity when it has above ~1000 visits

Conclusion: 40b (LZ181) is now the best LZ net for game analysis and long games!
(It's now better than LZ157, the best 15b net. But ELF may still be a bit better, that's my next test!)

Note: the last point on the right for LZ181 vs LZ157 is quite noisy, as I did only 50 games with 16K vs 3K visits because it is very slow (but the score was exactly 25/25, so it seems balanced :-)
And to conclude, the same graph on a log-log scale:
![image](https://user-images.githubusercontent.com/37338822/46564155-fce9cb80-c905-11e8-909b-0394bbc0b776.png)
An interesting observation: the ratio increase always quite linearly above ~1000 visits

MaurizioDeLeo:
Very nice work, and a lot of data. Kudos !

[...]

PR #2078, opened 2018-12-13
Add optional separate options for white in match game

Hersmunch:
Allows for the server to optionally add something like this to the JSON for match jobs:
```
   white_options : {
       playouts : "0",
       visits: "1601",
       resignation_percent : "5",
       noise : "false",
       randomcnt : "0"
    }
```
No separate hash has been added; I propose that the existing `options_hash` is calculated over both `options` and `white_options` if it is used.

Similar to #2052, I haven't been able to test this. @roy7 do you have any suggestions for how this could be done? I think I remember seeing similar tests done in the past by serving test jobs on a specifically set up test url.

### DIFF
--- a/autogtp/Job.cpp
+++ b/autogtp/Job.cpp
@@ -211,7 +211,7 @@ void ValidationJob::init(const Order &o) {
     m_engineFirst.m_network = "networks/" + o.parameters()["firstNet"] + ".gz";
     m_engineFirst.m_options = " " + o.parameters()["options"] + m_gpu + " -g -q -w ";
     m_engineSecond.m_network = "networks/" + o.parameters()["secondNet"] + ".gz";
-    m_engineSecond.m_options = " " + o.parameters()["options"] + m_gpu + " -g -q -w ";
+    m_engineSecond.m_options = " " + o.parameters()["optionsSecond"] + m_gpu + " -g -q -w ";
     m_sgf = o.parameters()["sgf"];
     m_moves = o.parameters()["moves"].toInt();
 }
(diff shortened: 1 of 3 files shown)


marcocalignano:
But why would the server want to do that?

Hersmunch:
It could be used for test match analysis of various networks to produce data as in #1914 or perhaps in the training pipeline as part of a gating strategy that involves playing against a pool of opponents of various strengths.
There is no immediate need I know of so apologies if I am wasting anyone's time with this.

I will do some testing on this (maybe later today?) by locally modifying the "order construction" like I have now done for #2052.

Hersmunch:
Tested match orders with and without the new optional `white_options` and fixed an issue with loading a stored match order. Appears to be working as intended.

gcp:
>But why would the server want to do that?

You could use this for scaling testing, for example? Or many things that currently have to be done by developers "by hand" on a single machine.

gcp:
This will need a rebase now that I pulled the other branch.

gcp:
Hmm, I have one question/observation about this, related to how currently matches work. We send white_hash and black_hash. Either white_ and black_ really just mean "first" and "second" player, and AutoGTP picks colors, or the color picking is done on the server side, and white does really mean white. In the latter case, this means that if we want non-standard options for one player, the server should flip the options around in such a case too, right. That is, it would send the normal options as white_options and the special ones as options.

Hersmunch:
As far as I can see currently in AutoGTP black == "first" and white == "second" so the server has to do the picking.

PR #2096, opened 2018-12-24
AutoGTP: Allow specifying initial GTP commands. 

Hersmunch:
Allows for the server to optionally add something like this to the JSON for self-play and match jobs:
```
    gtp_commands : [ "time_settings 600 30 1", "komi 0.5", "fixed_handicap 2" ],
```
and additionally optional for match jobs:
```
    white_gtp_commands : [ "time_settings 0 10 1", "komi 0.5", "fixed_handicap 2" ],
```

No separate hash has been added; I propose that the existing `options_hash` is calculated over `options [+ white_options] [+ gtp_commands [+ white_gtp_commands]]`.

Also add support for white taking the first move in handicapped job games.

Tested by locally modifying the the JSON inputs for orders to produce both self-play and match games. Confirmed they worked ok and saving and restoring still works.
I did pick up on one existing issue/missing feature that leelaz doesn't parse sgf time properties for a `loadsgf` GTP command. `printsgf` also doesn't write time or stones remaining for either side. I'll raise a separate issue for that though.

I think this may be useful for the server performing time controlled test matches (the range of hardware would mean a large variety for the number of playouts achieved but the two sides should have approximately equal strength on any given machine) and for producing training games with komi and/or handicap.

### DIFF
--- a/autogtp/Game.cpp
+++ b/autogtp/Game.cpp
@@ -26,6 +26,7 @@
 Game::Game(const Engine& engine) :
     QProcess(),
     m_engine(engine),
+    m_isHandicap(false),
     m_resignation(false),
     m_blackToMove(true),
     m_blackResigned(false),
@@ -172,36 +173,44 @@ bool Game::gameStart(const VersionTuple &min_version,
     // check any return values.
     checkVersion(min_version);
     QTextStream(stdout) << "Engine has started." << endl;
-    //Send any handicap GTP commands before potentially loading an SGF
-    //because handicap commands will fail if the board is not empty.
+    //If there is an sgf file to start playing from then it will contain
+    //whether there is handicap in use. If there is no sgf file then instead,
+    //check whether there are any handicap commands to send (these fail
+    //if the board is not empty).
     //Then send the rest of the GTP commands after any SGF has been loaded so
     //that they can override any settings loaded from the SGF.
-    auto commands = m_engine.m_commands;
-    const auto handicapCommands = commands.filter("handicap");
-    if (!handicapCommands.isEmpty()) {
-        QStringListIterator i(handicapCommands);
-        while (i.hasNext()) {
-            commands.removeAll(i.next());
-        }
-        for (auto command : handicapCommands) {
-            QTextStream(stdout) << command << endl;
-            if (!sendGtpCommand(command))
-            {
-                QTextStream(stdout) << "GTP failed on: " << command << endl;
-                exit(EXIT_FAILURE);
-            }
-        }
-        m_blackToMove = false;
-    }
     if (!sgf.isEmpty()) {
+        QFile sgfFile(sgf + ".sgf");
+        if (!sgfFile.exists()) {
+            QTextStream(stdout) << "Cannot find sgf file " << sgf << endl;
+            exit(EXIT_FAILURE);
+        }
+        sgfFile.open(QIODevice::Text | QIODevice::ReadOnly);
+        const auto sgfData = QTextStream(&sgfFile).readAll();
+        const auto re = QRegularExpression("HA\\[\\d+\\]");
+        const auto match = re.match(sgfData);
+        m_isHandicap = match.hasMatch();
+        sgfFile.close();
         if (moves == 0) {
             loadSgf(sgf);
         } else {
             loadSgf(sgf, moves);
         }
         setMovesCount(moves);
+    } else {
+        for (auto command : m_engine.m_commands.filter("handicap")) {
+            QTextStream(stdout) << command << endl;
+            if (!sendGtpCommand(command))
+            {
+                QTextStream(stdout) << "GTP failed on: " << command << endl;
+                exit(EXIT_FAILURE);
+            }
+            m_isHandicap = true;
+            m_blackToMove = false;
+        }
     }
-    for (auto command : commands) {
+    const auto re = QRegularExpression("^((?!handicap).)*$");
+    for (auto command : m_engine.m_commands.filter(re)) {
         QTextStream(stdout) << command << endl;
         if (!sendGtpCommand(command))
         {
@@ -227,11 +236,10 @@ void Game::move() {
 
 void Game::setMovesCount(int moves) {
     m_moveNum = moves;
-    //The game always starts at move 0 (GTP state that handicap stones are not part
+    //The game always starts at move 0 (GTP states that handicap stones are not part
     //of the move history), so if there is no handicap then black moves on even
     //numbered turns but if there is handicap then black moves on odd numbered turns.
-    const auto handicap_in_use = !m_engine.m_commands.filter("handicap").isEmpty();
-    m_blackToMove = (moves % 2) == (handicap_in_use ? 1 : 0);
+    m_blackToMove = (moves % 2) == (m_isHandicap ? 1 : 0);
 }
 
 bool Game::waitReady() {
--- a/autogtp/Game.cpp
+++ b/autogtp/Game.cpp
@@ -160,7 +160,9 @@ void Game::checkVersion(const VersionTuple &min_version) {
     }
 }
 
-bool Game::gameStart(const VersionTuple &min_version) {
+bool Game::gameStart(const VersionTuple &min_version,
+                     const QString &sgf,
+                     const int moves) {
     start(m_engine.getCmdLine());
     if (!waitForStarted()) {
         error(Game::NO_LEELAZ);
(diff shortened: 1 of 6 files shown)


autogtp/Game.cpp
```diff
@@ -196,7 +199,8 @@ void Game::move() {
 
 void Game::setMovesCount(int moves) {
     m_moveNum = moves;
-    m_blackToMove = (moves % 2) == 0;
+    const auto handicap_in_use = !m_engine.m_commands.filter("handicap").isEmpty();
+    m_blackToMove = (moves % 2) == (handicap_in_use ? 1 : 0);
```
marcocalignano: I think this should keep count of the handicap stones.
If White has 2 stones handicap the move 3rd is Black so `(moves % 2) = 1`, but if white has three stones handicap move 4th is Black so `(moves % 2) = 0`

autogtp/Job.cpp
```diff
@@ -158,25 +161,27 @@ Result ValidationJob::execute(){
...
+            if (!first.waitForMove()) {
+                return res;
+            }
+            first.readMove();
+            m_boss->incMoves();
+            if (first.checkGameEnd()) {
+                break;
+            }
+            second.setMove(bmove + first.getMove());
+            first.nextMove();
```
marcocalignano: I would put the if outside the do ... while loop. Something like that:

     if (first.getToMove() == Game::White) {
          second.move();
          if (!second.waitForMove()) {
              return res;
          }
          second.readMove();
          m_boss->incMoves();
          first.setMove(wmove + second.getMove());
    }
    do {
    ...
    } while (second.nextMove() && m_state.load() == RUNNING);

autogtp/Job.cpp
```diff
@@ -158,25 +161,27 @@ Result ValidationJob::execute(){
...
         if (!second.waitForMove()) {
             return res;
         }
         second.readMove();
-       m_boss->incMoves();
+        m_boss->incMoves();
         first.setMove(wmove + second.getMove());
-        second.nextMove();
-    } while (first.nextMove() && m_state.load() == RUNNING);
+    } while (second.nextMove() && m_state.load() == RUNNING);
```
marcocalignano: `second.nextMove()` is needed to keep the right Black White order. Or do you see a cool optimization I don't?

autogtp/Game.cpp
Hersmunch: According to GTP the handicap stones don’t count as moves so it is simply if there is handicap the white goes first

autogtp/Job.cpp
Hersmunch: I agree that will be easier to read. Will change.

autogtp/Job.cpp
Hersmunch: It is still there :) I’ve moved the call for first further up, keeping it with the rest of the stuff for black’s move, and put the call for second in the while test

autogtp/Game.cpp
marcocalignano: But then the game starts always at move 0 and black goes first anyway.

Hersmunch: I don't understand what you mean. From the [GTP spec](http://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html#SECTION00073000000000000000):
> The handicap stones are not included in the move history.

So as far as I can tell, white goes first if there is handicap. If I don't do this then I see that black gets the first move after the handicap stones are placed. The behaviour I have coded appears to work as I expect but do let me know if you observe otherwise.

Hersmunch:
I've noticed that the sgf files for match games only contain information about the options and result of the GTP commands used for the black side. I'll look at doing a separate PR for ensuring the `C[]` comment contains the command line options and GTP commands for both sides.

Hmm if #2097 is resolved such that time settings are loaded from an sgf file then the fact that I optimised away the second sgf for match games will mean both sides will use the time settings from the single sgf produced by the black side. Sigh. I guess the second sgf file will have to be reinstated.

autogtp/Game.cpp
marcocalignano: You are right Black has the handicap stones.
But looking at the formula there is something wrong: 
if the game start at move 1 then `moves` is `1` we have `(moves % 2)` equals to `1`  so with `handicap_in_use` true, `m_blackToMove` is also true.
But we just agreed to the opposite.

marcocalignano:
@Hersmunch We need some testing. I feel the change with the swap code is a major one and code review alone is not really enough. I would like to have a test with and without handicap and with and without command. Any idea how we can make that?

marcocalignano:
@Hersmunch Another thing in the swap code. Make a comment on the initialization of the pointers, please. I understood why you did that but in a year we won't remember.

autogtp/Game.cpp
Hersmunch: Ok, I think I understand now. Thanks for explaining.
The game always starts at move 0 (as per the pre-existing code and because the handicap stones are not part of the move history), so if there is no handicap then black moves on even numbered turns but if there is handicap then black moves on odd numbered turns.

marcocalignano:
@Hersmunch We (and I mean you ;) since I am not home no idea when I will be home with time to test something) need to test with a game that  is already started. To be sure on both selfpley and match. Also we need to test the save and restore feature.

Hersmunch:
@marcocalignano Thanks. I did try doing a save and restore for both self-play and match earlier and they seemed fine. Tomorrow I'll look at starting from saved self-play and match games produced from next branch build (but with a small JSON edit to start a match game as my machine is too slow to get them normally).

marcocalignano:
@Hersmunch Can you please start an autoplay with handicap, press `q-Enter` and post the storefile and the sfg file so I can look at them?
Thanks.

autogtp/Management.cpp
```diff
@@ -22,11 +22,13 @@
 #include <QThread>
 #include <QList>
 #include <QCryptographicHash>
+#include <QJsonArray>
 #include <QJsonDocument>
 #include <QJsonObject>
 #include <QLockFile>
 #include <QUuid>
 #include <QRegularExpression>
+#include <QVariant.h>
```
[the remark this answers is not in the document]
Hersmunch: autocomplete error.. will fix

Hersmunch:
Sorry I got distracted this past week. I'll have a look again tomorrow at this and I'll post those files.

Hersmunch:
@marcocalignano hopefully this is what you are after (:
[storefile10457191d6534f71b013f871cf4d2653.bin.txt](https://github.com/gcp/leela-zero/files/2730457/storefile10457191d6534f71b013f871cf4d2653.bin.txt)
[c9c4ab134eb145eda52d962d201dbee6.sgf.txt](https://github.com/gcp/leela-zero/files/2730458/c9c4ab134eb145eda52d962d201dbee6.sgf.txt)
[c9c4ab134eb145eda52d962d201dbee6.train.txt](https://github.com/gcp/leela-zero/files/2730459/c9c4ab134eb145eda52d962d201dbee6.train.txt)

Hersmunch:
OK, so to fix the issue of both sides of a match game loading the black side's komi (and time settings if #2097 is resolved) etc. from the sgf, the GTP commands can be sent after the sgf is loaded. Well, except any handicap GTP commands, which fail if sent whilst the board is not empty. This avoids having to reintroduce saving and restoring an sgf file for each side.
To do this I will attempt to move the loading of an sgf into `Game::gameStart()` and do it after handling any handicap GTP commands all before then sending all the other GTP commands in a loop as per the existing code.

Hersmunch:
OK, so I've done that and it works :)
Here's the storefile and sgf for a match game that has been saved, restored and then saved again showing that black's time settings are still recorded ok (rather than being lost because they were not loaded from the sgf):
[storefileb3cf4d3152ab4a67a4b768bbfdcb92f2.bin.txt](https://github.com/gcp/leela-zero/files/2730640/storefileb3cf4d3152ab4a67a4b768bbfdcb92f2.bin.txt)
[82c95125af4b430ab1c714776c4bcd25.sgf.txt](https://github.com/gcp/leela-zero/files/2730641/82c95125af4b430ab1c714776c4bcd25.sgf.txt)

marcocalignano:
IMHO `loadSgf()` should put the handicap stones reading the SGF file. That would help when someone else wants to analyze an handicap game. So we can suppress the `fixed_handicap ...` command if the game is started or avoid to save it altogether in the store file. All handicap information are in the SGF file.

Hersmunch:
@marcocalignano Done.
I've left the GTP commands in the store file so that they might be more easily copied into the sgf's comment in a later PR. I think this could be done by adding to `SGFTree::state_to_string()` and `Game::fixSgf()` to match (but I'm not sure about how leelaz should/shouldn't go about storing all(?) the GTP commands it has received).

Issue #2001, opened 2018-11-07
lz-analyze does not post output if search is restarted


gcp:
Reported over email, I can confirm the incorrect behavior.

```
./leelaz -w networks/febf1d9c230f9c6f8a7b83056faa87779abb09ae64b538b836f99c91af50b0b8.gz --gtp --playouts 100 --noponder

lz-analyze b 10
=
info move D16 visits 15 winrate 4652 prior 1935 order 0 pv D16 Q4 Q16 D4 R6 R14 info move Q16 visits 15 winrate 4650 prior 1971 order 1 pv Q16 D4 Q4 D16 F3 O3 info move Q4 visits 15 winrate 4649 prior 1912 order 2 pv Q4 D16 Q16 D4 F3 O3 info move D4 visits 11 winrate 4651 prior 1918 order 3 pv D4 Q16 Q4 D16 C14 C6


lz-analyze b 10
=
```

The info move does not post, I presume because it terminates instantly. I think we should always post it at the end of the search.

PR #2114, opened 2019-01-01
Make sure analysis for 'info move' line is printed at least once

dbosst:
For move analysis make sure to output the "info move ..." line at least once. This is useful for client programs expecting/parsing the info move line.

When you do an analysis with a limited number of visits it will finish and put it in the cache.  Re-running the analysis with the same move will finish the analysis before the interval to output the "info move ..."  line so the user/client never gets the easy to parse/read info move line from ::output_analysis.

This fix makes sure to output the analysis if it not outputted after the automatic interval period.


gcp:
This might also fix #2001?

dbosst:
Yeah, that's it.  That's how I discovered the 'bug' as well (turning on/off analysis in Sabaki multiple times for the same move).

PR #2120, opened 2019-01-02
Replace float with double where appropriate.

TFiFiE:
Following up on https://github.com/gcp/leela-zero/pull/2072#discussion_r243037050.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -61,7 +61,7 @@ int cfg_resignpct;
 int cfg_noise;
 int cfg_random_cnt;
 int cfg_random_min_visits;
-float cfg_random_temp;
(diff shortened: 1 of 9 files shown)


Ttl:
I would rather just use float everywhere. There isn't really enough precision in the neural net output to need double. I don't think we are fast enough for it to really matter, but chess people got few percent speedup on fast systems by replacing the float functions with faster approximations.

gcp:
I'd also prefer float, with the exceptions of the node accumulators for score, which need more than 23 bit of mantissa, i.e. this is why https://github.com/gcp/leela-zero/pull/2120/files#diff-28216771f77a09f4e0782ad5404fdd28L210 looks like it does.

PR #2121, opened 2019-01-03
Cuda

cjohnchen:


Ttl:
Why are you making a pull request of my code? There is a reason I haven't made it myself. 

This PR replaces OpenCL with CUDA, but we don't want to remove the OpenCL. Both should be supported. It's also slower than the current code so there isn't really any reason to use it.

PR #2071, opened 2018-12-07
Gather qt5 in main page for a full build compile all in once (leelaz+/-autogtp) + clearer main README + more detailed AUTOGTP readme

wonderingabout:
My next branch is now in sync with gcp/next

this branch is a copy of gcp/next with a few new commits

- fix qt5 missing dependency for contributing
- added cmake and curl in the compile dependencies too
- clearer readme again
- cleanup of the autogtp readme
- other clearer changes

general outlook 

- here : https://github.com/wonderingabout/leela-zero/tree/minor-clearer-next
- and here : https://github.com/wonderingabout/leela-zero/blob/minor-clearer-next/autogtp/README.md

you can compare it to : 

- old version : https://github.com/gcp/leela-zero/tree/next
- old version : https://github.com/gcp/leela-zero/blob/next/autogtp/README.md

PR #2122, opened 2019-01-03
Building for Mac OS X updated

toomasr:
I went through the Mac part today and had to brew install zlib and qt5 also to make cmake happy. I've added these two to the list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ by adding -DUSE_CPU_ONLY=1 to the cmake command line.
     git submodule update --init --recursive
 
     # Install build depedencies
-    brew install boost cmake
+    brew install boost cmake zlib qt5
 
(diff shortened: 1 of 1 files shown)


wonderingabout:
my pull request is related to yours : https://github.com/gcp/leela-zero/pull/2071

see mac building instructions

wonderingabout:
however as you can see in @gcp 's review, qt5 should be installed only if one person wants to contribute,

therefore it should be an optionnal step

wonderingabout:
is zlib needed for autogtp or for cmake ?

gcp:
>is zlib needed for autogtp or for cmake ?

zlib is needed for Leela itself, to decompress weights files on the fly.

I pulled the zlib part only, qt5 is not required as you explained.

Issue #2107, opened 2018-12-30
idea of slightly reduce weight size


l1t1:
if the difference of( round to 2 digit - round to 3 digit) is small enough, we can use round to 2 digit result.
I use the the ratio 0.01 and the result to many different size weight is following
```
D:\>c:\python37\python quantize_weights23.py 57.txt
Weight file difference L2-norm: 0.3201855506392345

D:\>c:\python37\python quantize_weights23.py 91.txt
Weight file difference L2-norm: 0.1640525086752053

D:\>c:\python37\python quantize_weights23.py 116.txt
Weight file difference L2-norm: 4.988491974049602

D:\>c:\python37\python quantize_weights23.py elf_converted_weights.txt
Weight file difference L2-norm: 0.00405287248899887

the orginal code result

D:\>c:\python37\python quantize_weights.py 57.txt
Weight file difference L2-norm: 0.31965124587016847

D:\>c:\python37\python quantize_weights.py 91.txt
Weight file difference L2-norm: 0.040919122536652267

D:\>c:\python37\python quantize_weights.py 116.txt
Weight file difference L2-norm: 0.04882264516397222

D:\>c:\python37\python quantize_weights.py elf_converted_weights.txt
Weight file difference L2-norm: 0.0035875174236782307 

```
the modifed code
```
#!/usr/bin/env python3
import sys, os, argparse

def format_n(x):
    x = float(x)
    x2 =float('{:.2g}'.format(x))
    if x!=0 and abs(x-x2)/abs(x) <0.01:
     x = '{:.2g}'.format(x)
    else: 
     x = '{:.3g}'.format(x)
    x = x.replace('e-0', 'e-')
    if x.startswith('0.'):
        x = x[1:]
    if x.startswith('-0.'):
        x = '-' + x[2:]
    return x

if __name__ == "__main__":

    parser = argparse.ArgumentParser(
            description='Quantize network file to decrease the file size.')
    parser.add_argument("input",
        help='Input file', type=str)

    parser.add_argument("-o", "--output",
        help='Output file. Defaults to input + "_quantized"',
        required=False, type=str, default=None)

    args = parser.parse_args()

    if args.output == None:
        output_name = os.path.splitext(sys.argv[1])
        output_name = output_name[0] + '_quantized23' + output_name[1]
    else:
        output_name = args.output
    output = open(output_name, 'w')

    calculate_error = True
    error = 0

    with open(args.input, 'r') as f:
        for line in f:
[...]

PR #2126, opened 2019-01-04
update quantize_weights.py with round  some numbers to 2 digits  reduce weight file size

l1t1:
see https://github.com/gcp/leela-zero/issues/2107#issuecomment-451390081

### DIFF
--- a/training/tf/quantize_weights.py
+++ b/training/tf/quantize_weights.py
@@ -3,7 +3,11 @@ import sys, os, argparse
 
 def format_n(x):
     x = float(x)
-    x = '{:.3g}'.format(x)
+    x2 =float('{:.2g}'.format(x))
+    if x!=0 and abs(x)<1 and abs(x-x2)/abs(x) <0.01:
+        x = '{:.2g}'.format(x)
+    else: 
+        x = '{:.3g}'.format(x)
     x = x.replace('e-0', 'e-')
     if x.startswith('0.'):
         x = x[1:]


gcp:
@Ttl any thoughts? Looks pretty safe to me?

Ttl:
Yes, this looks safe.

gjm11:
Why is this worth doing? I mean, it makes the weight files a little smaller, and it presumably may make the networks a little weaker. On what grounds is this thought to be a good tradeoff? In the discussion of #2107 I see some evaluations of e.g. the L2 norm of the difference in the weight vectors (which I guess is useful for finding big errors but tells us nothing about the impact on how well the network chooses or evaluates moves) and nothing that, e.g., plays more-quantized against less-quantized networks, or compares their move choices or win probability estimates.

What's the need for this? Why has it been committed without any evaluation of anything relevant to actual go-playing performance?

wonderingabout:
@gjm11 

i agree with you, i see little benefit to compress further

this is likely to do more harm than good

gcp:
Mostly because it makes sense and it's similar to trellis quantization. I didn't pay enough attention to the error bound in here though: our current quantization to 3 digits should suggest a bound like <0.005, not 0.01 which is 2 digits!

You are right this shouldn't be have been enabled without testing, and I'm going to back it out.

l1t1:
362,149,606 199.txt
346,808,778 199_quantized(err<0.005).txt
331,955,103 199_quantized(err<0.01).txt

l1t1:
i saw words such as likely and should. i didnt see an  evidence of it weaken the orignal weight.
recent ten matchs winrate are not bigger than  0.5, only one is quantized by the pr. 
pls give it a chance to have more tests.

l1t1:
some candidates maybe the quantized version of
1. current best lz201
2. the one has biggest winrate to lz201 ,ie. c3ed2789
3. each long steps new weights,ie +128 +192 +256

PR #2134, opened 2019-01-08
Fix GCC8 'catching polymorphic type by value' warning

akdtg:
Fixes the initial warning:

warning: catching polymorphic type ‘struct Management::NetworkException’ by value [-Wcatch-value=]
         } catch (NetworkException ex) {
                                   ^~


marcocalignano:
Then if you are at it catch a `const` reference!

PR #2138, opened 2019-01-10
Update README.md to add help for installing on debian with nvidia GPU.

leeschumacher:
Add directions for installing/troubleshooting on debian with nvidia GPU.


roy7:
Can you PR this against /next instead of /master please?

gcp:
The OpenCL drivers would be expected to pull those in, and putting specific instructions for every distro and card is out of scope for that readme. Maybe a generic remark in ## Requirements to make sure to install any OpenCL drivers if they're separate?

PR #2158, opened 2019-01-18
Change search stop condition

thynson:
This pull requests implements the behaviour required to compromise both visit limit and playout limit for self play.
See https://github.com/gcp/leela-zero-server/pull/216#issuecomment-455441274


thynson:
Note that if this change is applied, it won't change the behaviour of commanly used `-p 800 -v 0` or `-v 1600 -p 0`. 
And GTP command `lz-setoption name [visits|playouts] value ...` is roughly tested.

gcp:
It seems counter-intuitive to me that if you specify -v 800 and -p 1000, then it will stop when it reaches *both*, rather then when it reaches *any*. I'm not convinced about combining the limits for self-play either. Not going to pull.

PR #2160, opened 2019-01-19
AutoGTP: Added full engine options and starting GTP commands to sgf comments

Hersmunch:
Follow up to #2078 and #2096 to include full engine options and starting GTP commands are in sgf comments.

### DIFF
--- a/autogtp/Game.cpp
+++ b/autogtp/Game.cpp
@@ -392,13 +392,14 @@ bool Game::loadSgf(const QString &fileName, const int moves) {
     return sendGtpCommand(qPrintable("loadsgf " + fileName + ".sgf " + QString::number(moves + 1)));
 }
 
-bool Game::fixSgf(const QString& weightFile, const bool resignation) {
+bool Game::fixSgf(const Engine& whiteEngine, const bool resignation,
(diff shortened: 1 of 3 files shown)


autogtp/Game.cpp
```diff
@@ -392,23 +392,45 @@ bool Game::loadSgf(const QString &fileName, const int moves) {
...
+        comment += " White options:";
+        comment += whiteEngine.m_options + " " + whiteEngine.m_network;
+        comment += " Starting GTP commands:";
+        for (const auto command : whiteEngine.m_commands) {
+            comment += " " + command;
+        }
+    }
+    comment += "]";
+    comment.replace(QRegularExpression("\\s\\s+"), " ");
+    sgfData.replace(oldComment, comment);
```
marcocalignano: I would like to have a different member to change the comments: like 
```
 void fixSgfComment(const Engine& whiteEngine, const bool isSelfPlay)
```

Hersmunch: Are my latest changes what you meant?

marcocalignano:
@Hersmunch Yep!

PR #2170, opened 2019-01-23
Added several simple GTP commands useful for building interfaces to LZ

anoek:
Added the following GTP commands.

    last_move
    move_history

The output of these commands match the corresponding commands in GnuGo.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -381,6 +381,11 @@ const std::string GTP::s_commands[] = {
     "genmove",
     "showboard",
     "undo",
+    "redo",
+    "rewind",
+    "last_move",
+    "move_history",
+    "move_number",
     "final_score",
     "final_status_list",
     "time_settings",
@@ -734,6 +739,44 @@ void GTP::execute(GameState & game, const std::string& xinput) {
         } else {
             gtp_fail_printf(id, "cannot undo");
         }
+        return;
+    } else if (command.find("redo") == 0) {
+        if (game.forward_move()) {
+            gtp_printf(id, "");
+        } else {
+            gtp_fail_printf(id, "cannot redo");
+        }
+        return;
+    } else if (command.find("rewind") == 0) {
+        game.rewind();
+        gtp_printf(id, "");
+        return;
+    } else if (command.find("last_move") == 0) {
+        auto last_move = game.get_last_move();
+        if (last_move == FastBoard::NO_VERTEX) {
+            gtp_fail_printf(id, "no previous move known");
+            return;
+        }
+        std::string coordinate = game.move_to_text(last_move);
+        std::string color = game.get_to_move() == FastBoard::WHITE ? "black": "white";
+        gtp_printf(id, "%s %s", color.c_str(), coordinate.c_str());
+        return;
+    } else if (command.find("move_history") == 0) {
+        if (game.get_movenum() == 0) {
+            gtp_printf_raw("= \n");
+        } else {
+            gtp_printf_raw("= ");
+        }
+        for (size_t i=0; i < game.get_movenum(); ++i) {
+            auto state = game.get_past_state(i);
+            std::string coordinate = game.move_to_text(state->get_last_move());
+            std::string color = state->get_to_move() == FastBoard::WHITE ? "black": "white";
+            gtp_printf_raw("%s %s\n", color.c_str(), coordinate.c_str());
+        }
+        gtp_printf_raw("\n");
+        return;
+    } else if (command.find("move_number") == 0) {
+        gtp_printf(id, "%u", game.get_movenum());
         return;
     } else if (command.find("showboard") == 0) {
         gtp_printf(id, "");
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -42,7 +42,6 @@
 #include <string>
 #include <vector>
 #include <boost/algorithm/string.hpp>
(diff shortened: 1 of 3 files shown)


gcp:
In terms of specs, GTP2 with 2.13 Private Extensions dis-recommends the use of these non-standard commands without a ```gg-``` prefix.

But I think the fact that they're supported by GNU Go is a very good reason why the risk would be acceptable here. Anyone else have any input?

anoek:
I ended up removing `move_history` and the supporting modifications I made to `GameState`, I thought it would be useful for my case but it was not, so might as well not introduce the excess code.

As GnuGo only supported `last_move` and `move_history`, I went ahead and prefixed all the commands I added (including `last_move` since it's not technically part of the spec as you point out).

gcp:
I guess this is meant for interfacing when your interface isn't able to parse SGF itself? Then you can use loadsgf and go back and forth and still get meaningful analysis. 
 
I'm generally hesitant to extend GTP in a way that transfers too much responsibility to the engine, i.e. this functionality is redundant with loading the SGF in the interface and just sending sequences of clear_board + play commands (though it'd be nice if play could took a sequence of moves to avoid a lot of context switching), which is how UCI works.

To understand the problem with an extension like this (or general shift of responsibility to the engine), consider what happens when your SGF has variations. These commands immediately become useless. If you designed your interaction with the engine around them, you now have to redo everything.

That being said, these seem generally useful for a lot of simple analysis tools. So I'm quite willing to not die on this hill and just merge it.

gcp:
I think lz-move_number is quite useful too. At first it seemed like it was a sign that interface and engine had gotten de-synchronized (which is a bug to fix, not protocol to add!), but then I realized that if you do a loadsgf you have no way of knowing how many moves are in there, and consequently how many undos you can do. You can keep undoing until you get an error, but then you have to loadsgf again unless lz-redo is added. Pretty nasty. So if loadsgf is there, knowing the amount of moves is sensible.

I'd like to ponder a bit about how to best handle this in the protocol.

For example, would something like ```lz-gamehistory``` be more generic? That could then give you the game moves from the main branch, i.e. W C4 B Q16 etc up to the final position that loadsgf will get.

But then you'd still have to clear_board + play all moves. In the absence of a ```play``` that could take move sequences, ```lz-redo``` also makes a lot of sense.

What about ```lz-list_moves``` instead of ```lz-last_move``` ? That would then return a list of moves for this position. That would allow you to handle variations as well. ```lz-last_move``` would then be ```undo``` + ```lz-list_moves``` and taking the first one.

So my current thoughts are:
* ~~~lz-redo~~~
* lz-list_moves

Which gives:

```lz-move_number``` is then ```undo``` till fail + ```lz-redo```
```lz-rewind``` is ```undo``` till fail
```lz-last-move``` is ```undo``` + ```lz-list_moves``` + play first result.

I just realized that ```lz-redo``` is redundant because we can do ```lz-list_moves``` and ```play``` the first result.

Am I making sense here?

gcp:
I prefer not to introduce new commands if their effect can (reasonably) be achieved with existing ones. I know it's trivial in LZ to add these, but the idea is that if someone ever makes a new/better/different engine, and wants it to work with GUIs which are oriented towards LZ, then they have to replicate all those commands. I also think it's not really nice from a protocol design POV.

>about lz-redo equaling clear_board + replay

No, the idea was that if ```lz-list_moves``` exists, you can just ask what move was played in the current position and ```play``` it, so you don't have to go to the starting state.

>lz-list_moves seems like an over complication to me as it entails (I think?) adding code to the engine to track the various SGF branches so you can provide that data

Yeah, unfortunately that's true. SGFTree itself tracks them, but GTP.cpp does not keep the SGFTree and just copies the mainline into GameState. I think we can have GTP keep the SGFTree around, but it would have to track the current node in that tree while we play/undo around. Doable but not trivial.

>Intuitively it seems like caring about branches is a bit beyond the scope of what LZ would want to do, but you tell me.

If it were up to me then protocol wise even ```loadsgf``` would be out of scope and it'd be up to the client to do that parsing. But if we're allowing ```loadsgf```, and then pretend variations don't exist or give no way to the client to traverse them, that also seems a bit suboptimal.

It's possible I'm totally overthinking this and the reality is that a) GUI which want to deal with variations can load SGF themselves b) basic tools that can't read SGF would be happy with the very basic commands.

Let's do last_move and move_history? They don't even necessarily need a prefix, I think, as long as output is exactly as GNU Go would do. Defining a "standardized" version of those that is different from what GNU Go and Leela Zero do would be insane...

gcp:
Hmm, if ```lz-list_moves``` reported variations, you'd have to spec out what happens for example if you ```play``` a different move or ```genmove``` wants something different. Would these then be added or not. Seems like quite the can of worms. Let's forget about that idea for now.

Hersmunch:
gcp point out that:
> lz-redo is redundant because we can do lz-list_moves and play the first result.

This sounds preferable to me but do you foresee a problem with that approach?

Code looks good to me.

anoek:
@Hersmunch There would be nothing wrong with that per se, but as discussed, implementing something like `lz-list_moves` opens a can of worms as it would imply adding code to track (and probably manipulate) branches.

Hersmunch:
Sorry, I hadn't woken up yet and got confused with `move_history`. I suppose that could be used first and remembered before doing `undo` though.

anoek:
Yep that's true :)

PR #2172, opened 2019-01-26
Added extra support for "TM" and "OT" and other sgf time control properties on printsgf and loadsgf GTP commands

### LINKED ISSUE #2097: `printsgf' and 'loadsgf' have incomplete time property support
Noticed whilst testing #2096.
leelaz doesn't parse sgf time properties for a `loadsgf` GTP command. `printsgf` also doesn't write time or stones remaining for either side. See e.g. https://www.red-bean.com/sgf/properties.html#BL

I realise that [the GTP spec](http://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html) only mentions `loadsgf` and the effect given is met:
> Board size and komi are set to the values given in the sgf file. Board configuration, number of captured stones, and move history are found by replaying the game record up to the position before move_number or until the end if omitted.

However, I think that adding support for saving and parsing the other time properties would be desirable for both users playing against leelaz and for pausing and resuming any games with time controls set up as per #2096.

Hersmunch:
-Part of what is described in- Resolves #2097; allowing saving and loading of time control settings to be lossless within LeelaZ.
Note that this only supports "OT" syntax matching output from a printsgf GTP command and currently does not try to save or load remaining time and stones for either side (see https://www.red-bean.com/sgf/properties.html#BL). I'll look at adding that in another commit.

My changes do seem a bit awkward so please do let me know if it can be simplified. Also, if the overtime/byo-yomi syntax is unrecognised, is it reasonable to set up the time control with none? Should a message be printed in this case, and if so, how?

Edit: Added sgf saving and loading of remaining time and stones for both sides.

### DIFF
--- a/src/GameState.h
+++ b/src/GameState.h
@@ -71,7 +71,7 @@ public:
     const TimeControl& get_timecontrol() const;
     void set_timecontrol(const TimeControl& timecontrol);
     void set_timecontrol(int maintime, int byotime, int byostones,
-        int byoperiods);
+                         int byoperiods);
     void adjust_time(int color, int time, int stones);
 
     void display_state();
--- a/src/GameState.h
+++ b/src/GameState.h
@@ -69,8 +69,9 @@ public:
     void start_clock(int color);
     void stop_clock(int color);
     const TimeControl& get_timecontrol() const;
+    void set_timecontrol(const TimeControl& timecontrol);
     void set_timecontrol(int maintime, int byotime, int byostones,
-                         int byoperiods);
+        int byoperiods);
     void adjust_time(int color, int time, int stones);
 
     void display_state();
--- a/src/SGFTree.cpp
+++ b/src/SGFTree.cpp
@@ -191,8 +191,18 @@ void SGFTree::populate_states() {
         const auto maintime = it->second;
         it = m_properties.find("OT");
         const auto byoyomi = (it != end(m_properties)) ? it->second : "";
+        it = m_properties.find("BL");
+        const auto black_time_left = (it != end(m_properties)) ? it->second : "";
+        it = m_properties.find("WL");
+        const auto white_time_left = (it != end(m_properties)) ? it->second : "";
+        it = m_properties.find("OB");
+        const auto black_moves_left = (it != end(m_properties)) ? it->second : "";
+        it = m_properties.find("OW");
+        const auto white_moves_left = (it != end(m_properties)) ? it->second : "";
         m_timecontrol_ptr = std::make_shared<TimeControl>();
-        m_timecontrol_ptr->set_from_text_sgf(maintime, byoyomi);
+        m_timecontrol_ptr->set_from_text_sgf(maintime, byoyomi,
+                                             black_time_left, white_time_left,
+                                             black_moves_left, white_moves_left);
     }
 
     // handicap
--- a/src/SGFTree.cpp
+++ b/src/SGFTree.cpp
@@ -81,6 +81,10 @@ GameState SGFTree::follow_mainline_state(unsigned int movenum) const {
     // sets up the game history.
     GameState result(get_state());
 
+    if (m_loaded_timecontrol) {
+        result.set_timecontrol(m_timecontrol);
+    }
+
     for (unsigned int i = 0; i <= movenum && link != nullptr; i++) {
         // root position has no associated move
         if (i != 0) {
@@ -181,6 +185,16 @@ void SGFTree::populate_states() {
         }
     }
 
+    // time
+    it = m_properties.find("TM");
+    if (it != end(m_properties)) {
+        std::string maintime = it->second;
+        it = m_properties.find("OT");
+        std::string byoyomi = (it != end(m_properties)) ? it->second : "";
+        m_timecontrol.set_from_text_sgf(maintime, byoyomi);
+        m_loaded_timecontrol = true;
+    }
+
     // handicap
     it = m_properties.find("HA");
     if (it != end(m_properties)) {
@@ -268,6 +282,8 @@ void SGFTree::populate_states() {
 void SGFTree::copy_state(const SGFTree& tree) {
     m_initialized = tree.m_initialized;
     m_state = tree.m_state;
+    m_loaded_timecontrol = tree.m_loaded_timecontrol;
+    m_timecontrol = tree.m_timecontrol;
 }
 
 void SGFTree::apply_move(int color, int move) {
--- a/src/SGFTree.cpp
+++ b/src/SGFTree.cpp
@@ -151,7 +151,7 @@ void SGFTree::populate_states() {
     // board size
     it = m_properties.find("SZ");
     if (it != end(m_properties)) {
-        std::string size = it->second;
+        const auto size = it->second;
         std::istringstream strm(size);
         int bsize;
         strm >> bsize;
@@ -167,13 +167,13 @@ void SGFTree::populate_states() {
     // komi
     it = m_properties.find("KM");
     if (it != end(m_properties)) {
-        std::string foo = it->second;
+        const auto foo = it->second;
         std::istringstream strm(foo);
         float komi;
         strm >> komi;
-        int handicap = m_state.get_handicap();
+        const auto handicap = m_state.get_handicap();
         // last ditch effort: if no GM or SZ, assume 19x19 Go here
-        int bsize = 19;
+        auto bsize = 19;
         if (valid_size) {
             bsize = m_state.board.get_boardsize();
         }
@@ -188,9 +188,9 @@ void SGFTree::populate_states() {
     // time
     it = m_properties.find("TM");
     if (it != end(m_properties)) {
-        std::string maintime = it->second;
+        const auto maintime = it->second;
         it = m_properties.find("OT");
-        std::string byoyomi = (it != end(m_properties)) ? it->second : "";
+        const auto byoyomi = (it != end(m_properties)) ? it->second : "";
         m_timecontrol.set_from_text_sgf(maintime, byoyomi);
         m_loaded_timecontrol = true;
     }
@@ -198,7 +198,7 @@ void SGFTree::populate_states() {
     // handicap
     it = m_properties.find("HA");
     if (it != end(m_properties)) {
-        std::string size = it->second;
+        const auto size = it->second;
         std::istringstream strm(size);
         float handicap;
         strm >> handicap;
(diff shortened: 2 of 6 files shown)


src/SGFTree.cpp
```diff
@@ -181,6 +185,16 @@ void SGFTree::populate_states() {
         }
     }
 
+    // time
+    it = m_properties.find("TM");
+    if (it != end(m_properties)) {
+        std::string maintime = it->second;
```
gcp: Can be auto.

src/SGFTree.cpp
```diff
@@ -81,6 +81,10 @@ GameState SGFTree::follow_mainline_state(unsigned int movenum) const {
     // sets up the game history.
     GameState result(get_state());
 
+    if (m_loaded_timecontrol) {
```
gcp: Someday, when we mandate C++17, this can be a std::optional :-)

src/SGFTree.cpp
```diff
@@ -268,6 +282,8 @@ void SGFTree::populate_states() {
 void SGFTree::copy_state(const SGFTree& tree) {
     m_initialized = tree.m_initialized;
     m_state = tree.m_state;
+    m_loaded_timecontrol = tree.m_loaded_timecontrol;
```
gcp: Are there cases in which it's necessary to copy this?

src/SGFTree.cpp
```diff
@@ -268,6 +282,8 @@ void SGFTree::populate_states() {
...
+    m_timecontrol = tree.m_timecontrol;
```
gcp: Same question.

src/GameState.h
```diff
@@ -68,8 +68,9 @@ class GameState : public KoState {
     void start_clock(int color);
     void stop_clock(int color);
     const TimeControl& get_timecontrol() const;
+    void set_timecontrol(const TimeControl& timecontrol);
     void set_timecontrol(int maintime, int byotime, int byostones,
-                         int byoperiods);
+        int byoperiods);
```
gcp: Whitespace change. Also I think most of Leela's code aligns arguments whenever possible.

src/SGFTree.h
```diff
@@ -77,6 +78,8 @@ class SGFTree {
 
     bool m_initialized{false};
     KoState m_state;
+    bool m_loaded_timecontrol{false};
+    TimeControl m_timecontrol;
```
gcp: Hmm, so this is a bit annoying. Most of the SGFTree objects only have a KoState m_state, which has no associated timecontrol.

The GameStates are all constructed in follow_mainline_state. So, we don't really need every SGFTree node to have a copy of this.

The SGF spec says:
```
        A node containing game-info properties is called a game-info node.
        There may be only one game-info node on any path within the tree,
        i.e. if some game-info properties occur in one node there may not be
        any further game-info properties in following nodes:
        a) on the path from the root node to this node.
        b) in the subtree below this node.
```

And time control things are game-info properties.

Maybe SGFTree could be modified to hold shared pointers to a SGFGameInfo object or something similar instead of copying them everywhere. But there's also plenty of game-info properties which are encoded into KoState anyway, so it wouldn't gain us much.

So maybe, for now, make SGFTree take a shared_ptr to a TimeControl object? That will eliminate m_loaded_timecontrol too.

src/SGFTree.cpp
Hersmunch: Ah I had just copied the style of the surrounding code. Will give it all a once over with auto.

src/SGFTree.cpp
Hersmunch: I wasn’t sure so I put it in..

Hersmunch: I wasn’t sure so I put it in..

src/SGFTree.h
Hersmunch: @gcp I think I've done what you proposed. Let me know if I misunderstood or got it wrong somehow. sgf loading still works :)

Hersmunch:
So I added support for the rest of the sgf time properties which are for time and stones remaining for each side. You might want to review the last commit separately.

src/SGFTree.cpp
```diff
@@ -181,10 +185,30 @@ void SGFTree::populate_states() {
...
+        it = m_properties.find("BL");
+        const auto black_time_left = (it != end(m_properties)) ? it->second : "";
+        it = m_properties.find("WL");
+        const auto white_time_left = (it != end(m_properties)) ? it->second : "";
+        it = m_properties.find("OB");
+        const auto black_moves_left = (it != end(m_properties)) ? it->second : "";
+        it = m_properties.find("OW");
+        const auto white_moves_left = (it != end(m_properties)) ? it->second : "";
+        m_timecontrol_ptr = std::make_shared<TimeControl>();
+        m_timecontrol_ptr->set_from_text_sgf(maintime, byoyomi,
```
gcp: Does this need some sanity checks in case parts are missing or inconsistent?

gcp: Guess this could be a shared_ptr<TimeControl> factory as it's the only caller of set_from_text_sgf, i.e. make_from_text_sgf.

src/TimeControl.cpp
```diff
@@ -49,24 +51,89 @@ TimeControl::TimeControl(int maintime, int byotime,
...
-            s += std::to_string(m_byotime/100) + " byo-yomi]";
+            s += std::to_string(m_byotime / 100) + " byo-yomi]";
         }
+        s += stones_left_to_text_sgf(FastBoard::BLACK);
+        s += stones_left_to_text_sgf(FastBoard::WHITE);
     }
+    // Generously round up to avoid a remaining time of 0 triggering byo-yomi
+    // to be started when the sgf is loaded. This happens because byo-yomi
+    // stones have to be only written to the sgf when actually in byo-yomi
+    // and this is interpretted in adjust_time() as a special case
```
gcp: typo: interpreted

gcp:
Seems good to me. I think you want to add a unit test if saving and loading works and is lossless for various cases, as well as adding a few SGF with timecontrol info from other places / that is broken and seeing if the engine doesn't do anything too amusing.

src/SGFTree.cpp
Hersmunch: > Does this need some sanity checks in case parts are missing or inconsistent?

If we're only trying to be self-consistent and lossless then no. Note that there is no mention of leading time settings from an sgf in the GTP spec so the controller should be sending time settings separately... which sounds like a problem if we override previously sent time settings :(

I think the only case that could be inconsistent is if there is no byo-yomi but the sgf has remaining byo-yomi stones recorded. Other than there there is the case in which the byo-yomi syntax is unrecognised (which could lead to the preceding case). If the controller sends the time controls after then it would be bad to not load the sgf.

Hmm. Thoughts?

src/TimeControl.cpp
```diff
@@ -49,24 +52,91 @@ TimeControl::TimeControl(int maintime, int byotime,
...
     return s;
 }
 
+std::shared_ptr<TimeControl> TimeControl::make_from_text_sgf(
+    const std::string& maintime, const std::string& byoyomi,
+    const std::string& black_time_left, const std::string& white_time_left,
+    const std::string& black_moves_left, const std::string& white_moves_left) {
+    const auto maintime_centis = std::stoi(maintime) * 100;
+    auto byotime = 0;
+    auto byostones = 0;
```
Hersmunch: Note that defaulting `byostones` to 0 here matches the behaviour of the GTP time settings handling but is inconsistent with the `TimeControl` constructor's default argument of 25. Why is it 25?

Hersmunch:
Any tips for building and running the unit tests on Windows?

src/TimeControl.cpp
gcp: It's what Leela assumes the timecontrol is if you don't configure anything.

Hersmunch: Sure but is there a reason why 25? I'm trying to understand if it matters whether I use 0 here or if it should also be 25. As far as I can tell, it doesn't matter what the value is because there are no byo-yomi periods if nothing is configured or e.g. the byo-yomi syntax isn't recognised.
So I went for 0 to match the behaviour when time settings are given by a GTP command.

gcp: >Sure but is there a reason why 25?

That's a convenient thinking time for playing games against the computer if you have a primitive interface that can't even do a time_settings.

gcp: @Hersmunch I guess we should also verify that if you load an SGF without time_settings (because the saver didn't know about them), the engine has a "reasonable" default.

gcp:
@Hersmunch I'm willing to pull this without tests but see the remarks about defaults.

src/TimeControl.cpp
Hersmunch: > > Sure but is there a reason why 25
>
> That's a convenient thinking time for playing games against the computer if you have a primitive interface that can't and won't even do a time_settings.

Time? Here's the time control constructor default arguments:
https://github.com/leela-zero/leela-zero/blob/63af3687462ba27daeabc526da6bbfa3d6149389/src/TimeControl.h#L29-L34
With no time or periods does it matter how many stones there are in byo-yomi?
After launching leelaz, if a `time_left b 1 0` is sent (1 sec, 0 stones == not in byo-yomi) and then the time is run down (e.g. via `auto`) the result is that byo-yomi is entered with 25 stones but then the clock continues to count down, going negative. This resets to 0 when the stones hit 0 and then goes negative again.

> So basically test that if you just launch the program or reset_board or change board size etc and don't do any time_settings, you still get something sane.

These all result in the sgf containing `TM[3600]BL[3600]WL[3600]`. Notice that this doesn't include any byo-yomi. This means that if the sgf is loaded again that it doesn't have the behaviour described above and instead doesn't go into byo-yomi (as per `byostones = 0` in my code) but the clock goes negative from then on.
I think it would be good to make this behaviour consistent with the time settings used just after launching leelaz but perhaps that should be changed to have some time?

> @Hersmunch I guess we should also verify that if you load an SGF without time_settings (because the saver didn't know about them), the engine has a "reasonable" default.

This didn't behave as I expected, which was that no time settings would be loaded and the time settings would be whatever leelaz was using before loading the sgf. Instead, the time settings are reset to the default behaviour after launching leelaz. Am I right in thinking that this is because the shared_ptr will point to a time control based upon the default constructor?

gcp: Okay, probably byostones should just be 0 by default given that there was no time set.

gcp: >Am I right in thinking that this is because the shared_ptr will point to a time control based upon the default constructor?

No, it will just be null.

What happens is that this case is detected and the set_timecontrol isn't done here: https://github.com/leela-zero/leela-zero/pull/2172/files#diff-44ba5da097788068a1c8def978c3293cR84

And you indeed get the default constructor.

PR #2173, opened 2019-01-27
truncate to 4 precision and compress converted minigo weights.

sethtroisi:


l1t1:
7-zip's gz format is about 90% size of gzip, and @gcp uses it. pls use it so

sethtroisi:
I would be very supervised by that result.

I spot checked a single file

|tool|final_size|compression_ratio|
|----|----|-----|
|raw|229mb|1.0|
|gzip (this code) l=9|64mb|28%|
|gzip l=7|72mb|31% (but much faster)|
|7z a test.txt.gz|61mb|27%|

l1t1:
compress ratio is related with  precision, 3 is higher, 2 even higer, iuse mx5 in7z

gcp:
>iuse mx5

This was worse for me than mx9, so the official networks still use that.

PR #2176, opened 2019-01-27
Update thompson_sampling to /next

roy7:
I fixed the conflicts and compile time errors, and it will run and generate some moves. It's possible the ts-specific code might still do something invalid under modern /next code though, but nothing I easily spotted. But this should give you a head start I hope on updating the branch to modern code.


roy7:
I think the initial EXPANDING code I ported in probably needs to wrap the whole section to avoid all of that calculation.

gcp:
I suspect this would look a lot simpler if the thompson_sampling branch was rebased to next (instead of merged)? Then the only diff in here would be the changes from UCT to Thompson sampling.

As it is, it's essentially unreviewable.

roy7:
Ok I will try to give that a shot sometime and figure out how to rebase things. :)

PR #2184, opened 2019-02-01
Increase default thread count.

gcp:
Benchmarking on modern hardware (aka RTX 2080 card) indicates that we're
no longer saturing the GPU with 2 threads, far from it in fact. I see
substantial scaling past 2 threads, with even a small increase going
from 8 to 16, showing that our scaling improvements have been
successful.

On the other hand, older hardware and drivers might not scale as well,
and the search does get less efficient because MCTS scaling is not
perfect.

So change the default from 2 to 8 threads, capped by the number of
hardware threads.


gcp:
Curious about cases where this *decreases* performance!

roy7:
Should default thread count be conditional on RTX detection or is that too messy?

roy7:
#2188 might supersede this.

PR #2187, opened 2019-02-01
ts_update rebase attempt #1

roy7:
I assume I did this totally wrong? I rebased ts_update off of /next and it would replace the existing thompson_sampling. I don't want it merged into /next. But trying to merge it into /thompson_sampling seems to have 127 files that are changed, etc etc.

Can you advise on how to do this more correctly?


gcp:
>But trying to merge it into /thompson_sampling seems to have 127 files that are changed, etc etc.

The thompson_sampling branch here is out of date, so that won't work. Let's see if I can pull your branch onto mine, then update here.

roy7:
Oh right, because my rebase means I'm totally unrelated to your old Thompson sampling branch now.

Maybe if you reset your branch to match mine? The intent here, I think, is that my branch is now up to date with next so the only differences should be Thompson related.

gcp:
Yes, that should work and it's what I'm doing.

gcp:
You should also run git submodule update BTW.

roy7:
Thanks for the submodule tip, done.

PR #2188, opened 2019-02-02
Batched neural net evaluations

ihavnoid:
Group evaluations and run them in parallel.  Roughly 50% speedup on my setup, but there are a couple of points that is debatable.

- Thread / batch sizing heuristics : This PR changes how the default threads / default batch sizes are picked.  See Leela.cpp
- Batch-forming heuristic : See OpenCLScheduler.cpp for the batch forming heuristic : the heuristic exists so that we can wait for the rest of the engine to create more NN evaluations so that we can run larger batches.  We can't wait indefinitely since there are cases we enter 'serial' paths.  Since heuristics are heuristics, these might need some tests on a larger variety of types of systems.

Did make sure that winrate improves when running default vs. default command line `./leelaz -w (weight file)` on time parity.


roy7:
Not to distract from the awesome real work being done, but if someone might put in the back of their mind to eventually add support within get_output() to use batching compatible with this PR when ensemble == AVERAGE that would be awesome.

ihavnoid:
@roy7 so, what you are suggesting is to evaluate all eight symmetries in parallel.  Correct?

roy7:
Yes. In real life it isn't important, but I've been playing around with some changes to the Thompson Sampling branch where I'm using average symmetry for all lookups. So it's 1/8 the speed of normal. From a batching standpoint, there's always exactly NUM_SYMMETRIES lookups that could be batched together even in a single thread when doing that.

ihavnoid:
10-second per move with default command line and noponder

```
                   A     B
A leelaz-baseline       22-58
B leelaz-batch    58-22

leelaz-baseline v leelaz-batch (80/80 games)
board size: 19   komi: 7.5
leelaz-baseline     22 27.50%       9  22.50%     13 32.50%   1527.43
leelaz-batch        58 72.50%       27 67.50%     31 77.50%   1327.68
                                    36 45.00%     44 55.00%
```

src/Leela.cpp
```diff
@@ -274,15 +337,19 @@ static void parse_commandline(int argc, char *argv[]) {
         }
     }
 #endif
+    if (vm.count("cpu-only")) {
```
Hersmunch: Repeated on line 378

Ttl:
It's that the number of threads needs to be equal or greater to batch size. It's not possible to have single threaded search with batched NN evaluation with this approach. Cudnn requires much bigger batch size to be fast enough, either 16, 32 or 64 would be good choices. For example if user had 4 GPUs and batch size of 64 per GPU they would require at least 256 CPU search threads with this approach, which is kind of silly as it far exceeds the number of cores in common CPUs.

Lc0 does this the right way (https://github.com/LeelaChessZero/lc0/blob/master/src/mcts/search.cc#L690), it's possible to assemble arbitrary big batch even with single search thread. With this approach there are more possibilities in optimizing the batch gathering, for example some positions can be put only in the NN cache and not backpropagated. No waiting heurestics are needed either since the search thread knows when there are no more positions that can fit in the batch.

ihavnoid:
@Ttl - okay if my understanding is correct, it seems that what you are suggesting is NN precomputation.  I was also thinking whether that would help, but I couldn't convince myself that it was necessary, simply because I couldn't find much evidence that throwing more threads resulted in a weaker engine.

In other words, `-t 10 -v 10000` should be weaker than `-t 1 -v 10000`, but so far I couldn't find any evidence that it is true, at least for 19x19 go games.  Maybe it will start getting weaker when we go 200 or 2000 threads, but we aren't getting anywhere close to that.

The downside of lc0 approach is that we may end up computing NN evaluations that does not contribute to anything.  I wonder what is the 'hit rate' of the NN evaluations - that is, how many of the evaluations really end up contributing to the back propagation.  Instead of the waiting heuristics, now we need search heuristics on how to create NN evaluations that is likely to be used later on.

If massive number of threads are not your cup of tea, we can try using other approaches like coroutines...

Ttl:
Don't get me wrong, I think this is good enough for now and it is a clear improvement. OpenCL backend doesn't require very large batch size.

It's more about cudnn which requires larger batch size. Adding positions to minibatch that only go to cache was just an example. It's probably not useful with OpenCL backend that is fast even with small batch size, but cudnn Winograd convolution takes basically the same time for batch sizes 1 to 16 so when there would be less than 16 positions in batch it's almost free to add some more to fill it even if it's not clear if they will be used.

lp200:
K80 with 192x15 net 
2 sec per move

```
pr2188 v next   ( 30 games)
              wins        black       white
pr2188   19 63.33%    9 64.29%   10 62.50%
next     11 36.67%    5 35.71%    6 37.50%
                     14 46.67%   16 53.33%
```
faster and stronger:relaxed:

ihavnoid:
Added 2x more threads.  This results in saturating the GPU much better with tensor cores.  Seems that the faster GPU is resulting in the thread count being bottlenecked more.

src/OpenCLScheduler.cpp
```diff
@@ -254,29 +266,142 @@ template <typename net_t>
...
+        if (m_single_eval_in_progress.load()) {
+            m_waittime += 2;
+        }
+    }
+    m_cv.notify_one();
+    entry->cv.wait(lk);
+}
+
+#ifndef NDEBUG
+std::atomic<size_t> batch_stats[2];
```
gcp: Make this a struct so you can give meaningful names?

src/OpenCLScheduler.cpp
```diff
@@ -254,29 +266,142 @@ template <typename net_t>
...
+
+    // batch scheduling heuristic.
+    // Returns the batch picked up from the queue (m_forward_queue)
+    // 1) Wait for m_waittime milliseconds for full batch
+    // 2) if we don't have a full batch then just do a single eval
+    //
+    // The purpose of m_waittime is to prevent the system from deadlocking
+    // because we were waiting for a job too long, while the job is never
+    // going to come due to a control dependency (e.g., evals stuck on a
+    // critical path).  To do so:
```
gcp: Does this actually happen?

I guess maybe due to this logic? https://github.com/leela-zero/leela-zero/blob/master/src/UCTNode.cpp#L268

src/OpenCLScheduler.cpp
```diff
@@ -254,29 +266,142 @@ template <typename net_t>
...
+    //
+    // 1) if we couldn't form a batch after waiting m_waittime ms, it means 
+    // that we hit the critical path and should do scalar evals.
+    // Wait 1ms shorter next time.
```
gcp: I'm not a fan of time based batch forming but making it dynamic based on experience is a very, very nice trick.

src/OpenCLScheduler.cpp
```diff
@@ -254,29 +266,142 @@ template <typename net_t>
...
+    //
+    // 2) if we picked up a single eval, but were getting additional evals
+    // while that single eval was being processed, it means that we made
+    // the wrong decision.  Wait 2ms longer next time.
+
+    auto pickup_task = [this, gnum] () {
+        std::list<std::shared_ptr<ForwardQueueEntry>> inputs;
+        int count = 0;
```
gcp: auto count = size_t{0};

Would make the cast below disappear, though maybe we'd now get a complaint that cfg_batch_size is not a size_t...so maybe it should be?

src/OpenCLScheduler.cpp
```diff
@@ -254,29 +266,142 @@ template <typename net_t>
...
+                lk,
+                std::chrono::milliseconds(m_waittime),
+                [this] () {
+                    return !m_running || static_cast<int>(m_forward_queue.size()) >= cfg_batch_size;
+                }
+            );
+
+            if (!m_forward_queue.empty()) {
+                if (timeout && m_single_eval_in_progress.exchange(true) == false) {
+                    // waited long enoough but couldn't form a batch.
```
gcp: nit: typo enough

src/OpenCLScheduler.cpp
```diff
@@ -254,29 +266,142 @@ template <typename net_t>
...
+            }
         }
-        // If this failed, it means we ran out of contexts
-        // which should be more than or equal to the number of threads.
-        assert(ctx != nullptr);
-    }
+        // move 'count' evals from shared queue to local list
+        auto end = begin(m_forward_queue);
+        std::advance(end, count);
+        std::move(begin(m_forward_queue), end, std::back_inserter(inputs));
```
gcp: Instead of end you could just write m_forward_queue + count?

src/OpenCLScheduler.cpp
```diff
@@ -254,29 +266,142 @@ template <typename net_t>
...
+
+        if (!m_running) {
+            return;
+        }
+
+#ifndef NDEBUG
+        batch_stats[static_cast<int>(count) == cfg_batch_size ? 1 : 0]++;
+#endif
+
+        // parepare input for forward() call
```
gcp: typo: prepare

src/OpenCLScheduler.cpp
```diff
@@ -254,29 +266,142 @@ template <typename net_t>
...
+        batch_input.resize(in_size * count);
```
gcp: If we want to lessen the malloc overhead, these could be thread_locals.

I'm not sure the speed difference will be measurable.

src/OpenCLScheduler.cpp
```diff
@@ -254,29 +266,142 @@ template <typename net_t>
...
+        batch_output_pol.resize(out_pol_size * count);
+        batch_output_val.resize(out_val_size * count);
+
+        size_t index = 0;
```
gcp: auto index = size_t{0};

src/OpenCLScheduler.cpp
```diff
@@ -254,29 +266,142 @@ template <typename net_t>
...
+        for (auto it = begin(inputs); it != end(inputs); ++it) {
```
gcp: I think this can be a range based for that takes a (const) ref?

src/OpenCLScheduler.cpp
```diff
@@ -254,29 +266,142 @@ template <typename net_t>
...
+            index++;
+        }
+
+        // run the NN evaluation
+        m_networks[gnum]->forward(
+            batch_input, batch_output_pol, batch_output_val, context, count);
+
+        // Get output and copy back
+        index = 0;
+        for (auto it = begin(inputs); it != end(inputs); ++it) {
```
gcp: Same?

gcp:
> they would require at least 256 CPU search threads with this approach, which is kind of silly as it far exceeds the number of cores in common CPUs.

I understand that the split select-node and update-ancestors way is more flexible in that it allows large batches with a single thread, and the kind of update tricks as you describe, but it's also true that there's no real underlying reason why using hundreds of threads is inherently evil. My main concern there would be running in to more OpenCL bugs, and the fact that we use spinlocks for tree synchronization, which will *not* scale paste the amount of CPU threads.

>I couldn't find much evidence that throwing more threads resulted in a weaker engine. In other words, -t 10 -v 10000 should be weaker than -t 1 -v 10000, but so far I couldn't find any evidence that it is true, at least for 19x19 go games.

That's interesting and somewhat counter-intuitive, but the fact that lc0 runs with 256-sized batches was some indication already.

>there are more possibilities in optimizing the batch gathering, for example some positions can be put only in the NN cache and not backpropagate

Nice trick, this avoids the scores being pulled towards avg() instead of max() due to virtual loss widening the search...

For what it's worth, I would prefer tracking the ko-hash of every position we enqueue, and making the batching decision based on the same hash being enqueued twice, thus signaling us that we have run out of virtual-loss or (https://github.com/leela-zero/leela-zero/blob/next/src/UCTNode.cpp#L280) based parallelism, rather than a timed-based wait. (I haven't looked at lc0's search but I assume that is what they are also doing as they have parameters about collisions) But making the time for the wait based on dynamic statistics removes most of my concerns about this, and in any case we can just follow up on this to try to improve it.

gcp:
Code looks OK to me with minor nits.

Need some benchmarking if the defaults are also better on smaller and intermediate size GPU. Specifically, do the defaults not fail on cards with a small amount of RAM?

I'm also a bit curious wrt thread scaling, if there's no cases where we near-deadlock due to spinlock contention on the tree.

gcp:
Issues like #2197 suggest that perhaps batching or threading should be disabled if the amount of playouts is small, to avoid people shooting themselves in the foot.

Maybe factor 20 less than batch size or something?

ihavnoid:
I will wait for more comments until end of the week, address them, and get them merged.  Thanks for your suggestions.

src/Leela.cpp
```diff
@@ -104,6 +172,7 @@ static void parse_commandline(int argc, char *argv[]) {
                 "ID of the OpenCL device(s) to use (disables autodetection).")
         ("full-tuner", "Try harder to find an optimal OpenCL tuning.")
         ("tune-only", "Tune OpenCL only and then exit.")
+        ("batchsize", po::value<int>(), "Max batch size. Default is the number of threads divided by the number of GPUs")
```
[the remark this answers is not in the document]
Hersmunch: Replace “GPU” with “OpenCL device”

src/Leela.cpp
```diff
@@ -61,15 +61,83 @@ static void license_blurb() {
...
+            cfg_batch_size = vm["batchsize"].as<int>();
+        } else {
+            cfg_batch_size = 5;
+        }
+
+        cfg_num_threads = std::min(cfg_max_threads, cfg_batch_size * gpu_count * 2);
+    }
+
+    if (cfg_num_threads < cfg_batch_size) {
+        printf("Threads number = %d must be larger than batch size = %d\n", cfg_num_threads, cfg_batch_size);
```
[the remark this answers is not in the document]
alreadydone: Yes, it's possible to get full batches when they are equal, and impossible if num_threads is less.

Ttl:
Increasing batch size by one increases the memory allocation by about 2 MB with single precision and half of that with half precision. The transformed 40b weights for comparison are about 700 MB in single precision. Increased GPU memory use from batching shouldn't be a problem.

gcp:
Oh, I guess that's because the weights are the expensive part, and all that batching does is introduce more 256 x 19 x 19 x (2 | 4) sized intermediates, which are tiny in the scheme of things.

src/Leela.cpp
```diff
@@ -61,15 +61,83 @@ static void license_blurb() {
...
+
+#ifdef USE_OPENCL
+static void calculate_thread_count_gpu(boost::program_options::variables_map & vm) {
+    auto cfg_max_threads = MAX_CPUS;
+
+    // Default thread count : GPU case
+    // 1) if no args are given, use batch size of 5 and thread count of (batch size) * (number of gpus) * 2
+    // 2) if number of threads are given, use batch size of (thread count) / (number of gpus) / 2
+    // 3) if number of batches are given, use thread count of (batch size) * (number of gpus) * 2
+    auto gpu_count = static_cast<int>(cfg_gpus.size());
```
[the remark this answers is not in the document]
ihavnoid: Seems to need one, as it compile-errors below due to signed / unsigned mixed arithmetic.

src/OpenCL.cpp
```diff
@@ -172,9 +172,9 @@ void OpenCL_Network<net_t>::forward(const std::vector<float>& input,
         const auto n_ceil = ceilMultiple(ceilMultiple(tiles, nwg), vwn);
 
         const auto alloc_inSize =
-            MAX_BATCH * NUM_INTERSECTIONS * max_channels * sizeof(net_t);
+            getOpenCL().m_batch_size * NUM_INTERSECTIONS * max_channels * sizeof(net_t);
```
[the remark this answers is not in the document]
ihavnoid: Maybe as a follow-up, but as mentioned below, memory footprint doesn't change much as batch size gets larger, so unless somebody wants to try a grossly out-of-default value then it shouldn't be a problem.

src/OpenCLScheduler.cpp
ihavnoid: Not only that, but if we imagine what happens when we start from an empty tree, at least the root has to be expanded for any further evaluations.

src/OpenCLScheduler.cpp
ihavnoid: Doesn't seem to work since the iterator is not a RandomAccessIterator.

src/OpenCLScheduler.cpp
ihavnoid: Doesn't really need to be thread_local as this local variable will be alive until the thread terminates.  We can avoid resize() but I left it as-is since this isn't something big enough to affect performance.  What probably will happen is that the vector will keep memory large enough to keep the whole batch, while only the (internal) size variable is going to change.

ihavnoid:
Addressed comments so far.  Please let me know if I missed anything.

One thing in particular is related to signed/unsigned type casting.  I just changed the type of batchsize and thread count from (signed) int to unsigned int.   I see no reason why thread count should be a signed, and just changing things to unsigned made things simpler.

src/OpenCLScheduler.cpp
alreadydone: deque should have random access iterators and seems to fit the use case here (inserting/removing at end/start).

roy7:
Since the batching requires new tuning, I ran full tunes at a number of thread counts on my 1080Ti.

1 and 2 threads (3134 GFLOPS)
4 threads (4548.3 GFLOPS)
8 threads (5535.1 GFLOPS)
16 threads (6571.1 GFLOPS)
16 threads with --batchsize 16 (8228.6 GFLOPS)

I found it interesting that I can increase batch size over the recommend half the # of threads and, at least on the tuning GFLOPS, it seemed to provide a really big boost.

ihavnoid:
@roy7 The default number 5 seemed to be a reasonable selection among different GPUs.  I can definitely say that higher-performing GPUs will be more optimal on larger batches, but for that level of detail... I think we need more contributions from many people.

Ttl:
Tuner always runs single threaded. It's not a good idea to try to optimize parameters based on the tuner output since there are lot of other things that affect the performance in play.

Testing with `--benchmark` command `-t 16` is about 40% faster than`-t 16 --batchsize=16` on my PC even though `-t 16` gets worse results in tuner.

AncalagonX:
The tuner finishes very quickly for low batchsizes. Whereas previously a 40-block quicktune would have taken about 35 seconds for me with no extra arguments, now it takes about 6 seconds with no extra arguments.

With the extra argument `--batchsize 4` it takes 6 seconds, with `--batchsize 2` it takes 4 seconds, and with `--batchsize 1` it takes less than 2 seconds.

Is this intended tuner behavior for the quicktune?

AncalagonX:
The tuner finishes very quickly for low batchsizes. Whereas previously a 40-block quicktune would have taken about 35 seconds for me with no extra arguments, now it takes about 6 seconds with no extra arguments.

With `--batchsize 4` it takes 6 seconds, with `--batchsize 2` it takes 4 seconds, and with `--batchsize 1` it takes less than 2 seconds.

Is this intended tuner behavior for the quicktune?

roy7:
@Ttl The tuner runs now for any new combination of threads/batchsize. I assumed that meant the tuner itself was using the --threads provided? If the tuner uses 1 thread I'm confused  how it can tell --threads 2 from --threads 8, or is it just the batch sizes that matter?

If so, does that mean --threads 8 and --threads 4 --batchsize 4 would be the same actual underlying tuner result (only batchsize matters, not threads)?

Ttl:
Tuner tries to find the best parameters for matrix-matrix multiplication kernel on GPU. It's not even measuring the speed of single convolution, only a part of it. Batch size changes the size of the input matrices which affects the optimal parameters.

Number of threads affects how fast the batch can be assembled. If number of threads is too small GPU can be idle while search threads try to fill the batch. The evaluation of the network on GPU runs at the same speed no matter how many search threads there are.

So looking at only GPU evaluation speed high batch size and few threads might look good, but during the actual play it might take so much more time to assemble the whole batch that it ends up being slower.

roy7:
Makes a lot of sense, thanks. So in terms of my doing full-tunes in advance do I only need to worry about the actual batch sizes and not the actual thread counts?

src/Leela.cpp
```diff
@@ -61,14 +61,82 @@ static void license_blurb() {
...
+        if (num_threads > cfg_max_threads) {
+            myprintf("Clamping threads to maximum = %d\n", cfg_max_threads);
+            num_threads = cfg_max_threads;
+        }
+        cfg_num_threads = num_threads;
+
+        if (vm.count("batchsize")) {
+            cfg_batch_size = vm["batchsize"].as<unsigned int>();
+        } else {
+            cfg_batch_size = (cfg_num_threads + (gpu_count * 2) - 1) / (gpu_count * 2);
```
alreadydone: Why the `(gpu_count * 2) - 1`?

I think `cfg_num_threads / (gpu_count + 1)` is reasonable, since it's theoretically possible to saturate *n* gpus with *(n+1)* x batchsize search threads (if the gpus are of the same speed with evaluations scheduled in an evenly spaced manner, and if the search threads are fast enough to pick up a position in *1/n* of the time of a batch evaluation).

PR #2191, opened 2019-02-03
Mixed precision training support

godmoves:
Add mixed precision training support. In this method we do all computation in fp16 except loss calculation and the weight update/storage, this will speed up the training procedure and help us use larger batch size.

### DIFF
--- a/training/tf/mixprec.py
+++ b/training/tf/mixprec.py
@@ -1,0 +1,48 @@
+import tensorflow as tf
+
+
+def float32_variable_storage_getter(getter, name, shape=None, dtype=None,
+                                    initializer=None, regularizer=None,
+                                    trainable=True,
+                                    *args, **kwargs):
+    """Custom variable getter that forces trainable variables to be stored in
+    float32 precision and then casts them to the training precision."""
+    storage_dtype = tf.float32 if trainable else dtype
+    variable = getter(name, shape, dtype=storage_dtype,
+                      initializer=initializer,
+                      regularizer=regularizer,
+                      trainable=trainable,
+                      *args, **kwargs)
+    if trainable and dtype != tf.float32:
+        cast_name = name + '/fp16_cast'
+        try:
+            cast_variable = tf.get_default_graph().get_tensor_by_name(
+                cast_name + ':0')
+        except KeyError:
+            cast_variable = tf.cast(variable, dtype, name=cast_name)
+        cast_variable._ref = variable._ref
+        variable = cast_variable
+    return variable
+
+
+class LossScalingOptimizer(tf.train.Optimizer):
+    """An optimizer that scales loss and un-scales gradients."""
+
+    def __init__(self, optimizer,
+                 scale=None,
+                 name="LossScalingOptimizer",
+                 use_locking=False):
+        super(LossScalingOptimizer, self).__init__(
+            name=name, use_locking=use_locking)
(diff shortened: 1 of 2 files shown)


godmoves:
Here is the speed test. We can see there is a relatively large speedup for large networks. For GPU with TensorCore, it will use that for the fp16 computation automatically, so the speed improvement for smaller net is larger than the GPU without that.

1080ti (no TensorCore)

|Architecture|Batch Size|FP32 Speed (pos/s)|FP16 Speed (pos/s)|Speed Boost|
|------------|----------|------------------|------------------|-----------|
|10x128      |128       |1300              |1490              |14.6%      |
|            |256       |1195              |1397              |16.9%      |
|            |512       |OOM*              |1358              |           |
|20x256      |128       |205               |337               |64.4%      |
|            |256       |OOM*              |278               |           |

2080ti (TensorCore)

|Architecture|Batch Size|FP32 Speed (pos/s)|FP16 Speed (pos/s)|Speed Boost|
|------------|----------|------------------|------------------|-----------|
|10x128      |128       |1752              |2543              |45.1%      |
|            |256       |1810              |2753              |52.1%      |
|            |512       |OOM*              |2529              |           |
|20x256      |128       |381               |562               |47.5%      |
|            |256       |OOM*              |251               |           |

OOM: Out of memory.

gcp:
I'll try testing it on Windows & with an RTX.

roy7:
Any test results to report @gcp? :)

gcp:
I don't. I no longer manage to get TensorFlow working in Windows at all.

The problem is that I've upgraded to CUDA 10.0 to get Turing features, but the official TensorFlow releases, for god knows what reason, are still on CUDA 9.0 and MSVC2015.

Nightlies do seem to be on CUDA 10.0 but are too broken and crash with obscure error messages.

So I have no good way to test this right now.

apetresc:
> The problem is that I've upgraded to CUDA 10.0 to get Turing features, but the official TensorFlow releases, for god knows what reason, are still on CUDA 9.0 and MSVC2015.

This doesn't appear to be the case anymore since 1.13.1, so maybe it's worth a try now?

gcp:
I am pretty sure it was downloading 1.13.x.

godmoves:
For networks larger than 40b256f, the reg_term of the initial random weight might be too large and get out of the fp16 range. So I cast the l2 loss of each weight back to fp32 before sum them up to prevent the possible overflow.

gcp:
I'll try this again now, the docs seem to really claim CUDA 10 support for latest TensorFlow.

vipmath:
@gcp  https://github.com/fo40225/tensorflow-windows-wheel/tree/master/1.13.1/py37/GPU/cuda101cudnn75sse2

gcp:
I can only confirm what I already said: TensorFlow 1.13.1 does not support CUDA 10.

gcp:
Current TensorFlow Nightlies "work" in the sense that they start up, but they hang after loading the shufflebuffers and nothing seems to happen on the GPU.

TensorFlow 2.0 alpha is not backwards compatible and can't run the code.

gcp:
@vipmath That one won't work (wrong CUDA, wrong Python), but that repo does have a version of 1.12.0 that might be suitable.

gcp:
Works, about 33% speedup on a Turing card.

PR #2199, opened 2019-02-06
added: .gitignore: build

ncaq:
leela-zero default build directory is `build`.
It is very annoying when using leela as a git submodule that the repository updates whenever it builds.


gcp:
Backed out of master and transplanted to next.

PR #2206, opened 2019-02-10
Autogtp: Tune for batchsize 1

Ttl:
Self-play games specify `-t 1` for playing which implies batch size of 1, but tuning was done for default settings since number of threads was not specified.


marcocalignano:
I always wanted to do this!

PR #2232, opened 2019-02-18
ELF training data converter

alreadydone:
Usage: in GTP console, type `convert_elf`, followed the json training data file to convert, followed by a space, followed by the prefix of output file name; just like `dump_supervised`.
Accept decompressed .json files only, not .xz files.
I used @Ishinoshita's https://github.com/Ishinoshita/Leela-Zero-Training-Data-Viewer to correct the alignment between the policy probabilities in ELF and LZ (the correct conversion formula is i*19+j -> (18-i)*19+j) and made sure that the picked move has the most visits after the first 30 moves.

Decompressed file is 10x or more bigger than the .xz file, so it's very much desired to decompress as necessary. I found some libraries to do so, to be detailed in the issue https://github.com/leela-zero/leela-zero/issues/2220. However, I'm not sure how to use the libraries, and it'd be preferable that somebody familiar with the licenses have a look at them.


alreadydone:
I'd expect @Hersmunch to be the most appropriate to review this, but does other members have opinions about whether this should be merged?
I am not sure whether ELF's training data is of higher or lower quality; although ELF 20b are weaker than LZ 40b under the same visits, the data is generated with -p 1600 (so Dirichlet noise is always in effect for 1600 simulations for each move, and -p 1600 is also higher quality than -v 1600 by itself). At least some people would like to give it a try. The training data is grouped by model, and the strongest models can be read out from the paper (see diagram below).
image

src/Training.h
```diff
@@ -47,7 +47,7 @@ class TimeStep {
     using BoardPlane = std::bitset<NUM_INTERSECTIONS>;
     using NNPlanes = std::vector<BoardPlane>;
     NNPlanes planes;
-    std::vector<float> probabilities;
+    std::array<float, POTENTIAL_MOVES> probabilities{{0}};
```
gcp: Why change the datatype here, especially if you're going to try to std::move it?

src/Training.cpp
```diff
@@ -342,18 +341,73 @@ void Training::process_game(GameState& state, size_t& train_pos, int who_won,
         step.to_move = to_move;
         step.planes = get_planes(&state);
 
-        step.probabilities.resize(POTENTIAL_MOVES);
-        step.probabilities[move_idx] = 1.0f;
+        if (policies.size() == 0) {
+            step.probabilities[move_idx] = 1.0f;
+        } else {
+            step.probabilities = std::move(policies[counter]);
```
gcp: This does not do what you think it does because it's a std::array, not a std::vector.

src/SGFParser.cpp
```diff
@@ -38,6 +38,81 @@
 #include "SGFTree.h"
 #include "Utils.h"
 
+std::vector<SGFParser::ELF_Data> SGFParser::chop_elf(std::string filename,
```
gcp: If it's not actually parsing SGF then just put it into its own file?

src/SGFParser.cpp
```diff
@@ -38,6 +38,81 @@
...
+
+    while (result.size() <= stopat) {
+        while (true) {
+            ins.ignore(MAX, '\"');
+            if ((ins.rdstate() & std::ifstream::eofbit) != 0) return result;
+            std::getline(ins, str, ':');
+            if (str == "content\"") break;
+        }
+        result.emplace_back(ELF_Data());
+        auto& data = result.back();
```
gcp: Don't need this trick with a vector, they're almost free to move.

src/SGFParser.cpp
```diff
@@ -38,6 +38,81 @@
...
+        data.policies.reserve(num_move + 1);
+        while (true) {
+            ins >> c; // eat [ (assume there's at least one policy record)
+            str.clear();
+            std::getline(ins, str, ',');
+            data.policies.emplace_back(std::array<float, POTENTIAL_MOVES>());
+            auto& probs = data.policies.back();
+            probs[NUM_INTERSECTIONS] = total_policy = std::stoi(str);
+            for (auto i = 0; i < BOARD_SIZE; i++) ins.ignore(MAX, ',');
+            for (int i = BOARD_SIZE - 1; i >= 0; i--) {
```
gcp: Weird mix of auto and int here.

src/SGFParser.cpp
```diff
@@ -38,6 +38,81 @@
...
+        data.policies.clear();
+        data.policies.reserve(num_move + 1);
+        while (true) {
+            ins >> c; // eat [ (assume there's at least one policy record)
+            str.clear();
+            std::getline(ins, str, ',');
+            data.policies.emplace_back(std::array<float, POTENTIAL_MOVES>());
+            auto& probs = data.policies.back();
+            probs[NUM_INTERSECTIONS] = total_policy = std::stoi(str);
+            for (auto i = 0; i < BOARD_SIZE; i++) ins.ignore(MAX, ',');
```
gcp: Braces please.

src/SGFParser.cpp
```diff
@@ -38,6 +38,81 @@
...
+                ins.ignore(MAX, ',');
+                for (auto j = 0; j < BOARD_SIZE; j++) {
+                    str.clear();
+                    std::getline(ins, str, ',');
+                    auto entry = std::stoi(str);
+                    total_policy += entry;
+                    probs[i * BOARD_SIZE + j] = entry;
+                }
+            }
+            for (auto& p : probs) p /= total_policy;
```
gcp: Braces.

src/Training.cpp
```diff
@@ -82,10 +82,10 @@ std::istream& operator>> (std::istream& stream, TimeStep& timestep) {
     }
     int prob_size;
     stream >> prob_size;
-    for (auto i = 0; i < prob_size; ++i) {
+    for (auto i = 0; i < prob_size && i < POTENTIAL_MOVES; ++i) {
```
gcp: Why does it need the extra condition?

src/Training.cpp
```diff
@@ -342,18 +341,73 @@ void Training::process_game(GameState& state, size_t& train_pos, int who_won,
...
+        } else {
+            step.probabilities = std::move(policies[counter]);
+        }
 
         train_pos++;
         m_data.emplace_back(step);
 
         counter++;
-    } while (state.forward_move() && counter < tree_moves.size());
+    } while (state.forward_move() && counter < tree_moves.size() && counter < policies.size());
```
gcp: Does this not break the existing code, as you're passing an empty array so it will exit immediately?

gcp:
>However, I'm not sure how to use the libraries, and it'd be preferable that somebody familiar with the licenses have a look at them.

The thing you linked is BSD 2 clause so it's fine, but it requires adding a bunch of dependencies in again just to deal with ELF data conversion. I'd hide the code behind a define and not include it by default then.

You can also just recompress from xz to gz.

>other members have opinions about whether this should be merged?

It probably breaks the existing training, so those issues should be fixed first. My concern about including it would be that it's really only useful once, so that's not very compelling to keep carrying around.

src/Training.cpp
alreadydone: Good catch. I thought that std::array would behave similarly to std::vector (takes O(1) time to move) and didn't think how std::move would work for an aggregate type. Let me change it back to std::vector.

src/Training.h
alreadydone: NNPlanes and probabilities sizes are both fixed at compile time. Are there reasons to prefer std::vector over std::array here?

src/SGFParser.cpp
alreadydone: What about making it a member of `Training` as well? I originally started working in SGFParser.cpp because I would reuse some code there.

src/SGFParser.cpp
alreadydone: ELF_Data is a struct consisting of a std::string, a float, and a std::vector, so I think this is better. Any argument against this "trick"?

src/SGFParser.cpp
alreadydone: With auto `i` will be assigned an `unsigned` type, and will always be `>=0` leading to infinite loop. I could change it to `auto i = (int)BOARD_SIZE - 1;` if you wish...

src/Training.cpp
alreadydone: I somehow confused myself ... I really want `||`, not `&&` here. Will fix and test.

src/Training.cpp
alreadydone: I am just checking for out-of-bound because I'm now accessing by index. (I don't know what I can assume about prob_size.) Anyway it's moot now that I'm switching back to vector.

alreadydone:
Thanks for the review! Maybe I won't introduce the bunch of dependencies into this repo. I am not sure I have the right to redistribute the datasets as .gz (and it makes more sense to distribute the converted training data (already in .gz)) since they don't come with a license.

TFiFiE:
@alreadydone, if you still want to keep this pull request open for some reason, would you mind running the following on your end, so that it properly shows here what actual changes it makes?
```
git rebase --onto ac6c4ab64129710d718286d5ebfca3b6f3689686 efcec18e12b5dbf38a278a5a3dfb2f19ea72dfd1 convert_elf
git push -f
```
You don't have to resolve the would-be merge conflict.

alreadydone:
No need to keep it open but I'll keep the branch https://github.com/alreadydone/lz/commits/convert_elf and the relevant commits https://github.com/alreadydone/lz/commit/d9159552e1548b5a567023c0e33c92eb45970066 https://github.com/alreadydone/lz/commit/054ba13d8fe5d2654bfa6e35a48f566770d197f1

PR #2241, opened 2019-02-21
small tweak for ELF v2 weights

sethtroisi:
This was merged into leela-zero/next in https://github.com/leela-zero/leela-zero/pull/2213
it should have been merged to the master branch as it's doesn't affect the client code and doesn't require staging/releasing.


gcp:
I did a cherry-pick of the patch on ``next`` to avoid any merge issues.

Not too hot about merging stuff in master (instead of making sure next is releasable), but if this helps you out...

Issue #1850, opened 2018-09-19
Possibly negative winrate estimation


Vandertic:
https://github.com/gcp/leela-zero/blob/9e903eda6e93fb65375ea03127e4f62268913175/src/UCTNode.cpp#L264
In some cases fpu_eval may result negative, which I think is unintended behaviour, as this should be a pessimistic but reasonable estimate of an unknown winrate. 

I suggest maxing or softmaxing it with 0.

gcp:
This is intentional, IIRC, and clamping it to zero was no better or even worse. Could be rested.

Vandertic:
Thank you for the clarification. I don't think it could make much of a difference myself.

PR #2242, opened 2019-02-21
Clamp Q to 0 if fpu reduces below that value

sethtroisi:
I'm not sure how to properly test, I'd appreciate any insight people have.

Theoretical `Q` stands for expected winrate and probably shouldn't be negative.

From the literature you could argue this is closer to init-to-loss.

From a practical standpoint I think it would be best to load a bunch of games that are at 10% winrate and see what percent of the time LZ can convert them with / without this patch. I'm not quite sure how to do this with ringmaster but I can investigate if you think that's worthwhile.

I played 100 quick games (with 50 playouts) and logged how many times this changed fpu_eval and it wasn't very often (I can also try to measure what percent of select_child it changes but my guess is << 1%)

```
LZ_173_50 v LZ_173_50_clamp (100/100 games)
board size: 19   komi: 7.5
LZ_173_50           47 47.00%       23 46.00%     24 48.00%     32.54
LZ_173_50_clamp     53 53.00%       26 52.00%     27 54.00%     32.55
                                    49 49.00%     51 51.00%
```

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -76,6 +76,7 @@ float cfg_puct;
 float cfg_softmax_temp;
 float cfg_fpu_reduction;
 float cfg_fpu_root_reduction;
+bool cfg_fpu_clamp;
 std::string cfg_weightsfile;
 std::string cfg_logfile;
 FILE* cfg_logfile_handle;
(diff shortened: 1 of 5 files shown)


TFiFiE:
See #1850.

sethtroisi:
@TFiFiE Thanks for the link. The discussion in the original PR (#1850) is very good. (it's super weird seeing a [post by myself](https://github.com/leela-zero/leela-zero/pull/1565#issuecomment-397969368) that I'd totally forgotten)

This change is not likely to change self-play move selection very often but it may have some impact on handicap games. Given that gcp and ttl have both devalued it let me I'm closing this issue till I can provide better supporting data

PR #2243, opened 2019-02-22
[QoL] Command history in Leela

sethtroisi:
I made a [little demo](https://giant.gfycat.com/RelievedEntireCanvasback.webm) of it being used.

### DIFF
--- a/cmake/Modules/FindReadline.cmake
+++ b/cmake/Modules/FindReadline.cmake
@@ -1,0 +1,53 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
+# file Copyright.txt or https://cmake.org/licensing for details.
+
+#.rst:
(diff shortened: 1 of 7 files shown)


cmake/Modules/FindReadline.cmake
```diff
@@ -0,0 +1,53 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
sethtroisi: I've never written a cmake file before, this was created by slowly removing parts of FindOpenCL.cmake, I'd appreciate if someone familiar with cmake could take a harder look at this.

gcp:
Are there a lot of people intensively using the engine manually over the command line?

I'm OK with this either way.

PS. Maybe we can remove the old Makefile in /src, even I am no longer using it.

PR #2244, opened 2019-02-22
tree_stats_helper refactor

sethtroisi:


gcp:
@sethtroisi you need to update your submodules, this pull regresses Eigen to a previous version

PR #2246, opened 2019-02-23
Fixed Travis links in README.md

sethtroisi:


sethtroisi:
Fixed instead by https://github.com/leela-zero/leela-zero/pull/2229

PR #2248, opened 2019-02-23
Update .travis.yml with script names

sethtroisi:
Old: 
screenshot from 2019-02-22 17-13-12


sethtroisi:
I need to test this locally at home to make sure these are valid.

https://stackoverflow.com/questions/21053657/how-to-run-travis-ci-locally

PR #2253, opened 2019-02-25
Enable batching for self-play

Ttl:
Self-play games are still played with batch-size of 1. Larger batch size results in weaker games, but it should speed up the game generation. If the strength loss is small enough compared to speed-up it makes sense to enable it for self-play generation.

Some strength testing against `-t 1 --batch-size=1` with 1600 visits and Dirichlet noise enabled:

```
-t 10 --batch-size=5

lz_205_bs5 v lz_205_bs1 (400/400 games)
board size: 19   komi: 7.5
lz_205_bs5    126 31.50%       54  27.00%     72  36.00%    187.08
lz_205_bs1    274 68.50%       128 64.00%     146 73.00%    281.53
                               182 45.50%     218 54.50%
```

```
-t 6 --batch-size=5

lz_205_bs5 v lz_205_bs1 (400/400 games)
board size: 19   komi: 7.5
lz_205_bs5    162 40.50%       68  34.00%     94  47.00%    185.79
lz_205_bs1    238 59.50%       106 53.00%     132 66.00%    294.66
                               174 43.50%     226 56.50%
```

Using too many threads seems to result in decreased strength. I think 6 threads and batch-size of 5 gives most of the speed-up without hurting the strength too much.

With `--benchmark -v 10000` on RTX 2080 I get 434 n/s with `-t1`, 922 n/s with `-t6` and 1075 n/s with `-t10`. Speed-up from enabling batching is over 100% with only a slight decrease in strength. I think it's worth it even if visits would be increased to compensate for the lost strength.

### DIFF
--- a/autogtp/Management.cpp
+++ b/autogtp/Management.cpp
@@ -106,7 +106,7 @@ void Management::giveAssignments() {
     QTextStream(stdout) << "Starting tuning process, please wait..." << endl;
 
     Order tuneOrder = getWork(true);
-    QString tuneCmdLine("./leelaz -t 1 --tune-only -w networks/");
+    QString tuneCmdLine("./leelaz --batchsize=5 --tune-only -w networks/");
     tuneCmdLine.append(tuneOrder.parameters()["network"] + ".gz");
     if (m_gpusList.isEmpty()) {
         runTuningProcess(tuneCmdLine);
@@ -266,7 +266,8 @@ QString Management::getOptionsString(const QJsonObject &opt, const QString &rnd)
     options.append(getOption(opt, "visits", " -v ", ""));
     options.append(getOption(opt, "resignation_percent", " -r ", "1"));
     options.append(getOption(opt, "randomcnt", " -m ", "30"));
-    options.append(getOption(opt, "threads", " -t ", "1"));
+    options.append(getOption(opt, "threads", " -t ", "6"));
+    options.append(getOption(opt, "batchsize", " --batchsize ", "5"));
     options.append(getBoolOption(opt, "dumbpass", " -d ", true));
     options.append(getBoolOption(opt, "noise", " -n ", true));
     options.append(" --noponder ");


roy7:
What alternatives (if any practical ones exist) are there to find evals to batch that won't result in weaker strength?

Ttl:
Some theoretical ones maybe, but nothing easy to implement at the moment.

Strength decrease isn't probably a big deal. AGZ, minigo, ELF and lc0 all use batching during self-play are doing just fine.

ihavnoid:
There is one more alternative : I have a client-server implementation which does the NN evaluation on a separate process, while the clients connect to the server to get the NN evaluation results.  By launching one server per GPU with huge batches (eg., 10) and launching large number of concurrent clients, we can play a dozen single-threaded games concurrently with the NN evaluations being done in batches.  I tested it a couple of months ago and it definitely works, but I didn't bother to go into the details of integrating the client-server architecture into autogtp since it will be a huge amount of work that may be questionable of value (the setup will end up being very complicated that it will be hard to do the tech support).

If anybody is interested of the client-server implementation, please have a look here:

https://github.com/ihavnoid/leela-zero/tree/distributed

gcp:
>AGZ, minigo, ELF and lc0 all use batching during self-play are doing just fine.

Doesn't ELF batch over concurrent games? Same for lc0? Or do they do both?

Anyway, the speedup is much bigger than the downside so we can take this, I'd be curious if there'd be a way to get all clients more or less at the same strength though. Maybe it's just a matter of making a release with this.

Ttl:
lc0 uses batch size of 32 for single game and batches many games together. I think ELF paper also mentioned that they use batching for single games to gather big enough batch.

gcp:
Any thoughts about what to do with the mix of self-play games that will result? Just ignore?

Ttl:
Are there other options? I think it will work since the difference isn't that large, but who knows without trying it.

PR #2254, opened 2019-02-25
Allow configuring default komi at compile-time

apetresc:
In a similar vein to #928, I'm experimenting with various potential komi values on smaller boards. I'd like to be able to set komi in a single place when compiling, the same way I can now set `BOARD_SIZE`, rather than manually editing the `7.5f` literal in a half-dozen places throughout the code.

Seems to work for me, and tests pass (in tag v0.16 at least)

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -368,7 +368,7 @@ void GTP::execute(GameState & game, const std::string& xinput) {
     } else if (command.find("komi") == 0) {
         std::istringstream cmdstream(command);
         std::string tmp;
-        float komi = 7.5f;
+        float komi = KOMI;
         float old_komi = game.get_komi();
 
         cmdstream >> tmp;  // eat komi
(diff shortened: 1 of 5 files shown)


Hersmunch:
Most of this has been done in #1825 but I think it is fine to do as a separate commit if it is wanted earlier than that. I didn't change all the places as not all of them are dependent upon the trained Komi so take a look at mine and see if you agree. It also needs rebasing to the next branch.

sethtroisi:
@Hersmunch #1825 has been open for almost half a year. This seems like a compact, nice-to-have change that can be merged today. @apetresc has a history of commits and seems like a reasonable developer so unlocking him (and potentially other people playing with komi) seems good. if #1825 needs to change this a little I still think it will be better for having submitted this,  unless you plan to get merge #1825 in the next couple of days.

gcp:
I agree with seth that we can pull this until there's a decision about #1825, but it needs to be on the next branch.

apetresc:
My mistake - I keep forgetting to change the target branch. Reopened as #2257.

Hersmunch:
@sethtroisi thanks for clearing saying what I was actually trying to; I completely agree with it all :) I'll try to be clearer in future so sorry for the confusion.

Issue #2214, opened 2019-02-13
Tests can't pass in GPU mode


gcp:
Was broken by #2188.

Some of the internal variables are no longer set up correctly and the tests end up bypassing the expected initialization.

```
Tensor Core support: No.

Thread 1 "tests" received signal SIGFPE, Arithmetic exception.
0x0000555555610bf3 in OpenCLScheduler<float>::initialize(int) (this=0x5555569d72b0, channels=8)
    at ../src/OpenCLScheduler.cpp:116
116         auto num_worker_threads = cfg_num_threads / cfg_batch_size / (m_opencl.size() + 1) + 1;
(gdb) bt
#0  0x0000555555610bf3 in OpenCLScheduler<float>::initialize(int) (this=0x5555569d72b0, channels=8)
    at ../src/OpenCLScheduler.cpp:116
#1  0x00005555555dc519 in Network::init_net (this=this@entry=0x7ffff7e4f010, 
    channels=channels@entry=8, pipe=...) at ../src/Network.cpp:385
#2  0x00005555555e782c in Network::select_precision (this=this@entry=0x7ffff7e4f010, 
    channels=channels@entry=8) at ../src/Network.cpp:421
#3  0x00005555555e9cbc in Network::initialize (this=this@entry=0x7ffff7e4f010, 
    playouts=playouts@entry=1073741823, Python Exception <class 'gdb.error'> There is no member named _M_dataplus.: 
weightsfile=) at ../src/Network.cpp:566
#4  0x0000555555619c3b in LeelaEnv::SetUp() (this=<optimized out>) at ../src/tests/gtests.cpp:87
#5  0x000055555566aaae in testing::internal::SetUpEnvironment(testing::Environment*) ()
#6  0x0000555555686f47 in void (*std::for_each<__gnu_cxx::__normal_iterator<testing::Environment* const*, std::vector<testing::Environment*, std::allocator<testing::Environment*> > >, void (*)(testing::Environment*)>(__gnu_cxx::__normal_iterator<testing::Environment* const*, std::vector<testing::Environment*, std::allocator<testing::Environment*> > >, __gnu_cxx::__normal_iterator<testing::Environment* const*, std::vector<testing::Environment*, std::allocator<testing::Environment*> > >, void (*)(testing::Environment*)))(testing::Environment*) ()
[...]

PR #2256, opened 2019-02-26
Regression fix : Addresses #2214

ihavnoid:
I am not sure if this is the most reasonable fix, but this enables the option to provide zero threads / batchsize to let leela-zero pick a reasonable default.


gcp:
As described in #2214, I don't think this is useful. It should just display the autodetected defaults.

PR #2257, opened 2019-02-26
Allow configuring default komi at compile-time (next branch)

apetresc:
This is a reissue of #2254 because I accidentally opened that one against the wrong branch - sorry!

In a similar vein to #928, I'm experimenting with various potential komi values on smaller boards. I'd like to be able to set komi in a single place when compiling, the same way I can now set `BOARD_SIZE`, rather than manually editing the `7.5f` literal in a half-dozen places throughout the code.

Seems to work for me, and tests pass (in tag v0.16 at least)


roy7:
@gcp I don't feel qualified to click the merge button in the LZ repo. :) But could you peek at this, or maybe another admin could merge it?

Issue #2273, opened 2019-03-10
Double-passing pathology


gjm11:
Consider the following situation, which might arise anywhere in the search tree.

* Black is far ahead, perhaps because this is a handicap game.
* However, if the game ended here then the end-of-game scorer LZ uses would reckon it a win for white (e.g., because neither side has any territory yet and W has komi).
* The NN's policy head gives non-negligible weight to passing for whoever's move it is in this position. (Even though passing is actually a dumb idea.)
* It's white's move.

Then when MCTS visits this position, quite early in the search it may look at the line where W passes and then B passes -- which will be assessed as a win for _white_. After that, until this position has been visited many many more times, B's winrate here will look anomalously low.

In the long run, this is harmless. But the long run may be very long. E.g., if "normal" play yields a 99.5% winrate for black, then passing in this position is going to look like a good move for white until the position's seen about 200 visits. And if some of the positions further down the tree _also_ evoke non-negligible weight on passing from the policy head, it may be a whole lot more than 200 visits.

This hurts us (1) if passing is one of W's options from the root position, in which case it may get chosen because of this effect even though it's a silly move, and (2) in any position where this sort of thing occurs _somewhere_ in the tree, because then every position's evaluation gets some admixture of nonsense double-pass-and-win outcomes.

For an example, try putting 6 handicap stones on the board and letting white think for a while using Lizzie. Here is what you will see, or at least what I see using network 208, which is the latest right now. (Of course the details may vary depending on the order in which you place those handicap stones, and what orientation LZ happens to feed the board to the NN in.)

* Initially, LZ wants to play either a 3-3 invasion or a knight's-move approach. For the first ~5k visits, these moves are clear winners. Every move, though, has a black winrate in the 97%-99.5% range or thereabouts; typically LZ prefers a move whose winrate is ~97%.
* Every now and then, though, some (typically crazy) move will get considered and will be seen as having a much lower black winrate: maybe 80% or so. As LZ searches it more, that winrate will gradually climb up until it's in the same range as all the other moves' winrates, and then it will switch to a different move.
* Eventually (after ~5k to 15k visits) it will stumble across a move on the edge of the board whose B winrate starts off a little under 2/3 and climbs a bit more slowly; this becomes LZ's preferred move for some time.
[...]

PR #2277, opened 2019-03-13
Address double-passing pathology (#2273)

gjm11:
This PR is for a change to UCTNode that makes the search ignore passing moves in circumstances where they are almost certainly a terrible idea. See #2273 for a full explanation.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -108,10 +108,10 @@ bool UCTNode::create_children(Network & network,
     // Don't consider pass moves if they would be immediately losing and we "shouldn't"
     // accept that; see issue #2273. --dumbpass disables this heuristic.
     bool suppress_passes =
-        winrate_for_player >= 0.75         // we should be winning
-        && nodelist.size() >= 5            // there are surely good alternatives
-        && !cfg_dumbpass                   // allowed to be clever about passing
-        && state.final_score(to_move) < 0; // passing => immediate loss
+        winrate_for_player >= 0.75          // we should be winning
+        && nodelist.size() >= 5             // there are surely good alternatives
+        && !cfg_dumbpass                    // allowed to be clever about passing
+        && state.final_score(to_move) <= 0; // passing => immediate not-win
     if (!suppress_passes) {
         nodelist.emplace_back(raw_netlist.policy_pass, FastBoard::PASS);
         legal_sum += raw_netlist.policy_pass;
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -106,8 +106,6 @@ bool UCTNode::create_children(Network & network,
 
     // Add pass move. BUT don't do this if the following conditions
     // all obtain (see issue #2273, "Double-passing pathology"):
-    //   - The move played in order to reach this node was a pass.
-    //     (So another pass would end the game.)
     //   - The NN's evaluation of the current node is very good
     //     for the player whose move it is. (Say, 0.75 or better.)
     //     (So we don't want to end the game unless we win.)
@@ -126,11 +124,9 @@ bool UCTNode::create_children(Network & network,
     // The magic numbers 0.75 and 5 are somewhat arbitrary and it seems
     // unlikely that their values make much difference.
     // This check prevents some serious evaluation errors but has a cost:
-    // we make extra calls to final_score() at some nodes. But this is done
-    // only at nodes where the other player just passed despite having a
-    // really bad position; the cost should not be large.
-    if (state.get_passes() == 0
-        || (to_move == FastBoard::WHITE ? 1.0f - m_net_eval : m_net_eval) < 0.75
+    // we make extra calls to final_score() at some nodes. This is much
+    // cheaper than a single NN evaluation, and the cost should be negligible.
+    if (   (to_move == FastBoard::WHITE ? 1.0f - m_net_eval : m_net_eval) < 0.75
         || nodelist.size() < 5
         || cfg_dumbpass
         || (to_move == FastBoard::WHITE ? -state.final_score() : state.final_score()) > 0) {
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -117,6 +117,12 @@ bool UCTNode::create_children(Network & network,
     //   - We do have at least five other legal moves.
     //     (So it's not likely that all our available moves
     //     are actually disastrous.)
+    //   - The "dumbpass" option is not turned on.
+    //     (Because, as per GCP's comment at
+    //     https://github.com/leela-zero/leela-zero/issues/2273#issuecomment-472398802 ,
+    //     enabling this heuristic is un-Zero-like and enabling
+    //     dumbpass is meant to suppress any such things that
+    //     affect passing.)
     // The magic numbers 0.75 and 5 are somewhat arbitrary and it seems
     // unlikely that their values make much difference.
     // This check prevents some serious evaluation errors but has a cost:
@@ -125,8 +131,9 @@ bool UCTNode::create_children(Network & network,
     // really bad position; the cost should not be large.
     if (state.get_passes() == 0
         || (to_move == FastBoard::WHITE ? 1.0f - m_net_eval : m_net_eval) < 0.75
-        || (to_move == FastBoard::WHITE ? -state.final_score() : state.final_score()) > 0
-        || nodelist.size() < 5) {
+        || nodelist.size() < 5
+        || cfg_dumbpass
+        || (to_move == FastBoard::WHITE ? -state.final_score() : state.final_score()) > 0) {
         nodelist.emplace_back(raw_netlist.policy_pass, FastBoard::PASS);
         legal_sum += raw_netlist.policy_pass;
     }
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -103,8 +103,33 @@ bool UCTNode::create_children(Network & network,
             legal_sum += raw_netlist.policy[i];
         }
     }
-    nodelist.emplace_back(raw_netlist.policy_pass, FastBoard::PASS);
-    legal_sum += raw_netlist.policy_pass;
+
+    // Add pass move. BUT don't do this if the following conditions
+    // all obtain (see issue #2273, "Double-passing pathology"):
+    //   - The move played in order to reach this node was a pass.
+    //     (So another pass would end the game.)
+    //   - The NN's evaluation of the current node is very good
+    //     for the player whose move it is. (Say, 0.75 or better.)
+    //     (So we don't want to end the game unless we win.)
+    //   - Ending the game now would actually lose the game for
+    //     the player whose move it is.
+    //     (So we don't want to end the game.)
+    //   - We do have at least five other legal moves.
+    //     (So it's not likely that all our available moves
+    //     are actually disastrous.)
+    // The magic numbers 0.75 and 5 are somewhat arbitrary and it seems
+    // unlikely that their values make much difference.
+    // This check prevents some serious evaluation errors but has a cost:
+    // we make extra calls to final_score() at some nodes. But this is done
+    // only at nodes where the other player just passed despite having a
+    // really bad position; the cost should not be large.
+    if (state.get_passes() == 0
+        || (to_move == FastBoard::WHITE ? 1.0f - m_net_eval : m_net_eval) < 0.75
+        || (to_move == FastBoard::WHITE ? -state.final_score() : state.final_score()) > 0
+        || nodelist.size() < 5) {
+        nodelist.emplace_back(raw_netlist.policy_pass, FastBoard::PASS);
+        legal_sum += raw_netlist.policy_pass;
+    }
 
     if (legal_sum > std::numeric_limits<float>::min()) {
         // re-normalize after removing illegal moves.
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -107,13 +107,11 @@ bool UCTNode::create_children(Network & network,
 
     // Don't consider pass moves if they would be immediately losing and we "shouldn't"
     // accept that; see issue #2273. --dumbpass disables this heuristic.
-    float final_score_for_player = state.final_score();
-    if (to_move == FastBoard::WHITE) final_score_for_player = -final_score_for_player;
     bool suppress_passes =
-        winrate_for_player >= 0.75     // we should be winning
-        && nodelist.size() >= 5        // there are surely good alternatives
-        && !cfg_dumbpass               // allowed to be clever about passing
-        && final_score_for_player < 0; // passing => immediate loss
+        winrate_for_player >= 0.75         // we should be winning
+        && nodelist.size() >= 5            // there are surely good alternatives
+        && !cfg_dumbpass                   // allowed to be clever about passing
+        && state.final_score(to_move) < 0; // passing => immediate loss
     if (!suppress_passes) {
         nodelist.emplace_back(raw_netlist.policy_pass, FastBoard::PASS);
         legal_sum += raw_netlist.policy_pass;
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -88,8 +88,7 @@ bool UCTNode::create_children(Network & network,
     const auto to_move = state.board.get_to_move();
     if (to_move == FastBoard::WHITE) {
       m_net_eval = eval = 1.0f - winrate_for_player;
-    }
-    else {
+    } else {
       m_net_eval = eval = winrate_for_player;
     }
 
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -83,13 +83,15 @@ bool UCTNode::create_children(Network & network,
         &state, Network::Ensemble::RANDOM_SYMMETRY);
 
     // DCNN returns winrate as side to move
-    m_net_eval = raw_netlist.winrate;
-    const auto to_move = state.board.get_to_move();
     // our search functions evaluate from black's point of view
-    if (state.board.white_to_move()) {
-        m_net_eval = 1.0f - m_net_eval;
+    float winrate_for_player = raw_netlist.winrate;
+    const auto to_move = state.board.get_to_move();
+    if (to_move == FastBoard::WHITE) {
+      m_net_eval = eval = 1.0f - winrate_for_player;
     }
-    eval = m_net_eval;
+    else {
+      m_net_eval = eval = winrate_for_player;
+    }
 
     std::vector<Network::PolicyVertexPair> nodelist;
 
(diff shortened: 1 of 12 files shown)


Ttl:
This does seem to help in the starting position as described in the issue, but doesn't completely fix the problem,

For example in the position below LZ with network 210 tries to pass at low playouts since it would be winning with Tromp-Taylor rules.

At over 10k playouts it sometimes plays on the first line. Looking at the black's move after the first line move, the pass policy is high and even with this PR black is allowed to consider passing. Passing then allows white to get a winning evaluation making the first line move look good.

pr2273_first_line


<details>
<summary>sgf</summary>

```
(;DT[2019-03-13]KM[7.5]PB[]PW[]AP[Lizzie: 0.6];B[pp];W[tt];B[pj];W[tt];B[pd];W[tt];B[dd];W[tt];B[dj];W[tt];B[dp];W[cq];B[cp];W[dq];B[fq];W[eq];B[ep];W[fr];B[bq];W[br];B[er];W[dr];B[gr];W[es];B[gq])
```
</details>

gjm11:
@Ttl Fair point: to prevent all double-passing-related craziness we'd want to apply the same check everywhere, not only at nodes where the last move was a pass. That would increase the cost of the change, though. Perhaps it would still be negligible; I'm not sure how expensive a score-check is. (It feels like it shouldn't be too expensive; it shouldn't have to do very much work per board point.)

There might be intermediate possibilities: e.g., this probably only matters when the policy head's score for passing is fairly high, so we could do the test after a non-passing move only if (1) NN eval is very good and (2) policy puts a high weight on passing.

None of this will prevent the situation where white wants to pass because the amount by which passing makes its situation worse is smaller than the random errors in evaluation. I suppose there could be a heuristic saying "never consider passing anywhere in the tree unless the number of moves played so far is >= 1/3 the number of points on the board" or something of the kind. But that feels more like building human knowledge -- that it's almost always a bad idea to pass early in the game -- into the program, which is rather non-Zero-ish. (I don't think the patch above does. The idea that when making a move that could end the game immediately one should check whether the result of doing so would be favourable is neither subtle nor Go-specific.)

alreadydone:
Yes, I think scoring is very cheap (certainly O(NUM_INTERSECTIONS) with the implemented standard bread-first search). I estimate it to be cheaper than a call of uct_select_child (which involves floating point arithmetic). So I recommend checking it everywhere in the tree.

Ideally the score can be updated after placing each stone, just like how the number of liberties is updated.

The current relevant code is:
https://github.com/leela-zero/leela-zero/blob/4fd6e694ead45b7cbb14ea76c5bdfcbfa662304b/src/FastBoard.cpp#L278-L282
https://github.com/leela-zero/leela-zero/blob/4fd6e694ead45b7cbb14ea76c5bdfcbfa662304b/src/FastBoard.cpp#L246-L275

TFiFiE:
This is just too ugly of a hack; I'd rather see #2273 be addressed by certainty propagation.

gjm11:
I agree that another way to deal with this issue would be via smarter searching (e.g., by pruning nodes that provably lead to a loss), but that's a decidedly more drastic change, albeit one that has the potential to make a bigger difference to LZ's strength.

Given that final-score evaluation is cheap enough, this change can be simplified and made less hackish, the simple rule being "don't consider passing if doing so would lose immediately and you have any other legal moves". ... Except that this would make LZ insist bloody-mindedly on continuing to play at the end of the game even when it's clearly hopeless, so let's make it "don't consider passing if doing so would lose immediately, you have any other legal moves, and this looks like an advantageous position". This is _better_ as well as _simpler_ because it deals with Ttl's observation about some double-passing pathologies remaining after implementing the current version of the fix.

gjm11:
I've made that simplification. The resulting code shows no inclination to play silly moves on the edge in Ttl's position.

src/UCTNode.cpp
```diff
@@ -92,8 +92,36 @@ bool UCTNode::create_children(Network & network,
...
+    //     (Because, as per GCP's comment at
+    //     https://github.com/leela-zero/leela-zero/issues/2273#issuecomment-472398802 ,
+    //     enabling this heuristic is un-Zero-like and enabling
+    //     dumbpass is meant to suppress any such things that
+    //     affect passing.)
+    // The magic numbers 0.75 and 5 are somewhat arbitrary and it seems
+    // unlikely that their values make much difference.
+    // This check prevents some serious evaluation errors but has a cost:
+    // we make extra calls to final_score() at some nodes. This is much
+    // cheaper than a single NN evaluation, and the cost should be negligible.
```
gcp: This is a long comment and it doesn't add much, as it's literally reading off the conditions of the if that follows. I think it's better to put this in the commit message or just refer to the issue here.

Comments are very good if they're explaining something that is not obvious from the code, but here it's just repeating what the code does. I'd rather make the code clearer.

src/UCTNode.cpp
```diff
@@ -92,8 +92,36 @@ bool UCTNode::create_children(Network & network,
...
+    if (   (to_move == FastBoard::WHITE ? 1.0f - m_net_eval : m_net_eval) < 0.75
```
gcp: Pull out this computation into an appropriately-named local variable.

src/UCTNode.cpp
```diff
@@ -92,8 +92,36 @@ bool UCTNode::create_children(Network & network,
...
+        || nodelist.size() < 5
+        || cfg_dumbpass
+        || (to_move == FastBoard::WHITE ? -state.final_score() : state.final_score()) > 0) {
```
gcp: You can also move this computation out into a named variable if you nest the ifs.

training/elf/elf_convert.py
```diff
@@ -41,7 +41,10 @@ def write_block(f, b):
 with open('elf_converted_weights.txt', 'w') as f:
     # version 2 means value head is for black, not for side to move
     f.write('2\n')
-    b = convert_block(state, 'init_conv')
+    if 'init_conv.0.weight' in state:
```
gcp: You need to rebase.

src/UCTNode.cpp
gjm11: I wanted to keep my change as self-contained as possible, but I think the Right Way to do this is a small reorganization of the code earlier headed by the comment "DCNN returns winrate as side to move", which _already has_ the thing you're suggesting pulling out but throws it away. I'll do that.

src/UCTNode.cpp
gjm11: There's some actual explanation there as well, but I agree we can get by with substantially less.

training/elf/elf_convert.py
gjm11: Evidently. Sorry, I'm rather a git novice. I've pushed a change that addresses the other review comments; I'll fix this one once I'm confident of not breaking everything when I try :-).

src/UCTNode.cpp
```diff
@@ -92,8 +93,20 @@ bool UCTNode::create_children(Network & network,
...
     }
-    nodelist.emplace_back(raw_netlist.policy_pass, FastBoard::PASS);
-    legal_sum += raw_netlist.policy_pass;
+
+    // Don't consider pass moves if they would be immediately losing and we "shouldn't"
+    // accept that; see issue #2273. --dumbpass disables this heuristic.
+    float final_score_for_player = state.final_score();
+    if (to_move == FastBoard::WHITE) final_score_for_player = -final_score_for_player;
+    bool suppress_passes =
+        winrate_for_player >= 0.75     // we should be winning
```
Ttl: Is there any reason to check the winrate now that number of pass condition was removed? I don't see the point in allowing pass at low winrates if it's losing.

src/UCTNode.cpp
```diff
@@ -92,8 +93,20 @@ bool UCTNode::create_children(Network & network,
             legal_sum += raw_netlist.policy[i];
         }
     }
-    nodelist.emplace_back(raw_netlist.policy_pass, FastBoard::PASS);
-    legal_sum += raw_netlist.policy_pass;
+
+    // Don't consider pass moves if they would be immediately losing and we "shouldn't"
+    // accept that; see issue #2273. --dumbpass disables this heuristic.
+    float final_score_for_player = state.final_score();
```
Ttl: Calculating the score could be done only after other checks have passed. Even though it's cheap compared to NN evaluation it's much more expensive than the other checks. Having to count the score at every node can affect performance with fast GPU and small net.

training/elf/elf_convert.py
gjm11: I _think_ I've fixed this now. But, again, git novice, may have broken everything without noticing.

src/UCTNode.cpp
gjm11: It's common for a human game to end with two consecutive passes even though one of the players knows that they're going to lose. I confess I haven't checked how often that actually happens in LZ games rather than LZ resigning once it's sure it's lost. I think I've seen no-resign selfplay games, at least, end with two consecutive passes.

I'm happy to defer to others who actually know what they're doing here. Would anything bad happen if LZ literally never passed when ending the game would result in a loss?

src/UCTNode.cpp
gjm11: Ah, yes, you're right. The score check _was_ done at the end in an earlier version of this PR, but then I pulled it out into a separate variable for clarity without thinking carefully enough about the implications.

So I think the right way to do this is for FastState to grow _either_ a new method like `final_score` except that it returns the score from the current player's perspective _or_ a new argument to `final_score` that lets the caller decide whether they want that. I think the former is slightly better and will do that, but I am open to persuasion.

gjm11: I changed my mind about the right way to do it. `final_score` can now be called with no args (meaning the same as it always did) or with one arg (meaning you tell it which player's perspective you want). I think this is clearer than e.g. making up some name meaning "final score from the perspective of the current player".

roy7:
Humans often double pass to end, even the losing player, but LZ never passes into a loss. LZ will resign under the -r setting, but play forever otherwise.

src/UCTNode.cpp
Ttl: Passing loses instantly, but any other move would have possibly non-zero chance to win. In self-play I don't think the losing player ever passes second, but it can pass first due to tree search averaging giving pass greater than zero winning chance. Self-play uses dumbpass option so this wouldn't affect it.

Removing this condition should be beneficial for handicap play since no playouts are spent for unnecessary passing moves. In match play resign should be enabled and the engine would resign before passing is relevant.

gjm11: The winrate check does serve one other possibly useful purpose: it avoids bothering with the final-score check in some cases (i.e., those where the player to move is not comfortably ahead). So if the cost of the final-score check isn't entirely negligible -- as in another comment you suggested it might be -- then the winrate check brings some benefit by skipping roughly half the final-score checks (if one player is comfortably ahead) or most of them (if the position is roughly equal).

On the other hand, if there are many nodes where winrate isn't high, the policy on passing is highish, but passing would lose, then the search will be a bit more efficient for the reason you mention.

It's not obvious to me how these considerations balance out. I suspect there's very little difference overall.

src/UCTNode.cpp
```diff
@@ -103,8 +104,18 @@ bool UCTNode::create_children(Network & network,
...
-    nodelist.emplace_back(raw_netlist.policy_pass, FastBoard::PASS);
-    legal_sum += raw_netlist.policy_pass;
+
+    // Don't consider pass moves if they would be immediately losing and we "shouldn't"
+    // accept that; see issue #2273. --dumbpass disables this heuristic.
+    bool suppress_passes =
+        winrate_for_player >= 0.75         // we should be winning
+        && nodelist.size() >= 5            // there are surely good alternatives
+        && !cfg_dumbpass                   // allowed to be clever about passing
+        && state.final_score(to_move) < 0; // passing => immediate loss
```
alreadydone: Suggest `<=0` as we want to support draws.

gjm11: Yeah, probably. I'll make that change. Some overengineering bit of my brain wants to consider whether we want a higher winrate threshold for drawn-if-passed positions than for lost-if-passed ones, but I think that's a wrinkle that can be added later in the unlikely event that it ever makes a difference.

gjm11:
I'm not sure whether there's more I'm supposed to be doing at this point. (E.g., is the protocol around here that I'm supposed to "resolve" review comments when I think they've been addressed, or not?) The only thing that looks to me like an open question is whether to keep the winrate condition or not (see discussion above with Ttl); my guess is that it doesn't make very much difference one way or the other.

gcp:
>Is there any reason to check the winrate now that number of pass condition was removed? I don't see the point in allowing pass at low winrates if it's losing.

I'm wondering the same thing. It seems that the only condition we care about is what happens on the score after passing.

>The winrate check does serve one other possibly useful purpose: it avoids bothering with the final-score check in some cases (i.e., those where the player to move is not comfortably ahead). 

Seems like the sensible checks are then a) are enough legal moves left that we don't end up killing seki groups b) is *anyone* winning by a large enough margin that one side can afford to pass?

Note that (b) makes the score counting almost free as it's only done in tight situations.

gcp:
``
    bool suppress_passes =
        winrate_for_player >= 0.75          // we should be winning
        && nodelist.size() >= 5             // there are surely good alternatives
        && !cfg_dumbpass                    // allowed to be clever about passing
        && state.final_score(to_move) <= 0; // passing => immediate not-win
    if (!suppress_passes) {
``

So we allow passing if a) we are not winning b) there's very few good moves c) we win on the board count.

I have problems imagining situations where a+b+c would *ever* be true.

src/UCTNode.cpp
```diff
@@ -83,13 +83,14 @@ bool UCTNode::create_children(Network & network,
...
     // DCNN returns winrate as side to move
-    m_net_eval = raw_netlist.winrate;
-    const auto to_move = state.board.get_to_move();
     // our search functions evaluate from black's point of view
-    if (state.board.white_to_move()) {
-        m_net_eval = 1.0f - m_net_eval;
+    float winrate_for_player = raw_netlist.winrate;
+    const auto to_move = state.board.get_to_move();
+    if (to_move == FastBoard::WHITE) {
+      m_net_eval = eval = 1.0f - winrate_for_player;
```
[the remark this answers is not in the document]
gcp: Oh oops that seems to be a parameter that is now superfluous.

gcp:
@gjm11 @Ttl 

I rewrote this into a compact patch with logic that seems saner to me: only try passes if we're a) net-winning and (after checking) winning on the board or b) running out of legal moves.

Can you take a look?

gjm11:
@gcp Will take a look at your revised version shortly. I think your "can't imagine how it would ever be true" comment is incorrect: we allow passing if we are not winning OR there are very few legal moves OR we would win if the game were counted now. (Not AND.)

gjm11:
So, GCP's proposal differs from mine in a few ways. (I'm focusing on the logic of it; there are also differences in how the code is organized.)

First and least interestingly, we have some slightly different thresholds: 0.75 versus 0.80 for the eval, and 5 versus 20 legal non-passing moves. I don't think these matter one way or the other, and would trust GCP's intuition over mine for them in any case.

Second, the corner-case very-few-moves situations. If we're in such a situation, I say "always allow passing" and GCP says "allow passing if `final_score` says we wouldn't lose here".

These corner cases can only arise in weird endgame positions where the board is almost completely filled. My concern was that it might happen that the player's only legal non-passing moves are crazy ones -- e.g., self-atari or filling a group's last eye -- in which case we don't want passing to be forbidden. And, even more extremely, that it's possible to have a position where you have literally no legal moves other than passing, and we don't want to find ourselves with an empty list of candidate moves then.

In the second (admittedly kinda absurd) case we _mustn't_ forbid passing, but GCP's proposal does. In the first, my feeling is that we should allow passing because playing a move will only make things _even worse_. At the moment LZ doesn't care whether it loses by 0.5 points or >200, but it's not impossible that some future version might, and then in these weird situations the passing constraint will hurt. So in both cases I prefer my proposal. Also, though I doubt it matters, in these cases my proposal avoids needing to call `final_score` so it saves a small amount of work.

None of this should matter much -- these are very strange positions we're talking about -- but no-resign games do sometimes get into very strange positions :-).

Third, ignoring the differences already mentioned, I think the difference between GCP's proposal and mine is that when the eval is _not_ good, we both propose not bothering to check `final_score` ... but GCP says "never allow passing" and I say "always allow passing".

I don't think this matters much in practice. Recall that the point here isn't to stop LZ _actually passing_ when playing a game (there is other logic that handles that) but to stop MCTS evals being polluted by considering stupid pass-moves. I believe the NN's policy on passing is only non-negligible in "extreme" positions (and maybe also when very few moves are available, but that's covered by the discussion above). The eval-pollution from a losing pass-move that makes the winrate 0 instead of 0.01 is negligible.

I think GCP's proposal is _slightly better strategically_ than mine (if it ever happens that the NN puts non-negligible policy weight on passing in an even-ish position then with my proposal the MCTS evals will get polluted and with GCP's they won't) and _slightly less Zero-ish_ than mine (in cases where passing is less blatantly crazy because it doesn't involve passing up a very likely win, my proposal leaves the search unaltered and GCP's prunes the pass move).

There might be a different in _training effects_ (unless "dumbpass" is, and will be, always on for training games, in which case obviously it doesn't matter). If I've correctly understood how our training is done (I might well not have) GCP's proposal will make training more strongly push _almost all_ pass moves' policy towards zero. (Because we train the policy network on how much moves get explored in search.) I doubt this matters much, but I have a handwavy intuition that it is better to actually _prefer_ passing once everything's resolved, both so that no-resign games have a chance of actually finishing in finite time and to reduce the danger that LZ learns to fill in its territory and every now and then fills in a bit too much and kills something that used to be alive.

If we train without "dumbpass", for me those effects suffice to make me prefer my policy to GCP's. If not, meh, whichever.

(As to whether GCP's rewrite is an improvement in other respects -- obviously it would be easy to adjust my patch to use his logic, or his to use mine -- I confess I don't see that it is, and in particular I think my extension to `final_score` makes it easier and clearer to use, but it's GCP's codebase not mine.)

src/UCTNode.cpp
```diff
@@ -103,8 +104,18 @@ bool UCTNode::create_children(Network & network,
             legal_sum += raw_netlist.policy[i];
         }
     }
-    nodelist.emplace_back(raw_netlist.policy_pass, FastBoard::PASS);
-    legal_sum += raw_netlist.policy_pass;
+
+    // Don't consider pass moves if they would be immediately losing and we "shouldn't"
+    // accept that; see issue #2273. --dumbpass disables this heuristic.
```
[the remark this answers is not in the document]
gjm11: By "the latter part" you mean "see issue #2273"?

If you've got the code open in your text editor or IDE and are wondering what it's for then it's an order of magnitude less work to see "see issue #2273" in the comments than to run "git blame". Especially if e.g. at some point in the future this bit of code has been touched by some more superficial modifications, e.g. a change of indentation.

I find that one of the most common questions I ask myself when reading code is "_why_ is this the way it is?". In this case, only an exceptionally astute reader will have any chance of figuring out that the issue being addressed is MCTS eval pollution by crazy passes; it's not important enough to be worth having a full explanation in comments -- I think -- but a few words showing where to find a full explanation seem well worth the space they take up, to me.

gcp:
Training never uses any of this logic, that's what the dumbpass flag is for.

gcp:
>I think my extension to final_score makes it easier and clearer to use, but it's GCP's codebase not mine

I don't necessarily disagree, but that change is orthogonal to this patch.

gcp:
>Second, the corner-case very-few-moves situations. If we're in such a situation, I say "always allow passing" and GCP says "allow passing if final_score says we wouldn't lose here".

I think it's a good point that if there's few legal moves, we probably always want to allow passing regardless of board score. Obviously we'll never win but it can also avoid games from going on and on while the losing player self-destructs all his groups.

See e.g.: https://github.com/leela-zero/leela-zero/pull/2301#issuecomment-476275326 but the logic to do it entirely right quickly gets larger.

I'll do a follow-up to always allow pass if there's few remaining moves.

PR #2290, opened 2019-03-19
Pick chosen move based on normal distribution LCB

Ttl:
Similar to `lcb_max_root` branch by @roy7, but uses normal distribution instead of binomial.

Some tests:
```
15x192 with 1600 visits:
"-g -d -r 5 -v 1600 --noponder -w f438268ef88e083aaf7a08fba885552818abc08cd0aefce9671954a8df3cf707.gz --precision half --timemanage off", "time_settings 0 1 0".

lz_gauss v lz_next (400/400 games)
board size: 19   komi: 7.5
lz_gauss    268 67.00%       118 59.00%     150 75.00%     81.32
lz_next     132 33.00%       50  25.00%     82  41.00%     81.26
                             168 42.00%     232 58.00%
```

```
15x192 with 800 visits:
"g -d -r 5 -v 800 --noponder -w 438268ef88e083aaf7a08fba885552818abc08cd0aefce9671954a8df3cf707.gz --precision half --timemanage off" "time_settings 0 1 0".

lz_gauss v lz_next (400/400 games)
board size: 19   komi: 7.5
lz_gauss    294 73.50%       136 68.00%     158 79.00%     35.05
lz_next     106 26.50%       42  21.00%     64  32.00%     35.13
                             178 44.50%     222 55.50%

```

```
15x192 with 30+1s on GTX 1080 Ti:
"-g -d -r 5 --noponder -w f438268ef88e083aaf7a08fba885552818abc08cd0aefce9671954a8df3cf707.gz --precision half", "time_settings 30 1 1".

lz_gauss v lz_next (400/400 games)
board size: 19   komi: 7.5
lz_gauss    250 62.50%       115 57.50%     135 67.50%     42.16
lz_next     150 37.50%       65  32.50%     85  42.50%     42.16
                             180 45.00%     220 55.00%
```

```
Latest 40x256 with 120+2s on RTX 2080:
"-g -d -r 5 --noponder -w e5dd6019c73a853466abb2fb9cf502d80d85179619f8ca76676f4eec16a13468.gz --precision half", "time_settings 180 2 1".

lz_gauss v lz_next (400/400 games)
board size: 19   komi: 7.5
lz_gauss    234 58.50%       97  48.50%     137 68.50%    312.18
lz_next     166 41.50%       63  31.50%     103 51.50%    288.99
                             160 40.00%     240 60.00%
```

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -330,7 +330,8 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
 class NodeComp : public std::binary_function<UCTNodePointer&,
                                              UCTNodePointer&, bool> {
 public:
-    NodeComp(int color) : m_color(color) {};
+    NodeComp(int color, float lcb_min_visits) : m_color(color),
+        m_lcb_min_visits(lcb_min_visits){};
 
     // WARNING : on very unusual cases this can be called on multithread
     // contexts (e.g., UCTSearch::get_pv()) so beware of race conditions
@@ -340,7 +341,7 @@ public:
         auto b_visit = b.get_visits();
 
         // Calculate the lower confidence bound for each node.
-        if (a_visit && b_visit) {
+        if ((a_visit > m_lcb_min_visits) && (b_visit > m_lcb_min_visits)) {
             auto a_lcb = a.get_eval_lcb(m_color);
             auto b_lcb = b.get_eval_lcb(m_color);
 
@@ -365,10 +366,11 @@ public:
     }
 private:
     int m_color;
+    float m_lcb_min_visits;
 };
 
-void UCTNode::sort_children(int color) {
-    std::stable_sort(rbegin(m_children), rend(m_children), NodeComp(color));
+void UCTNode::sort_children(int color, float lcb_min_visits) {
+    std::stable_sort(rbegin(m_children), rend(m_children), NodeComp(color, lcb_min_visits));
 }
 
 UCTNode& UCTNode::get_best_root_child(int color) {
@@ -376,8 +378,13 @@ UCTNode& UCTNode::get_best_root_child(int color) {
 
     assert(!m_children.empty());
 
+    auto max_visits = 0;
+    for (const auto& node : m_children) {
+        max_visits = std::max(max_visits, node.get_visits());
+    }
+
     auto ret = std::max_element(begin(m_children), end(m_children),
-                                NodeComp(color));
+                                NodeComp(color, cfg_lcb_min_visit_ratio * max_visits));
     ret->inflate();
 
     return *(ret->get());
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -227,7 +227,8 @@ float UCTNode::get_eval_lcb(int color) const {
     // Lower confidence bound of winrate.
     auto visits = get_visits();
     if (visits < 2) {
-        return 0.0f;
+        // Return large negative value if not enough visits.
+        return -1e6f + visits;
     }
     auto mean = get_raw_eval(color);
 
@@ -339,6 +340,11 @@ public:
                     const UCTNodePointer& b) {
         auto a_visit = a.get_visits();
         auto b_visit = b.get_visits();
+
+        // Need at least 2 visits for LCB.
+        if (m_lcb_min_visits < 2) {
+            m_lcb_min_visits = 2;
+        }
 
         // Calculate the lower confidence bound for each node.
         if ((a_visit > m_lcb_min_visits) && (b_visit > m_lcb_min_visits)) {
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -232,7 +232,15 @@ float UCTNode::get_lcb(int color) const {
     if (visits < 2) {
         return mean - 1.0f;
     }
-    float stddev = get_stddev(1.0f) / std::sqrt(visits);
+
+    // Value head output is correlated, if a position is misevaluated it's
+    // likely for subsequent moves to also be misevaluated.
+    // We overestimate the confidence bounds if we don't correct for it.
+    // The problem is that we don't really know how it's correlated.
+    // Just empirically make the bound decrease slower by not counting full
+    // visits.
+    // Maybe this should apply to z calculation too?
+    float stddev = get_stddev(1.0f) / std::sqrt(cfg_value_correlation * visits);
 
     boost::math::students_t dist(visits - 1);
     auto z = boost::math::quantile(boost::math::complement(dist, cfg_ci_alpha));
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -180,8 +180,12 @@ void UCTNode::virtual_loss_undo() {
 }
 
 void UCTNode::update(float eval) {
+    // FIXME: possible race issues
+    float old_delta = m_visits > 0 ? eval - m_blackevals / m_visits : 0.0f;
     m_visits++;
     accumulate_eval(eval);
+    float new_delta = eval - m_blackevals / m_visits;
+    m_squared_diff = m_squared_diff + old_delta * new_delta;
 }
 
 bool UCTNode::has_children() const {
@@ -205,6 +209,10 @@ float UCTNode::get_policy() const {
 
 void UCTNode::set_policy(float policy) {
     m_policy = policy;
+}
+
+float UCTNode::get_variance() const {
+    return m_squared_diff / (m_visits + 1);
 }
 
 int UCTNode::get_visits() const {
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -40,6 +40,7 @@
 #include <numeric>
 #include <utility>
 #include <vector>
+#include <boost/math/distributions/students_t.hpp>
 
 #include "UCTNode.h"
 #include "FastBoard.h"
@@ -224,14 +225,6 @@ int UCTNode::get_visits() const {
     return m_visits;
 }
 
-float approx_t(float v, float z) {
-    // Fast approximation to inverse CDF of student-t distribution.
-    if (v > 100) {
-        return z;
-    }
-    return z + 26.0f / v + 3.50f / (v * v * v);
-}
-
 float UCTNode::get_lcb(int color) const {
     // Lower confidence bound of winrate.
     int visits = get_visits();
@@ -240,7 +233,11 @@ float UCTNode::get_lcb(int color) const {
         return mean - 1.0f;
     }
     float stddev = get_stddev(1.0f) / std::sqrt(visits);
-    return mean - approx_t(visits - 1, cfg_conf_z) * stddev;
+
+    boost::math::students_t dist(visits - 1);
+    auto z = boost::math::quantile(boost::math::complement(dist, cfg_ci_alpha));
+
+    return mean - z * stddev;
 }
 
 float UCTNode::get_raw_eval(int tomove, int virtual_loss) const {
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -180,11 +180,12 @@ void UCTNode::virtual_loss_undo() {
 }
 
 void UCTNode::update(float eval) {
-    // FIXME: possible race issues
+    // FIXME: possible race issues.
     float old_delta = m_visits > 0 ? eval - m_blackevals / m_visits : 0.0f;
     m_visits++;
     accumulate_eval(eval);
     float new_delta = eval - m_blackevals / m_visits;
+    // See Welford's online algorithm for calculating variance for explanation.
     m_squared_diff = m_squared_diff + old_delta * new_delta;
 }
 
@@ -211,12 +212,35 @@ void UCTNode::set_policy(float policy) {
     m_policy = policy;
 }
 
-float UCTNode::get_variance() const {
-    return m_squared_diff / (m_visits + 1);
+float UCTNode::get_variance(float default_var) const {
+    return m_visits > 1 ? m_squared_diff / (m_visits - 1) : default_var;
+}
+
+float UCTNode::get_stddev(float default_stddev) const {
+    return m_visits > 1 ? std::sqrt(get_variance()) : default_stddev;
 }
 
 int UCTNode::get_visits() const {
     return m_visits;
+}
+
+float approx_t(float v, float z) {
+    // Fast approximation to inverse CDF of student-t distribution.
+    if (v > 100) {
+        return z;
+    }
+    return z + 26.0f / v + 3.50f / (v * v * v);
+}
+
+float UCTNode::get_lcb(int color) const {
+    // Lower confidence bound of winrate.
+    int visits = get_visits();
+    float mean = get_raw_eval(color);
+    if (visits < 2) {
+        return mean - 1.0f;
+    }
+    float stddev = get_stddev(1.0f) / std::sqrt(visits);
+    return mean - approx_t(visits - 1, cfg_conf_z) * stddev;
 }
 
 float UCTNode::get_raw_eval(int tomove, int virtual_loss) const {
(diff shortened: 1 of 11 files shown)


l1t1:
is the lcb method the unique method or one option to choose move after merge the pr?

roy7:
This would make the change for all play, not set by an option.

l1t1:
it would be better to test on low visits

src/UCTNode.cpp
```diff
@@ -207,10 +215,32 @@ void UCTNode::set_policy(float policy) {
     m_policy = policy;
 }
 
+float UCTNode::get_variance(float default_var) const {
```
Hersmunch: Suggest making it clear what the variance is of. E.g. get_eval_variance()

src/UCTNode.cpp
```diff
@@ -207,10 +215,32 @@ void UCTNode::set_policy(float policy) {
...
+    return m_visits > 1 ? m_squared_diff / (m_visits - 1) : default_var;
+}
+
+float UCTNode::get_stddev(float default_stddev) const {
```
Hersmunch: get_eval_stddev()?

src/UCTNode.cpp
```diff
@@ -207,10 +215,32 @@ void UCTNode::set_policy(float policy) {
...
+    return m_visits > 1 ? std::sqrt(get_variance()) : default_stddev;
+}
+
 int UCTNode::get_visits() const {
     return m_visits;
 }
 
+float UCTNode::get_lcb(int color) const {
+    // Lower confidence bound of winrate.
```
Hersmunch: No need for the comment if named get_winrate_lcb()

src/UCTNode.cpp
```diff
@@ -207,10 +215,32 @@ void UCTNode::set_policy(float policy) {
...
+    auto visits = get_visits();
+    if (visits < 2) {
+        return 0.0f;
+    }
+    auto mean = get_raw_eval(color);
+
+    auto stddev = std::sqrt(get_variance(1.0f) / visits);
```
Hersmunch: Does it matter if the visits used to divide by is not the same as m_visits used within get_variance() due to a different thread incrementing m_visits?

src/UCTNode.cpp
```diff
@@ -313,6 +343,17 @@ class NodeComp : public std::binary_function<UCTNodePointer&,
         auto a_visit = a.get_visits();
         auto b_visit = b.get_visits();
 
+        // Calculate the lower confidence bound for each node.
+        if (a_visit && b_visit) {
+            auto a_lcb = a.get_lcb(m_color);
+            auto b_lcb = b.get_lcb(m_color);
```
Hersmunch: Similarly, does it matter that the number of visits used to calculate the lcbs could be different to a_visits/b_visit?
I know we’ve had a similar race condition issue here previously.

src/UCTNode.cpp
```diff
@@ -313,6 +343,17 @@ class NodeComp : public std::binary_function<UCTNodePointer&,
...
+
+            // Sort on lower confidence bounds
+            if (a_lcb != b_lcb) {
+                return a_lcb < b_lcb;
+            }
+}
```
Hersmunch: Indentation

src/UCTNode.h
```diff
@@ -122,6 +125,9 @@ class UCTNode {
     float m_policy;
     // Original net eval for this node (not children).
     float m_net_eval{0.0f};
+    // Initialize to prior of variance. Avoids accidental zero variances
+    // at low visits.
+    std::atomic<float> m_squared_diff{0.01f};
```
Hersmunch: Perhaps m_eval_squared_diff?

src/UCTNodePointer.cpp
```diff
@@ -137,6 +137,18 @@ float UCTNodePointer::get_policy() const {
     return read_policy(v);
 }
 
+float UCTNodePointer::get_stddev(float default_stddev) const {
```
Hersmunch: Similar name suggestion for clarity.

src/UCTSearch.cpp
```diff
@@ -615,28 +616,39 @@ int UCTSearch::est_playouts_left(int elapsed_centis, int time_for_move) const {
...
     // during a multithreaded search, so it is safe to walk it here without
     // taking the (root) node lock.
     for (const auto& node : m_root->get_children()) {
         if (node->valid()) {
-            Nfirst = std::max(Nfirst, node->get_visits());
+            const auto visits = node->get_visits();
+            if (visits > 0) {
+                lcb_max = std::max(lcb_max, node->get_lcb(color));
+            }
+            Nfirst = std::max(Nfirst, visits);
```
Hersmunch: Could move this within the if case above.

src/Utils.cpp
```diff
@@ -49,6 +50,27 @@
...
+    }
+}
+
+float Utils::cached_t_quantile(int v) {
+    if (v < 1) {
+        return z_lookup[0];
+    }
+    if (v < z_entries) {
+        return z_lookup[v - 1];
+    } else {
```
Hersmunch: No need for the else

src/UCTNode.cpp
roy7: I tried figuring out atomic structs last year to avoid this concern but wasn't sure exactly how to do it. Discussion was in this thread. https://github.com/leela-zero/leela-zero/issues/1431#issuecomment-389603386

From exploring google at the time, it appears you could make a struct atomic and create a helper function for it that would do all of updating/increments needed as a single atomic operation. But I just couldn't quite figure it out on my own, and feedback in the thread was against doing so. I ended up just throwing a lock into each of these functions, but later on pretty much all locks were removed. To stay in sync with the new lockless setup, I decided to just delete all of my Normal supporting code since I was concerned about the threading.

roy7: An interesting idea from here, if you needed for example two integers, is to pack them both into a long int and separate them/combine them as needed. This way it's one a single atomic action to get/set "two" integers. Our situation is a bit more complex though. Using a mutex is probably too slow.

https://stackoverflow.com/questions/27247973/best-way-to-atomically-update-two-members-of-a-struct

src/UCTNode.cpp
Ttl: Removed the `get_stddev` function instead since it wasn't used.

src/UCTNode.cpp
roy7: Some discussion of atomic structs.

https://developers.redhat.com/blog/2016/01/14/toward-a-better-use-of-c11-atomics-part-1/

https://preshing.com/20150402/you-can-do-any-kind-of-atomic-read-modify-write-operation/

src/UCTNode.cpp
Ttl: I left the comment there to explain what lcb means.

src/UCTNode.cpp
Ttl: It slightly overestimates the error if visits is increased by other thread. Not really a problem, overestimating the error is better than underestimating it.

src/UCTNode.cpp
Ttl: I don't think it's an issue. a_visit and b_visit variables are just checked that they are both bigger than zero. lcb check isn't really any different from other checks in the function.

src/UCTSearch.cpp
Ttl: I ended up leaving this as it was, but leave a comment if you really disagree.

src/UCTNode.cpp
```diff
@@ -215,27 +215,23 @@ void UCTNode::set_policy(float policy) {
     m_policy = policy;
 }
 
-float UCTNode::get_variance(float default_var) const {
-    return m_visits > 1 ? m_squared_diff / (m_visits - 1) : default_var;
-}
-
-float UCTNode::get_stddev(float default_stddev) const {
-    return m_visits > 1 ? std::sqrt(get_variance()) : default_stddev;
+float UCTNode::get_eval_variance(float default_var) const {
```
OmnipotentEntity: Have you considered using a `std::optional` here?  That way you don't need the `default_var` parameter.  You can just use `.value_or(0)`, this would also allow you to distinguish between cases where the variance is exactly the same as the passed `default_var` vs when the variance is merely undefined and taking the `default_var`.

Ttl: I don't think the default is even used anywhere right now. `get_eval_variance` is only needed in `get_eval_lcb` that only calls it when node has at least two visits. Probably not much point in using optional when it's never needed.

Ttl:
I collected some stats on how often move with highest LCB doesn't have also the most visits and it seems that the answer is around 10% of time:

highest_move_lcb

At very low nodes the uncertainty is so big that the move with the best LCB also very likely has also the most visits.

I also logged the ratio of highest LCB move visits to move with the most visits and got the following plot:

lcb_ratio_0


There seems to be an unnatural looking spike at very low visits. I suspect the reason for that is that the confidence bounds are underestimated. The reason being is that the evaluation distribution isn't really a normal distribution. If NN misevaluates a position there is a high probability that it also misevaluates nearby positions making the samples correlated. The distribution can also change if during tree search some better option is found deeper in the tree. This violates the i.i.d assumption making the estimated confidence bounds too tight. I have plans to find a better formula for the confidence bounds based on measurements, but that might take a while.

I added a parameter for enforcing a minimum visit ratio of the chosen move. I set it to 10% for now based on the above plot. After the change and with more samples the ratio distribution visit distribution looks like this:

lcb_ratio_after

Friday9i:
Nice.
If someone has a windows .exe, I can generate comparison matches at various visits against a recently compiled next LZ : -)

langit7:
> Nice.
> If someone has a windows .exe, I can generate comparison matches at various visits against a recently compiled next LZ : -)

http://fira.pw/lz/leelazX64.zip

iopq:
Wouldn't there be the same problem just one ply deeper? Our variation will have the max LCB because the "obvious" policy move didn't refute our variation. It's refuted twenty moves later, but we never visit that line again because it's buried in win rates of other branches

Don't we need to choose the correct refutation by doing min UCB for our opponent? (In other words we want the max black score, our opponent wants the min black score)

OmnipotentEntity:
@iopq AFAIK, this PR does not affect the tree search, only the chosen move.  (Though, all evaluations are done from the perspective of the player to move, so a max LCB from black's perspective is a min UCB from white's, no?)

iopq:
It doesn't have to be.

Let's say the best black move is putting opponent in atari. Now, after a lot of searching, we find that it has the best win rate. The reason is that if the opponent doesn't run his group out, we win big. Now, if he does run his group out, his win rate increases. But we can give ANOTHER atari and still increase our win rate, etc.

After a few forced moves if our opponent responds correctly he might have 55% max LCCB win rate. But in ALL of the other variations he'll have like 30%, so when we average his value for this node it looks like a winner. Let's call this max LCB node.

But if we look at the visits for our opponent's node, it might ALREADY have 50% min UCB for the correct response, just not total overall. It makes sense to assign our max UCB as the min UCB of our opponent, as long as the cutoff for visits is met (10% right now)

If that works, you can do this iteratively until you no longer meet the cut off 10% (max LCB, min UCB, max LCB, etc.)

then each node's score would actually be dependent on things deeper in the tree for stronger tactical fighting

Ttl:
The latest commit should fix the issue with incorrect sorting when there are very few visits.

src/UCTNode.cpp
gcp: std::optional is C++17 btw so not usable for now.

gcp:
Seems to be a win in every situation, including fixed time with time-manage on (IIRC a point where it previously failed)?

Should I pull it for 0.17?

roy7:
Using the normal does seem stronger than using the Binomial, and @Ttl did a cleaner job than my test branch. And he even avoids pruning the max_lcb move during timemanage calculations. :)

This won't hurt training/selfplay in any way right, since the only change is to actual move selection and that just means the resulting selfplay games have stronger moves overall.

src/UCTSearch.cpp
```diff
@@ -58,9 +58,11 @@ constexpr int UCTSearch::UNLIMITED_PLAYOUTS;
...
-                       float winrate, float policy_prior, std::string pv)
+                       float winrate, float policy_prior, std::string pv,
+                       float lcb, bool lcb_ratio_exceeded)
     : m_move(move), m_visits(visits), m_winrate(winrate),
-      m_policy_prior(policy_prior), m_pv(pv) {};
+      m_policy_prior(policy_prior), m_pv(pv), m_lcb(lcb),
+      m_lcb_ratio_exceeded(lcb_ratio_exceeded) {};
 
     std::string get_info_string(int order) const {
         auto tmp = "info move " + m_move
```
gcp: I think you want to add "lcb" here. Without it it's also going to be confusing why the engine plays the "wrong" move.

(After it, it will still be confusing because of the min_visits condition, but that seems unfixable)

src/UCTNode.cpp
gcp: You can use a signaling NaN to let things blow up if it's accidentally used.

src/UCTNode.h
```diff
@@ -122,6 +124,10 @@ class UCTNode {
     float m_policy;
     // Original net eval for this node (not children).
     float m_net_eval{0.0f};
+    // Variable used for calculating variance of evaluations.
+    // Initialized to small non-zero value to avoid accidental zero variances
+    // at low visits.
+    std::atomic<float> m_squared_eval_diff{0.01f};
```
gcp: This seems like a pretty high starting value considering examples of say 2 visits with 51% vs 52% winrates?

src/Utils.h
```diff
@@ -40,6 +40,8 @@
 
 extern Utils::ThreadPool thread_pool;
 
+auto constexpr z_entries = 1000;
```
gcp: z_entries is an implementation detail and could go in the .cpp file.

gcp:
>This won't hurt training/selfplay in any way right, since the only change is to actual move selection and that just means the resulting selfplay games have stronger moves overall.

Right.

The amount of gain this gives is signaling that our primitive time allocation could also use a lot of improvement by detecting max_lcb != max_winrate, but that wouldn't help fixed time. I think we can and should do both, eventually.

src/UCTNode.h
Ttl: This isn't that important anymore with minimum visit ratio. I set it to a smaller value, but it could probably be even set to 0 now.

gcp:
LGTM.

src/UCTNode.cpp
Ttl: Signaling NaN didn't seem to do anything. It seems that it would require changing how floating point exceptions are handled for it to raise exception. Didn't want to go messing with that.

Ttl:
I'm not sure if this is better at very high playouts, but probably not worse either. I ran some tests with 192x15 network and 180+5s time controls on GTX 1080 Ti. This is around 10k playouts / move.

```
lz_gauss v lz_next (99/400 games)
board size: 19   komi: 7.5
lz_gauss     50 50.51%       18 36.00%     32 65.31%    852.98
lz_next      49 49.49%       17 34.69%     32 64.00%    797.97
                             35 35.35%     64 64.65%
```

I think that the normal distribution confidence bounds are too tight at very high playouts, but it's not very easy to figure out how it should be modified. Testing with high playouts takes too long.

I guess it's okay to merge since it's clearly better at more common visit counts.

l1t1:
please post the full comand line of validation.exe of two versions leelaz.exe
@Ttl

iopq:
I ran a 400 game match on 9x9 with a strong 9x9 network with 7.5 komi at 1600 visits. In non-duplicate games:

LCB won with Black 7 times, with White 36 times. Standard Leela won with Black 7 times, with White 71 times.

Total record LCB-standard is 208-192

l1t1:
cp from lifein19x19 
C:\APPS\l0gpu16\validation.exe -n C:\APPS\net\35824222.gz -o "-g -v 1601 --gpu 0 --gpu 1 --noponder -t 24 -q -d --timemanage off --precision single -w" -n C:\APPS\net\35824222.gz -o "-g -v 1601 --gpu 0 --gpu 1 --noponder -t 24 -q -d --timemanage off --precision single -w" -- C:\APPS\l0gpu16\leelaz -- C:\APPS\l0gpu17beta\leelaz -k 215-215

roy7:
@Ttl Is https://github.com/leela-zero/leela-zero/blob/next/src/UCTNodePointer.cpp#L142 a bug that the is_inflated check is different than every other function in here?

I actually "fixed" that in the thompson_sampling branch but I'm not 100% certain it was a bug to fix. https://github.com/leela-zero/leela-zero/pull/2352

Ttl:
`is_inflated()` is equal to `is_inflated(m_data.load())` so it's functionally the same. It could cause the `m_data.load()` to be called twice though.

PR #2297, opened 2019-03-22
lagbuffer fix #2296

### LINKED ISSUE #2296: lagbuffer and TimeControl
The following applies to a client not sending constant updates on every move using the GTP command time_left, and does not affect pure byo-yomi time controls.

The current implementation for lag buffer makes LZ often lose on time when using Canadian time control (and for Absolute time too). This is mainly because **LZ does not take the lag into account to correct it's time remaining when a move is made.**

This makes LZ clock get out of sync with the client clock as more and more moves are made. This doesn't affect byo-yomi with zero main time, but it does affect Canadian and Absolute timing control.  Additionally, when LZ calculates the maximum move time it calculates using the lag over an entire period, but when the above is true, the lag really should add up for each move in a period; this results  in moves later in the period get less and less time than they should.

At the moment, with the current implementation the client either has to set a very large lagbuffer, or, try to guess and adjust the lagbuffer based on the number of stones.

dbosst:
Better time control when using lagbuffer.  This fixes a significant portion of problems with the time going out of sync with the client when using Canadian/Absolute timing.

Fixes the issues outlined in #2296.

### DIFF
--- a/src/TimeControl.cpp
+++ b/src/TimeControl.cpp
@@ -92,6 +92,9 @@ void TimeControl::stop(int color) {
 
     assert(elapsed_centis >= 0);
 
+    // because of GUI/network lag, cfg_lagbuffer_cs centisecond is lost on each move
+    elapsed_centis += cfg_lagbuffer_cs;
+
     m_remaining_time[color] -= elapsed_centis;
 
     if (m_inbyo[color]) {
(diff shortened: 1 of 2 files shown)


src/TimeControl.cpp
```diff
@@ -92,6 +92,9 @@ void TimeControl::stop(int color) {
 
     assert(elapsed_centis >= 0);
 
+    // because of GUI/network lag, cfg_lagbuffer_cs centisecond is lost on each move
+    elapsed_centis += cfg_lagbuffer_cs;
```
gcp: I'm not clear whether we should pull this. It indicates an interface that is completely broken.

src/TimeControl.cpp
```diff
@@ -192,10 +195,11 @@ int TimeControl::max_time_for_move(int boardsize,
...
+    // always keep a cfg_lagbuffer_cs centisecond margin
     // for network hiccups or GUI lag
-    auto base_time = std::max(time_remaining - cfg_lagbuffer_cs, 0) /
-                     std::max(moves_remaining, 1);
+    auto base_time = std::max(
+        (std::max(time_remaining, 0) / std::max(moves_remaining, 1)) -
+        cfg_lagbuffer_cs, 0);
     auto inc_time = std::max(extra_time_per_move - cfg_lagbuffer_cs, 0);
 
     return base_time + inc_time;
```
gcp: Note that with an increment, you're now subtracting the lagbuffer twice.

dbosst:
I'm closing this pull request since I suppose updating LZ with time_left is expected to be mandatory before sending genmove commands.

PR #2299, opened 2019-03-24
Fix qt5 minor clearer

wonderingabout:
continuation of #2071 

rebased on top of latest devel

as said in commit message, i dont have mac, so mac instructions 
may be inaccurate

but this PR adds a few meaningful changes like adding qt5 and cmake
dependencies on main readme page

also, autogtp readme is now only documentation : for compile help, 
referr to main readme page rather

@gcp @sethtroisi @Hersmunch @ihavnoid @marcocalignano @Ttl


wonderingabout:
meant to be be merged on next, will re PR

Issue #2275, opened 2019-03-12
0.17 release


gcp:
I think we should consider a 0.17 release with the Tensor Core support, batching, and the whole pile of bug-fixes and extensions to GTP support, as well as AutoGTP support for more experimentation.

Any reason why or why not? (If you have a massively important and useful pull you were about to request for example...)

If OK, I'll start working on it this week.

l1t1:
it is better to release a beta version first. after all, many people dont compile the next branch. they just use the binary version.

herazul:
Will it be possible with the release to have a little rundown (maybe in the readme ?) of what we know about how batching and tensor core impact performance, what to use and what corresponding options to use on what graphic cards, ...

roy7:
/next basically is the beta test, but if we are able to get nightly build going for Windows binaries that would be awesome. @bood set something up to give us the OpenBLAS binaries needed that are newer than NuGet's modules as I understand it. So we'd just need LZ in AppVeyor configured to copy over his binaries and build the program and push the latest binaries to GitHub?

gcp:
>what to use and what corresponding options to use on what graphic cards, ...

If the answer is not "don't do anything, use the defaults" then this is a serious bug that should be addressed first and foremost, not a burden to shift upon the user. 

I'm very strongly against shifting the burden of configuring for good performance to the user by exposing a bunch of settings (well, they're already exposed, I want to not even give the impression that this is something you have to touch) that they necessarily must tweak, with some risk they make the program weaker by not understanding how batching and threads influence search efficiency, etc. Note that documenting does nothing to stop that problem. The program should just work. And yes, you may get a tiny bit more speedup by finetuning provided you know *exactly* what you're doing, but there really shouldn't even be a temptation to do that. 

Currently if you launch with the defaults you'll get a nice speedup, something like 25% on mid-range cards, and a factor 2 or 3 with Tensor Cores.

And you don't need to do anything to get that.

herazul:
Yeah i get that, makes a lot of sense. So nothing is needed in the readme then =)

With cards with tensor cores support, is there a big difference in performance between middle-end and high-end cards ? (2060 vs 2080ti for example ?)

l1t1:
i suggest fix this thread at the top, and add a link of latest build link in the main post, such as https://ci.appveyor.com/project/zakki/ray/builds/21043153/artifacts

gjm11:
[...]

PR #2300, opened 2019-03-24
Fix qt5 minor clearer readme

wonderingabout:
continuation of #2071 

rebased on top of latest devel

as said in commit message, i dont have mac, so mac instructions 
may be inaccurate

but this PR adds a few meaningful changes like adding qt5 and cmake
dependencies on main readme page

also, autogtp readme is now only documentation : for compile help, 
referr to main readme page rather

@gcp @sethtroisi @Hersmunch @ihavnoid @marcocalignano @Ttl


wonderingabout:
this meant to hopefully be reviewed before 0.17 release in #2275

README.md
```diff
@@ -124,8 +129,16 @@ by adding -DUSE_CPU_ONLY=1 to the cmake command line.
...
 
-    # Install build depedencies
-    brew install boost cmake zlib
+    # Install leelaz build depedencies
+    brew install boost cmake zlib curl
+    
+    # Optional : if you want to contribute to the distributed effort, you need these dependencies as well
+    brew install qt5
+    # after qt5 install, then you also need to do the post install for mac, replace 5.x.x with your qt5 version
+    export QT5DIR=/usr/local/Cellar/qt/5.x.x
```
gcp: None of this should be needed aside from fixing the path.

autogtp/README.md
```diff
@@ -30,10 +30,42 @@ directly.
...
+# Help 
+
+For more details about AutoGTP, or for specific use (for example how to use multi GPU, 
+or any other setting), you can run, while in the directory where your autogtp executable is :
+- `./autogtp --help` on linux/mac
+- `autogtp.exe --help` on windows
+
+This will display a list of all possible settings and how to use them
+
+For your convenience, and for easier reference, a copy is provided below :
```
gcp: Please don't do this, or it has to be updated each time one of the command line options changes.

README.md
```diff
@@ -124,8 +129,16 @@ by adding -DUSE_CPU_ONLY=1 to the cmake command line.
     cd leela-zero
     git submodule update --init --recursive
 
-    # Install build depedencies
-    brew install boost cmake zlib
+    # Install leelaz build depedencies
+    brew install boost cmake zlib curl
```
gcp: curl is not a build dependency

README.md
```diff
@@ -93,6 +94,7 @@ If you do not have a GPU, add the define "USE_CPU_ONLY", for example
 by adding -DUSE_CPU_ONLY=1 to the cmake command line.
 * Optional: BLAS Library: OpenBLAS (libopenblas-dev) or Intel MKL
 * The program has been tested on Windows, Linux and macOS.
+* autogtp and leelaz binaries need to be in the same directory. The below instructions can compile leelaz, and optionally autogtp at the same time.
```
gcp: The instructions already do this, this serves no purpose.

autogtp/README.md
wonderingabout: this would add maintainance indeed, but i thought it would serve some purpose

ok then, will remove this :+1: 
we may link to specific github issues everytime it is needed rather then

README.md
wonderingabout: ok, will fix

README.md
wonderingabout: the purpose was in case end user somehow forgets to copy leelaz binary in same folder as autogtp binary

but it is right that if followed correctly, this information should not be needed, and may add confusion, will remove then

README.md
wonderingabout: @gcp 
i dont understand your explanation : 
- so only path export may be needed for mac ?
- qt is factory installed on all macs, or somehow shipped as a dependency while installing other needed packages ?
- to sum up, only lines 137 and 138 are needed then ?

thanks

wonderingabout:
@gcp 

just rebased on top of latest upstream next branch
is it ok now ?

thanks

wonderingabout:
edit : forgot to commit again

wonderingabout:
i guess this needs to be once again rebased on top of newest next

wonderingabout:
i also may look at this again when i have more time, closing for now too not to pollute here :+1:

PR #2301, opened 2019-03-25
Ignore passing moves unless they make sense.

gcp:
See discussion in #2277.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -83,11 +83,13 @@ bool UCTNode::create_children(Network & network,
         &state, Network::Ensemble::RANDOM_SYMMETRY);
 
     // DCNN returns winrate as side to move
-    m_net_eval = raw_netlist.winrate;
+    const auto stm_eval = raw_netlist.winrate;
     const auto to_move = state.board.get_to_move();
     // our search functions evaluate from black's point of view
     if (to_move == FastBoard::WHITE) {
-        m_net_eval = 1.0f - m_net_eval;
+        m_net_eval = 1.0f - stm_eval;
+    } else {
+        m_net_eval = stm_eval;
     }
     eval = m_net_eval;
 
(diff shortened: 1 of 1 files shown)


src/UCTNode.cpp
```diff
@@ -103,8 +103,24 @@ bool UCTNode::create_children(Network & network,
...
     }
-    nodelist.emplace_back(raw_netlist.policy_pass, FastBoard::PASS);
-    legal_sum += raw_netlist.policy_pass;
+
+    // Always try passes if we're not trying to be clever.
+    auto allow_pass = cfg_dumbpass;
+
+    // If we're clever, only try passing if we're winning on the
+    // net score and on the board count.
+    if (!allow_pass && (m_net_eval > 0.8f || nodelist.size() <= 10)) {
```
Ttl: `m_net_eval` is from black's point of view. It needs to be inverted if the player is white.

gcp: That's what the code above this already does.

Ttl: NN returns value from the point of the current player and the above code makes it black's view for the search. It needs to be inverted back for the current players view for this check.

gcp:
Wasn't sure what to put the minimum legal moves at. It needs to "reasonably" cover all seki points on the board. 

If there are only eye-points left, then get_nopass_child will already prevent disaster at the root, but the engine could get pretty confused while searching ahead. So it tried to put it high enough, i.e. to count for like 7 or 8 groups and some seki points.

Ttl:
This solves the issue and could be even better solution since it doesn't consider passing at all now. In handicap games with very low eval it did still consider passing with the other patch and could in some situations even play pass.

There could be potential issue where passing would win and it would be sensible to pass, but the NN doesn't realize it and doesn't give high enough value output. At least some of the earlier nets had issues with scoring close games, but I don't think the latest networks really make this mistake anymore. Maybe not worth worrying about?

gcp:
>Maybe not worth worrying about?

It's tricky to do right. The logic here avoids counting in most cases. Even if you'd always count, you would still have the problem that the engine would start adding passes whenever the board is winning in Trump Taylor, which could happen in the middle of tight middlegames.

Note that we will pass anyway when we start running out of moves.

alreadydone:
What about marking the pass move as invalid when it is selected and if the position is net-winning but score-losing (and if at least one visited sibling isn't marked invalid), like in https://github.com/leela-zero/leela-zero/blob/master/src/UCTSearch.cpp#L227-L237? This involve score checks but should reduce them to a minimum.

gcp:
>and if at least one visited sibling isn't marked invalid

Sounds much more complicated. The only thing it would save is score checks if the score is >80% winrate, and I think once the score is there we don't care about the tiny performance impact - if any - any more.

PR #2306, opened 2019-03-28
Add NixOS support

OmnipotentEntity:
This commit uses [direnv](https://direnv.net/) to automatically set up a [nix-shell](https://nixos.org/nix/manual/#name-2) environment to one that is suitable for compiling and running Leela Zero, and sets up CMake to use the unusual environment in NixOS.

To use:
- Have direnv already installed.
- git clone leela-zero recursively
- cd to your leela-zero directory
- Allow direnv to run this script by using `direnv allow`
- Compile as normal, all dependencies ought to be installed


OmnipotentEntity:
This will also probably work if you merely have `nix` installed on Debian for instance.  But I have not tested that.

gcp:
Seems like this should go into nix's portage tree or whatever equivalent, I don't see the point of merging a distribution specific build setup here?

We have a few lines in the CMakefile for .debs but that's not quite the same thing.

OmnipotentEntity:
What on earth is going on in this pull request?

Makes sense, I can put it in nixpkgs, and I didn't really expect this to get merged.  Mostly, I wanted to document the process so other people wouldn't have to figure it out.

gcp:
Yeah sorry about the force pushing, I wanted next to fast-forward onto master and had to reshuffle some commits for that.

PR #2307, opened 2019-03-28
I had problems with the build when qtdir wasn't set

iopq:
Found out I needed not only to install Qt, but also tell VS where it is

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ by adding -DUSE_CPU_ONLY=1 to the cmake command line.
     Double-click the leela-zero2015.sln or leela-zero2017.sln corresponding
     to the Visual Studio version you have.
     # Build from Visual Studio 2015 or 2017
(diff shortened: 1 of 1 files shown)


AncalagonX:
Perhaps a custom Build Error Message could be generated that informs the user when this occurs?

iopq:
Yes, that would probably be ideal. It should be mentioned in the readme anyway.

gcp:
>It should be mentioned in the readme anyway.

https://github.com/leela-zero/leela-zero/tree/master/autogtp

However, the leela-zero build shouldn't fail if you don't have Qt, that's a bug.

>Perhaps a custom Build Error Message could be generated that informs the user when this occurs?

I think this is sensible, pointing out autogtp won't be compiled because the requirements aren't met.

iopq:
@gcp it's not clear, is the main page telling you how to build auto-gtp AND leela-zero or just one of them?

Because it switches between telling you how to compile auto-gtp on Linux back to just doing leela-zero on Windows fooling you into thinking that's the only instructions

PR #2332, opened 2019-04-11
TS update

roy7:
I think I've updated this branch to be as compatible as possible with latest /next code? 

Or perhaps git has betrayed me once again. :)

I couldn't seem to get my local thompson_sampling to rebase or reset into the one from upstream. So I'm using the same ts_update from my original thompson_sampling update attempt I think you forced into there.


rooklift:
The readme has some instances of `github.com/leela_zero` (underscore instead of dash) which I don't think is right?

roy7:
@fohristiwhirl Thanks! Fixed.

gcp:
This didn't quite look like I would've expected a proper rebase to look, so I did one myself. thompson sampling branch should look Ok now?

roy7:
Great thank you. Somehow my local copy had gotten messed up. Maybe I will just delete my local download of that branch and do a fresh clone or something to sync up with server.

PR #2337, opened 2019-04-13
Increase policy for moves with good evaluation

Ttl:
This is one attempt to solve the issue where NN policy affects the search too much even at very high playouts. The idea is to give bonus policy to moves with good evaluation in proportion to the probability that the mean of the evaluation distribution is better than the move with the best evaluation.

Some tests:

```
2s / move on RTX2080
lz_pol_lcb v lz_next (400/400 games)
board size: 19   komi: 7.5
lz_pol_lcb    226 56.50%       99  49.50%     127 63.50%    679.66
lz_next       174 43.50%       73  36.50%     101 50.50%    673.71
                               172 43.00%     228 57.00%
```

```
-v 1600

lz_pol_lcb v lz_next (400/400 games)
board size: 19   komi: 7.5
lz_pol_lcb    227 56.75%       105 52.50%     122 61.00%    103.39
lz_next       173 43.25%       78  39.00%     95  47.50%    103.74
                               183 45.75%     217 54.25%
```

<details>
<summary>gomill command</summary>

```
Player("leelaz -g -r 5 -v 1600 --noponder -w f438268ef88e083aaf7a08fba885552818abc08cd0aefce9671954a8df3cf707.gz --precision half --timemanage off",
    startup_gtp_commands=["time_settings 0 1 0"])
```
</details>

This only has effect at high playouts and it does seem to be an improvement even at large playouts. 2s gets about 6000 playouts on RTX2080 with the 15b network and with tree reuse there is often over 10k visits at the root.

`lz_next` is using the current next branch with commit 3f297889563bcbec671982c655996ccff63fa253.

The original idea was to make policy dynamic by treating it as probability that move is the best. Some sort of bayesian update rule would have been ideal. On big issue is that the probability that one normal distribution mean is higher than several other normal distributions doesn't have nice closed form solution and deriving any good bayseian update formula doesn't really seem possible. This PR instead calculates the probability that mean is bigger than mean of the move with the best evaluation. The policy bonus is proportional to the probability, but the range is limited to the policy of the sibling node with the highest policy.

`cfg_lcb_pol_min_visits` is not optimized, just my guess at the moment. Low visit behaviour could also be smarter than just a fixed visit cutoff, but I'll make a PR now as there is evidence that this is an improvement. I'm sure that this can be improved still.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -91,6 +91,8 @@ float cfg_fpu_reduction;
 float cfg_fpu_root_reduction;
 float cfg_ci_alpha;
 float cfg_lcb_min_visit_ratio;
+float cfg_lcb_pol;
+int cfg_lcb_pol_min_visits;
 std::string cfg_weightsfile;
 std::string cfg_logfile;
 FILE* cfg_logfile_handle;
@@ -351,6 +353,8 @@ void GTP::setup_default_parameters() {
     cfg_fpu_root_reduction = cfg_fpu_reduction;
     cfg_ci_alpha = 1e-5f;
     cfg_lcb_min_visit_ratio = 0.10f;
+    cfg_lcb_pol = 0.66f;
+    cfg_lcb_pol_min_visits = 100;
     cfg_random_cnt = 0;
     cfg_random_min_visits = 1;
     cfg_random_temp = 1.0f;
--- a/src/GTP.h
+++ b/src/GTP.h
@@ -116,6 +116,8 @@ extern float cfg_fpu_reduction;
 extern float cfg_fpu_root_reduction;
 extern float cfg_ci_alpha;
 extern float cfg_lcb_min_visit_ratio;
+extern float cfg_lcb_pol;
+extern int cfg_lcb_pol_min_visits;
 extern std::string cfg_logfile;
 extern std::string cfg_weightsfile;
 extern FILE* cfg_logfile_handle;
--- a/src/Leela.cpp
+++ b/src/Leela.cpp
@@ -204,6 +204,7 @@ static void parse_commandline(int argc, char *argv[]) {
         ("softmax_temp", po::value<float>())
         ("fpu_reduction", po::value<float>())
         ("ci_alpha", po::value<float>())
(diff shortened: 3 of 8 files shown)


l1t1:
could you upload a binary windows exe for test.

Ttl:
```
1s / move with 40b net on RTX2080:

lz_lcb_pol v lz_next (400/400 games)
board size: 19   komi: 7.5
lz_lcb_pol    208 52.00%       98  49.00%     110 55.00%    170.64
lz_next       192 48.00%       90  45.00%     102 51.00%    170.83
                               188 47.00%     212 53.00%
```
Probably requires more nodes to have significant effect.

I did a quick clop for min_visits and it could probably be lowered a little bit, but I don't think there is much Elo in it.

I don't use Windows and can't generate Windows binaries.

Ttl:
I'm not sure if the tests in the first posts were very lucky or I did some mistake setting up the tests but it looks like that in the further tests Elo gain is much smaller. I ran the v1600 tests again and got 200-200 result against next branch, clopping also shows that this is probably smaller gain.

wonderingabout:
@Ttl  if its not weaker at low visits (1600-), but stronger at higher visits (10k+), we may still want to go with it i think, because leela zero currently has many problematic blindspots as i saw again and again on countless sgf, especially at higher visits

you may want to try at higher visits again, like : 

10k visits
if 10k is significantly stronger, you may try 5k as well
else, maybe
25k
else, maybe
50k

you may also want the testing scale to lz-minigo v15 as an opponent rather than selfplay (if both AI have blindspots, in selfplay, we may fail to be able to quantify the punishment of mistakes that should have  happened) for example, which i found very very much more accurate in my testing against native phoenixgo

gcp:
If we're not confident it's an improvement I would prefer iterating on the idea until there's clear gains.

gcp:
FWIW at least in chess policy flattening appears to work: https://github.com/LeelaChessZero/lc0/issues/743#issuecomment-473357829

Ttl:
I'll maybe look at this later again when I have more time. Closing this for now.

PR #1565, opened 2018-06-17
FPU reduction proportional to root eval

Ttl:
I did some statistics about the optimal FPU reduction. I first changed FPU to 1 to expand all the first level nodes, added code to save child node net evaluations and then played some self play games to gather enough data. In the analysis script I calculated the difference between child node net evaluation and root net evaluation for every node, this would be the optimal FPU reduction constant to use for that node. Code is available at: https://github.com/Ttl/leela-zero/tree/fpu_estimation

Here is histogram of the evaluation difference of child and root:

fpu_hist

As expected most of the moves are worse than the root which is why root evaluation minus constant is a more accurate initialization than just root evaluation.

Here is L2 loss plot with different FPU reduction constants:

fpu_constant_l2

The minimum is very close to what was previously determined to be strongest using CLOP. Estimated maximum was -0.3 in the last tuning run (#1211) and in this analysis -0.29 has the lowest L2 loss.

Plotting FPU reduction vs. root evaluation gives the following plot:

fpu_winrate

As the winrate can't go lower than 0 it makes sense to use FPU reduction that is proportional to the root evaluation instead of just subtracting a constant. In the plot green line is the least squares fitted FPU reduction vs. winrate (This pull request).

Plotting the FPU reduction vs. policy head output gives:

fpu_policy

As would be expected moves with high policy output usually have value network evaluation close to the root and low policy moves are often much worse, although there are exceptions. This plot does give some explanation why `sqrt(total_visited_policy)` works: It lowers the reduction for moves with high policy that are expanded first and that are usually close to the root evaluation. In green lines are the FPU reductions with 0.1, 0.5 and 0.9 root evaluations using the equation in this pull request.

There are many different functions that could be used to initialize nodes from policy and root evaluation, but I decided to keep it simple and avoid adding any more tunable variables. The equation in this pull request gives lower L2 loss than simple subtraction. I didn't try to find any better formula for taking the policy output into account and used the current `sqrt(total_visited_policy)` for that.

L2 loss for this equation in prediction the child evaluation is 0.033 using the latest network compared to 0.051 using the current equation. -0.6 is the best reduction using this formula with this network. As was earlier observed when FPU reduction was first introduced it does depend on the network. Minigo-303 optimum is at -0.29 and ELF is at -0.7. The new formula is more accurate also for them by about the same amount as for LZ net.

[...]

PR #2342, opened 2019-04-16
Scaling fpu by policy according to normal distribution

wentaol:
Assume that network eval on a parent node gives us a normal prior on the true value, and this is also the prior on the value of the best child:

V_parent ~ N(parent_eval, sigma^2)
V_bestchild ~ N(parent_eval, sigma^2)

Some other child which is not the best child could be then assumed to be distributed with a lower expected value:

V_child ~ N(parent_eval - fpu_reduction_child, sigma_child^2)

sigma and sigma_child are unknown standard deviation that can be empirically determined.
We want to find fpu_reduction_child, which is the difference in the means of V_bestchild and V_child. We can set P(V_bestchild > V_child) = psa_bestchild / (psa_best_child + psa_child) and derive a formulation for fpu reduction scaled by relative policy:

fpu_reduction_child = cfg_fpu_reduction * erf_inverse(1 - 2 * psa_child / (psa_best_child + psa_child))

cfg_fpu_reduction is the stddev of V_bestchild - V_child.  I set it to 0.2 without much tuning but it seems to work better than 0.25. 

image
[colab notebook with derivation and plots](https://colab.research.google.com/drive/1QKOVkW8cOHxizW-BrcoEvzCwA5tZQxez)

Results thus far:

800 playouts:
```
lz-next v lz-fpuscaling (200/200 games)
board size: 19   komi: 7.5
lz-next           78 39.00%       34 34.00%     44  44.00%    164.13
lz-fpuscaling    122 61.00%       56 56.00%     66  66.00%    172.58
                                  90 45.00%     110 55.00%
```

200 visits:
```
lz-0.17 v lz-fpuscaling (400/400 games)
board size: 19   komi: 7.5
lz-0.17          160 40.00%       63  31.50%     97  48.50%     44.49
lz-fpuscaling    240 60.00%       103 51.50%     137 68.50%     44.53
                                  166 41.50%     234 58.50%
```

1600 visits:
```
lz-0.17 v lz-fpuscaling (200/200 games)
board size: 19   komi: 7.5
lz-0.17           81 40.50%       34 34.00%     47  47.00%    253.23
lz-fpuscaling    119 59.50%       53 53.00%     66  66.00%    259.31
                                  87 43.50%     113 56.50%
```
EDIT: Update 1600v matches
EDIT: Attached windows binary
EDIT: Updated windows binary for new param value
[leelaz.zip](https://github.com/leela-zero/leela-zero/files/3098204/leelaz.zip)

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -301,7 +301,7 @@ float erfinv_approx(float x)
 {
     float sign = x > 0 ? 1 : -1;
     float tmp = (1 - x) * (1 + x);
-    float tt1 = 2 / (M_PI * 0.147) + 0.5 * log(tmp);
+    float tt1 = 4.330747 + 0.5 * log(tmp);
     float tt2 = log(tmp) / 0.147;
     return sign * sqrt(-tt1 + sqrt(tt1 * tt1 - tt2));
 }
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -316,9 +316,10 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
 
     const auto numerator = std::sqrt(double(parentvisits) *
             std::log(cfg_logpuct * double(parentvisits) + cfg_logconst));
-    const auto policyratio = max_unvisited_policy / (max_unvisited_policy + max_policy);
+    const auto policyratio = (max_policy - max_unvisited_policy)
+                            / (max_policy + max_unvisited_policy);
     const auto fpu_reduction = (is_root ? cfg_fpu_root_reduction : cfg_fpu_reduction)
-                                * Utils::erfinv_approx(1 - 2 * policyratio);
+                                * Utils::erfinv_approx(policyratio);
     // Estimated eval for unknown nodes = original parent NN eval - reduction
     const auto fpu_eval = get_net_eval(color) - fpu_reduction;
 
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -301,15 +301,15 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     wait_expanded();
 
     // Count parentvisits manually to avoid issues with transpositions.
-    auto total_visited_policy = 0.0f;
     auto parentvisits = size_t{0};
     auto max_policy = 0.0f;
+    auto max_unvisited_policy = 0.0f;
     for (const auto& child : m_children) {
         if (child.valid()) {
             parentvisits += child.get_visits();
             max_policy = std::max(max_policy, child.get_policy());
-            if (child.get_visits() > 0) {
-                total_visited_policy += child.get_policy();
+            if (child.get_visits() == 0) {
+                max_unvisited_policy = std::max(max_unvisited_policy, child.get_policy());
             }
         }
     }
@@ -317,6 +317,10 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     const auto numerator = std::sqrt(double(parentvisits) *
             std::log(cfg_logpuct * double(parentvisits) + cfg_logconst));
 
+    const auto policyratio = max_unvisited_policy / (max_unvisited_policy + max_policy);
+    //cfg_fpu_reduction is now interpreted as parent node stddev
+    const auto fpu_reduction = (is_root ? cfg_fpu_root_reduction : cfg_fpu_reduction)
+                                * Utils::erfinv_approx(1 - 2 * policyratio);
 
     auto best = static_cast<UCTNodePointer*>(nullptr);
     auto best_value = std::numeric_limits<double>::lowest();
@@ -328,10 +332,6 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
         }
         const auto psa = child.get_policy();
 
-        const auto policyratio = psa / (psa + max_policy);
-        //cfg_fpu_reduction is now interpreted as parent node stddev
-        const auto fpu_reduction = (is_root ? cfg_fpu_root_reduction : cfg_fpu_reduction)
-                                    * Utils::erfinv_approx(1 - 2 * policyratio);
         // Estimated eval for unknown nodes = original parent NN eval - reduction
         auto winrate = parent_net_eval - fpu_reduction;
 
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -297,36 +297,54 @@ void UCTNode::accumulate_eval(float eval) {
     atomic_add(m_blackevals, double(eval));
 }
 
+float erfinv_approx(float x)
+{
+    float sign = x > 0 ? 1 : -1;
+    float tmp = (1 - x) * (1 + x);
+    float tt1 = 2 / (M_PI * 0.147) + 0.5 * log(tmp);
+    float tt2 = log(tmp) / 0.147;
+    return sign * sqrt(-tt1 + sqrt(tt1 * tt1 - tt2));
+}
+
+
 UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     wait_expanded();
 
     // Count parentvisits manually to avoid issues with transpositions.
     auto total_visited_policy = 0.0f;
     auto parentvisits = size_t{0};
+    auto max_policy = 0.0f;
     for (const auto& child : m_children) {
         if (child.valid()) {
             parentvisits += child.get_visits();
+            max_policy = std::max(max_policy, child.get_policy());
             if (child.get_visits() > 0) {
-                total_visited_policy += child.get_policy();
+                total_visited_policy += child.get_policy();                
             }
         }
     }
 
     const auto numerator = std::sqrt(double(parentvisits) *
             std::log(cfg_logpuct * double(parentvisits) + cfg_logconst));
-    const auto fpu_reduction = (is_root ? cfg_fpu_root_reduction : cfg_fpu_reduction) * std::sqrt(total_visited_policy);
-    // Estimated eval for unknown nodes = original parent NN eval - reduction
-    const auto fpu_eval = get_net_eval(color) - fpu_reduction;
+
 
     auto best = static_cast<UCTNodePointer*>(nullptr);
     auto best_value = std::numeric_limits<double>::lowest();
+    auto parent_net_eval = get_net_eval(color);
 
     for (auto& child : m_children) {
         if (!child.active()) {
             continue;
         }
+        const auto psa = child.get_policy();
 
-        auto winrate = fpu_eval;
+        const auto policyratio = psa / (psa + max_policy);
+        //cfg_fpu_reduction is now interpreted as parent node stddev
+        const auto fpu_reduction = 2 * cfg_fpu_reduction 
+                                * erfinv_approx(1 - 2 * policyratio);                              
+        // Estimated eval for unknown nodes = original parent NN eval - reduction
+        auto winrate = parent_net_eval - fpu_reduction;
+
         if (child.is_inflated() && child->m_expand_state.load() == ExpandState::EXPANDING) {
             // Someone else is expanding this node, never select it
             // if we can avoid so, because we'd block on it.
@@ -334,7 +352,6 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
         } else if (child.get_visits() > 0) {
             winrate = child.get_eval(color);
         }
-        const auto psa = child.get_policy();
         const auto denom = 1.0 + child.get_visits();
         const auto puct = cfg_puct * psa * (numerator / denom);
         const auto value = winrate + puct;
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -319,7 +319,7 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
             parentvisits += child.get_visits();
             max_policy = std::max(max_policy, child.get_policy());
             if (child.get_visits() > 0) {
-                total_visited_policy += child.get_policy();                
+                total_visited_policy += child.get_policy();
             }
         }
     }
@@ -340,8 +340,7 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
 
         const auto policyratio = psa / (psa + max_policy);
         //cfg_fpu_reduction is now interpreted as parent node stddev
-        const auto fpu_reduction = cfg_fpu_reduction 
-                                * erfinv_approx(1 - 2 * policyratio);                              
+        const auto fpu_reduction = cfg_fpu_reduction * erfinv_approx(1 - 2 * policyratio);
         // Estimated eval for unknown nodes = original parent NN eval - reduction
         auto winrate = parent_net_eval - fpu_reduction;
 
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -297,16 +297,6 @@ void UCTNode::accumulate_eval(float eval) {
     atomic_add(m_blackevals, double(eval));
 }
 
-float erfinv_approx(float x)
-{
-    float sign = x > 0 ? 1 : -1;
-    float tmp = (1 - x) * (1 + x);
-    float tt1 = 4.330747 + 0.5 * log(tmp);
-    float tt2 = log(tmp) / 0.147;
-    return sign * sqrt(-tt1 + sqrt(tt1 * tt1 - tt2));
-}
-
-
 UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     wait_expanded();
 
@@ -340,7 +330,8 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
 
         const auto policyratio = psa / (psa + max_policy);
         //cfg_fpu_reduction is now interpreted as parent node stddev
-        const auto fpu_reduction = cfg_fpu_reduction * erfinv_approx(1 - 2 * policyratio);
+        const auto fpu_reduction = (is_root ? cfg_fpu_root_reduction : cfg_fpu_reduction)
+                                    * Utils::erfinv_approx(1 - 2 * policyratio);
         // Estimated eval for unknown nodes = original parent NN eval - reduction
         auto winrate = parent_net_eval - fpu_reduction;
 
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -316,25 +316,21 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
 
     const auto numerator = std::sqrt(double(parentvisits) *
             std::log(cfg_logpuct * double(parentvisits) + cfg_logconst));
-
     const auto policyratio = max_unvisited_policy / (max_unvisited_policy + max_policy);
-    //cfg_fpu_reduction is now interpreted as parent node stddev
     const auto fpu_reduction = (is_root ? cfg_fpu_root_reduction : cfg_fpu_reduction)
                                 * Utils::erfinv_approx(1 - 2 * policyratio);
+    // Estimated eval for unknown nodes = original parent NN eval - reduction
+    const auto fpu_eval = get_net_eval(color) - fpu_reduction;
 
     auto best = static_cast<UCTNodePointer*>(nullptr);
     auto best_value = std::numeric_limits<double>::lowest();
-    auto parent_net_eval = get_net_eval(color);
 
     for (auto& child : m_children) {
         if (!child.active()) {
             continue;
         }
-        const auto psa = child.get_policy();
 
-        // Estimated eval for unknown nodes = original parent NN eval - reduction
-        auto winrate = parent_net_eval - fpu_reduction;
-
+        auto winrate = fpu_eval;
         if (child.is_inflated() && child->m_expand_state.load() == ExpandState::EXPANDING) {
             // Someone else is expanding this node, never select it
             // if we can avoid so, because we'd block on it.
@@ -342,6 +338,7 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
         } else if (child.get_visits() > 0) {
             winrate = child.get_eval(color);
         }
+        const auto psa = child.get_policy();
         const auto denom = 1.0 + child.get_visits();
         const auto puct = cfg_puct * psa * (numerator / denom);
         const auto value = winrate + puct;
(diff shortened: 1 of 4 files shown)


src/Utils.cpp
```diff
@@ -74,6 +74,15 @@ float Utils::cached_t_quantile(int v) {
     return z_lookup[z_entries - 1];
 }
 
+float Utils::erfinv_approx(float x)
+{
+    float sign = x > 0 ? 1 : -1;
```
Hersmunch: The coding style guidelines say to use `auto` if possible so probably need to use `0.0f` etc.
Same for rest of function.

src/Utils.cpp
```diff
@@ -74,6 +74,15 @@ float Utils::cached_t_quantile(int v) {
...
+    float tmp = (1 - x) * (1 + x);
+    float tt1 = 4.330747 + 0.5 * log(tmp);
```
Hersmunch: Should this be `std::log()`?

src/Utils.cpp
```diff
@@ -74,6 +74,15 @@ float Utils::cached_t_quantile(int v) {
...
+    float tt2 = log(tmp) / 0.147;
+    return sign * sqrt(-tt1 + sqrt(tt1 * tt1 - tt2));
```
Hersmunch: Should this be `std::sqrt()`?

Hersmunch:
In case it is helpful here, see #1565 for some previous analysis of FPU stats. It was a while ago though.

wentaol:
Those are interesting plots, the Fpu vs root Eval plot are particularly surprising. 

If the true distributions are beta , then when root eval approaches 1, variance shrinks so we should see a smaller spread of FPU reduction. Maybe I'm missing something, otherwise it seems beta distribution would have been a worse choice.

I also just realized boost has built in erf inverse functions, any advice if I should switch to that instead?

Ttl:
Very nice PR. I like that is has some theory why it works.

I don't think I posted it to the PR #1565, but I remember proportional FPU being worse in handicap games. If evaluation can't go under 0, lot of low policy moves are expanded when evaluation is low.

Multiple sqrt and log calls during FPU evaluation are expensive as it's called very many times. In 1000 visit benchmark there seems to be around 4 million additional calls to log which starts to have some measurable effects. Even simpler approximation could be used since good precision isn't needed.

I have clop running with the latest 40b net and self-play settings. Should take around day or two to finish.

oscardssmith:
In LC0, we have made some special 1st order approximate functions for sqrt and log calls. For a use like this, it should be very safe to use these approximations. One other optimization worth considering is you only need to compute fpu for highest policy unvisited node since fpu will be increasing with policy. That should bring the speed much higher.

wentaol:
> I don't think I posted it to the PR #1565, but I remember proportional FPU being worse in handicap games. If evaluation can't go under 0, lot of low policy moves are expanded when evaluation is low.

That's quite an important observation, and leads me to think that puct and fpu are two interlocked parts of the puzzle that cannot be tackled independently. If fpu shrinks at low winrate, then puct should behave similarly. Meanwhile, the right choice for puct formulation probably also depends on the fpu reduction applied.  Perhaps there is a way to get winrate-scaled fpu to work, but we'd have to also make puct winrate-scaled. Following that line of thought, maybe the right way is to unify the two and fold them into the same expression?

> Multiple sqrt and log calls during FPU evaluation are expensive as it's called very many times. In 1000 visit benchmark there seems to be around 4 million additional calls to log which starts to have some measurable effects. Even simpler approximation could be used since good precision isn't needed.

The boost erf inverse turns out to be even slower in our regime, so I fit a piecewise quadratic. I've pushed it into a [side branch](https://github.com/wentaol/leela-zero/tree/scaled-fpu-approxerf), will merge it into this one after I verify results are not broken.

Ttl:
Optimization suggested by @oscardssmith where FPU is calculated only for the unvisited move with the biggest policy should reduce computation a lot. That is probably much more significant than simplifying the formula.

This should work:
<details>
<summary>diff</summary>

```diff
diff --git a/src/UCTNode.cpp b/src/UCTNode.cpp
index 3db3bb9..1cac0e2 100644
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -301,15 +301,15 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     wait_expanded();
 
     // Count parentvisits manually to avoid issues with transpositions.
-    auto total_visited_policy = 0.0f;
     auto parentvisits = size_t{0};
     auto max_policy = 0.0f;
+    auto max_unvisited_policy = 0.0f;
     for (const auto& child : m_children) {
         if (child.valid()) {
             parentvisits += child.get_visits();
             max_policy = std::max(max_policy, child.get_policy());
-            if (child.get_visits() > 0) {
-                total_visited_policy += child.get_policy();
+            if (child.get_visits() == 0) {
+                max_unvisited_policy = std::max(max_unvisited_policy, child.get_policy());
             }
         }
     }
@@ -317,6 +317,10 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     const auto numerator = std::sqrt(double(parentvisits) *
             std::log(cfg_logpuct * double(parentvisits) + cfg_logconst));
 
+    const auto policyratio = max_unvisited_policy / (max_unvisited_policy + max_policy);
+    //cfg_fpu_reduction is now interpreted as parent node stddev
+    const auto fpu_reduction = (is_root ? cfg_fpu_root_reduction : cfg_fpu_reduction)
+                                * Utils::erfinv_approx(1 - 2 * policyratio);
 
     auto best = static_cast<UCTNodePointer*>(nullptr);
     auto best_value = std::numeric_limits<double>::lowest();
@@ -327,11 +331,6 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
             continue;
         }
         const auto psa = child.get_policy();
-
-        const auto policyratio = psa / (psa + max_policy);
-        //cfg_fpu_reduction is now interpreted as parent node stddev
-        const auto fpu_reduction = (is_root ? cfg_fpu_root_reduction : cfg_fpu_reduction)
-                                    * Utils::erfinv_approx(1 - 2 * policyratio);
         // Estimated eval for unknown nodes = original parent NN eval - reduction
         auto winrate = parent_net_eval - fpu_reduction;

```
</details>

It gives the same output as before when using single thread and same RNG seed. The number of log calls is reduced from 4M to 11k.

oscardssmith:
The 100x reduction is nice.

l1t1:
good job, could you upload a windows exe?

wentaol:
@oscardssmith thanks for suggesting the optimization

@Ttl I'm probably missing something, but won't we also need a check to skip non max unvisited children? Otherwise all unvisited children would get the same fpu reduction? Or are they sorted by policy order already so that won't happen?

@l1t1 will upload a windows binary later today

Ttl:
No there isn't a need to skip them. Only one node with the largest score will be chosen. All unvisited nodes will have the same Q, but the node with the largest policy will have the largest U term so it has the largest score and is the one that will be chosen.

src/UCTNode.cpp
```diff
@@ -301,40 +301,47 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
...
+    auto parent_net_eval = get_net_eval(color);
 
     for (auto& child : m_children) {
         if (!child.active()) {
             continue;
         }
+        const auto psa = child.get_policy();
+
+        // Estimated eval for unknown nodes = original parent NN eval - reduction
+        auto winrate = parent_net_eval - fpu_reduction;
```
Hersmunch: It looks like the lop innards can be reverted, calculating `fpu_eval` beforehand.

Hersmunch:
Just wanted to say thanks for the insightful contribution. Nice work.

wentaol:
@Hersmunch Thanks, I've just pushed a commit to address the style suggestions.

Also many thanks to @Ttl for running CLOP on cfg_fpu_reduction, and the optimal parameter for the latest network seems to be around 0.27. I've verified this at 200 visits:

latest(221) weights, -v200 --fpu_reduction 0.27:
```
lz-0.17 v lz-fpuscaling (400/400 games)
board size: 19   komi: 7.5
lz-0.17          144 36.00%       56  28.00%     88  44.00%     48.56
lz-fpuscaling    256 64.00%       112 56.00%     144 72.00%     47.49
                                  168 42.00%     232 58.00%
```
I've updated GTP.cpp with this value.

Note that on 220 weights, -v200 --fpu_reduction 0.27:
```
lz-0.17 v lz-fpuscaling (400/400 games)
board size: 19   komi: 7.5
lz-0.17          170 42.50%       79  39.50%     91  45.50%     49.57
lz-fpuscaling    230 57.50%       109 54.50%     121 60.50%     48.08
                                  188 47.00%     212 53.00%
```
Which is a little bit worse than the results I got with --fpu_reduction 0.20, suggesting that the optimal value is slightly different depending on the network.

petgo3:
I'm veryfing fpuscaling with 1600 visits (net lz222) on a RTX 2080 Ti (hope it helps):
gogui-twogtp.exe -black "C:\temp\lz\leelaz.exe -t 10 -g -w C:\temp\lz\40x256 --noponder --visits 1600" -white "C:\temp\pet\leelaz.exe -g -t 10 -w C:\temp\lz\40x256 --noponder --visits 1600" -games 500 -sgffile "C:\temp\log\sgf" -auto -komi 7.5 -alternate -force

Complete results will be available this evening.
As interim result we have (with 132 games):
43.9% (±4.3) for lz-0.17 so 56.1% for lz-fpuscaling, witch is quite fitting with previous results :-)
Thx to @wentaol @Hersmunch!

src/UCTNode.cpp
```diff
@@ -301,20 +301,24 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
...
             }
         }
     }
 
     const auto numerator = std::sqrt(double(parentvisits) *
             std::log(cfg_logpuct * double(parentvisits) + cfg_logconst));
-    const auto fpu_reduction = (is_root ? cfg_fpu_root_reduction : cfg_fpu_reduction) * std::sqrt(total_visited_policy);
+    const auto policyratio = max_unvisited_policy / (max_unvisited_policy + max_policy);
+    const auto fpu_reduction = (is_root ? cfg_fpu_root_reduction : cfg_fpu_reduction)
+                                * Utils::erfinv_approx(1 - 2 * policyratio);
```
Hersmunch: `1.0f` & `2.0f`

src/Utils.cpp
```diff
@@ -74,6 +74,15 @@ float Utils::cached_t_quantile(int v) {
     return z_lookup[z_entries - 1];
 }
 
+float Utils::erfinv_approx(float x)
+{
```
Hersmunch: Move `{` to end of previous line.

src/Utils.cpp
```diff
@@ -74,6 +74,15 @@ float Utils::cached_t_quantile(int v) {
...
+    const auto sign = x > 0 ? 1.0f : -1.0f;
```
Hersmunch: `0.0f`

Hersmunch: Hmm it's probably an insignificant saving but with current function usage `x` will never be negative. Actually I see you've accounted for this in the piecewise approximation https://github.com/wentaol/leela-zero/commit/b8c2b0bf69273e9a0d3d7427bdc496ecffe279eb. If this approach is taken, then perhaps suffix the function name with something like a `_pos` and add an assert to ensure `x` isn't negative?

src/UCTNode.cpp
Hersmunch: Or avoid the multiplication with:
`1 - 2 * a / (a + b) == (b - a) / (b + a)`

Hersmunch (commented):
I picked up a few more things on another read through. Sorry to be a pain!

Hersmunch:
Oh and this needs to be rebased on `next` before it can be merged.

wentaol:
Should I create a separate branch, rebase to next and create a new pull request off of that? Otherwise if I rebase directly would github handle it correctly?

petgo3:
After 250 games it's 45.6% (±3.2) for lz-0.17 so 54.3% for lz-fpuscaling. I used fpu_reduction 0.27.
I tried also about 100 games with -v 100 with quite the same result.
Perhaps Lz 222 net needs different fpu_reduction?

wentaol:
> After 250 games it's 45.6% (±3.2) for lz-0.17 so 54.3% for lz-fpuscaling. I used fpu_reduction 0.27.
> I tried also about 100 games with -v 100 with quite the same result.
> Perhaps Lz 222 net needs other fpu_reduction?

That's one possibility. I speculate it might also have something to do with the -t 10 setting. We compute fpu_reduction for the max policy unvisited node, and apply it  across all unvisited nodes, so with too many threads we might end up exploring low policy unvisited nodes over high policy nodes.

petgo3:
> I speculate it might also have something to do with the -t 10 setting.
>
I'll have a go with  -t 3 -v 100 and check if proportion changes ...

petgo3:
I finished some extented tests on lz 222 with fpu_scaling. The results show some correlation between tasks (-t), visits (-v) and the proportion (of winning):
grafik
A high number of tasks supports the effectiveness (especially at 25 tasks) of fpu_scaling.
More visits also seems to help a bit (not as significant). I have no idea about the reason of this behavior.

wentaol:
No idea what's going on either. I'll run some matches at -t2 -v1600...

petgo3:
Just to get a gut instinct, i ran about 15 games at kgs with golois2:
http://kgs.gosquares.net/index.rhtml.en?id=petgo3&id2=Golois2&r=0
the recent games on 2019/04/21 with 25 threads/tasks and got a quite good result.

wentaol:
It looks like 222 isn't a regular promoted net. That could be the reason.

l1t1:
can you try cb2aca23 weight? (cb2aca23 VS 0407e5b5 209 : 189 (52.51%))
it is regular weight, but it used part of lz 222 data to traing

Ttl:
Getting enough games for clop at 1600 visits took too long, so I ran another clop with 200 visits. After 1800 games the optimal `fpu_reduction` is around 0.22.

Running test matches with 1s / move on RTX 2080 using net 222 I got the following results:

```
0.27:
lz_scaled_fpu v lz_next (400/400 games)
board size: 19   komi: 7.5
lz_scaled_fpu    199 49.75%       76  38.00%     123 61.50%    187.46
lz_next          201 50.25%       77  38.50%     124 62.00%    185.43
                                  153 38.25%     247 61.75%
```

```
0.22:
lz_scaled_fpu v lz_next (316/400 games)
board size: 19   komi: 7.5
lz_scaled_fpu    182 57.59%       85  53.80%     97  61.39%    187.97
lz_next          134 42.41%       61  38.61%     73  46.20%    186.40
                                  146 46.20%     170 53.80%
```

I'll also do a clop for the `fpu_reduction` with the current code as it's been a while since it was set.

petgo3:
@l1t1 I tried cb2aca23 ("LZ222,5") (took fpu_reduction 0.27):
With 300 games v100 t3 i got 56,4% for scaled_fpu.
With 62 games v100 t25 i got 60,3 for scaled_fpu.
So effect seems mainly the same as for 222 (see above).

l1t1:
```
validation.exe -k elfv2_17 -s "0:1" -g 1 -n \elfv2.gz -o "-g -v 10 -t 1 -q -d --timemanage off  -w" -n \elfv2.gz -o "-g -v 10 -t 1 -q -d --timemanage off  -w " -- leelaz(0.17 master) -- leelaz171(@wentaol version)
```
67 wins, 94 losses
161 games played.

l1t1:
```
109 wins, 155 losses
264 games played.
Status: 0 LLR -0.134018 Lower Bound -2.94444 Upper Bound 2.94444
*ERROR*: Could not talk to engine after launching.
```

Ttl:
Clopping `fpu_reduction` in the next branch at 200 visits gives optimum around 0.30 with estimate of around 35 Elo better than `fpu_reduction=0.25`. I think that eats part of the strength gain from this PR, but this should still be improvement.

l1t1:
```
validation.exe -k elfv2_17 -s "0:1" -g 1 -n \elfv2.gz -o "-g -v 100 -t 1 -q -d --timemanage off  -w" -n \elfv2.gz -o "-g -v 100 -t 1 -q -d --timemanage off  -w " -- leelaz -- leelaz171
```

32 wins, 30 losses
62 games played.

wentaol:
@l1t1 thanks for running the games, wins is from the perspective of 0.17 or the PR?

l1t1:
wins should be of first engine(0.17 master)

gcp:
I understood I should pull with the parameter set to 0.22? Correct?

wentaol:
Looks like 0.22 works better on latest weights(222) but haven't been tested as much as 0.27. 

I can run some -v200 matches to check that it works reasonably on 221 (which is where we got 0.27) , and maybe 157?

gcp:
>Clopping fpu_reduction in the next branch at 200 visits gives optimum around 0.30 with estimate of around 35 Elo better than fpu_reduction=0.25. I think that eats part of the strength gain from this PR, but this should still be improvement.

Probably we'd need a test between 0.30 next vs 0.22 this branch then?

CLOP estimates for Elo gain are overly optimistic (unless you're looking at "mean", which is overly pessimistic), you need an independent test.

l1t1:
from my test of elfv2, the pr vs 0.17, visit 10, pr wins 57% at 264 games, visits 100, pr wins0.5 at 120games

Ttl:
1s / move on RTX2080. This PR with 0.22 fpu_reduction vs. next with 0.30 fpu_reduction:

```
lz_scaled_fpu v lz_next (400/400 games)
board size: 19   komi: 7.5
lz_scaled_fpu    215 53.75%       82  41.00%     133 66.50%    185.18
lz_next          185 46.25%       67  33.50%     118 59.00%    183.77
                                  149 37.25%     251 62.75%
```

Smaller win rate than before https://github.com/leela-zero/leela-zero/pull/2342#issuecomment-485624904. Still more likely to be better although the confidence isn't quite 95% yet.

Same with `-v 1600`:

```
lz_scaled_fpu v lz_next (400/400 games)
board size: 19   komi: 7.5
lz_scaled_fpu    223 55.75%       100 50.00%     123 61.50%    257.59
lz_next          177 44.25%       77  38.50%     100 50.00%    257.79
                                  177 44.25%     223 55.75%
```

wentaol:
Unfortunately the strength gain were less clear in my tests. 
Both sets of matches below were run on -t1 -v200, with `--fpu_reduction 0.3` for master and `--fpu_reduction 0.22` for PR.

Using 221 weights:
```
lz-0.17 v lz-fpuscaling (400/400 games)
board size: 19   komi: 7.5
lz-0.17          197 49.25%       87  43.50%     110 55.00%     48.60
lz-fpuscaling    203 50.75%       90  45.00%     113 56.50%     49.02
                                  177 44.25%     223 55.75%
```

Using 222 weights:
```
lz-0.17 v lz-fpuscaling (400/400 games)
board size: 19   komi: 7.5
lz-0.17          200 50.00%       84  42.00%     116 58.00%     47.44
lz-fpuscaling    200 50.00%       84  42.00%     116 58.00%     48.24
                                  168 42.00%     232 58.00%
```

Using newly promoted 223 weights:
```
lz-0.17 v lz-fpuscaling (400/400 games)
board size: 19   komi: 7.5
lz-0.17          179 44.75%       85  42.50%     94  47.00%     48.39
lz-fpuscaling    221 55.25%       106 53.00%     115 57.50%     48.89
                                  191 47.75%     209 52.25%
```

Also it looks like the dependence on the number of threads as observed by @petgo3  is a bigger effect than I expected. The same 222 weight test as above, but with 10 threads (default) instead of 1:

```
lz-0.17 v lz-fpuscaling (400/400 games)
board size: 19   komi: 7.5
lz-0.17          176 44.00%       71  35.50%     105 52.50%     37.63
lz-fpuscaling    224 56.00%       95  47.50%     129 64.50%     38.26
                                  166 41.50%     234 58.50%
```

After checking the logs I realized that all my earlier tests with the higher strength gains were obtained with 10 threads. My speculation is that low policy nodes are visited a lot more with more threads, and with low thread count the search is usually not wide enough for the difference introduced by the PR to matter.

Ttl:
There seems to be a bug with high number of threads. FPU reduction is currently calculated as sqrt(sum of visited policy), but visits are only increased after NN evaluation is done. All threads assembling a batch use the same FPU reduction. The same issue is with this PR due to the optimization mentioned earlier.

When adding a check that `virtual_loss` is also zero in https://github.com/leela-zero/leela-zero/pull/2342/files#diff-28216771f77a09f4e0782ad5404fdd28R311 I get the following results with 10 threads and 800 visits:

```
lz_scaled_fpu_vloss v lz_scaled_fpu (400/400 games)
board size: 19   komi: 7.5
lz_scaled_fpu_vloss    223 55.75%       98  49.00%     125 62.50%    213.18
lz_scaled_fpu          177 44.25%       75  37.50%     102 51.00%    213.30
                                        173 43.25%     227 56.75%
```

wentaol:
Some results on 225, which was the newly promoted net when I started these matches: 

It looks like fpu reduction of 0.22 is better than 0.27 for the PR, at -t10, -v200:
```
lz-0.17-0.3 v lz-fpuscaling-0.22 (400/400 games)
board size: 19   komi: 7.5
lz-0.17-0.3           162 40.50%       71  35.50%     91  45.50%     31.71
lz-fpuscaling-0.22    238 59.50%       109 54.50%     129 64.50%     32.31
                                       180 45.00%     220 55.00%

lz-0.17-0.3 v lz-fpuscaling-0.27 (400/400 games)
board size: 19   komi: 7.5
lz-0.17-0.3           179 44.75%       80  40.00%     99  49.50%     31.00
lz-fpuscaling-0.27    221 55.25%       101 50.50%     120 60.00%     31.18
                                       181 45.25%     219 54.75%
```

Also, 0.3 fpu reduction doesn't seem to be any better than 0.25 for 0.17 on the same settings...
```
lz-0.17-0.3 v lz-0.17-0.25 (400/400 games)
board size: 19   komi: 7.5
lz-0.17-0.3     196 49.00%       82  41.00%     114 57.00%     31.16
lz-0.17-0.25    204 51.00%       86  43.00%     118 59.00%     31.88
                                 168 42.00%     232 58.00%
```

Some matches at higher visits (-v10000) which were missing from the previous round of tests
```
lz-0.17-0.25 v lz-fpuscaling-0.22 (188/400 games)
board size: 19   komi: 7.5
lz-0.17-0.25           70 37.23%       30 31.91%     40  42.55%    911.75
lz-fpuscaling-0.22    118 62.77%       54 57.45%     64  68.09%    889.85
                                       84 44.68%     104 55.32%
```

Here's my own interpretation of the results so far:
This PR only differs significantly from the current implementation for low policy children. Also, it assumes that there is meaningful information to be extracted from the magnitude of these low policy outputs (e.g. a 1% child is meaningfully better than 0.5% child) Perhaps when running on low thread counts at low visits, we don't put enough visits into these children for a significant difference.

Issue #2323, opened 2019-04-07
Is there a command to do a memory clean without exiting the program?


zhanzhenzhen:
We'd like to use Leela command to do some automatic test matches. Originally, I used the GTP commands to clean the board and then generate moves to play a new game after the old game ended. But this makes it "remember" some positions in the old game, therefore makes the new game unfair. So now I exit and restart Leela for every game. This makes the network loading time a huge problem (our test is just 1 visit per move, so the playing lasts less than 5 seconds but the loading time is 10 seconds). So I have a question: Is there a command to do a memory clean so that we can have a pristine state as if it's just restarted?

l1t1:
you can use a bat file instead of one command

wpstmxhs:
I suggest to make a GTP command that clears search tree and NNCache.
`clear_board` command resets search tree but it does not reset NNCache.
For some purposes(e.g. testing strength), this may lead a biased search.

zhanzhenzhen:
@wpstmxhs Thanks! So I also hope to add a GTP command to clear the NNCache!

alreadydone:
See https://github.com/gcp/leela-zero/pull/2046 for what you need to do.

zhanzhenzhen:
Thanks! That looks pretty good!

PR #2046, opened 2018-11-26
Clear NNCache when clear_board or loadsgf is issued.

alreadydone:
Hopefully fixes lack-of-variability or even duplication issues observed in https://github.com/gcp/leela-zero/issues/2044 and https://github.com/gcp/leela-zero/issues/2006#issuecomment-441488218 when using gogui-twogtp. Technically twogtp isn't using loadsgf to start games with specified openings, though, but instead issue clear_board followed by a bunch of play commands, AFAIK.

PR #2344, opened 2019-04-17
Add "clear_cache" command.

TFiFiE:
As requested in #2323 and based on #2046.
The command does not start with `lz-` despite not being part of the GTP standard, because it likewise exists non-prefixed in GNU Go, just like `last_move`, `move_history` and `printsgf`.

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -399,6 +399,7 @@ const std::string GTP::s_commands[] = {
     "fixed_handicap",
     "last_move",
     "move_history",
+    "clear_cache",
     "place_free_handicap",
     "set_free_handicap",
     "loadsgf",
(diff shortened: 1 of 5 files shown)


wpstmxhs:
I think we also have to clear NNCache statistics, too.
And we should use the mutex lock to avoid race condition (i.e. pondering).
So code should be something like this:

```c++
void NNCache::clear() {
    std::lock_guard<std::mutex> lock(m_mutex);

    m_cache.clear();
    m_order.clear();

    m_lookups = 0;
    m_inserts = 0;
    m_hits = 0;
}
```

`NNCache::resize(int)` also has no mutex lock! It's another problem...

gcp:
>And we should use the mutex lock to avoid race condition (i.e. pondering).

Pondering is done in the main thread and stopped whenever a GTP command arrives, so I don't think there is a risk of racing.

gcp:
>I think we also have to clear NNCache statistics, too.

@TFiFiE any opinion on that? I can see it either way.

TFiFiE:
Not sure. Maybe make clearing the cache stats a separate command.

l1t1:
how to use the command in gogui-twogtp command line?

PR #2348, opened 2019-04-22
Update README with Homebrew instructions for macOS

apetresc:
Happy to report that after over a year of trying, [my latest attempt](https://github.com/Homebrew/homebrew-core/pull/38743) to get leela-zero into Homebrew has finally succeeded. So macOS users don't need to compile LZ by themselves anymore (or even if they do, they can just do it through `brew install --build-from-source leela-zero`). 

This just updates the README to reflect this fact under the "I just want to play with Leela Zero" section (of course, the compilation instructions for macOS are still there in the contributing section)

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,13 @@ a (weaker) network trained from human games [here](https://sjeng.org/zero/best_v
 If you are on Windows, download an official release from [here](https://github.com/leela-zero/leela-zero/releases) and head to the [Usage](#usage-for-playing-or-analyzing-games)
 section of this README.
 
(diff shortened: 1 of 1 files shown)


apetresc:
(I know pull requests are supposed to go against `next` not `master` but since this is a documentation change that applies to the currently-released version, and `master` is still for some reason the default branch, it seems more appropriate this way. Let me know if this, too, should be into `next`)

gcp:
Fixed the default branch, thanks for pointing that out.

Issue #2347, opened 2019-04-19
Double free or corruption (out), with Tensor Cores + Small networks


zliu1022:
Today, I compile leela-zero master branch under Tesla V100 in google cloud.
I tried some weights, 6b128f, 10b128f, 15b192f, 19b256f, 40b256f works well. But 
1b8f, 4b32f, 5b64f,  during the tuning engine says ```double free or corruption (out)```

Would you like give me some help? :-)

```
$ ./leelaz-master -g -w 4b32f.gz
Using OpenCL batch size of 5
Using 10 thread(s).
RNG seed: 13540007540573633150
BLAS Core: built-in Eigen 3.3.7 library.
Detecting residual layers...v1...32 channels...4 blocks.
Initializing OpenCL (autodetecting precision).
Detected 1 OpenCL platforms.
Platform version: OpenCL 1.2 CUDA 10.0.132
Platform profile: FULL_PROFILE
Platform name:    NVIDIA CUDA
Platform vendor:  NVIDIA Corporation
Device ID:     0
Device name:   Tesla V100-SXM2-16GB
Device type:   GPU
Device vendor: NVIDIA Corporation
Device driver: 415.27
Device speed:  1530 MHz
Device cores:  80 CU
Device score:  1112
Selected platform: NVIDIA CUDA
Selected device: Tesla V100-SXM2-16GB
with OpenCL 1.2 capability.
Half precision compute support: No.
Tensor Core support: Yes.
OpenCL: using fp16/half or tensor core compute support.

Started OpenCL SGEMM tuner.
Will try 380 valid configurations.
(2/380) KWG=16 KWI=8 MDIMA=8 MDIMC=8 MWG=32 NDIMB=8 NDIMC=8 NWG=16 SA=1 SB=1 STRM=0 STRN=0 TCE=0 VWM=4 VWN=2 0.0118 ms (782.6 GFLOPS)
(18/380) KWG=32 KWI=8 MDIMA=8 MDIMC=8 MWG=32 NDIMB=16 NDIMC=16 NWG=32 SA=1 SB=1 STRM=0 STRN=0 TCE=0 VWM=2 VWN=2 0.0115 ms (800.0 GFLOPS)
[...]

PR #2360, opened 2019-04-28
Fix tuner buffer overflow when dealing with small nets

ihavnoid:
Fixes issue #2347


ihavnoid:
Since this is a straightforward fix I will merge it in if there is no objection for the next 48 hours.

roy7:
@zliu1022 Would you like to give this a test?

PR #2361, opened 2019-04-29
Take network size in TFProcess constructor

TFiFiE:
It would still be the responsibility of the user to supply the correct size when restoring a model.


TFiFiE:
To make things less error prone, perhaps the number of blocks and filters should not have default values in `parse.py` and be required arguments in its call.

PR #2370, opened 2019-05-03
Support SE nets

Ttl:
Squeeze excitation network support for engine and training. This should be better than the current resnet architecture. I trained one 128x10 net for testing and it seems decent for a small net but much weaker than big nets. Minigo SE nets should be compatible with this but I haven't written any conversion scripts for it.

We might want to wait for bigger better trained SE nets before merging this, so that we don't end up supporting this without having any usable networks. I'll try training 256x20 SE network next, it's probably going to take around two weeks to finish.

### DIFF
--- a/src/CPUPipe.cpp
+++ b/src/CPUPipe.cpp
@@ -49,6 +49,12 @@
 #ifndef USE_BLAS
 // Eigen helpers
 template <typename T>
+using EigenVectorMap =
+    Eigen::Map<Eigen::Matrix<T, Eigen::Dynamic, 1>>;
+template <typename T>
+using ConstEigenVectorMap =
+    Eigen::Map<const Eigen::Matrix<T, Eigen::Dynamic, 1>>;
+template <typename T>
 using EigenMatrixMap =
     Eigen::Map<Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>>;
 template <typename T>
@@ -356,6 +362,7 @@ void batchnorm(const size_t channels,
                std::vector<float>& data,
                const float* const means,
                const float* const stddevs,
+               const bool relu = true,
                const float* const eltwise = nullptr) {
     const auto lambda_ReLU = [](const auto val) { return (val > 0.0f) ?
                                                           val : 0.0f; };
@@ -367,14 +374,92 @@ void batchnorm(const size_t channels,
         if (eltwise == nullptr) {
             // Classical BN
             for (auto b = size_t{0}; b < spatial_size; b++) {
-                arr[b] = lambda_ReLU(scale_stddev * (arr[b] - mean));
+                auto val = scale_stddev * (arr[b] - mean);
+                if (relu) {
+                    val = lambda_ReLU(val);
+                }
+                arr[b] = val;
             }
         } else {
             // BN + residual add
             const auto res = &eltwise[c * spatial_size];
             for (auto b = size_t{0}; b < spatial_size; b++) {
-                arr[b] = lambda_ReLU((scale_stddev * (arr[b] - mean)) + res[b]);
+                auto val = scale_stddev * (arr[b] - mean) + res[b];
+                if (relu) {
+                    val = lambda_ReLU(val);
+                }
+                arr[b] = val;
             }
+        }
+    }
+}
+
+std::vector<float> innerproduct_(const size_t inputs,
+                                const size_t outputs,
+                                const bool ReLU,
+                                const std::vector<float>& input,
+                                const std::vector<float>& weights,
+                                const std::vector<float>& biases) {
+    std::vector<float> output(outputs);
+
+#ifdef USE_BLAS
+    cblas_sgemv(CblasRowMajor, CblasNoTrans,
+                // M     K
+                outputs, inputs,
+                1.0f, &weights[0], inputs,
+                &input[0], 1,
+                0.0f, &output[0], 1);
+#else
+    EigenVectorMap<float> y(output.data(), outputs);
+    y.noalias() =
+        ConstEigenMatrixMap<float>(weights.data(),
+                                   inputs,
+                                   outputs).transpose()
+        * ConstEigenVectorMap<float>(input.data(), inputs);
+#endif
+    const auto lambda_ReLU = [](const auto val) { return (val > 0.0f) ?
+                                                          val : 0.0f; };
+    for (unsigned int o = 0; o < outputs; o++) {
+        auto val = biases[o] + output[o];
+        if (ReLU) {
+            val = lambda_ReLU(val);
+        }
+        output[o] = val;
+    }
+
+    return output;
+}
+
+void global_avg_pooling(const size_t channels,
+                        const std::vector<float>& input,
+                        std::vector<float>& output) {
+
+    for (auto c = size_t{0}; c < channels; c++) {
+        auto acc = 0.0f;
+        for (auto i = size_t{0}; i < NUM_INTERSECTIONS; i++) {
+            acc += input[c * NUM_INTERSECTIONS + i];
+        }
+        output[c] = acc / NUM_INTERSECTIONS;
+    }
+}
+
+void apply_se(const size_t channels,
+              const std::vector<float>& input,
+              const std::vector<float>& res,
+              const std::vector<float>& scale,
+              std::vector<float>& output) {
+
+    const auto lambda_ReLU = [](const auto val) { return (val > 0.0f) ?
+                                                          val : 0.0f; };
+
+    const auto lambda_sigmoid = [](const auto val) { return 1.0f/(1.0f + exp(-val)); };
+
+    for (auto c = size_t{0}; c < channels; c++) {
+        auto gamma = lambda_sigmoid(scale[c]);
+        auto beta = scale[c + channels];
+        for (auto i = size_t{0}; i < NUM_INTERSECTIONS; i++) {
+          output[c * NUM_INTERSECTIONS + i] = lambda_ReLU(gamma * input[c * NUM_INTERSECTIONS+ i] +
+                                                 beta + res[c * NUM_INTERSECTIONS + i]);
         }
     }
 }
@@ -402,8 +487,11 @@ void CPUPipe::forward(const std::vector<float>& input,
                                  m_weights->m_batchnorm_stddevs[0].data());
 
     // Residual tower
+    auto pooling = std::vector<float>(output_channels);
     auto conv_in = std::vector<float>(output_channels * NUM_INTERSECTIONS);
     auto res = std::vector<float>(output_channels * NUM_INTERSECTIONS);
+    auto block = 0;
+    auto se = m_weights->m_se_fc1_w.size() != 0;
     for (auto i = size_t{1}; i < m_weights->m_conv_weights.size(); i += 2) {
         auto output_channels = m_input_channels;
         std::swap(conv_out, conv_in);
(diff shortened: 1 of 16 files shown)


src/kernels/apply_se.opencl
```diff
@@ -0,0 +1,55 @@
+/*
+    This file is part of Leela Zero.
+    Copyright (C) 2017-2018 Gian-Carlo Pascutto and contributors
```
gcp: Put your name and the current year here.

src/kernels/clblast/xgemv.opencl
```diff
@@ -0,0 +1,155 @@
...
+    // Synchronizes all threads in a workgroup
+    barrier(CLK_LOCAL_MEM_FENCE);
+
+    // Loops over the work per thread, and checks whether in bounds
+    #pragma unroll
+    for (int _w = 0; _w < WPT1; _w += 1) {
+      const int gid = _w*get_global_size(0) + get_global_id(0);
+      if (gid < m) {
+
+        // The multiply-add function for the main part (divisable by WGS1)
```
gcp, line 110: Something went wrong with the indentation here.

src/kernels/pooling.opencl
```diff
@@ -0,0 +1,57 @@
+/*
+    This file is part of Leela Zero.
+    Copyright (C) 2017-2018 Gian-Carlo Pascutto and contributors
```
gcp: Same remark.

training/tf/tfprocess.py
```diff
@@ -544,28 +537,58 @@ def batch_norm(self, net):
...
+        self.weights.append(b_fc2)
 
-        self.add_weights(W_conv)
+        net = tf.add(tf.matmul(net, W_fc2), b_fc2)
+        net = tf.reshape(net, [-1, 2 * channels, 1, 1])
+
+        # Split to scale and bias
+        gammas, betas = tf.split(net, 2, axis=1)
+
+        out = tf.nn.sigmoid(gammas) * x + betas
```
gcp, line 598: These are SE-FiLM units rather than the ones from the original paper, correct?

training/tf/tfprocess.py
```diff
@@ -534,7 +527,7 @@ def add_weights(self, var):
             assert var.dtype.base_dtype == tf.float32
             self.weights.append(var)
 
-    def batch_norm(self, net):
+    def batch_norm(self, net, scale=True):
```
gcp, line 532: Yay for finally fixing this. But there's a format description in the README.md, so we'll have to decide what to do with that.

src/Network.cpp
```diff
@@ -280,19 +282,99 @@ std::pair<int, int> Network::load_v1_network(std::istream& wtfile) {
                     linecount + 2); //+1 from version line, +1 from 0-indexing
             return {0, 0};
         }
-        if (linecount < plain_conv_wts) {
-            if (linecount % 4 == 0) {
+        if (linecount < input_weights && version == 3) {
```
gcp, line 285: Maybe this part of the parsing should be factored out into a separate file.

TFiFiE:
Instead of just increasing the version number again, wouldn't this be a good opportunity to introduce a more dynamic metadata header describing the general format of the network?

Grant-Tao:
@Ttl why not using the ready 00961-atermis？ changing a file's format is much easier to training a new net.

src/CPUPipe.cpp
```diff
@@ -367,18 +374,96 @@ void batchnorm(const size_t channels,
...
+void apply_se(const size_t channels,
+              const std::vector<float>& input,
+              const std::vector<float>& res,
+              const std::vector<float>& scale,
+              std::vector<float>& output) {
+
+    const auto lambda_ReLU = [](const auto val) { return (val > 0.0f) ?
+                                                          val : 0.0f; };
+
+    const auto lambda_sigmoid = [](const auto val) { return 1.0f/(1.0f + exp(-val)); };
```
TFiFiE: `std::exp`

src/CPUPipe.cpp
```diff
@@ -367,18 +374,96 @@ void batchnorm(const size_t channels,
...
+
+    for (auto c = size_t{0}; c < channels; c++) {
+        auto gamma = lambda_sigmoid(scale[c]);
+        auto beta = scale[c + channels];
```
TFiFiE: These and some other variables elsewhere can be made `const`.

src/CPUPipe.cpp
```diff
@@ -367,18 +374,96 @@ void batchnorm(const size_t channels,
...
         }
     }
 }
 
+std::vector<float> innerproduct_(const size_t inputs,
+                                const size_t outputs,
+                                const bool ReLU,
+                                const std::vector<float>& input,
+                                const std::vector<float>& weights,
+                                const std::vector<float>& biases) {
```
TFiFiE, line 402: Alignment.

src/CPUPipe.cpp
```diff
@@ -367,18 +374,96 @@ void batchnorm(const size_t channels,
...
+                                                          val : 0.0f; };
+
+    const auto lambda_sigmoid = [](const auto val) { return 1.0f/(1.0f + exp(-val)); };
+
+    for (auto c = size_t{0}; c < channels; c++) {
+        auto gamma = lambda_sigmoid(scale[c]);
+        auto beta = scale[c + channels];
+        for (auto i = size_t{0}; i < NUM_INTERSECTIONS; i++) {
+          output[c * NUM_INTERSECTIONS + i] = lambda_ReLU(gamma * input[c * NUM_INTERSECTIONS+ i] +
+                                                 beta + res[c * NUM_INTERSECTIONS + i]);
```
TFiFiE: Either align this or make the indentation 4 spaces.

wpstmxhs:
> Instead of just increasing the version number again, wouldn't this be a good opportunity to introduce a more dynamic metadata header describing the general format of the network?

I agree on this opinion. It's not efficient to make new version number whenever new network structure comes out. and management of version number is also hard work.

Instead of that, It seems good to make descriptive format of network structure as network weight file format (XML? JSON?). If so, we don't need a version number. because LZ makes network directly from described network structure of weight file, instead of using version number to distinguish.

For backward compatibility, we can use `-1` or `0` or some nonsense number like `999` as version number in old format. and after version tag, we can use completely different weight format, as Leela Zero gives up to read when it finds different version number:
```c++
    auto line = std::string{};
    auto format_version = -1;
    if (std::getline(buffer, line)) {
        auto iss = std::stringstream{line};
        // First line is the file format version id
        iss >> format_version;
        if (iss.fail() || (format_version != 1 && format_version != 2)) {
            myprintf("Weights file is the wrong version.\n");
            return {0, 0};
        } else {
//... omitted ...
```

Ttl:
This is the first time network version has been increased since this project started, so it doesn't look like there is much demand for a better network file format. I don't have anything against it though if you want to implement it.

Lc0 has used protobuf with pretty good success. It works pretty well and it is easy to add new fields to it.

wpstmxhs:
@Ttl Yes, protobuf is also one of possible candidates.
It's exactly second time. PhoenixGo had slightly different network structure. when we were trying to convert PhoenixGo into Leela Zero format, we used version 2 and gave up compatibility with Leela Zero weights, but in fact, we also had to use different version number for it.

l1t1:
https://github.com/leela-zero/leela-zero/issues/2380 the solution of https://github.com/leela-zero/leela-zero/files/3164660/leela-zero-mg-0.17-win64.zip is very good.

wpstmxhs:
@l1t1 I don't think my approach is good. I assumed that MiniGo is always in 19b or 39b. So I could check what network type it is. Because both of (19 * 12) % 8 and (39 * 12) % 8 are not zero.

Without this assumption, let's say SE with 40 blocks, the program will not determine correctly.

So we should consider new version tag or new weights format to distinguish MiniGo SE format for general use.

gcp:
>Instead of that, It seems good to make descriptive format of network structure as network weight file format (XML? JSON?). If so, we don't need a version number. because LZ makes network directly from described network structure of weight file, instead of using version number to distinguish.

There's already an issue here with a ton of bike-shedding about possible new network formats: https://github.com/leela-zero/leela-zero/issues/1751

In my opinion the advantages of protobuf are mostly theorethical. You can see that also in lc0, the *engine has to be adapted to actually support the new things in the network format* anyway. There's also things like ONNX, which are a bit better that at least for some backends the latter part wouldn't be true.

As long as there's no complete patches for engine + training this is all theory.

wpstmxhs:
@gcp I see. But I think making new format is actually not so hard.

Of course, It's hard to support fully in the first step, so let's not think about full dynamic support of engine/training side at first.

However, for this project's future, we have to consider about it someday.

training/tf/tfprocess.py
Ttl: Yes, they have a bias term calculated from pooling. This form was better in lc0 and setting biases to zero gives the original SE form. Performance difference is essentially zero.

Ttl:
It doesn't look like the 20x256 SE network I've been training is going anywhere. It's been trained about 300M positions and there hasn't been much progress in the last 100M positions.  The strength is only slightly stronger than net 157 at equal nodes. Not sure what's the issue. Does it require >1B positions to get good or was the issue that I didn't go through the training data like in RL? Anyway, I don't think I'm continuing it.

I updated the minigo converter for this format, the latest v17 run nets are pretty strong. But other than those there isn't really any good nets to use with this PR yet.

remdu:
There is a recent paper that claims to improve over SE https://arxiv.org/abs/1904.11492

TFiFiE:
Would anyone be willing to test this code with smaller networks to see whether it is at least significantly effective there? Otherwise, there might be a bug here.

bubblesld:
After

git clone -b se https://github.com/TFiFiE/leela-zero

What to do next to compile?
I did the following:

cd leela-zero
git submodule update --init --recursive
sudo apt install libboost-dev libboost-program-options-dev libboost-filesystem-dev opencl-headers ocl-icd-libopencl1 ocl-icd-opencl-dev zlib1g-dev
mkdir build && cd build
cmake ..
cmake --build .

But the leelaz is not working with the senet I trained.
I do now know if the following is a problem. When I do " cmake ..", there is an error message:

-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed

Ttl:
Building should be identical to the normal leela-zero. Does compiling the normal leela-zero work? Make sure that you don't have existing `leela-zero` folder where you are cloning.

cmake shouldn't give error. You can try `sudo apt install build-essential` to fix it.

bubblesld:
Yes, compiling the normal leela-zero works. I do  use separate folders.

When I ran leelaz with lizze, the error message is:

Detecting residaul layers...v3...256 channels...
Inconsistent number of weights in the file.

bubblesld:
> cmake shouldn't give error. You can try `sudo apt install build-essential` to fix it.

I did this. And it says
build-essential is already the newest version (12.4ubuntu1).

Ttl:
Ok. Sounds like the compiling works, but there is something wrong with the weight file and it doesn't have the right amount of weights. If you extract it there should be 300 lines in the text file if it's 20x256 weight. Net with `b` blocks should have `20+14*b` lines. The last line should always have only a single number.

Here is one 20x256 weight file that should be working so you can test the binary: http://hforsten.com/leelaz/leelaz-model-swa-2824000.txt.gz .

bubblesld:
> Here is one 20x256 weight file that should be working so you can test the binary: http://hforsten.com/leelaz/leelaz-model-swa-2824000.txt.gz .

Yes, it works with your file. Then what should I do to my file?

bubblesld:
[This one](https://drive.google.com/open?id=19EQrSdlqNBGU6dcDRyucORbltCEGQKzp) is now about the same strength vs v173 (198:202). I will keep training to see if it can catch up the other 20b.

src/OpenCL.cpp
```diff
@@ -265,25 +298,60 @@ void OpenCL_Network<net_t>::forward(const std::vector<float>& input,
...
-                      use_inout, skip_next_in_trans, true,
-                      batch_size);
+
+            if (niter->is_se_unit) {
+                // Residual connection and relu are applied in SE-unit.
+                convolve3(opencl_context,
+                          layer.channels,
+                          layer.outputs,
+                          inBuffer2,
+                          inBuffer2,
```
TFiFiE, line 315: @Ttl, is `inBuffer2` actually supposed to be passed twice here?

PR #2401, opened 2019-05-24
Improvement to the UCT search in Leela.

hauensteina:
I did a validation run pitting the release search 0.17 against my modified version,
both running network d42f950a. I will repeat with a20c31da.

Command line and outcome:

validation/validation -g 1 -k sgfs
                      -n best-network -o '-g -m 0 -p 0 -v 1600 --noponder -t 6 --batchsize 5 -r 5 -w'
		      -n best-network -o '-g -m 0 -p 0 -v 1600 --noponder -t 6 --batchsize 5 -r 5 -w'
		      leelaz leelaz-x

Stopping engine.
47 wins, 71 losses
best-net v best-net ( 118 games)
              wins        black       white
best-net   47 39.83%   21 38.89%   26 40.63%
best-net   71 60.17%   33 61.11%   38 59.38%
                       54 45.76%   64 54.24%

It would be great if somebody could take the time to reproduce this to make sure it's not a fluke.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -324,9 +324,8 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     for (auto& child : m_children) {
         idx++;
         if (visits[idx] > 0) {
-            smallest_winrate = std::min( smallest_winrate, winrates[idx]);
-        }
-        else {
+            smallest_winrate = std::min(smallest_winrate, winrates[idx]);
+        } else {
             winrates[idx] = smallest_winrate;
         }
         if (child.valid()) {
@@ -355,8 +354,7 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
         if (value > best_value) {
             if (child.is_inflated() && child->m_expand_state.load() == ExpandState::EXPANDING) {
                 // Someone else is expanding this node, never select it
-            }
-            else {
+            } else {
                 best_value = value;
                 best = &child;
             }
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -302,20 +302,26 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     wait_expanded();
 
     // Children are sorted by descending policy.
-    // An unvisited child gets the lowest winrate to the left.
+    // An unvisited child gets the lowest winrate *to the left* .
     // We need the array because an unvisited child can be followed by visited ones.
-    // Also, cache everything to avoid potential threading inconsistencies.
+    // Also, cache everything to minimize threading inconsistencies.
     // This is not paranoia. I tried.
     float winrates[361]; // On the stack. No mallocs.
     float visits[361];
-    float smallest_winrate = get_net_eval(color); // Important. Do no start with anything else.
-    auto parentvisits = size_t{0};
     int idx = -1;
+    // Do this as quickly as possible. Ideally we should lock the node.
     for (auto& child : m_children) {
         idx++;
         visits[idx] = child.get_visits();
+        if (visits[idx]) { winrates[idx] = child.get_eval(color); }
+    }
+
+    float smallest_winrate = get_net_eval(color); // Important. Do no start with anything else.
+    auto parentvisits = size_t{0};
+    idx = -1;
+    for (auto& child : m_children) {
+        idx++;
         if (visits[idx] > 0) {
-            winrates[idx] = child.get_eval(color);
             smallest_winrate = std::min( smallest_winrate, winrates[idx]);
         }
         else {
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -324,8 +324,15 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     for (auto& child : m_children) {
         idx++;
         if (visits[idx] > 0) {
-            smallest_winrate = std::min(smallest_winrate, winrates[idx]);
-        } else {
+            if (child.active()) {
+                if (child.is_inflated() && child->m_expand_state.load() == ExpandState::EXPANDING) {
+                    // // Someone else is expanding this node, do nothing
+                } else {
+                    smallest_winrate = std::min( smallest_winrate, winrates[idx]);
+                }
+            }
+        }
+        else {
             winrates[idx] = smallest_winrate;
         }
         if (child.valid()) {
@@ -354,7 +361,8 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
         if (value > best_value) {
             if (child.is_inflated() && child->m_expand_state.load() == ExpandState::EXPANDING) {
                 // Someone else is expanding this node, never select it
-            } else {
+            }
+            else {
                 best_value = value;
                 best = &child;
             }
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -328,7 +328,7 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
                 if (child.is_inflated() && child->m_expand_state.load() == ExpandState::EXPANDING) {
                     // // Someone else is expanding this node, do nothing
                 } else {
-                    smallest_winrate = std::min( smallest_winrate, winrates[idx]);
+                    smallest_winrate = std::min(smallest_winrate, winrates[idx]);
                 }
             }
         } else {
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -301,15 +301,28 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     (void)is_root;
     wait_expanded();
 
-    // Count parentvisits manually to avoid issues with transpositions.
-    auto total_visited_policy = 0.0f;
+    // Children are sorted by descending policy.
+    // An unvisited child gets the lowest winrate to the left.
+    // We need the array because an unvisited child can be followed by visited ones.
+    // Also, cache everything to avoid potential threading inconsistencies.
+    // This is not paranoia. I tried.
+    float winrates[361]; // On the stack. No mallocs.
+    float visits[361];
+    float smallest_winrate = get_net_eval(color); // Important. Do no start with anything else.
     auto parentvisits = size_t{0};
-    for (const auto& child : m_children) {
+    int idx = -1;
+    for (auto& child : m_children) {
+        idx++;
+        visits[idx] = child.get_visits();
+        if (visits[idx] > 0) {
+            winrates[idx] = child.get_eval(color);
+            smallest_winrate = std::min( smallest_winrate, winrates[idx]);
+        }
+        else {
+            winrates[idx] = smallest_winrate;
+        }
         if (child.valid()) {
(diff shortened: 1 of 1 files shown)


src/UCTNode.cpp
```diff
@@ -298,6 +298,7 @@ void UCTNode::accumulate_eval(float eval) {
 }
 
 UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
+    (void)is_root;
```
roy7, line 301: This line doesn't do anything does it? :)

Ishinoshita:
@hauensteina Hi! Can you please summarize in a few words the idea behind your changes ?
I see in comment "An unvisited child gets the lowest winrate seen so far". So, I understand you propose to replace, for unvisited children, (original parent NN eval - fpu_reduction) by min(Q) over visited children. Is that correct ? Is there something else ? Thanks !

src/UCTNode.cpp
hauensteina: It suppresses the unused var warning. I wanted to keep the method signature.
We could alternatively get rid of the is_root parameter in this version.

hauensteina:
> @hauensteina Hi! Can you please summarize in a few words the idea behind your changes ?
> I see in comment "An unvisited child gets the lowest winrate seen so far". So, I understand you propose to replace, for unvisited children, (original parent NN eval - fpu_reduction) by min(Q) over visited children. Is that correct ? Is there something else ? Thanks !

You are correct, and there is nothing else.
I will try the second smallest next to see if the added stability helps.
I will also try the median(Q over visited children).

wonderingabout:
@hauensteina 

interesting work :+1: 
regardless of what your final results are, keep up ! :1st_place_medal:

l1t1:
pls attach a windows exe file

hauensteina:
Can somebody explain why any checks failed? 
This was a minor change inside one function only.
Since I don't build on Windows, I cannot attach a windows executable.

hauensteina:
Anybody interested?
The new code is much simpler, eliminates two domain specific constants, and is
reproducibly stronger.

These two go away: 
cfg_fpu_root_reduction and cfg_fpu_reduction .

So does this formula:
fpu_reduction = (is_root ? cfg_fpu_root_reduction : cfg_fpu_reduction) * std::sqrt(total_visited_policy);

What's not to like?

src/UCTNode.cpp
```diff
@@ -314,37 +315,52 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
...
-    // Estimated eval for unknown nodes = original parent NN eval - reduction
-    const auto fpu_eval = get_net_eval(color) - fpu_reduction;
 
     auto best = static_cast<UCTNodePointer*>(nullptr);
     auto best_value = std::numeric_limits<double>::lowest();
 
+    // Get winrates for visited children.
+    // This goes from best to worst policy value.
+    // An unvisited child gets the lowest winrate seen so far.
+    double winrates[m_children.size()];
```
alreadydone: This will only compile in gcc. C++ standard doesn't allow arrays whose size is not a compile time constant.
get_eval is cheap so there is no need to cache it.

src/UCTNode.cpp
```diff
@@ -314,37 +315,52 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
...
+        if (child.get_visits() > 0) {
+            winrates[idx] = child.get_eval(color);
+            if (winrates[idx] < smallest_winrate) {
+                smallest_winrate = winrates[idx];
+            }
         }
+        else {
+            winrates[idx] = smallest_winrate;
+        }
+    } // for
```
alreadydone: Does this prevent your code from passing style checks? It's currently failing travis and appveyor.

src/UCTNode.cpp
```diff
@@ -314,37 +315,52 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
...
+    double winrates[m_children.size()];
+    int idx = -1;
+    double smallest_winrate = get_net_eval(color);
     for (auto& child : m_children) {
-        if (!child.active()) {
-            continue;
+        idx++;
+        if (child.get_visits() > 0) {
+            winrates[idx] = child.get_eval(color);
+            if (winrates[idx] < smallest_winrate) {
```
alreadydone: `smallest_winrate = std::min(smalleste_winrate, winrates[idx]);` would be more straightforward (and maybe faster)

src/UCTNode.cpp
```diff
@@ -314,37 +315,52 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
...
-            winrate = -1.0f - fpu_reduction;
-        } else if (child.get_visits() > 0) {
-            winrate = child.get_eval(color);
+    idx = -1;
+    for (auto& child : m_children) {
+        idx++;
+        if (!child.active()) {
+            continue;
         }
+        auto winrate = winrates[idx];
```
alreadydone: No need to use cached value, just make another `get_eval` call.

src/UCTNode.cpp
```diff
@@ -314,37 +315,52 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
...
         assert(value > std::numeric_limits<double>::lowest());
 
         if (value > best_value) {
-            best_value = value;
-            best = &child;
+            if (child.is_inflated() && child->m_expand_state.load() == ExpandState::EXPANDING) {
+                // Someone else is expanding this node, never select it
+            }
+            else {
+                best_value = value;
```
alreadydone, line 364: Change this to `best_value = smallest_winrate + puct` would do what you want. (Edit: that doesn't quite work.)

alreadydone (changes requested):

src/UCTNode.cpp
hauensteina: Thanks for the review. Will fix.

src/UCTNode.cpp
hauensteina: I need the array because sometimes unvisited nodes are followed by visited ones. 
I want the minimum to the *left* of the unvisited node, not the overall minimum.
Will use a vector.

src/UCTNode.cpp
hauensteina: I do not trust repeated get_eval() calls when multithreading. 
It gets updated as descendants are expanded.

hauensteina:
Added three commits to fix the Windows issue and reduce threading issues.
The caching is necessary. Huge strength difference.

hauensteina:
Do I need to do anything else to get my recent changes reviewed?
Not sure about the process.

src/UCTNode.cpp
```diff
@@ -301,15 +301,34 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
...
     wait_expanded();
 
-    // Count parentvisits manually to avoid issues with transpositions.
-    auto total_visited_policy = 0.0f;
+    // Children are sorted by descending policy.
+    // An unvisited child gets the lowest winrate *to the left* .
+    // We need the array because an unvisited child can be followed by visited ones.
+    // Also, cache everything to minimize threading inconsistencies.
+    // This is not paranoia. I tried.
+    float winrates[361]; // On the stack. No mallocs.
```
alreadydone: `POTENTIAL_MOVES` instead of 361 (should be 362, there's also the pass move).

src/UCTNode.cpp
```diff
@@ -301,15 +301,34 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
...
+    float visits[361];
+    int idx = -1;
+    // Do this as quickly as possible. Ideally we should lock the node.
+    for (auto& child : m_children) {
+        idx++;
+        visits[idx] = child.get_visits();
+        if (visits[idx]) { winrates[idx] = child.get_eval(color); }
```
alreadydone: won't pass style check

src/UCTNode.cpp
alreadydone: If the strength improvement is sensitive to multithreading, maybe the change should be tested with more varied thread count and batch size settings? Other opinions from @gcp and other leela-zero members?

alreadydone (changes requested):

src/UCTNode.cpp
alreadydone: Even though self-play uses fixed batch size and thread count settings, the number of parallel games could affect things. Notice that FPU reduction was introduced when we use only one thread in self-play, and I don't think it has been tested against any other choice of FPU in multithreading setting.

hauensteina:
Threading issues might have affected the old implementation, too.
It uses get_visits() in a loop with quite some computation.
That count changes all the time AFAIK.
It might be worth while running the old method with caching against itself
without caching to get a feeling for this.

hauensteina:
I've done many experiments, but nothing beats the simple minimum of the children to the left.
Things I've tried:
median, percentiles, average, partial average, weighted average, second smallest,
average times a discount factor (that one came close), the minimum times 0.95,
and exponential decay.
Nothing was better.

Finally I added the caching arrays to the old algorithm and tested again,
and the minimum is still a lot better.

If you can improve on this, let me know. I need a break.

src/UCTNode.cpp
alreadydone: `POTENTIAL_MOVES` is a macro defined in config.h and allows the code to be used for any odd board size, even greater than 19.

hauensteina:
The bottom loop has conditions child.active() and state != EXPANDING.
I added the same conditions to the loop over the left siblings, and it resulted in 
another measurable improvement.

hauensteina:
> The bottom loop has conditions child.active() and state != EXPANDING.
> I added the same conditions to the loop over the left siblings, and it resulted in
> another measurable improvement.

Actually, with the conditions, it beat the min version without the conditions.
When I let it play the original Leela Zero, I did not see a further improvement.
It appears things are not transitive. 

Anybody else done any testing yet?

hauensteina:
Some preliminary experiments seem to indicate that it will get even stronger if you multiply the min by 0.8 or so. Anybody wants to tweak that?

alreadydone:
Recently in another attempt https://github.com/leela-zero/leela-zero/pull/2342#issuecomment-484746753  to modify the FPU formula, cfg_fpu_reduction has been tuned with CLOP (it's exposed through command line when `USE_TUNER` is defined). You could do the same thing for your parameter.

hauensteina:
I wasted some more time on that parameter. Nothing reproducible or significant, I think. Also,
it is a lot nicer to just not have another magic number. So 1.0 is fine.

src/UCTNode.cpp
```diff
@@ -298,51 +298,72 @@ void UCTNode::accumulate_eval(float eval) {
...
-            parentvisits += child.get_visits();
-            if (child.get_visits() > 0) {
-                total_visited_policy += child.get_policy();
+    // Children are sorted by descending policy.
+    // An unvisited child gets the lowest winrate *to the left* .
+    // We need the array because an unvisited child can be followed by visited ones.
+    // Also, cache everything to minimize threading inconsistencies.
+    // This is not paranoia. I tried.
+    float winrates[362]; // On the stack. No mallocs.
+    int visits[362];
```
TFiFiE, line 310: Use the `POTENTIAL_MOVES` constant.

Issue #2389, opened 2019-05-16
ROCm OpenCL error


iopq:
Fresh Ubuntu 18 install. I installed padoka PPA to run my DXVK Wine stuff. Then I installed ROCm to get performance back.

```
/tmp/AMD_2357_18/t_2357_20.cl:4:5: error: implicit declaration of function 'asm' is invalid in OpenCL
    asm(
    ^
/tmp/AMD_2357_18/t_2357_20.cl:6:96: error: expected ')'
        "wmma.load.a.sync.aligned.m16n16k16.shared.row.f16 {a0,a1,a2,a3,a4,a5,a6,a7}, [%0];\n" : : "l"(ptr)
                                                                                               ^
/tmp/AMD_2357_18/t_2357_20.cl:4:8: note: to match this '('
    asm(
       ^
2 errors generated.
```

Now, it still runs, and it's actually faster than the AMD GPU PRO driver.

ihavnoid:
Hmm, that part is for the nvidia tensor core.  leelaz detects whether the GPU has tensor cores by trying a short inline assembly code, and if it fails to compile, it assumes it has no tensor core.  Usually the error message shouldn't be printed out (leelaz should catch and discard them) but I wonder whether ROCm works differently?

ihavnoid:
@iopq can you track down which line of code triggers this message?, and which SHA you are using?

iopq:
I'm on commit db5569ce8d202f77154f288c21d3f2fa228f9aa3

I have no idea how to track down where this message is coming from

ihavnoid:
The simplest way would be build the whole thing in debug mode, hook up a debugger, and run it step by step.  Or, you can put printf messages everywhere and try to pinpoint where it happens - this probably is somewhere in OpenCL.cpp or Tuner.cpp.  I am not sure if you have the right skillset to do something like this...?

iopq:
It's first in OpenCL.cpp, then it calls .build() from cl2.hpp as far as I can tell

then that calls ::clBuildProgram and returns with buildErrHandler which may or may not throw an exception based on whether they are enabled or not

ihavnoid:
I tried to come up with the simplest solution I can imagine, and just put in a PR that simply skips Tensor Core support check routines unless it's an NVIDIA GPU.  We don't expect any non-NVIDIA GPU to support Tensor Cores, right?

PR #2422, opened 2019-06-05
Don't bother trying tensor core support unless it's an NVIDIA GPU

ihavnoid:
This is to try address #2389 - strictly speaking this shouldn't be necessary but this is to prevent any oddities


ihavnoid:
I wished there to be some people testing this on Radeons... but the commit looks harmless enough.

PR #2423, opened 2019-06-05
Do not play too quickly when main time is going to exhaust

thynson:
Current time manager tries to accumulate time as much as possible when there is no alternative moves, but this make leelaz plays too quickly when the main time is going to exhaust. 
This PR implements a policy that ensure each move takes a minimal thinking time, which grows up as main time is used.

### DIFF
--- a/src/TimeControl.cpp
+++ b/src/TimeControl.cpp
@@ -324,9 +324,9 @@ int TimeControl::get_moves_expected(int boardsize, size_t movenum) const {
 }
 
 // Returns true if we are in a time control where we
-// can save up time. If not, we should not move quickly
+// should save up time. If not, we should not move quickly
 // even if certain of our move, but plough ahead.
-bool TimeControl::can_accumulate_time(int color) const {
+bool TimeControl::should_accumulate_time(int color) const {
     if (m_inbyo[color]) {
         // Cannot accumulate in Japanese byo yomi
         if (m_byoperiods) {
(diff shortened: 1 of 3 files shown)


TFiFiE:
I don't think this change is correct, because `can_accumulate_time` only comes into play whenever the prospect of wasting unused time is the only thing that could possibly stop the engine from immediately playing a move. That applies regardless of how little main time is left.
If you believe LZ plays too quickly under certain circumstances, you should address it in other parts of the time control code.

thynson:
> I don't think this change is correct, because `can_accumulate_time` only comes into play whenever the prospect of wasting unused time is the only thing that could possibly stop the engine from immediately playing a move. That applies regardless of how little main time is left.
> If you believe LZ plays too quickly under certain circumstances, you should address it in other parts of the time control code.

I think that's totally a time strategy and not related to search or other part. If the default time manage behavior need to be preserved, the TimeManager itself need to be refactered to support different policy rather than modify the other parts.

P.S. The change was made to support the team represents leela zero which attended 2019 “CHINA SECURITIES Cup” World AI Weiqi Open Competition. In my test it out perform the default one in a large margin in high playout match. And I think the it also fit general usage like trains with AI or fight with other AI.

TFiFiE:
If you are of the opinion that the maximum time the engine decides to take for a move is sometimes too little, but this can get "masked" by the fact that it wants to avoid wasting time, then the answer is not to change the nature of what is currently a completely objective function to determine whether stopping is a waste, so that it can "catch" more cases, but rather to work on the right calculation of the maximum time in the first place.
Here this means you should instead try to fine-tune the output of `TimeControl::max_time_for_move` so that the value for `time_for_move` in `UCTSearch::have_alternate_moves` is larger in those situations you consider problematic.

thynson:
> If you are of the opinion that the maximum time the engine decides to take for a move is sometimes too little

This PR does not want to increase the maximum time to think, but to solve the issue the original author has stated in the [comment](https://github.com/leela-zero/leela-zero/pull/2423/files#diff-92babf732b3178e5ba6873368d29e38cL342). 

As I renamed the function `can_accumulate_time` to `should_accumulate_time` implys, my solution is that, based on main time, elpased time and byoyomi, the time manager decides whether search *should* stop. 

Such change won't affect the time usage in the opening too much, main time is still accumulated as much as possible in the opening stage, and the accumulated main time will be properly used when byoyomi is comming.

TFiFiE:
Your change would have the engine use up time that would otherwise be available in the future, to effectively ponder on its own turn after it has already effectively decided what it's going to play. Surely you would agree that that is obviously less optimal than instead using that time after receiving the opponent's reply? If, however, this change does actually lead to a strength improvement, then that means there is a legitimate issue with the time management, but you should rather want to increase the maximum thinking time for what would be the next move in such a situation.

PR #2451, opened 2019-07-17
Fix a github security warning

roy7:


roy7:
I don't know how to get rid of all the garbage history but I assume "squash and merge" will solve that if we merge this.

roy7:
Ug I'll remove those wrong README changes.

Done.

PR #2474, opened 2019-08-13
Assert UCTNode pointer alignment requirement.

TFiFiE:


TFiFiE:
There are probably no (relevant) systems for which this assert would fail; it is primarily added to aid the reader in understanding this hack by exposing an otherwise implicit precondition.
Alternatively, a compile-time assert might also be an option:
`static_assert(alignof(UCTNode) % 4 == 0, "message");`

PR #2485, opened 2019-09-03
Tensor core performance optimization

ihavnoid:
Reduce number of loads by loading data within the loop

Roughly 20% faster on my system with i7-7700 and RTX 2080 Super


Ttl:
I see also about 20% improvement on RTX 2080 when comparing with the new tuning. Compared to the old non-exhaustive tuning with smaller NWG and KWG difference is closer to 30%. Seems to work fine in some quick testing.

PR #2496, opened 2019-09-17
Logit Q keeps exploration focused when winrate near 0 or 1

Naphthalin:
This PR is the equivalent of the corresponding PR925 of LeelaChessZero [https://github.com/LeelaChessZero/lc0/pull/925](https://github.com/LeelaChessZero/lc0/pull/925).

The basic idea is that when the game (or some part of the tree) is near decided, search goes super wide because meaningful score differences get smaller. This is counteracted by calculating ``winrate + puct`` in the logit space which unfolds the bounded interval (only ``[0,1]`` instead of ``[-1,1]`` since winrates are used in leelaz instead of q values).

### DIFF
--- a/src/GTP.cpp
+++ b/src/GTP.cpp
@@ -97,6 +97,7 @@ FILE* cfg_logfile_handle;
 bool cfg_quiet;
 std::string cfg_options_str;
 bool cfg_benchmark;
+bool cfg_use_logitQ;
 bool cfg_cpu_only;
 AnalyzeTags cfg_analyze_tags;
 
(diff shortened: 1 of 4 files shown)


Naphthalin:
Command line option is ``--use-logitQ``.

Some comment on ``UCTNode.cpp:323``: It would be preferrable to treat fpu in logit space as well. However the currently chosen ``fpu_reduction`` is too small for that, resulting in a big number of 1 visit moves. Either the value gets increased, or the logic of ``fpu_reduction`` near decided evals should be reconsidered if that is relevant to anything.

dbosst:
So, does this have the effect of eliminating random play in the endgame for decided, no resign games?

Naphthalin:
> So, does this have the effect of eliminating random play in the endgame for decided, no resign games?

I can't tell yet if it actually eliminates the random play for these games since that still depends on evals being somehow meaningful, but since it amplifies small differences it makes the search focus more on high policy moves if their evals are not worse than expected. I just let lz play 2 games to the end with resign disabled. Apart from the occasional dame stone the winning side played very sensible until the end while the losing side tried a few hopeless things and in the end refused to pass for ~20 moves because passing would lose despite having pass as the top move.

PR #2498, opened 2019-09-18
Use get_raw_eval() instead of get_net_eval() for fpu

Naphthalin:
Currently we use ``fpu_eval = get_net_eval(color) - fpu_reduction``. This however has a big downside: Whenever ``net_eval`` is either higher or lower than the best child eval by a substantial amount, this will result in undesired behavior:

1. ``net_eval`` is higher than best child eval + fpu: in this case, all unvisited moves will receive a single visit, irrespective of their policy since ``puct=0`` would be enough
2. ``net_eval`` is lower than best child eval: in this case, ``puct`` does not only have to overcome fpu but also the difference to the best child, which makes the exploration more dependent on the NN eval.

Both cases appear more frequently when trying things like logitQ #2496 but also appear sometimes with regular search, especially in unbalanced situations.

Basing fpu on the current eval of the parent (which should be close to best child) ensures that fpu has to be counteracted by the puct term only, which nearly directly translates ``(fpu, visits)`` into ``smallest visited policy``, in our case with ``fpu=0.25``and ``visits=800`` around 1%, depending on the share of high policy moves.

Doing this would result both in more accurate evals since low policy moves don't get visits too fast and in more sensible (since policy-driven) play when ahead/behind.

### DIFF
--- a/src/UCTNode.cpp
+++ b/src/UCTNode.cpp
@@ -303,21 +303,31 @@ UCTNode* UCTNode::uct_select_child(int color, bool is_root) {
     // Count parentvisits manually to avoid issues with transpositions.
     auto total_visited_policy = 0.0f;
     auto parentvisits = size_t{0};
+    // Get highest child eval for fpu_reduction
+    auto eval = 0.0f;
+    auto best_eval = std::numeric_limits<double>::lowest();
     for (const auto& child : m_children) {
         if (child.valid()) {
             parentvisits += child.get_visits();
             if (child.get_visits() > 0) {
                 total_visited_policy += child.get_policy();
+                eval = child.get_eval(color);
+                if (eval > best_eval) {
+                    best_eval = eval;
+                }
             }
         }
+    }
+    if (parentvisits == 0) {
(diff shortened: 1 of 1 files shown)


Naphthalin:
Simple alternative which would probably do roughly the same might be just replacing ``get_net_eval(color)`` by ``get_raw_eval(color)`` which would already counteract the randomness i mentioned above. ~~However this doesn't work in the intended way since the newly discovered nodes pull down the eval, forcing more new nodes to be evaluated.~~

22nsuk:
In my test, PR2401 is more powerful than the latest version. How about combining PR2401 with PR2498? If that's possible.

Naphthalin:
I agree that there will be probably more sophisticated fpu strategies as #2401 is attempting; I think though that #2401 would be a replacement for the current fpu strategy, so they aren't to be combined. 
I just want to argue that the change of this PR is how the current relative fpu strategy was meant to be implemented in the first place (and is also implemented in Lc0) but wasn't done for some reason.

PR #2597, opened 2020-05-21
More const correctness.

TFiFiE:


wonderingabout:
adding const is great!!!

PR #2641, opened 2021-08-24
Update recompute link

MaxVanDijck:
Previous link is dead / redirects to some spam advertising
Changed to link proposed in https://github.com/leela-zero/leela-zero/issues/2490


roy7:
Thanks!

PR #2672, opened 2025-09-21
Updated gtest submodule to fix compilation error of unintialized variable using GCC 13.3.0 on Ubuntu 24.04 LTS

Schmitty2005:
Updated gtest to latest commit.  Build now compiles without having to manually fix gtest file.  All tests pass as well.

https://github.com/google/googletest/blob/cafa9e52285d743d1394900fa95662950f712a72/googletest/src/gtest-death-test.cc#L1008C1-L1016C1

Line 1008 and on of commit $cafa9e5 of gtest :
```
GTEST_ATTRIBUTE_NO_SANITIZE_ADDRESS_
static bool StackGrowsDown() {
 //  int dummy;  ///causes initialiazation error when trying to compile
 int dummy = 0 ; // compiles and works  
bool result;
  StackLowerThanAddress(&dummy, &result);
  return result;
}
#  endif  // GTEST_HAS_CLONE
```


bood:
You may want to change the commit message style to align with the existing ones, i.e. capitalize the first letter

